PR #53, opened 2014-09-12
Changes to make bits in packets accessable to help protocol detection

jcarduino:
Changed callbacks to include bits in packet, also shown in output.

Now we can also check bits to determine if this is a packet we want or bail out.

For weather stations it still is a drag as it seams that they use some alike format but different interpretation of the fields. So no proper detection possible there :-(. What we can do however is perhaps include a config file reading what unitname/devicenumber has protocol x.... but that presumes you know what unit numer it holds...

So please comment on this pull request....


merbanan:
I need to spend some time to re-architect the whole thing. I just dont have the time to do it ;/

PR #60, opened 2014-10-16
acurite5n1 fixes

zerog2k:
* fix raincounter logic
 * fix wind directions
 * note/todo about wind speed inaccuracy

PR #73, opened 2015-01-11
simple fix to rain gauge support for Acurite 5n1 sensors

helgew:
Pulled the setting of the raincounter variable out of the conditional block. As it was, acurite_raincounter was never set to anything other than 0.

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -521,10 +521,9 @@ static int acurite5n1_callback(uint8_t bb[BITBUF_ROWS][BITBUF_COLS]) {
             // wind speed, wind direction, rainfall
 
             float rainfall = 0.00;
-            int raincounter = 0;
+            int raincounter = acurite_getRainfallCounter(buf[5], buf[6]);
             if (acurite_raincounter > 0) {
                 // track rainfall difference after first run
-                raincounter = acurite_getRainfallCounter(buf[5], buf[6]);
                 rainfall = ( raincounter - acurite_raincounter ) * 0.01;
             } else {
                 // capture starting counter


helgew:
btw - as is, the code will report cumulative rainfall since the first rain gauge report received. I modified my own code to report cumulative rainfall since the last time a rain gauge report was received. Let me know if you want that patch as well.

zerog2k:
@helgew 
I had fixed this and more submitted in PR #60 a long while back, but looks it never got merged.
So I guess I'll have to find some time to resolve conflicts and resubmit PR.

helgew:
@zerog2k never saw your PR, sorry!

What are your thoughts on resetting acurite_raincounter to the value of the station's counter to report incremental rain measurements rather than cumulative?

zerog2k:
I would say report both cumulative rainfall amount since startup and since last measurement - however the last measurements are frequent (and can be missed), so perhaps it would be better to keep a sliding window and report either the rainfall amount over some time N, or better yet report it as a rate (e.g. 0.10 inch/hr for last 15 mins, etc).
Ideally most of this behaviour would be configurable via command line switches, but these options can be highly device dependent and currently rtl_433 looks like it would need a good deal of refactoring to make this clean and maintainable. Current organization would make scaling this messy.
This is a separate topic, but I had actually started some work to factor out core logic from device-specific code to create a cleaner separation but I never really kept it up to date. I'm wondering if Ben @merbanan would be interested in supporting this (through reviewing/merging) if a few folks work on streamlining it.

helgew:
@zerog2k Is [this](https://github.com/helgew/rtl_433/tree/refactor) what you had in mind? I thought about doing the same thing earlier and just went ahead with it. My C/C++ is mighty rusty, so please take a close look at the header files. Obviously, this is just a first step but maybe it is useful. 

As for command line options that could be passed to individual devices, those could probably be added as a call back to the r_device struct.

merbanan:
@zerog2k sure I'm interested in that. Almost everything needs a rewrite though. All this was mostly a quick hack.

PR #57, opened 2014-10-05
Add support for the Acurite temperature and humidity sensor

pbfy0:
This adds support for the Acurite temperature and humidity sensor (http://amzn.com/B001B35APA).
Output looks like this:
```
Temperature event:
protocol      = Acurite Temp&Humidity
temp          = 12.1°C
humidity      = 50%
```

PR #54, opened 2014-09-20
Added KlikAanKlikUit protocol

jcarduino:
Added protocol for switches, on/off and also DIM commands.
This protocol is the new version, old version has different handler.

Please add to repository.

PR #52, opened 2014-09-09
WS3500 AlectoV1 working now

jcarduino:
Hi,

This code is for all Alecto V1 protocol weather stations. Prologue is detecting these device also, but representation of the fields is not correct.

AlectoV1 is used by Alecto, Ventus, Oregon and more.
tested with:
Oregon Scientific THN132N (temp/hym sensor) 
Alecto WS3500, WS4500 (weatherstation)
Ventus W155 (weatherstation) 
Ventus W044 (temp/hym sensor) 
Ventus W155 (weatherstation)
Alecto DKW2012 (weatherstation on 868 MHZ!) 

Please comment/edit

PR #50, opened 2014-09-06
Added support for Mebus Wireless sensors

toke:
Added support for mebus wireless Sensors and basic support for intertechno remote control.

I've merged your upstream with my changes and also fixed a format string warning from the last commits.

PR #74, opened 2015-01-17
Refactoring of device support and combined pull requests for PRs #34, #50, #52, #54, #57, and #60

helgew:
I have factored out the device support into separate files and merged in the commits from all other pull requests.


rct:
It's great to see you taking on this work.

Have you seen @Batilan's plugin work:  https://github.com/Batilan/rtl_433/tree/plugins?

There was a discussion under a different issue.  The decode routines are run (iterating over every sample) multiple times for device defined.

helgew:
@rct I had not seen @Batilan's work, otherwise I would have gone off of it since it is much more mature already. What do you think, @merbanan?

merbanan:
Looks good too me from a brief look.

rct:
We need a better place than an issue to discuss rtl_433.  Issues get closed and the information gets hidden.

The structure I've been thinking about is to do the edge detection that analyze mode does and then to let the call backs work with that to determine if it is their signal or not rather than having every callback iterate over every sample each time to find the edges.

Issue #76, opened 2015-01-17
rtl_433.h:48:30: error: unknown type name ‘uint8_t’


0xFACE:
Can't build rtl_433 now:
/rtl_433.h:48:30: error: unknown type name ‘uint8_t’

I added #include<stdint.h> to workaround.

helgew:
It compiles fine for me... what platform are you compiling on and what is the include you added?

0xFACE:
Linux raspberrypi 3.12.35+ #730 PREEMPT Fri Dec 19 18:31:24 GMT 2014 armv6l GNU/Linux
I included "stdint.h".

helgew:
I'll submit a patch... stdint.h is included via rtl-sdr.h, which is included in rtl_433.c. This works for my setup, but apparently, different compilers handle that differently.

0xFACE:
Ok, works now.

PR #77, opened 2015-01-17
added include to satisfy unresolved dependencies on some systems

helgew:
fixes issue #76


rct:
I couldn't compile on an older version of Linux with out this change.

Tehrasha:
I needed this for Ubuntu 14.04LTS as well.

helgew:
Made redundant by #82

PR #89, opened 2015-01-21
Update oregon_scientific.c with missing bracket 

pclov3r:
Fails to compile without the opening bracket.


clienthax:
+1

rct:
To be clear, it's the closing brace that appears to be missing to make this compile, not the opening one.  No idea if once compiled the oregon scientific actually works but rtl_433 is compiling and other protocols are working.

pclov3r:
I'm not that much of a programmer. Thanks for that comment.

Yes, it does work. Did for me at least.

PR #93, opened 2015-01-27
Add Owl CMR180 support

onlinux:
Energy Owl CMR180 sensor support added.


merbanan:
The patch is quite messy, type to make cleaner diffs the next time.

PR #95, opened 2015-01-27
First draft of unified reporting infrastructure -- DO NOT MERGE

helgew:
This is just for illustration and discussion of specific implementation details. I think we should continue to discuss the overall philosophy in the google groups thread.

The easiest entrance point for code review might be 4ca713c. Please let me know what you think!


Batilan:
Hi Helge,

As mentioned on the forum: a really nice achievement. One minor thing I noticed: it seems there are a few memory leaks in here? (float_to_str, get_device_details, get_weather_report results are never freed?). But I guess as a Java programmer you are allowed to create memory leaks in a terrible language like C ;-)

helgew:
Details, details, details... :)

Seriously, this is why I wanted someone else to have a look at my code. I suspected already that a variable going out of scope would not free the memory previously allocated to it. That said, I am not sure what to do about it from a design point of view. Should it be left to the client of of the device_reports functions to free that memory, or should the print_..._report functions take care to that? float_to_str needs to be re-written to accept a char pointer with pre-allocated, fixed-sized memory.

merbanan:
I will close this now.

PR #108, opened 2015-03-29
Cardin improved

Lefinnois:
Hi,

Getting false positive with Oregon Scientific and/or Lacrosse Sensors, i've added some tests in Cardin remote control decoding.

Bits 4&5 of byte 2 must allways be 0 and 4 and bits 0-3 can only be 3, 9, 12 or 6. 24 bits Cardin remote messages are so simples that's there is nothing else to test for. The rest of the message (18 bits on 24) is the encoded dipswitchs positions.

Regards


merbanan:
Next time look at the diff. Right now you add lots of tabs. Patch merged.

PR #109, opened 2015-03-29
added Oregon Scientific UVN800 UV sensor

Lefinnois:
Hi,  
me again :)
Here is small addition to Oregon Scientific support with UVN800 ultra-violet sensor (sensor ID 0xD874 protocol version 3). Decoded as described here : http://wmrx00.sourceforge.net/Arduino/OregonScientific-RF-Protocols.pdf (page 11).
Regards


merbanan:
Thanks

PR #111, opened 2015-03-30
Nexus sensor from Clas Ohlsen

ukshady:
Very similar to the Rubiscon/Auriol sensor but required changes to extract the humidity readings.  This code also writes the last readings to a one line file called nexus.csv in the format of: 
Nexus date time, Unix timestamp, temp, humidity
i.e Nexus 29/03/15 18:53, 1427651617,8.3,72

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -19,7 +19,8 @@
 		DECL(newkaku) \
 		DECL(alectov1) \
 		DECL(cardin) \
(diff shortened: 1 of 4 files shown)


merbanan:
I will merge all code but the csv writing when I get home. There is a discussion on the mailing list with how the sensor data should be output to cvs and the likes.

merbanan:
Can you submit an example signal to rtl433_tests ?

PR #116, opened 2015-05-01
Disable buffering for stdout and stderr, workaround for data redirection...

Serveurperso:
... and logging

To help some users want to make sensors logging ie. /usr/local/bin/rtl_433 | cut / grep / awk last sensors values to text file and to Cacti "Data Input Methods" template and make RRDtool weather graph...

Pascal

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -987,6 +987,9 @@ int main(int argc, char **argv) {
     char vendor[256], product[256], serial[256];
     int have_opt_R = 0;
 
+    setbuf(stdout, NULL);
+    setbuf(stderr, NULL);
+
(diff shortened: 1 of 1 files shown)


merbanan:
How does this help ? Can't this be fixed some cleaner way ?

Serveurperso:
The code need a complete rewrite for data output & plugin management. my workaround is a Quick'n'Dirty trick and it's the only way for now to get realtime data (look http://www.serveurperso.net/meteo it's a raspberry pi meteo server receiving oregon scientific data over this RTL-SDR tool). There is some beautifull work from other users pending merge.
Sorry for my english. Pascal

PR #124, opened 2015-06-29
New packet based pulse demodulation infrastructure

vestom:
Currently each added device adds its own sample based demodulator, which is running at the full sample rate. This adds a significant CPU overhead and becomes prohibitive when running e.g. on a Raspberry Pi.

This new infrastructure adds a single packet based pulse detector, which converts the samples to a list of pulses/gaps. This list of pulses is then processed once per packet per device. This should reduce CPU overhead significantly as only one function (the common pulse detector) is running at full sample rate.

The fineoffset driver is converted to the new infrastructure and tested OK with both test data and live data. If accepted I will add a new sensor to the infrastructure and start converting other drivers.

Pros:
- Potential for significant reduction in CPU overhead (only one device converted for now...)
- Simpler stateless demodulators
- Pulse analyser giving more statistics and making guesses at modulation type (with -D switch)
- Some common functions re-factored into own c files
- Designed to co-exist with existing decoders for a transition period
- Designed to have no (or little) impact on device drivers

Cons:
- Packet based processing has to declare End Of Package in a generic way and may require changes to device drivers assuming multi-packet data.
- Currently has some code redundancy (bitbuffer.c), which will be cleaned up soon.


rct:
This looks very interesting.  Thanks for working on this.  I've been wanting to do something based on detecting the pulses for a while.

I suggest you also send this to the google group, rtl_433, for wider distribution and for perhaps a better discussion.

How to detect the end of a message seems a bit tricky, especially to keep it somewhat generic.   What if you have two signals that are close together without enough silence in between?  

Also, I've been thinking that the device blocks instead of using an int to specify which demodulator, should instead take a function pointer to allow more freedom in specifying device specific demodulators if needed.

vestom:
I have already sent it to the google group https://groups.google.com/forum/#!topic/rtl_433/wkb-R7lT1G8 

I have addressed some of your comments there.

PR #132, opened 2015-07-15
Overall cleanup - fix most compile warnings

vestom:
- Removed debug_callback() from drivers and added consistent bitbuffer_print() in decoders
- Cleaned up drivers struct initialization
- Converted all drivers to bitbuffer for more future flexibility


merbanan:
For some reason github doesn't send out an email for the pull requests. That is why I didn't pull earlier.

Issue #58, opened 2014-10-10
Decode DSC security contact transmitters (odd OOK)


rct:
I'd like decode DSC security contact transmitters.   The encoding is OOK, but is a bit odd for rtl_433 detection, so -a isn't helping that much. 

Packets are 26.5 mS long
- Packets start with 2.5 mS of constant modulation
- The length of a bit is 500 uS, broken into two 250 uS segments.  
- A logic 0 is 500 uS (2 x 250 uS) of no signal.  
- A logic 1 is 250 uS of no signal followed by 250 uS of signal/keying
- Then there are 4 sync logic 1 bits.
- There is a sync/start 1 bit in between every 8 bits.
  - A zero byte would be 8 x 500 uS of no signal (plus the 250 uS of silence for the first half of the next 1 bit) for a maximum total of 4,250 uS (4.25 mS) of silence.
  - The last byte is a CRC with nothing after it, no stop/sync bit, so if there was a CRC byte of 0, the packet would wind up being short by 4 mS.
- There are 48 bits in the packet including the leading 4 sync 1 bits.   This makes the packet 48 x 500 uS bits long plus the 2.5 mS preamble for a total packet length of 26.5 ms.

Note: Logan Lamb found the protocol description in one of the FCC filings in prep for his 2014 DEFCON talk.   The document is here: https://apps.fcc.gov/eas/GetApplicationAttachment.html?id=100988   The FCC ID is F5300NB912

**\* signal_start = 167898276, signal_end = 167918759
signal_len = 20483,  pulses = 4
Iteration 1. t: 37    min: 22 (3)    max: 53 (1)    delta 4
Iteration 2. t: 37    min: 22 (3)    max: 53 (1)    delta 0
Pulse coding: Short pulse length 22 - Long pulse length 53

Short distance: 102, long distance: 0, packet distance: 120

p_limit: 37

[00] {1} 80 : 10000000
[01] {1} 00 : 00000000
[02] {1} 00 : 00000000
[03] {1} 00 : 00000000

**\* signal_start = 167959220, signal_end = 167979385
signal_len = 20165,  pulses = 2
Iteration 1. t: 27    min: 19 (1)    max: 36 (1)    delta 0
Distance coding: Pulse length 27

Short distance: 109, long distance: 0, packet distance: 109

p_limit: 27

[00] {0} 00 : 00000000
[01] {0} 00 : 00000000

merbanan:
[...]

PR #135, opened 2015-07-28
Initial DSC security parsing and debugging improvments

rct:
- Implements #58, Decodes DSC security contact decoding
- Multiple debug levels now supported by specifying -D multiple times
- View detected pulse/gap timing list with debug level 2 (-D -D)


merbanan:
Better levels can be added later.

PR #137, opened 2015-08-01
used dup2(1, 2); to redirect all output to stdout

jcarduino:
If there is any reason to keep err messages please let me know. Propose to use stdout.
As rtl libs use err forcing err to std is needed imho


merbanan:
The output from the sensors are to stdout, all other messages are to stderr. This is for all that sends data through a pipe and not wanting stderr messages.

PR #138, opened 2015-08-09
support Brennstuhl RCS 2044 remote control

ortylp:
This is implementation of decoder for Brennstuhl RCS 2044 remote control.

The modulation triggers unfortunately false positive in waveman.  This FP has been explicitly excluded.

I would suggest some way of testing the signal type already at demodulation.  Report demodulation quality, or some way of detect the correct one among competing decoders.  It means also, that output would have to be buffered and only the most likely demodulator+decoder pair should report to the console.

I have tried to generate enough data for rtl_433_test, but files recorded with ./rtl_433 -a -t produce no output when called as rtl_433 -r gfile001.data.  Tests with data recorded with rtl_sdr do work.  What should I do to contribute working signals to rtl_433_tests.  It is as far importar as we get fals positives with waveman (workarounded in this commit), ambient pollutes output with CRC error reports as soon as the network code for brennstuhl is set to 11111,   "Acurite 896 Rain Gauge" has also many false positive reports (not necessarily with brennstuhl remote signal).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Running:
            [21] Calibeur RF-104 Sensor
            [22] X10 RF
            [23] DSC (Digital Security Controls)
+           [24] Brennstuhl RCS 2044
 
 
 Examples:
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -26,7 +26,8 @@
 		DECL(ambient_weather) \
 		DECL(calibeur_RF104) \
 		DECL(X10_RF) \
-		DECL(DSC)
+		DECL(DSC) \
+		DECL(brennstuhl_rcs_2044)
 
 
 typedef struct {
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -42,6 +42,7 @@ add_executable(rtl_433
 	devices/calibeur.c
 	devices/x10_rf.c
 	devices/dsc.c
+	devices/brennstuhl_rcs_2044.c
 )
 
 target_link_libraries(rtl_433
--- a/src/devices/brennstuhl_rcs_2044.c
+++ b/src/devices/brennstuhl_rcs_2044.c
@@ -18,15 +18,56 @@
  * https://github.com/xkonni/raspberry-remote
  */
 
+static int brennstuhl_rcs_2044_process_row(int row, const bitbuffer_t *bitbuffer);
 static int brennstuhl_rcs_2044_callback(bitbuffer_t *bitbuffer)
 {
-  uint8_t *b = bitbuffer->bb[0];
+  int counter = 0;
+  for(int row=0; row<bitbuffer->num_rows; row++)
+    counter += brennstuhl_rcs_2044_process_row(row, bitbuffer);
+  return counter;
+}
+
+static int brennstuhl_rcs_2044_process_row(int row, const bitbuffer_t *bitbuffer)
+{
+  const uint8_t *b = bitbuffer->bb[row];
+  const int length = bitbuffer->bits_per_row[row];
   /* Two bits map to 2 states, 0 1 -> 0 and 1 1 -> 1 */
   int i;
   uint8_t nb[3] = {0};
 
+#if 0
+  {
+    // print raw bit squence for debug purposes (before exclusion of invalid sequenced is executed)
+    time_t time_now;
+    char time_str[LOCAL_TIME_BUFLEN];
+    time(&time_now);
+    local_time_str(time_now, time_str);
+    fprintf(stdout, "%s Brennstuhl RCS 2044: received RAW bit sequence (%d bits): ", time_str, length);
+    for(int i=0; i<4; i++)
+    {
+      for(int p=7; p>=0; p--)
+        fprintf(stdout, "%d", (b[i]>>p ) & 0x1);
+      fprintf(stdout, " ");
+    }
+    fprintf(stdout, "\n");
+
+    fprintf(stdout, "%s Brennstuhl RCS 2044: received RAW bit sequence (%d bits): ", time_str, length);
+    for(int i=0; i<4; i++)
+    {
+      for(int p=7; p>=0; p--)
+        if (p%2)
+          fprintf(stdout, ".");
+        else
+          fprintf(stdout, "%d", (b[i]>>p ) & 0x1);
+      fprintf(stdout, " ");
+    }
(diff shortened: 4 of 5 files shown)


merbanan:
You can select the protocols you want from the command line. There will always be collisions because not all signals have unique ways to be identified. And try with rtl_433 -m 0 gfile001.data. When you have sorted out the test data and you sent a pull for that I will merge this code also.

ortylp:
With rtl_433 -m 0 gfile001.data I see no proper data output.  This is the problem I was writing about, specifying explicitly -m 0 does not change anything.  Should I upload this not working test data anyway?

merbanan:
Yes, please upload it.

ortylp:
I am at the moment analysing why it does not work with the test data.  Can it be, that the LP filter is too aggresive.  Short pulses seem to bee extremely narrow.

merbanan:
Try with https://github.com/vestom/rtl_433/tree/experimental. Exactly that should be fixed.

ortylp:
Nope, it did not help much. I'll stay on it and report as soon as I find the problem and solution if possible.

ortylp:
I give up for today.

The demodulator is extremely sensitive to the changes in the filter implementation AND reception level.  Some kind of "intelligent", forward looking AGC would be probably better.  In my tests with the implementation from todays master I get more stable results when sending 3 meters away from the RX antenna instead of 1 meter distance.

Maybe I'll try to implement some kind of proof of concept computing on floats...

ortylp:
With the last commit I have fixed the instability (demodulator parameters were not properly chosen)

merbanan:
Thanks

vestom:
Yes, thanks for the good work - and also for making the waveman less chatty :+1: 

Regarding the problematic reception at close distance, could you try to make a test file for that?

ortylp:
The problems with reception at close distance has been solved by the last commit, so this is not an issue any more.

I got back home today and tested the setup with my raspi transmitter.  It looks like the raspi is sending some other code, but the receiver (the brennstuhl hardware switch == the authorised receiver) is still accepting it.  The signal is also much weaker, as it gas to go through two ceilings of reinforced concrete.  There is still work needed.

Does anybody know how the internal amplifier of the receiver is to be controlled?  The best solution would be to turn on the hardware AGC and then extract the signal using Fourier Transform.  Then it gets much more expensive computationally, but it gives orders of magnitude better sensitivity.  I have tens of signals I can receive at home (see them clearly in waterfall in gqrx), but none of them gets recognised by rtl_433.

I have also to change the name to some more generic name, as I have RCS 2044SN and RCS 1000N and they interoperate, so maybe simply "Brennstuhl RCS"

One more question on the side: are we sticking to C99?  If yes, I'll update my code to use it fully.

merbanan:
C99 yes, at least the parts which MSVC can compile. With regards to sensitivity you can lower the detection limit to -l 2000 or so.

merbanan:
Btw, the rtl sdr agc is active. Can you describe the demodulation algorithm based on a fft ? There are lots of tricks we can apply to get better reception. Right now the code is written for "embedded" type of systems.

ortylp:
Which version of MSVC? 2008 did not even fully support C89.  MSVC 2013 is more or less up to date.
I have no M$ products at home, so I cannot really test...

I was afraid "embedded" is the key word here, otherwise I would never want to use pure ANSI C any more...

What I describe below is a brute force, proof of concept, and in first implementation potentially really computationally expensive:
* our signals we want to demodulate are OOK (and FSK?)
* the main signal energy is in a very small frequency range (compared to our sampling rate of 250kS/s)
* the signal energy is never on expected frequency, it is 433.92MHz +/- 80kHz

Brute force demodulation could look like this:
* compute short FFTs (64 or 128 samples at a time), without overlap, maybe even with small gaps between
* for each FFT block detect of we have (flat) noise or a few pins contain most of the signal energy 
* using this method detect the start and end of transmission (energy and constant frequency)
* for the complete data block compute (long) FFT (use libfftw!) (I am talking desktop PC or Raspi2)
* set all frequency pins with noise to 0, leave pins containing 90% of the energy untouched
* move pins containing energy to baseband
* decimate the signal a bit (ignore half, or 3/4, or 7/8 zeroed frequency pins)
* compute IFFT to get our  (decimated) time signal again
* find maximum energy level and set threshold to the half of it
* feed the signal to the current set of demodulators (or implement new ones with proper clock recovery and symbol decision)

This approach with FFT has three functions: detect signal energy (discriminate it from noise), detect on which frequency is the signal we want to process, and the final one, that gives us the (proper) sensitivity is filtering noise and distortions out of our signal.  Alternatively one could just use mixer and low pass filter, but once we have the complete signal in FFT it is much easier to do that in frequency domain.

What we could do is to enable multiple preprocessing frontends as an "advanced" option and/or depending on the CPU resources we are willing to (ab)use.   It will never be as fast as the current implementation, but will provide vastly better demodulation performance.

[small disclaimer: I am developing and implementing demodulators in C++ in my day job]

merbanan:
I am thinking of the latest version of MSVC. I sort of like your ideas but I'm not convinced about going back from the frequency domain. I'm thinking a lapped 128 fft transform should give enough time resolution to make actual demodulation possible on the fft data. This is backed by nothing more then a hunch.

ortylp:
This would be another approach... if these short FFTs are at least two times shorter than the shortest pulse, the you could  just use the power of the strongest FFT pin, and go from there.

The approach with the single long FFT would still be much better in terms of sensitivity, It would remove much more noise and could adapt to the symbol rate of the received signal.

rct:
@ortylp Interesting stuff.  Glad to see your contributions.   This issue will likely get closed soon. Especially, now that this remote is supported.  This thread will be obscured.

I suggest moving the discussion to the rtl_433 google group,  rtl_433@googlegroups.com.

@vestom is currently discussing adding FSK demodulation and reworking the baseband processing.

ortylp:
FYI: I have a working proof of concept (detection+long FFT + IFFT) working for a single rtlsdr_callback().
I still need to run regression tests, put a lot of refactoring work into making the algorithm working across data blocks and optimise it for speed, but I am already reliably receiving previously undetected sensors :D

The only drawback: false positive rate went also up: "AcuRite Rain Gauge" and "Acurite 5n1 sensor".
I replaced calls to envelope_detect() and low_pass_filter(). If demodulator steps can work with reliably with decimated signal, the decimation would be almost for free (and actually speeding up IFFT).

PR #139, opened 2015-08-12
Adjust low pass filtering and add hysteresis to pulse detection.

vestom:
Low pass filter was very aggressive to avoid glitches in pulse detection. That would make pulse edges soft and pulse widths dependent on detection limit/signal level.
Now decodes Acurite test files correctly and all DSC test files. No regression found for other test files (including the new Brennstuhl)

...and a bit of clean up and FSK decoding preparation.


merbanan:
Man tackar

Issue #117, opened 2015-05-01
Help to implement Sensor GT-WT-02 (Aldi, Globaltronics..)


ludwich66:
Hi, 
i need help to implement a "new" Sensor.
I am not a hacker, but i had tested it with an excel code and it calculate the data of the Sensor correct.
See my example.
Who can help me here?

Sorry for my bad english!

ludwich

outputfie
*** signal_start = 113970967, signal_end = 114212857
signal_len = 241890,  pulses = 240
Iteration 1. t: 135    min: 129 (239)    max: 141 (1)    delta 9
Iteration 2. t: 135    min: 129 (239)    max: 141 (1)    delta 0
Distance coding: Pulse length 135
Short distance: 519, long distance: 1028, packet distance: 2257
p_limit: 135
[00] {0} 00 : 00000000 
[01] {37} 34 00 ed 47 60 : 00110100 00000000 11101101 01000111 01100000 
[02] {1} 00 : 00000000 
[03] {37} 34 00 ed 47 60 : 00110100 00000000 11101101 01000111 01100000 
[04] {1} 00 : 00000000 
[05] {37} 34 00 ed 47 60 : 00110100 00000000 11101101 01000111 01100000 
[06] {1} 00 : 00000000 
[07] {37} 34 00 ed 47 60 : 00110100 00000000 11101101 01000111 01100000 
[08] {1} 00 : 00000000 
[09] {37} 34 00 ed 47 60 : 00110100 00000000 11101101 01000111 01100000 
[10] {1} 00 : 00000000 
[11] {37} 34 00 ed 47 60 : 00110100 00000000 11101101 01000111 01100000 
[12] {0} 00 : 00000000

/* 2 examp les
/* [01] {37} 34 00 ed 47 60 : 00110100 00000000 11101101 01000111 01100000
/* code, BatOK,not-man-send, Channel1, +23,7°C, 35%
/*
/* [01] {37} 34 8f 87 15 90 : 00110100 10001111 10000111 00010101 10010000
/* code, BatOK,not-man-send, Channel1,-12,1°C, 10%

/* SENSOR: GT-WT-02 (ALDI Globaltronics..)
/* TYP AAAAAAAABCDDEFFFFFFFFFFFGGGGGGGxxxxx
/* BIT 0000000001111111111222222222233333333
/* BIT 1234567890123456789012345678901234567
/* 
/* TYPDescriptian
/* A = Rolling Device Code, Change after battarie
[...]

PR #140, opened 2015-08-16
initial implementation for GT_WT_02 decoder

ortylp:
This is initial implementation that works with the data we have.

We need more examples.  We need to reverse engineer the last set of bits, hopefully, some kind of CRC.

See also Issue  #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Running:
            [22] X10 RF
            [23] DSC (Digital Security Controls)
            [24] Brennstuhl RCS 2044
+           [25] GT-WT-02 Sensor
 
 
 Examples:
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -27,7 +27,8 @@
 		DECL(calibeur_RF104) \
 		DECL(X10_RF) \
 		DECL(DSC) \
-		DECL(brennstuhl_rcs_2044)
+		DECL(brennstuhl_rcs_2044) \
+		DECL(gt_wt_02)
 
 
 typedef struct {
--- a/include/util.h
+++ b/include/util.h
@@ -39,4 +39,11 @@ uint8_t crc8(uint8_t const message[], unsigned nBytes, uint8_t polynomial);
 /// @param buf: output buffer, long enough for YYYY-MM-DD HH:MM:SS
 void local_time_str(time_t time_secs, char *buf);
 
+/**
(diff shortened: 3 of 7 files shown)


merbanan:
+static int gt_wt_02_process_row(int row, const bitbuffer_t *bitbuffer);
+static int gt_wt_02_callback(bitbuffer_t *bitbuffer)

Change the order so you don't need to forward declare the function.

+    printf("%s GT-WT-02 Sensor %02x: battery %s, channel %d, button %d, temperature %3.1f C / %3.1f F, humidity %2d%%\n"

use fprintf

+    time_t time_now;

Timestamp printout does not belong in the decoder. We should add some global handling of timestamp, not sure how yet so you can keep it if you add a comment about it.

+float celsius2fahrenheit(float celsius)

The output should be what the user configures it to be, but as we don't have the supporting code yet just add a comment about it.

If you fix these issues I will then merge the code.

ortylp:
I am on it.

ortylp:
I do agree, that we need infrastructure supporting the following:
1/ printing should be put outside of the decoder
2/ time stamps should be also outside of the decoder but based on the time stamp provided by the decoder (bitbuffer must support timestamps)

In my use case the time stamp is essential and a single package must fit its information on a single line, so that unix text processing tools (grep/awk/sed/perl) can work with it without additional code.

merbanan:
I agree on all points. I think the thread about the output format has some relevant info regarding this.

ortylp:
Can you point me to the thread?  Is it an issue on Github, do we have wiki to document requirements on this?

merbanan:
It's a google groups thread:

Proposal: Output to stdout and adding timestamps to the output

https://groups.google.com/forum/#!topic/rtl_433/2mTMjihXbRg

ortylp:
"You do not have permission to view this content."

merbanan:
https://groups.google.com/forum/#!forum/rtl_433

merbanan:
Search for the topic "Proposal: Output to stdout and adding timestamps to the output".

merbanan:
I have now changed the setting so anyone should be able to read topics in this group. I just didn't know this setting was active.

ortylp:
@merbanan do you want me to change anything more, or we "wait" until we figure out the CRC stuff?

merbanan:
Nope thanks for the update.

PR #143, opened 2015-08-19
Waveman: reduce false positives by checking the bit length of the mes…

ortylp:
…sage


merbanan:
Thanks.

ortylp:
Sorry I was not thinking clearly... it should have been 
(b[3] & 0xe)

PR #145, opened 2015-08-19
GT_WT_02: fix gap length threshold as in latest data file

ortylp:
See #117 

The test file provided in test repository seems not be the correct one?!??!?


merbanan:
There are other transmitters sending all the time. Could be that.

ortylp:
But the bit sequence is OK for old and new recording.  Even the checksum is correct.

merbanan:
Thanks.

merbanan:
Can you remove the old sample and send the new ones ?

PR #146, opened 2015-08-19
FSK demodulation

vestom:
OK, this is a big one :-)
It adds support for FSK demodulation and adds a couple of stub device drivers (these FSK devices have long bit sequences and might take some time to decipher).
The FM demodulated data can be output to a test file by using -m 2.
It has been successfully regression tested against the existing test corpus and should not make changes for the OOK devices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,15 +15,15 @@ Installation using cmake:
     cd build
     cmake ../
     make
-	make install
+    make install
 
 Installation using autoconf:
 
     cd rtl_433/
     autoreconf --install
-	./configure
-	make
-	make install
+    ./configure
+    make
+    make install
 
 
 Running:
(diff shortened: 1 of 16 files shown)


merbanan:
Add an option to not perform fsk demodulation as I suspect that it uses some cpu.

rct:
Ideally (i think) the devices selected for decoding would trigger the type of demodulation that needs to be performed. So if only an OOK devices or two was enabled (-R) no FSK demodulation would be done.

vestom:
Good idea. I suspected though that the CPU increase was not that big, so I made some benchmarks on my Raspberry Pi Model B (version 1) 800 MHz. As you can see the old sample based decoders are the biggest cause of CPU usage. However for debugging and the very resource constrained CPU it does make sense. I will make a patch later.
________________
Git c614ce91 (2015-02-03) (Before pulse based demodulation and cmd line selection of devices)
- As is, with 17 active devices: ~55% CPU
- Modified source to only one device (Fine Offset): ~10% CPU
___________________
Git d9026dde (2015-08-20): (With FSK and pulse demodulation)
- As is, with 27 active devices: ~53% CPU
- rtl_433 -R 18 (Fine Offset): ~15% CPU
- rtl_433 -R 3 -R 8 -R 12 -R 17 -R 18 -R 20 -R 21 -R 22 -R 23 -R 24 -R 25 -R 26 -R 27
  (All pulse based devices and no sample based devices): ~15%
- rtl_433 -R 18, source modified to exclude FSK: ~10% CPU

vestom:
Just out of curiosity I tried to modify the code to use the floating point atan2() for FM demodulation
- rtl_433 -R 18, source modified to atan2() FM demod: ~30% CPU

So it is a good idea to avoid floating point ;-)

PR #147, opened 2015-08-20
Disable FM demodulation when no FSK devices are enables

vestom:
CPU usage on a Raspberry Pi 1 B (800 MHz):
- rtl_433 -R 18 (Fine Offset) : ~10% CPU
- rtl_433 -R 26 (Danfoss) : ~15% CPU

(Plus a bugfix...)

### DIFF
--- a/include/baseband.h
+++ b/include/baseband.h
@@ -36,6 +36,7 @@ typedef struct {
 /// FM_Demod state buffer
 typedef struct {
 	int16_t	br, bi;		// Last I/Q sample
+	int16_t xlp, ylp;	// Low-pass filter state
 } DemodFM_State;
 
 /// Lowpass filter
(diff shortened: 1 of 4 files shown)


vestom:
Just for info. When changing compiler optimization setting to -O 2:
- rtl_433 -R 18 (Fine Offset) : ~4% CPU
- rtl_433 -R 26 (Danfoss) : ~6% CPU
- rtl_433 (All sensors) : ~21% CPU

We should really do something about those remaining sample based demodulators...

ortylp:
A few hints you probably know, but as reminder for scratching the optimisation itch:
1. turn compiler optimisations on
2. analyse with profiler (oprofile, or valgrind as first approximation)
3. compare results from different architectures (intel/arm, CPU cache size)
Important: do not try to optimise beforehand (usually waste of time), use plenty of regression testing

merbanan:
thank you

PR #148, opened 2015-08-20
GT-WT-02: exclude explicitly all zero sequence as false positive

ortylp:
bugfix for computation of nibbles
code simplification


merbanan:
Looks good.

PR #151, opened 2015-08-22
Add Valeo, Chuango devices. Improve pulse analyzer.

vestom:
- Valeo Car Key (Detect key presses...)
- Chuango Security Technology (Home alarm system)
- Pulse analyzer: Add automatic demodulation attempt
- Pulse analyzer: Add separate option '-A' to enable it

### DIFF
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -56,10 +56,10 @@ struct protocol_state {
     unsigned int modulation;
 
     /* demod state */
-    int pulse_length;
+    unsigned int pulse_length;
     int pulse_count;
     int pulse_distance;
-    int sample_counter;
+    unsigned int sample_counter;
     int start_c;
 
     int packet_present;
@@ -67,9 +67,9 @@ struct protocol_state {
     int real_bits;
     int start_bit;
     /* pwm limits */
-    int short_limit;
-    int long_limit;
-    int reset_limit;
+    unsigned int short_limit;
+    unsigned int long_limit;
+    unsigned int reset_limit;
     char *name;
     unsigned long demod_arg;
 };
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -30,7 +30,8 @@
 		DECL(brennstuhl_rcs_2044) \
 		DECL(gt_wt_02) \
 		DECL(danfoss_CFR) \
-		DECL(ec3k) 
+		DECL(ec3k) \
+		DECL(valeo) 
 
 
 typedef struct {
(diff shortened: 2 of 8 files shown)


merbanan:
Thanks

ortylp:
Valeo Car Key recognises my keys... so I would assume it works...

Chuango Security Technology gives me false positives for Cardin remote

vestom:
Yeah, the Chuango uses a really crappy protocol. There is no CRC or other magic identifiers to accept/reject the messages. Only 25 bits, where the 20 are sensor ID and 4 are command and the last bit seems to be always 1. I check for 25 bits and the last bit being one. I excluded sensor ID 0xFFFFF in a later commit to avoid the "all bits 1" case. Any other ideas for how to reduce false positives are welcome.
Maybe we should consider making PWM/PPM decoding more timing selective instead of the coarse threshold. That is a larger change though...

merbanan:
You can add a check for how many rows you get in a message. (We need to define a common terminology so we know what we talk about.) But other then that we can add the signal length and bit lengths to make signals more unique.

vestom:
I cannot use row number as a discriminator. First of all the chuango sensors seem to send different number of packets per message. E.g. the push-button remote may only send 1 or 2 packets per button press. Secondly the gap between packets is just above the current limit of 10 times the longest pulse. We can raise this limit of course - but it doesn't solve the problem.

When looking at the timing in Audacity the Chuango and Cardin are actually very similar - only the Cardin seem to start with an extra long pulse (4-5 times normal). To have better discrimination we should probably change the decoders from using short/long timing thresholds to have the actual timing (width of short pulse, long pulse, min gap, max gap etc.). Then we could add some tolerance to that and filter out anything outside the correct timing.

merbanan:
Try and see if you get more packets if you hold the button longer.

merbanan:
And we should somehow extract all signal properties. Bit lengths, preamble, sync bits etc. Then collisions should be minimal.

nebman:
Look up the [datasheet](http://akizukidenshi.com/download/ds/sctech/SC2272.pdf) for SC2272 or PT2272 for the full encoding. It uses tri-state bits for addressing (Bits 0/1/F).
My keyfob remote has the SC2262 counterpart (encoder IC) and got detected as Chuango, so I think this ICs are used. 

Different counts of packets result from pressing the keys not long enough, since the remote's circuits are only powered while keypress.

ortylp:
* I would prefer specifying for demodulator/decoder enough information about the waveform, that based on this information implementation of a generator would be possible.
* if we store the timing information (and tolerance intervals) for each transmission type, then we can use it for creating a proxy value for the "quality" of demodulation, which would also help reduce false positives.

PR #152, opened 2015-08-23
Update README. Improve Pulse Analyzer

vestom:
Fooling around with wiki formatting - but should be easier to update in future.
pulse_analyzer(): Ternary PWM detection works better. Code simplified by sorting histograms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Running:
     rtl_433 -h
 
 ```
- Usage:	= Tuner options =
+Usage:	= Tuner options =
 	[-d <device index>] (default: 0)
 	[-g <gain>] (default: 0 for auto)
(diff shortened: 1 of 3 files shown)


merbanan:
Thanks

merbanan:
Can you add code that will generate a template file for a new sensor. Ie calculate the correct pulse values and add placeholder functions?

vestom:
Interesting idea :-) That kind of metaprogramming is a bit above my head though. Getting the whole tool-chain working is probably a bigger problem than making a stub-driver by copying an existing driver.

ortylp:
If we assume, that pulse_analyser() detects modulation type and measures timing, then it should be possible to spit out the stub for the decoder.  Just a longer printf...

PR #155, opened 2015-08-29
Implement a PWM decoder with precise timing for increased selectivity

vestom:
Makes Chuango less noisy
Has provision for replacing ternary encoder (implements sync)
Can be expanded to include more selectivity (e.g. gap timing)
Can be expanded to include start bit removal (for replacing RAW)


merbanan:
Thanks.

vestom:
The reason for tightening the timing is the decoder having a lot of false positives. The chip can be configured to a timing difference of more than a factor of 8! That would make the driver very noisy!
I have elaborated more in https://github.com/merbanan/rtl_433/pull/156, please continue the discussion there.

PR #156, opened 2015-08-29
chuango: invert coding, display SC2262 8bit/4bit (16bit/8bit in binary) address/data model and tristate

nebman:
in SC2262 encoding it is possible to have any combination of 6-12 address bit, using the remaining of the 12 bits for 0-6 data bits. popular designs use 10+2 and 8+4 bit ICs. both are printed now.  
tristate encoding is used where 00 is 0, 11 is 1 and 01 is F (floating input, only possible in address part).

### DIFF
--- a/src/devices/chuango.c
+++ b/src/devices/chuango.c
@@ -21,38 +21,68 @@
 static int chuango_callback(bitbuffer_t *bitbuffer) {
 	bitrow_t *bb = bitbuffer->bb;
 	uint8_t *b = bb[0];
+	b[0] = ~b[0];
+	b[1] = ~b[1];
+	b[2] = ~b[2];
+
 	unsigned bits = bitbuffer->bits_per_row[0];
 
 	// Validate package
 	if ((bits == 25)
-	 && (b[3] && 0x80)	// Last bit is always 1
-	 && (b[0] != 0xFF) && (b[1] != 0xFF) && (b[2] != 0xFF)	// Reduce false positives. ID 0xFFFFF not supported
+	 && (b[3] && 0x7F)	// Last bit is always 0
+	 && (b[0] != 0x00) && (b[1] != 0x00) && (b[2] != 0x00)	// Reduce false positives. ID 0x00000 not supported
 	) {
 		uint32_t ID = (b[0] << 12) | (b[1] << 4) | (b[2] >> 4); // ID is 20 bits (Ad: "1 Million combinations" :-)
 		char *CMD;
 		switch(b[2] & 0x0F) {
-			case 0x0:	CMD = "0x0 (?)";	break;
-			case 0x1:	CMD = "0x1 (?)";	break;
-			case 0x2:	CMD = "0x2 (Low Battery)";	break;
-			case 0x3:	CMD = "0x3 (?)";	break;
-			case 0x4:	CMD = "0x4 (24H Zone)";	break;
-			case 0x5:	CMD = "0x5 (Single Delay Zone)";	break;
-			case 0x6:	CMD = "0x6 (?)";	break;
-			case 0x7:	CMD = "0x7 (Arm)";	break;
-			case 0x8:	CMD = "0x8 (Normal Zone)";	break;
-			case 0x9:	CMD = "0x9 (Home Mode Zone)";	break;
-			case 0xA:	CMD = "0xA (?)";	break;
-			case 0xB:	CMD = "0xB (Home Mode)";	break;
-			case 0xC:	CMD = "0xC (Tamper)";	break;
-			case 0xD:	CMD = "0xD (Alarm!)";	break;
-			case 0xE:	CMD = "0xE (Disarm)";	break;
-			case 0xF:	CMD = "0xF (Test)";	break;
+			case 0xF:	CMD = "0xF (?)";	break;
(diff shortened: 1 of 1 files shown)


vestom:
I don't really think we should output the tri-state values. It is only a way to encode the necessary bits by using fewer I/Os on the actual chip (apparently only some chips). The only thing that matters for output is the actual ID code itself (for differentiating between transmitters).

We should also decide whether we will make vendor specific  driver or a "generic" driver. Right now the Chuango is a vendor specific driver outputting vendor specific encodings/meaning, and we are now reworking it towards a more generic driver.

The biggest problem, is that the protocol is very simple/crappy and doesn't really make it possible to discern between vendors. The only true differentiation is in the timing selected by the vendor.
As there can be a factor of more than 8 between the timing selected, I think we cannot make a truly generic driver without making the decoder _very_ sloppy and having a lot of false positives! There are no CRC or magic markers to use for filtering.

My belief is that we should try to make vendor specific (to the OEM level) drivers with tight timing requirements, outputting meaningful decoded data. We could probably share some code between the drivers, but being so simple it doesn't really matter. For vendors with timing overlap, we must handle that specifically - but that at least gives us some vendor differentiation. It is not really useful for users of this program to just output "Message received: ID, CMD"

nebman:
Good input. Now that you added the specific tight timing, we can make the chuango driver output only vendor specific data. There should be little overlap with other vendors or generic devices anymore.  

For other devices, we can spin it off to a new, more generic driver. I have two generic devices at hand (remote and door sensor) and will buy more in the next weeks. There may be many combinations of timings, but I think only a few will be common (i.e. devices only use only either 3V or 12V batteries and specific oscillation resistors, otherwise they wont be able to work together). Of course we will have to make measures, to reduce false positives and not get the generic driver to be too noisy.  

I think it can be very useful for users to output "Msg received: ID, CMD", especially if you want to receive such events with Raspberry PI and build some scripts/projects on top of it.  

I can submit a pull request for code and test samples if you agree to continue this way.

vestom:
I think it is the right way forward, so go have fun :-)

I agree an ID, CMD is better than nothing, but we shouldn't throw away the additional information the setting of the timing resistor provides us. It seems this resistor is the only kind of "Vendor code" possible, and devices with different timing probably can't work together anyway.  So we should also make the other device drivers timing selective to the highest extent possible. 

But as you say, it is interesting to see how many combinations we will meet in reality, and it may be that for some timing values there is so much overlap, that it doesn't make sense to make a vendor specific decoding beyond the simple ID, CMD anyway... E.g. if many other vendors use the same timing as Chuango (142/426) we should probably strip out the Chuango specific decoding and leave that up to the user.

BTW when making new drivers with the "precise" decoder you can determine the polarity up front and avoid inverting the bits afterwards. I wasn't aware it was inverted before you showed us the datasheet :-)

merbanan:
@vestom we might come to the state where some decoders are not active by default. You have to enable it via a command line flag. The PIR I talked about has jumpers I can configure as I want. No way to get that working reliable without having configurable options in the receiver.

nebman:
@merbanan could you please take a picture please of the jumpers? or describe the possible settings? it must be something along 8bit address (H/N/L) when it's based on SC2260 and/or for choosing the oscillation resistor. when put to the right resistor, the generic_remote device should already be able to pick it up. there may be a timing drift though if you use a different voltage (mine uses a 3V coin cell which results in 116/351 timing, not sure about the resistor though - it's fixed smd).

merbanan:
https://drive.google.com/file/d/1yHxtclID0nQ0YPiw-1CLAbK5ozWuCjVqEg/view?usp=sharing

It is based on PT2262. Your code does not actually pick up the PIR.

nebman:
Does it work when you set it to 3.3M? That could be the right timing range, if I read the tiny tiny SMD resistor right (should be 10M on my SC2260) and comparing it right to [that table](http://blog.solidremote.com/post/identify-rf-remote-control.aspx).  
If not, can you please submit a recorded sample, so we can collect as much data as possible? :-)

merbanan:
tests/PT2262/* should have 3 signals and relevant info.

merbanan:
None of the settings worked.

vestom:
Interesting with such a configurable device.

However could we please call it e.g. "Generic SC266x" instead of "Generic Remote"? There are probably other "generic" remotes out there...

The datasheet says the receiver should have a clock of 2.5-4 times the transmitter - preferably in the middle. That corresponds to a ratio of 1.6 or a rough receiver tolerance of +/- 26% (sqrt(1.6)). We should probably have something similar, corresponding to a pulse tolerance of ~50 for your generic settings.

It looks like there could be maybe up to 5 different "channels" we can discern by making 5 devices. We should have these these 5 devices in the same file, but with different timing.

I can see we are approaching a point, where we shouldn't enable everything by default...

You should really insert your copyright - it is your code :-)

nebman:
@merbanan Thanks for the signals, if I up the tolerance to 50, your test file 2 and 3 are working. File 1 only shows zeros in "-a" mode. Even with higher tolerance we have 0 false positives in all tests.  
@vestom Yes, we can put different timings, but in the same file. But let's gather some data first. Maybe we can use the previous demod and have better checks to reduce false positives? Just an idea, I don't know the code very good yet ;)

PR #157, opened 2015-08-30
revert additional chuango output, add new device "generic_remote" to deal with SC226x and EV1527 coding

nebman:
So after the discussion in pull request #156 I reverted the additional output in chuango device, thus the device stays very vendor specific.  
I created a new device to deal with other devices using the SC/PT226x and EV1527 chips. For now the timing is tight and fits the remote I submitted in the test samples. Running all other tests resulted in no single false positive.


merbanan:
I think that the copyright is actually yours. Feel free to change that if you care.

PR #160, opened 2015-08-31
GT_WT_02: adjust display of humidity to mimic the authorised receiver, count channels from 1

ortylp:
@ludwich66 please test it, you gave me no relevant samples, so I cannot test it myself :(

See #117


merbanan:
Thanks

Issue #141, opened 2015-08-16
TFA Twin Plus 30.3049


iotzo:
Hi!
I've created little bash script to support TFA Twin Plus 30.3049 thermo/hygrometer protocol:

 rtl_433 -f 433920000 -l 500 -a -g 18 2>&1 |mawk -W interactive -F ":" '/\[01\] {36}/{print $2}' | sed -u 's/ //g'| stdbuf -oL rev | mawk -W interactive '{ printf"ibase=2; %s\n",substr($0,6,7) ; printf"ibase=2; %s\n", substr($0,16,9) }' | bc | mawk -W interactive 'NR==1{ print"----------------------------\n" ; system("date") ; printf"Humidity\t: %s\n",$1-28 } NR==2{ printf"Temperature\t: %s\n",$1/10 ; NR=0 }'

Here is some sample data:
[04] {36} e4 4b 70 73 00 : 111001000100 101101110 000 0111001 10000    ---> temp/hum:23.7/50
temp num-->13-21bit(9bits) in reverse order in this case "011101101"=237
positive temps ( with 000 in bits 22-24) : temp=num/10 (in this case 23.7 C)
negative temps (with 111 in bits 22-24) :  temp=(512-num)/10
negative temps example:
[03] {36} e4 4c 1f 73 f0 : 111001000100 110000011 111 0111001 11111 temp: -12.4 

Humidity:
hum num-->25-32bit(7bits) in reverse order : in this case "1001110"=78
humidity=num-28 --> 78-28=50

It would be nice if we have official support for that model/protocol!
10x!

merbanan:
Hi, add some sample files, look at:

https://github.com/merbanan/rtl_433_tests

This will help people to write the code for your protocol. And I will not accept new protocols without test files.

iotzo:
OK!
Done!
Sending pull request for the sample files!

ortylp:
I have working implementation based on your description.  I need to clean the code a bit, add some more tests and push to github.  I hope I get to it tomorrow at the evening.

We need to decode the purpose of other bits and kind of CRC, if any.  THIS IS IMPORTANT, otherwise we get false positives and that means disaster.  Can you please test if there are things like "channel" or "battery status", do bits change after battery replacement?

iotzo:
It works now, thank You!
The last 4bits maybe is some kind of CRC. Here is some samples:
{36} e4 25 c8 87 50 : 111001000010 010111001 000 1000011 1 0101
[...]

PR #161, opened 2015-08-31
TFA-Twin-Plus-30.3049

ortylp:
Decoder for TFA-Twin-Plus-30.3049

Almost all bits are decoded, checksum is working.
Negative temperature formula is too complex to be the right one, more samples and simplification is needed, the same for humidity.  

Please test it with the hardware, compare temperature AND humidity.  Humidity usually goes pretty low at negative temperatures.

See also #141

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -33,7 +33,7 @@
 		DECL(ec3k) \
 		DECL(valeo) \
 		DECL(chuango) \
-		DECL(generic_remote) 
+		DECL(generic_remote) \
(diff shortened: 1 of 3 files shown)


merbanan:
Can you use this bit reverse instead ?

/* bit swap */
static uint8_t bs(uint8_t x) {
    x = (x & 0xF0) >> 4 | (x & 0x0F) << 4;
    x = (x & 0xCC) >> 2 | (x & 0x33) << 2;
    x = (x & 0xAA) >> 1 | (x & 0x55) << 1;
    return x;
}

ortylp:
done

PR #164, opened 2015-09-01
Changed outputformat to be more linux like

jcarduino:
Changed output to be more conform logging rules. e.g. 2015-09-01 21:19:12 Prologue Sensor b8 Channel 03: Button 0:Temperature 25.7 C: Humidity 48 %: Battery Ok

### DIFF
--- a/src/devices/alecto.c
+++ b/src/devices/alecto.c
@@ -159,6 +159,7 @@ static int alectov1_callback(bitbuffer_t *bitbuffer) {
             reverse8(bb[1][0]),
             (bb[1][0] & 0xc) >> 2
             );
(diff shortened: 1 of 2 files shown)


merbanan:
Send a pull for the humidity check that I want. But I wont merge the logging change. If you read up on an old thread in the rtl_433 group we came to the conclusion that all decoders should output the same kind of object and then you should be able to decide what to transform the output in a later stage based on command line parameters. So what you want will be accepted just not this way. If you want to work on it you can start a new thread in the rtl_433 group and we can continue the discussion there.

jcarduino:
Will do..

Will however keep output change in private code for now as I have a script running which convert output of all sensors I receive in rtl_433 to weewx weatherstation software. Same format simplifies handling of evens. 

Mvg Johan Corstjens

Verstuurd via mobile

PR #165, opened 2015-09-02
Extra check false positive

jcarduino:
sometimes packages from prologue sensors come in detected with correct crc,
humidity is > 100 in these cases


merbanan:
Thanks.

Issue #166, opened 2015-09-05
New device (Digitech XC0348 weather station) at 433MHz


viraptor:
I'm trying to receive data from my weather station (Digitech XC0348). Documentation confirms it sends data at 433MHz. From running analyzer for some time, I believe I get consistent results. Skipping duplicate entries (many), here are some unique rows:

Common header: (always 88 pulses)

```
Detected OOK package
Analyzing pulses...
Total number of pulses: 88
Pulse width distribution:
 [ 0] mean:  134 [133;135],  count:  32
 [ 1] mean:  379 [378;381],  count:  56
Gap width distribution:
 [ 0] mean:  246 [245;249],  count:  87
Pulse period distribution:
 [ 0] mean:  381 [380;383],  count:  31
 [ 1] mean:  625 [624;627],  count:  56
Guessing modulation: Pulse Width Modulation with fixed gap
Attempting demodulation... short_limit: 256, long_limit: 250, reset_limit: 250, demod_arg: 0
pulse_demod_pwm(): Analyzer Device 
bitbuffer:: Number of rows: 1 
```

Data:

```
[00] {88} ff a8 51 f2 52 00 01 04 5d 06 0e 
[00] {88} ff a8 51 f2 52 00 01 04 5d 02 ca 
[00] {88} ff a8 51 f2 52 00 01 04 5d 00 a8 
[00] {88} ff a8 51 f2 52 01 02 04 5d 02 85 
[00] {88} ff a8 51 f2 52 00 00 04 5d 00 33
```

Current conditions: 9.6C, 83% hum, 0 wind, 0 rain.

viraptor:
It is the weather station - I just noticed the panel that came with it blinks the receiving icon exactly when rtl_433 reports new packet.

viraptor:
And some early data analysis, all big endian:
```
[00] {88} ff a8 XX XX YY ZZ 01 04 5d UU UU
```
`XX XX`: temperature, likely in 0.1⁰C steps (`51 e7` == 8.7⁰C, `51 ef` == 9.5⁰C)
`YY`: percent in a single byte (for example `54` == 84%)
`ZZ`: wind speed (`00` == 0, `01` == 1.1km/s, ...)
`UU UU`: wind direction: `00 00` is N, `02..` is NE, `04..` is E, etc. up to `0f ff` is seems (highest I got was `0e 08` for NW)

viraptor:
[...]

PR #167, opened 2015-09-05
Add support for digitech weather station

viraptor:
Currectly works: wind, temperature, humidity
To do: rain, pressure, checksum validation

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -34,7 +34,8 @@
 		DECL(valeo) \
 		DECL(chuango) \
 		DECL(generic_remote) \
-		DECL(tfa_twin_plus_303049)
+		DECL(tfa_twin_plus_303049) \
+		DECL(digitech_ws)
 
 
 typedef struct {
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -51,6 +51,7 @@ add_executable(rtl_433
 	devices/chuango.c
 	devices/generic_remote.c
 	devices/tfa_twin_plus_30.3049.c
+	devices/xc0348.c
 )
 
 target_link_libraries(rtl_433
--- a/src/devices/xc0348.c
+++ b/src/devices/xc0348.c
@@ -1,0 +1,83 @@
+/*
+ * Digitech XC0348 weather station
+ * Reports 1 row, 88 pulses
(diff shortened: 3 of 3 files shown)


merbanan:
Can you send some test signals to rtl_433_tests ?

viraptor:
Preparing them right now

merbanan:
The code looks good. Can you also log x amount of packets so we can take a stab at figuring out the checksum ?

viraptor:
Do you need more than what's in https://github.com/merbanan/rtl_433/issues/166 ? (5 packets)

viraptor:
Also, what can I do about samples which require specific detection level? Everything I have passes only on level 1000..1800.

merbanan:
Can't have too many. Some patterns can only be figured out if you have enough samples. I would say 30 should be enough.

merbanan:
The -l limit parameter is just lowering the bit detection threshold. If you need to lower the limit the signal is not so strong. So either keep the limit low of increase the signal somehow. Better antenna, better placement etc.

viraptor:
I meant for the tests - right now tests are run with the default limit, so my samples won't pass. But I can't get anything better out of my setup unfortunately :(

viraptor:
Not sure if you like the solution, but added samples in https://github.com/merbanan/rtl_433_tests/pull/36

rct:
In terms of the level limit, the greater the distance between the dongle's center frequency and the actual device frequency the lower the signal will be.  you might want to try figuring out what your dongle's frequency error is (PPM) and then specify a --p ppm number.  I have some dongles where the error is as much as 70 ppm - so without correction it is off by 30 Khz.   Your device will likely have some offset too.  Note: you should have some offset to make sure your device isn't too close to the center DC spike.

Of course the things merbanan mentioned, antenna, placement (get the dongle away from the computer) will also play a factor.

Once you have that all figured out, you might want to revisit the limits to see if you have the right timings at reasonable signal levels.

viraptor:
I've already done that. The -f for capture was set exactly to the frequency I got from gqrx's waterfall graph. That's also after ppm correction.

rct:
you need at least a little offset from the center spike.

merbanan:
Forgot to merge, sorry.

merbanan:
Thanks for making rtl_433 better.

PR #170, opened 2015-09-18
generic remote sc/pt226x ev1527: increase the tolerance to 50

nebman:
very short follow-up to pull #156 three weeks ago: I increased the tolerance to 50, now it works with all the devices we can test so far (my SC2260 remote, merbanan's PT2262 sensor, plus I got a new EV1527 remote that also works fine with this setting).

sorry this is a bit late - but here it is anyway ;)

the code did not produce any false-positives with the samples repo.

### DIFF
--- a/src/devices/generic_remote.c
+++ b/src/devices/generic_remote.c
@@ -5,6 +5,7 @@
  * EV1527
  *
  * Copyright (C) 2015 Tommy Vestermark
+ * Copyright (C) 2015 nebman
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation; either version 2 of the License, or
@@ -66,7 +67,7 @@ static int generic_remote_callback(bitbuffer_t *bitbuffer) {
 
 
 PWM_Precise_Parameters pwm_precise_parameters_generic = {
-	.pulse_tolerance	= 20,
+	.pulse_tolerance	= 50,
 	.pulse_sync_width	= 0,	// No sync bit used
 };
 


merbanan:
Thank you very much mr nebman.

merbanan:
I am waiting for batteries for some sensors, after that I will add some more signals based on SCXXXX and PTXXXX.

nebman:
Thanks, more samples will allow to further improve the code! Unfortunately the ID on my new EV1527 remote is fixed, so for privacy reasons I cant share a sample of it, otherwise I would have already commited one ;)
I'll get more sensors (PIR, door, etc) of various ICs soon, so lets see :)

jcarduino:
Can you also have a look at the cresta sample?

And would love to have some comments about how you change the demodulator!

I have some signals using lower frequencies eg 433.4 blinds from uk company. Is it that much work for changing demod stuff to have more bandwidth? Would raspi still work? Cpu usage has been down lately so coding has improved!!!

Mvg Johan Corstjens

Verstuurd via mobile

merbanan:
The way it works now is that signals should alias in to the detection range as long as they are close enough. I tested this now and I do see signals when tuned to 433.4MHz. I don't know if the signals demodulates well, but I do see them.

Regarding increasing the bandwidth, I initially built support for that. To reintroduce that we need to change the limit values in the protocols to mean microseconds instead of samples. When that is done we can just scale depending on the sample rate. Regarding cpu usage the max the rtl sdr can deliver is 2.56Ms. That is 10 times more data then what is used now thus it might take 10 times more cpu. So I think it would still work on a rpi.

PR #176, opened 2015-10-06
New Lacrosse Weather Station Decoding

ygator:
Tested with my WS-2310TWC weather station.

### DIFF
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -30,7 +30,7 @@
 #define DEFAULT_LEVEL_LIMIT     8000		// Theoretical high level at I/Q saturation is 128x128 = 16384 (above is ripple)
 #define MINIMAL_BUF_LENGTH      512
 #define MAXIMAL_BUF_LENGTH      (256 * 16384)
-#define MAX_PROTOCOLS           33
+#define MAX_PROTOCOLS           34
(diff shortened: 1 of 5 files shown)


merbanan:
Only thing missing are test recordings for the rtl_433_tests repository and a copyright notice. Can you add that ?

rct:
This looks good.  However, you've changed the device numbering for  people who are using devices with numbers higher than 9.  Anyone using -R will have to change the command lines they use.  I can see that grouping the LaCrosse devices together in the list also makes sense, but I think for compatibility new devices should be added to the end.

To make the change in include/rtl_433_devices.h, move DECL(lacrossews) to the end, which will make it device number 34, instead of 9.

Please submit a set of test files to rtl_433_tests.  The messages for the 433 Mhz TX temperature/humidity and the 2310 weather station are pretty close.  I have to look in more detail, but  I'm tempted to merge them into a single device file. The message content is obviously somewhat different, but the timing, checksum, etc. are all pretty close.

I'd like to get my hands on one of the 433 mhz weather station thermo-hygro units for testing, but it doesn't look like they make them any longer.  I have the 915 Mhz version but it uses an entirely different encoding.

ygator:
I originally started putting it in lacrosse.c, but decided to make it separate as to not step on what you are doing.  It is basically lacrosse.c with different decoding.  Feel free to combine if you want.

Issue #149, opened 2015-08-21
implementing Esperanza EWS


Alberts00:
Hello! 
I would like to thank all developers who have made this tool. Thank you!

A few days ago I remembered that I have a temperature sensor which runs on 433Mhz. I have successfully found the most relevant pieces of information which were given to me using the analyse output (-a flag)
```
*** signal_start = 800122, signal_end = 1062693
signal_len = 262571,  pulses = 266
Iteration 1. t: 152    min: 151 (23)    max: 153 (243)    delta 9
Iteration 2. t: 152    min: 151 (2)    max: 153 (264)    delta 0
Distance coding: Pulse length 152

Short distance: 485, long distance: 962, packet distance: 1922

p_limit: 152
bitbuffer:: Number of rows: 14 
[00] {0} : 
[01] {0} : 
[02] {42} 0b 9b b5 a4 86 00 : 00001011 10011011 10110101 10100100 10000110 00
[03] {0} : 
[04] {42} 0b 9b b5 a4 86 00 : 00001011 10011011 10110101 10100100 10000110 00
[05] {0} : 
[06] {42} 0b 9b b5 a4 86 00 : 00001011 10011011 10110101 10100100 10000110 00
[07] {0} : 
[08] {42} 0b 9b b5 a4 86 00 : 00001011 10011011 10110101 10100100 10000110 00
[09] {0} : 
[10] {42} 0b 9b b5 a4 86 00 : 00001011 10011011 10110101 10100100 10000110 00
[11] {0} : 
[12] {42} 0b 9b b5 a4 86 00 : 00001011 10011011 10110101 10100100 10000110 00
[13] {0} :
```

After that I found where and how the humidity, channel and temperature is stored.
These are the settings I'm currently using for my sensor when implementing callback.

```
r_device esperanza_ews = {
  .name          = "Esperanza EWS-103",
  .modulation    = OOK_PULSE_PPM_RAW,
  .short_limit   = 152,
  .long_limit    = 2500,
  .reset_limit   = 5000,
  .json_callback = &esperanza_ews_callback,
  .disabled      = 0,
  .demod_arg     = 0,
};
```

There are a few issues here:
1) The callback doesn't always get called, even though debug does show the information.
Here's an example

```
Pulse width distribution:
 [ 0] mean:  152 [150;155],	 count: 266
Gap width distribution:
 [ 0] mean: 1923 [1918;1939],	 count:  13
 [ 1] mean:  486 [479;493],	 count: 126
[...]

PR #178, opened 2015-10-06
Adding Esperanza EWS

Alberts00:
I've figured out the values for limits and successfully implemented the Esperanza EWS temperature/humidity sensor.

I'm not an C expert therefore I will be very thankful if you could point out any mistakes I've made in the commit. 
If there are no issues issues with the code I hope you don't mind if I mark #149 as solved. :)


merbanan:
Looks ok, and yes you can close the issue.

PR #183, opened 2015-10-08
DSC Security Contact added CRC-8 calc/check

rct:
* Added "little-endian" CRC-8 calculation
* Moved reverse8() to util.c
* Quieted WT450 callback false positive messages


merbanan:
Thank you

PR #185, opened 2015-10-09
Added quiet mode (-q) to suppress non-data related messages

rct:
-q will suppress many messages that don't contain any useful data. 
* There is still useful info that goes to STDERR, so just sending STDERR to /dev/null isn't a solution.
* Quiet mode is most helpful when using -r gfileNNN.data for testing. 

Note: the registering protocol messages should probably be made a debug message at this point.


merbanan:
Stray tab but ok. Thanks.

PR #187, opened 2015-10-10
Single line output for Generic and Waveman

ygator:


merbanan:
The idea is to change all protocol decoders to output yaml objects. These objects can then be transformed via a command line parameter to the desired output format (xml, json, yaml, cvs, etc). So I wont accept any patch that switches the output format from the initial implementation. So if you want this feature you have to either implement a yaml parser or just keep the change locally until someone implements it.

ygator:
It does not switch anything.  I just took out the new lines so that the sensor output transmission is on a single line consistent with other protocols that I am using.  The output of a PIR sensor I have is below, without the change the output would have consisted of 30 lines instead of 7 as shown making it more readable.

Generic remote keypress / sensor ID 16bit = 0x5440 CMD 8bit = 0x55 TRISTATE = FFF0F000FFFF
Generic remote keypress / sensor ID 16bit = 0x5440 CMD 8bit = 0x55 TRISTATE = FFF0F000FFFF
Generic remote keypress / sensor ID 16bit = 0x5440 CMD 8bit = 0x55 TRISTATE = FFF0F000FFFF
Remote button event: model   = Waveman Switch Transmitter id      = H channel = 1 button  = 2 state   = off
Generic remote keypress / sensor ID 16bit = 0x5440 CMD 8bit = 0x55 TRISTATE = FFF0F000FFFF
Generic remote keypress / sensor ID 16bit = 0x5440 CMD 8bit = 0x55 TRISTATE = FFF0F000FFFF
Generic remote keypress / sensor ID 16bit = 0x5440 CMD 8bit = 0x55 TRISTATE = FFF0F000FFFF

merbanan:
Right now there is random output i each decoder. The way forward is to pick some markup style and use it for all decoders. Accepting output changes to accommodate random needs is not maintainable in the long run.

PR #188, opened 2015-10-14
Added missing wt450.c for autoconf compile

smallsam:
The recent addition of wt450 support was missing w450.c for autoconf based compile


merbanan:
Thank you.

PR #190, opened 2015-10-16
Include the last sample when saving samples between filter runs

robho:
I'm not sure about this one, but since the lowpass calculations use the previous input/output in the calculations I think these values should be saved for the next pass.

(currently the second last input/output is saved (ie an off-by-one error))


merbanan:
Looks correct.

PR #192, opened 2015-10-17
Share memory between fm buffer and temporary buffer

robho:
The temporary buffer and the fm buffer are not used at the same
time, so the allocated memory can be shared between the two buffers.

This reduces memory usage by 8 MB

---

This may not be acceptable since it's a pretty ugly hack, but I thought I'd post it anyway since it saves some memory.

(I've looked into removing the temporary buffer, but I need to do some more investigations)


merbanan:
Little bit hackish but I'll allow it. I think that the whole buffer handling can be reworked to avoid this. But as that wont happen in the foreseeable future this will do fine.

PR #196, opened 2015-11-02
WT450: Fix parity computation, error messages only in debug mode

foldynl:
Hello,

this patch fixes a parity computation for WT450 device. The error messages are shown only in debug mode for WT450.


merbanan:
Looks ok, thanks.

PR #197, opened 2015-11-02
Wip

AlexandreCo:
Hello,

this patch add inovalley kw9015b rain and temperature  wheater sensor and an other rf433 noname wheater station

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -56,6 +56,7 @@ add_executable(rtl_433
 	devices/lacrossews.c
 	devices/esperanza_ews.c
 	devices/efergy_e2_classic.c
+	devices/inovalley-kw9015b.c
 )
 
 target_link_libraries(rtl_433
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -57,6 +57,7 @@ add_executable(rtl_433
 	devices/esperanza_ews.c
 	devices/efergy_e2_classic.c
 	devices/inovalley-kw9015b.c
+	devices/rfwireless.c
 )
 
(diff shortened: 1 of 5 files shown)


src/CMakeLists.txt
```diff
@@ -56,6 +56,8 @@ add_executable(rtl_433
 	devices/lacrossews.c
 	devices/esperanza_ews.c
 	devices/efergy_e2_classic.c
+	devices/inovalley-kw9015b.c
```
robho: You should update src/Makefile.am also

merbanan:
Can you submit sample data to the rtl_433_tests repository ? And fix the Makefile.am code ?

rct:
Ugh.  Maybe it's just me.  Naming is very subjective, but  rfwireless as a device / file name?  

I understand it is a "no name" weather station, but the source file name "rfwireless" seems unnecessarily vague and confusing.  It looks like it is temperature only, not a full weather station or only temperature messages are being decoded for now.

How about "generic temperature sensor 1", or "generic weather station 1"?

This device might generate a number of false positives for anything that will put a 3 byte row into the bit buffer and has timing anywhere near close.

It would be helpful if:
* The source / callback were renamed.
* The source file contained a little more of a comment than "Unknown Temperature weather station".

Also for the other device, inovalley-kw9015b.c, change r_device->name from ""KW9015B Transmitter" to "Inovalley kw9015b rain and Temperature weather station" since that's what gets printed in the usage info.

rct:
(btw, my comment is in the context of rtl_433 now decoding 38 wireless devices, so things need to be a bit clearer if possible.)

PR #198, opened 2015-11-05
fix Makefile.am and rename rfwireless  device and file name

AlexandreCo:
Hello,
those patch fix makefile.am and rename rfwireless into generic_temperature_sensor as rct advised

Thanks


merbanan:
Thanks

Issue #199, opened 2015-11-07
Add support for Acurite 592TXR "Tower" Temperature and Humidity Sensor.


rct:
Alternate Model Numbers: 00592TXR, 11126-592TXR, 06002RM

http://www.acurite.com/temperature-and-humidity-sensor-06002rm-592txr.html

PR #200, opened 2015-11-07
Add support for Acurite 592TXR "Tower" Temperature/Humidity Sensor.

rct:
Implements #199.

- Adds support for 592TXR "Tower" temperature/humidity sensor
- Alternate 5n1 weather station shared device decode with 592TXR
 - Wind direction decoding matches Aculink internet bridge and Acurite 01512 console
 - More reliable decoding/detection with PWM_TERNARY demodulator  (over a wider range of level limits)
 - Timestamps on 5n1 output
 - Original 5n1 (-R 9) implementation left unchanged for compatibility

The protocol and signal of the 592TXR and 5n1 weather station are very similar.  The 5n1 weather station is one byte longer.  The message type is in the same location.  The ID bits are interpreted slightly differently.

The original 5n1 wind direction decoding doesn't match my results when compared to the Acurite 01512 console and Aculink internet bridge.  The original decoding is a straight linear/circular mapping.  However some (or all) of the 5n1's seem to use an encoding that jumps around and is probably a modified Gray Code. 

Due to the differences I've left the original 5n1 decoding unchanged as this is working for some people Jens Jensen and @helgew.  David Ray Thompson's @draythomp Desert-home-rtl_433 fork uses the "modified Gray Code" mapping.  

```
// Mapping 5n1 raw RF wind direction values to aculink's values
//    RF, AcuLink
//     0,  6,   NW,  315.0
//     1,  8,  WSW,  247.5
//     2,  2,  WNW,  292.5
//     3,  0,    W,  270.0
//     4,  4,  NNW,  337.5
//     5,  A,   SW,  225.0
//     6,  5,    N,    0.0
//     7,  E,  SSW,  202.5
//     8,  1,  ENE,   67.5
//     9,  F,   SE,  135.0
//     A,  9,    E,   90.0
//     B,  B,  ESE,  112.5
//     C,  3,   NE,   45.0
//     D,  D,  SSE,  157.0
//     E,  7,  NNE,   22.5
//     F,  C,    S,  180.0
```

Note: The current wind speed decoding for both is incorrect.  It assumes that the values are KmPH.  At low wind speeds it's a close match, however they diverge as the wind speed increases.  I still need to look into this a bit more.  @aweatherguy suggested a formula in another issue that I need to try.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,9 @@ Supported devices:
 	[34] LaCrosse WS-2310 Weather Station
 	[35] Esperanza EWS
 	[36] Efergy e2 classic
+	[37] Inovalley kw9015b rain and Temperature weather station
+	[38] Generic temperature sensor 1
+	[39] Acurite 592TXR Temperature/Humidity Sensor and 5n1 Weather Station
 ```
 
 
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -30,7 +30,7 @@
 #define DEFAULT_LEVEL_LIMIT     8000		// Theoretical high level at I/Q saturation is 128x128 = 16384 (above is ripple)
 #define MINIMAL_BUF_LENGTH      512
 #define MAXIMAL_BUF_LENGTH      (256 * 16384)
-#define MAX_PROTOCOLS           38
+#define MAX_PROTOCOLS           39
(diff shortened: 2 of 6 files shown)


merbanan:
Hi, I see a few cases of printf change those to fprintf and if needed change makefile.am also.

Issue #71, opened 2014-12-15
Log records


deennoo:
Hello,

is this possible to had the possibility to records logs ?

On a CSV format, this way we can use this file as source for any app we want.

deennoo:
i send you a mail, did you recieve it ?

Serveurperso:
You can disable the buffering for stdin and stderr, to get grep/awk working

```
(root|~/rtl_433/src) git diff
diff --git a/src/rtl_433.c b/src/rtl_433.c
index fc85f0d..48406ef 100644
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -987,6 +987,9 @@ int main(int argc, char **argv) {
     char vendor[256], product[256], serial[256];
     int have_opt_R = 0;

+    setbuf(stdout, NULL);
+    setbuf(stderr, NULL);
+
     demod = malloc(sizeof (struct dm_state));
     memset(demod, 0, sizeof (struct dm_state));
```

Serveurperso:
```
#!/bin/bash

/usr/local/bin/rtl_433 \
| tee >(stdbuf -o0 fgrep 'Weather Sensor THGR810  Channel 1' > TH1.txt) \
| tee >(stdbuf -o0 fgrep 'Weather Sensor THGR810  Channel 2' > TH2.txt) \
| tee >(stdbuf -o0 fgrep 'Weather Sensor THGR810  Channel 3' > TH3.txt) \
| tee >(stdbuf -o0 fgrep 'Weather Sensor UVN800 Channel 1' > UV.txt) \
| tee >(stdbuf -o0 fgrep 'Weather Sensor WGR800' > WG.txt)
```

(root|/var/www/meteo) cat TH1.txt
Weather Sensor THGR810  Channel 1 Temp: 9.1C  48.4F   Humidity: 90%
Weather Sensor THGR810  Channel 1 Temp: 9.1C  48.4F   Humidity: 90%
Weather Sensor THGR810  Channel 1 Temp: 9.1C  48.4F   Humidity: 90%

merbanan:
CSV support is added now which should take care of this issue.

PR #201, opened 2015-11-18
Structural data & JSON support

eras:
Hi,

This PR provides the ability to construct structural data and a way to dump it in JSON. It implements the particular support only for wt450.c for now, but others should be easy to merge if the first of these two commits gets in. Incorporating the patch for wt450 of course breaks scripts that depend on the output, but this would hopefully happen only once.

While this addresses the problems of issue #71 (modules just use fprintf), it doesn't provide CSV output because the schema of the data is not known up-front. It seems most if not all of the modules use a fixed format, so it might be a possible continuation idea for them to declare the fields they are going to use before the first data arrives. This would allow the data_acquired_handler-function to construct CSV from the same data.

This patch set may not be the most memory-efficient way to construct structured data, but it isn't perhaps that a horrible way either, given the low rate of data being transmitted. The interface is convenient:

```
data_t *data = data_make("time"       , DATA_TYPE_STRING, time_str,
                         "house code" , DATA_TYPE_INT, house_code,
                         "channel"    , DATA_TYPE_INT, channel,
                         "battery"    , DATA_TYPE_STRING, battery_low ? "LOW" : "OK",
                         "temp"       , DATA_TYPE_DOUBLE, temp,
                         "humidity"   , DATA_TYPE_INT, humidity,
                         NULL));
data_print_json(data, stdout);
data_free(data);
```

It supports not only strings and ints and stuff but also arrays and itself, so it should be suitable for generating arbitrary JSON. Perhaps in the future someone wants to write an XML or binary output module and this is possible as well ;).

It also comes with a test to exert the basic functions. Certainly not with great coverage. The code tries to deal with out-of-memory situations as well as rare they might be, but those paths are completely untested.

Keep up the good work!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,7 @@ add_custom_target(uninstall
 ########################################################################
 add_subdirectory(include)
 add_subdirectory(src)
+add_subdirectory(tests)
 
 # use space-separation format for the pc file
 STRING(REPLACE ";" " " RTL433_PC_CFLAGS "${RTL433_PC_CFLAGS}")
--- a/include/data.h
+++ b/include/data.h
@@ -1,0 +1,98 @@
+/*
+ * A general structure for extracting hierarchical data from the devices;
+ * typically key-value pairs, but allows for more rich data as well.
+ *
+ * Copyright (C) 2015 by Erkki Seppälä <flux@modeemi.fi>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef INCLUDE_DATA_H_
+#define INCLUDE_DATA_H_
+
+typedef enum {
+	DATA_DATA,		/* pointer to data is stored */
+	DATA_INT,		/* pointer to integer is stored */
+	DATA_DOUBLE,		/* pointer to a double is stored */
+	DATA_STRING,		/* pointer to a string is stored */
+	DATA_ARRAY,		/* pointer to an array of values is stored */
+	DATA_COUNT		/* invalid */
+} data_type_t;
+
+typedef struct data_array {
+	int	     num_values;
+	data_type_t  type;
+	void	    *values;
+} data_array_t;
+
+typedef struct data {
+	char	    *key;
+	char	    *pretty_key; /* the name used for displaying data to user in with a nicer name */
+	data_type_t  type;
+	void	    *value;
+	struct data* next; /* chaining to the next element in the linked list; NULL indicates end-of-list */
+} data_t;
+
+/** Constructs a structured data object.
+
+    Example:
+    data_make("key", "Pretty key", DATA_INT, 42,
+	      "others", "More data", DATA_DATA, data_make("foo", DATA_DOUBLE, 42.0, NULL),
+	      "zoom", NULL, data_array(2, DATA_STRING, (char*[]){"hello", "World"}),
+	      "double", "Double", DATA_DOUBLE, 10.0/3,
+	      NULL);
+
+    Most of the time the function copies perhaps what you expect it to. Things
+    it copies:
+    - string contents for keys and values
+    - numerical arrays
+    - string arrays (copied deeply)
+
+    Things it moves:
+    - recursive data_t* and data_array_t* values
+
+    The rule is: if an object is boxed (look at the dmt structure in the data.c)
+    and it has a array_elementwise_import in the same structure, then it is
+    copied deeply. Otherwise, it is copied shallowly.
+
+    @param key Name of the first value to put in.
+    @param pretty_key Pretty name for the key. Use "" if to omit pretty label for this field completely,
+                      or NULL if to use key name for it.
+    @param type Type of the first value to put in.
+    @param ... The value of the first value to put in, follwed by the rest of the
+               key-type-values. The list is terminated with a NULL.
+
+    @return A constructed data_t* object or NULL if there was a memory allocation error.
+*/
+data_t *data_make(const char *key, const char *pretty_key, data_type_t type, ...);
+
+/** Constructs an array from given data of the given uniform type.
+
+    @param ptr The contents pointed by the argument are copied in.
+
+    @return The constructed data array object, typically placed inside a data_t or NULL
+            if there was a memory allocation error.
+*/
+data_array_t *data_array(int num_values, data_type_t type, void *ptr);
+
+/** Releases a data array */
+void data_array_free(data_array_t *array);
+
+/** Prints a structured data object as JSON to the given stream */
+void data_print_json(data_t *data, FILE* file);
+
+/** Releases a structure object */
+void data_free(data_t *data);
+
+#endif // INCLUDE_DATA_H_
--- a/include/data.h
+++ b/include/data.h
@@ -27,7 +27,8 @@ typedef enum {
 	DATA_DOUBLE,		/* pointer to a double is stored */
 	DATA_STRING,		/* pointer to a string is stored */
 	DATA_ARRAY,		/* pointer to an array of values is stored */
-	DATA_COUNT		/* invalid */
+	DATA_COUNT,		/* invalid */
+	DATA_FORMAT		/* indicates the following value is formatted */
 } data_type_t;
 
 typedef struct data_array {
@@ -40,6 +41,7 @@ typedef struct data {
 	char	    *key;
 	char	    *pretty_key; /* the name used for displaying data to user in with a nicer name */
 	data_type_t  type;
+	char        *format; /* if not null, contains special formatting string */
 	void	    *value;
 	struct data* next; /* chaining to the next element in the linked list; NULL indicates end-of-list */
 } data_t;
@@ -79,7 +81,7 @@ extern struct data_printer data_kv_printer;
 
     @return A constructed data_t* object or NULL if there was a memory allocation error.
 */
-data_t *data_make(const char *key, const char *pretty_key, data_type_t type, ...);
+data_t *data_make(const char *key, const char *pretty_key, ...);
 
 /** Constructs an array from given data of the given uniform type.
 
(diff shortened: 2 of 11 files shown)


merbanan:
I sort of had preferred to use YAML for the decoders. But how does a decoded packet look like ?

eras:
A decoded packet might look like (as the second patch implements it for WT450):

    {"time" : "2015-11-20 08:12:32", "house_code" : 1, "channel" : 1, "battery" : "OK", "temp" : 22.200001, "humidity" : 23}

Though the exact structure is easily changed. One idea might be to put "meta data" into its own structure to be filled by rtl_433 core code instead of the device drivers themselves. Then the output might look something like:

    {"time" : "2015-11-20 08:12:32", sensor:"WR450", data: { "house_code" : 1, "channel" : 1, "battery" : "OK", "temp" : 22.200001, "humidity" : 23}}

Extending this to support also YAML (and for example XML) would be quite easy, as the construction of the data and converting it to final representation are separate. Basically it would be implementing function data_print_yaml similar to the function data_print_json. There would be some refactoring involved to reduce code duplication in that case.

In practice I do believe JSON is:
- sufficient: JSON is more than able to express the data received from the sensors
- more practical: each JSON blob can be its own line, so there is no need to have additional message separators if the data is going to be just written to stdout. Newlines will do. People can use rtl_433 | while read msg; .. to do the processing as they have probably done so far :).
- de facto standard: it's going to be easier to use because JSON is much more widely supported.

merbanan:
The object should be linear, no nested structures.
I don't mind if the data from the decoders are json.
But what I really want is human readable output by default. And the option to transform it to cvs/plaintext.

merbanan:
Hmm, why is the temp value 22.200001 ?

eras:
Because of formulas involving floating point numbers :).

I can add support for formatted doubles.. They would be nice for the human readable part.

I'll also add support for an output that resembles the previous key-value output, but as currently every module just has their own fprintfs, their output quite likely won't conform to any single format..

merbanan:
We need to formalize and document the object type. Add a type member (sensor, remote, etc). Change house_code to id. Time to time_stamp. If possible make the human readable output to YAML. With that added I will accept and merge the patch.

merbanan:
Can you also supply some test signals of this device to the rtl_433_tests repo ?

merbanan:
And we need to rework the time stamp feature. The time stamp should be generated when the first signal is taken not in the decoders.

enavarro222:
Hi,

I'm really interested in such a feature ! It will be much more easiest to pipe rtl_433 to an adhoc parsing script that push readings to a database.

For such usage, it is simpler to have only one line by sensor reading. In JSON it is possible. I don't think will be possible with YAML.

However a multi-lines YAML output may be nicer for a direct command line usage...

One can add an option to choose output format ?

I hope to see that soon in rtl_433 (by the way: thanks for this soft !)
I may help to change some decoder to this new output system.

merbanan:
@enavarro222 you could help by adding sensor recordings to the rtl_433_tests repository.

rct:
First, I think it would be helpful to move the discussion of reworking rtl_433's architecture to the rtl_433 google group.  https://groups.google.com/d/forum/rtl_433

I have a strong preference for single line output formats.  This lets me easily manipulate the output especially for diagnosis using tools such as grep to pull lines I need out of the log file.  I have currently have over 25 devices that are transmitting on 433 mhz - Lacrosse TX, DSC security contacts, and 3 types of Acurite devices.

What advantages does YAML have over JSON?  I could see that it would be more readable for a low volume of output.  However, if the RF environment is busy, the output will scroll off the screen too quickly and become a hindrance.

Finally, given the number of existing devices (now 40 with my last pull request) and the scripts/systems we all have for putting the output to use, I think it an output switch is required before converting existing devices.

rct:
FYI, the Desert Home fork of rtl_433, emits JSON for the Acurite 5n1 weather station:

{"sensorId":{"SID":"92","t":"1434292619"},"channel":{"CH":"A","t":"1434292619"},"messageCaught":{"MC":"0","t":"1434292619"},"battLevel":{"BAT":"7","t":"1434292619"},"windSpeed":{"WS":"3.5","t":"1434292619"},"windDirection":{"WD":"NNE","t":"1434292619"},"temperature":{"T":"89.8","t":"1434292601"},"humidity":{"H":"27","t":"1434292601"},"rainCounter":{"RC":"445","t":"1434292619"}}

The 5n1 weather station sends two different message types. so the above output has aggregated them into a single message, which is why each element is time stamped.  

Personally, I think the output should be more raw. I'd rather see the two individual messages that the device sends separately, rather than have rtl_433 keep state and do some processing that might obscure information that downstream processing could use.

eras:
I added human-readable key-value output, which looks like this (implemented for wt450 only so far):

	% rtl_433 -R 33 2>/dev/null
	2015-11-21 18:01:51 WT450 sensor: House Code 1, Channel 1, Battery OK, Temperature 23.20 C, Humidity 22 %
	2015-11-21 18:01:57 WT450 sensor: House Code 1, Channel 2, Battery OK, Temperature 5.25 C, Humidity 35 %
	2015-11-21 18:01:58 WT450 sensor: House Code 1, Channel 2, Battery OK, Temperature 5.28 C, Humidity 35 %

And the JSON output looks like this:

	% rtl_433 -R 33 -j 2>/dev/null
	{"time" : "2015-11-21 18:02:51", "label" : "WT450 sensor", "house_code" : 1, "channel" : 1, "battery" : "OK", "temp" : 23.400000, "humidity" : 22}
	{"time" : "2015-11-21 18:02:51", "label" : "WT450 sensor", "house_code" : 1, "channel" : 1, "battery" : "OK", "temp" : 23.420000, "humidity" : 22}
	{"time" : "2015-11-21 18:02:51", "label" : "WT450 sensor", "house_code" : 1, "channel" : 1, "battery" : "OK", "temp" : 23.450001, "humidity" : 22}

And finally the CSV output:

	% rtl_433 -R 33 -c 2>/dev/null
	time;label;house_code;channel;battery;temp;humidity
	2015-11-21 18:03:25;WT450 sensor;1;1;OK;22.639999;21
	2015-11-21 18:03:25;WT450 sensor;1;1;OK;22.660000;21
	2015-11-21 18:03:26;WT450 sensor;1;1;OK;22.690001;21

Why no truncated decimals? Because there is only one formatting function for 'human readable' output, and as it contains also the "C" suffix, I decided not to use the same formatter for the other cases. This is all a driver needs to do:

	data = data_make(
		"",           "time" ,       DATA_STRING, time_str,
		"",	      "label" ,      DATA_STRING, "WT450 sensor",
		"House Code", "house_code",  DATA_INT, house_code,
		"Channel",    "channel",     DATA_INT, channel,
		"Battery",    "battery",     DATA_STRING, battery_low ? "LOW" : "OK",
		"Temperature","temp",        DATA_FORMAT, "%.02f C", DATA_DOUBLE, temp,
		"Humidity",   "humidity",    DATA_FORMAT, "%u %%", DATA_INT, humidity,
		NULL);
	data_acquired_handler(data);

And also for the benefit of CSV support:

	static char *output_fields[] = { "time", "label", "house_code", "channel", "battery", "temp", "humidity", NULL };
	
	r_device wt450 = {
	..other fields..
	   .json_callback = &wt450_callback,
	   .fields        = output_fields
	};

Pretty bearable.

If there's something to add (ie. common fields, they can be done in the rtl_433.c) in the data_acquired_handler function. data doesn't yet support adding additional fields to it (so hierarchies would be eaiser ;)), but if some format is decided upon, that's probably easy to implement to apply for all protocols.

rct:
@eras - this all looks like a great start!

As I said previously,  I think it would be helpful to move the discussion to the rtl_433 google group. https://groups.google.com/d/forum/rtl_433.

A few comments:
- Labels such as "temp" are ambiguous, is it C of F?   Off the top of my head, I currently don't think we need a separate unit type for each field. I think the label should be sufficient, tempc vs. tempf, wind-mph, wind-mps, etc.
- The device's "native" units should be reported at a minimum. If a device reports in Fahrenheit, it shouldn't be converted to Celsius and back to Fahrenheit again to avoid loss of precision especially after rounding has occurred to make pretty printed floats. If it is decided that message formats should always include metric units, then the original units should be included if they weren't metric.
- The device manufacturer and model number should both be included, preferably as separate fields. Downstream parsers might want want to identify such as Acurite and Oregon Scientific.

Temperature and humidity sensors are pretty straightforward data wise. It would be helpful to look at a wider range of devices before baking the protocol too much. Even with those there is a bit of a range of identifiiers, serial number, channel, house code, etc.

jcarduino:
Is it an option to make a switch for metric so that switch indicated inch f c meters etc

Mvg Johan Corstjens

Verstuurd via mobile

merbanan:
@rct I think that decoders should pass down native data from the sensors. The receiving layer can normalize the data later if selected. So if there is temperature reading in Celsius a parameter with unit Celsius should be used. The normalizing system could then transform it to Kelvin later.

eras:
Btw, adding units could work like this:

	data = data_make(
	    "Temperature", "temperature", DATA_UNIT, "C", DATA_DOUBLE, 12.0.
	    NULL);

Some aspects of this approach:

- The JSON output is not quite self-evident. Some options:

        { "temperature_C" : 12.0 } <- my favorite I think
	{ "temperature" : { "value": 12.0, "unit" : "C" } }
	{ "values": [{name : "temperature", "value" : 12.0, "unit" : "C" }] }

- works with the current key-value-output format easily (just append the unit after the value)
- it would be immediately useful even without an elaborate subsystem to do conversions between units
- it's opt in: existing code could be converted as-is without unit-support
- the key names are ambigious in the sense that underscore can be a unit separator as well as a symbol in the original label

merbanan:
@eras CSV should be separated by comma.

Variable names with the included unit is preferred. It is not the job of the decoders to normalized data. They should just report what they decode. The JSON structure should be possible to modify by a normalizing step. So that Fahrenheit can be converted to Celsius if needed. This can wait until later, but we need the decoders to use different names for units so they can be normalized later. temperature_C would be fine in this case.

The human readable output should be placed on several lines for easy parsing by the eyes.

Regarding id and channel, those parameters need to be separate as channel is a fixed variable parameter. While id can be a random value that i generated when batteries are inserted. Maybe we need to expose it as random_id and fixed_id. The main idea behind all this is that an event parser system should be able to do a 1:1 mapping from generated events.

eras:
Rewrote the whole bunch for cleaner commit history ;).

- reordered the arguments so pretty_key can be omitted (so it defaults to the key then)
- renamed house_code->id, label->model, temp->temperature_C
- used , for CSV separator (but really, there's no standards and ; can be mighty useful if strings (names) contain comma themselves)
- changed the command line switches -j for json and -c for csv to -F kv|json|csv defaulting to kv
- added support for waveman as well. This changes how the default waveman output looks like, though, as it is completely different from what wt450 did..

merbanan:
Ok, I think I am happy enough with the patch. I will leave it for some time before I merge it.

@rct any comments ? I think that we have almost we need. The details can be ironed out later. I think this is good step forward.

eras:
Oops, I had somehow broken the waveman output slightly while adjusting the kv output for wr450. For the record, it now looks like this (line-oriented or YAML output is still unimplemented):

    model Waveman Switch Transmitter, id F, channel 1, button 4, state off

merbanan:
Ok, this looks good. Thanks for the patch.

Issue #202, opened 2015-11-21
Add support for Acurite 986 Refrigerator/Freezer Thermometer


rct:
Manufacturer link: http://www.acurite.com/kitchen-gadgets/digital-freezer-refrigerator-thermometer-with-temperature-alarms-00986.html

Two temperature sensor transmitters includes a receiver display.

rct:
PR #203 implements this, however, the existing PPM decoder looses the last bit in the CRC byte.  The data is sent twice.  If the CRC is even both copies will decode correctly.  If odd, one copy will fail.

This could be improved by implementing PPM demodulators that are similar to the newer PWM demods, ternary and precise.   There are some short and long start/sync pulses that should ideally not be included in the data (should be in separate rows in the bitbuffer.)

This is working well enough for now. If I get time I'll look at adding the new PPM demods.

Cabalist:
I realize this issue is very old but I was curious if the new PPM demods were implemented?

zuckschwerdt:
PWM has be updated and PPM should be next. I'll look into it.

Cabalist:
Thanks!  Do you have a donate button or anything I can send a beer to?

zuckschwerdt:
Thanks for the kind offer. I guess it wouldn't be fair to all the other contributors.
But perhaps we'll do a fundrasing for some bigger feature some day.
Cheers!

merbanan:
Well it would be nice to gather a war chest for bigger feature requests.

Cabalist:
Let me know!  This is neat work and I’d like to show some appreciation. :)

rct:
I can't believe it's been over two years.  I never got back to this because the Acurite 986 has been working well enough for me as is even with the occasional lost bit.  I think at the time vestom was doing some work on the demodulators.  I've been meaning to convert it to the structured output (JSON, etc.).

trlafleur:
I would also like to see this fix, but its not high priority...

trlafleur:
I have done some work on decode this sensor....

This is the data that you will see in the capture below. (Note: bits are LSB first)
0Xd2, 0X96, 0X1e, 0X00, 0Xb4  -->    11010010, 10010110, 11110, 0, 10110100
Acurite 986 sensor: 0X6978 - 1R: 75 F

1) Sometime it sends one block of data, sometime two
2) 1st sync pulse is ~1700us most of the time, but it stretches to ~2300us on occasion.
[...]

PR #203, opened 2015-11-21
Add Acurite 986 Refrigerator/Freezer Thermometer.

rct:
Implements #202.

Decodes with existing PPM demodulator, which looses 1 bit on some messages.


merbanan:
Thanks

PR #209, opened 2015-11-24
ambient: Migrate to using common method for output

punitagrawal:
Migrate the Ambient sensor driver to use common methods to output
sensor values.


merbanan:
Looks good.

PR #210, opened 2015-11-25
Migrate "prologue" to new output handler

enavarro222:
Tested with data recently pushed to rtl_433_test (https://github.com/merbanan/rtl_433_tests/tree/master/tests/prologue/02)

Maybe an "unification" of id/rid naming over different sensors should be done ?

By the way, there is no checksum with this protocol. However one may test if the 6 transmitted packets are equals to better validate reception. Do you think it's a good idea ?

### DIFF
--- a/src/devices/prologue.c
+++ b/src/devices/prologue.c
@@ -6,7 +6,7 @@
  * the data is grouped in 9 nibles
  * [id0] [rid0] [rid1] [data0] [temp0] [temp1] [temp2] [humi0] [humi1]
  *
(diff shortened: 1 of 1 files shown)


merbanan:
Patch looks good, thanks. Testing several packets is a good idea also, but you must do it in a smart way. Maybe check a packets (rows) and use the one with most matches.

merbanan:
Naming of parameters need to be formalized also. Feel free to start such a document.

PR #213, opened 2015-11-29
Prevent data files from being overwritten

rct:
Got tired of occasionally overwriting data files.


merbanan:
Can you add a command line parameter that will overwrite the files. By default it should not overwrite the files.

merbanan:
The code is fine otherwise.

rct:
Ok, if that's what you want. I didn't think it really needed yet another command line option as getting rid of the files is easy.

(Personally I'd just use   `rm -rf gfile*data && rtl_433 -a -t` if I wanted to overwrite the files with each repetitive run for some reason.)

One of these days we'll need to switch to getopt_long.

merbanan:
Can you split this pull request so I can merge the documentation changes ?

rct:
Added -W to overwrite files.
* rebased to current master.

PR #214, opened 2015-12-01
waveman: added missing comma to fix CSV output

eras:
The missing comma at the end of the line concatenates these two strings.


merbanan:
thanks

PR #216, opened 2015-12-01
Convert users of OOK_PWM_P to pulse demodulator and remove pwm_p_decode()

vestom:
Tested with test data


merbanan:
Yey, thanks.

PR #217, opened 2015-12-02
Update 2 drivers to new output format

tazounet:
Update 2 drivers to new output format and correct GNU style (-> C99)


merbanan:
Thanks.

Issue #218, opened 2015-12-04
Watchman Sonic / Apollo Ultrasonic oil monitor


dwmw2:
These two are basically the same device. The receiver uses a Si4320, the same as the popular RFM01/RFM11/RFM12/etc FSK transceivers. The framing, FSK and Manchester encoding are all handled by the chip and so probably the same as plenty of other devices.

I've got a proof-of-concept hack which takes output from rtl_fm and decodes the packets: http://david.woodhou.se/watchman-demod.c

Obviously having got this far, I need to replace my horrid hack with some proper signal processing, and adding it as a supported device to rtl_433 seems like the right approach. I could do with some guidance with that task though, please.

I have two devices to test with. One appears to transmit at 433906MHz ± 34kHz, while the other is 433896MHz ± 29kHz. Looking at the Si4320/RFC01 datasheets, it seems the bandwidth is supposed to be 67kHz.

Here's the waveform I get from rtl_fm: http://david.woodhou.se/watchman.png

There are three high periods, followed by three low periods. After that, a 0 bit is indicated by high-then-low, while 1 is indicated by low-then-high. After the 64th bit, there are two more periods remaining at the same level (so remaining low after a final 0 bit, or remaining high after a final 1 bit).

The period is about 1ms, and the complete transmission 136ms (6ms preamble, 128ms data, 2ms postamble).

This results in 8 bytes of data. The final byte is a 1Wire CRC8, and the actual sensor data (a depth reading from the ultrasonic data) is in the byte before that.

cf. http://openenergymonitor.org/emon/node/10107

So, where do I start? :)

dwmw2:
Test capture from rtl_433 uploaded to http://david.woodhou.se/rtl433_testcap_watchman.cap

merbanan:
Nice domain :) Anyway pretty please document the device(s), take some signal samples (if they actually are useful, in your case we might need a higher sample rate) and send a pull request to https://github.com/merbanan/rtl_433_tests.

Read the rtl_433 mailing list the thread "Lutron Motion Detection". Hopefully everything will be clear after that. Most likely the actual fsk demodulation should be supported and that you only need to write the actual packet parsing code. But start with this and raise issues on the bug tracker or the mailing list.

dwmw2:
For a packet which ends with a 1 bit, I end up with this:

```
[...]

PR #221, opened 2015-12-07
Add support for RFMxx/Si4320 FSK encoding and Watchman Sonic oil tank monitor

dwmw2:
Fixes Issue #218

The encoding (and the RFMxx/SI chip) is also used by a number of other devices, such as the WH1080 weather system.


merbanan:
Thank you.

PR #222, opened 2015-12-07
hideki: Remove overzealous warning about invalid parity

dwmw2:
It's warning about this even for other kinds of packets.


enavarro222:
Thanks ! sorry for that, I forget to remove it before to push... (more automatic test needed ;) )

Issue #224, opened 2015-12-09
CurrentCost current sensor & generic RFM01 decoding


enavarro222:
The next not-yet-decoded-by-rtl_433 device at home is a CurrentCost (CC) current sensor (http://www.currentcost.com/product-envi.html). (Next and last for now, will be the door bell).

I push to rtl_433_test some samples + a audacity project with data analysed with tag.
https://github.com/merbanan/rtl_433_tests/pull/53

I push a quick and dirty (but mostly working) decoding in my fork:
https://github.com/enavarro222/rtl_433/tree/currentcost

I do not propose a PR yet because, for now, I just savagely patch `pulse_demod_manchester_framed` proposed by @dwmw2. The thing is CC seems to use RFM01... However the msg's sync&init seems somehow different. My "patch" no more work for "Oil Watchman" sensors.

So the ideal PR will provide a decoding that work for both devices.

I didn't dig further for now. But maybe there is someone here with a stronger knowledge of RFM01 who can help !

I may tear-down my sensor to take pics and check it is actually a RFM01.

dwmw2:
Hm, I'm looking back at the RFM01 documentation and I see this:

> BR is the bit rate set in the receiver and ΔBR is bit rate difference between the transmitter and the receiver. N is the maximal number of consecutive ones or zeros in the data stream. It is recommended for long data packets to include enough 1/0 and 0/1 transitions, and be careful to use the same division ratio in the receiver and in the transmitter

This implies that perhaps the framing and the Manchester encoding *aren't* done by the hardware, and it's actually part of the higher-level protocol. Perhaps we should parameterise the decoder in a fashion similar to `pulse_demod_pwm_precise()` so that we can pass in the details of the pre/postamble we expect on different devices?

merbanan:
Sounds like a good idea. I remember writing something in the previous ticket how it could be done.

dwmw2:
Hm, this is not as simple as a different length of preamble and postamble. They all seem to start like this...
![currentcost](https://cloud.githubusercontent.com/assets/911239/11685212/40df20e8-9e6f-11e5-9b18-580e03c88417.png)
In rtl_433's analysis, the low frequency is 'pulse' and the high is 'gap':
```
[  1] Pulse: 8279, Gap:  126, Period: 8405
[  2] Pulse:  126, Gap:  125, Period:  251
[  3] Pulse:  124, Gap:  126, Period:  250
[  4] Pulse:  125, Gap:  125, Period:  250
[...]

PR #226, opened 2015-12-10
Handle Manchester decoding in bitbuffer domain

dwmw2:
As discussed in issue #224, this kills the `pulse_demod_manchester_framed()` function which turns out to be an application-specific special case, and not as generic as originally thought.

New `bitbuffer_search()` and `bitbuffer_manchester_decode()` functions, based on the code written by Emmanuel, are added. The oil watchman code is converted to use those, instead.


merbanan:
Looks good.

PR #237, opened 2015-12-14
Update CMakeLists.txt

ua3nbw:
Add support for sensors RST 025100, RST 02500


merbanan:
Resend as one pull request.

PR #238, opened 2015-12-14
Consistent ouput for reading sample files

dwmw2:
When comparing results on the test packet corpus I was offended by the number of false mismatches I got on the current time. This makes it print the offset within the packet when reading from the file, rather than the time of day.

It's not perfect, because in most cases there's only one block of data within the file, so we always print @0.00000s. Perhaps we could actually report the offset within that block at which the pulse stream was found. But this achieves my real objective (consistent output when reading sample files) well enough for now.


merbanan:
Thanks.

PR #240, opened 2015-12-15
Add temperature output in °C to Watchman decode

dwmw2:
We're fairly convinced we have a reasonable estimate of the what the
temperature values mean.

See https://hackaday.io/project/1548-rocket-receiver-teardown for all
the gory details.

Keep outputting the raw value too, just in case we change it later; a
consumer of this data which thinks it knows better can still reconvert
for itself.

### DIFF
--- a/src/devices/oil_watchman.c
+++ b/src/devices/oil_watchman.c
@@ -28,6 +28,7 @@ static int oil_watchman_callback(bitbuffer_t *bitbuffer) {
 	uint16_t binding_countdown = 0;
 	uint8_t flags;
 	uint8_t maybetemp;
+	double temperature;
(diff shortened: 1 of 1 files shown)


merbanan:
Looks good, gory details are nice.

cas--:
I was curious about the gory details but seems the Hackaday user and project have been deleted recently. Luckily found a cached version so saved it as PDF and attached for posterity.

[Project | Rocket Receiver Teardown | Hackaday.io.pdf](https://github.com/merbanan/rtl_433/files/9123550/Project.Rocket.Receiver.Teardown.Hackaday.io.pdf)

zuckschwerdt:
Appreciated, thanks!

PR #241, opened 2015-12-15
Automatic level detection

vestom:
Adds automatic level detection, which is enabled by `-l 0` (not yet enabled by default).
When testing the complete test corpus in rtl_433_tests, it decodes many more files (e.g. Acurite, Chuango, Efergy E2, TFA Twin Plus and Oil Watchman). Only one regression found (current_cost/01/cc_2542watt.data).

Works by estimating the noise level during idle periods, high level when pulse is detected and setting threshold in between. Initial high level estimate is 8 times noise level - corresponding to a threshold of 4.5 times the noise level ((8-1)/2+1). To avoid too many false noise triggers, initial pulses of less than 10 samples is suppressed. Furthermore coalescing of too short FSK pulses is added to improve the noisy Oil Watchman samples.

Level estimates are now printed in the pulse analyzer (`-A`) (Could probably need some more explanation...)

It should be stable enough for use, but should probably be tested by others before enabled by default.


vestom:
For fun or info is here a graph of the signal/noise levels of all signals detected in the test corpus. (It is made before the lead-in noise estimation, so it might improve slightly if redone)
level_estimates

merbanan:
Thank you.

PR #249, opened 2016-01-11
add a converter for cf32 (gnuradio output format)

ckuethe:


merbanan:
Thanks.

vestom:
This merge generates a Segmentation Fault. rtlsdr_callback() does not like being called with n_read=0. I have traced it to baseband_low_pass_filter() - probably the memcpy() functions.
We should either fix this commit or make all of rtlsdr_callback() tolerant to being called with len=0.

ckuethe:
Thanks for the report - I'll look into fixing this.

On Tue, Jan 12, 2016 at 11:24 AM, vestom <notifications@github.com> wrote:

> This merge generates a Segmentation Fault. rtlsdr_callback() does not like
> being called with n_read=0. I have traced it to baseband_low_pass_filter()
> - probably the memcpy() functions.
> We should either fix this commit or make all of rtlsdr_callback() tolerant
> to being called with len=0.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/merbanan/rtl_433/pull/249#issuecomment-171023867>.
>



-- 
GDB has a 'break' feature; why doesn't it have 'fix' too?

vestom:
I have made a quick fix in 6e28a9a1f2a5d66ef5af965f68cf775560a21875, which is pushed for merging as part of #250 
However, please look into whether this is the proper way to fix it. 
Also the code will currently attempt to load a file as floats for _all_ other file modes than 0. This breaks the possibility to read AM demodulated data... (mode 1)

merbanan:
Yes, add an explicit check for the mode number.

PR #250, opened 2016-01-12
Convert remaining uses of sample based decoder (OOK_PWM_D) and cleanup

vestom:
Tested with test corpus.
Files missing for: Steffen, Mebus, Intertechno, Newkaku and some Acurite


merbanan:
Thank you.

PR #253, opened 2016-01-13
newkaku: support APA3-1500R

bwesterb:
Just bought a KlikAanKlikUit APA3-1500R remote.  `rtl_433` did not seem to support it at first.  I reverse-engineered the protocol, just to find out that the existing `newkaku` code did exactly the same, except for accepting the starting byte.

### DIFF
--- a/src/devices/newkaku.c
+++ b/src/devices/newkaku.c
@@ -18,7 +18,7 @@ static int newkaku_callback(bitbuffer_t *bitbuffer) {
     uint8_t bitcount = 0;
     uint32_t kakuid = 0;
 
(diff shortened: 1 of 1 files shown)


vestom:
Hi,
I believe we don't have any test samples for the Newkaku/KlikAanKlikUit in the test repository (https://github.com/merbanan/rtl_433_tests). This means it is difficult to support this device driver and there is a risk we might break something along the way. 

Is it possible you could add a couple of test recordings to the test repository?

Thanks
Tommy

bwesterb:
Done: merbanan/rtl_433_tests#66

vestom:
Excellent! It is now officially supported ;-)
Thanks.

PR #255, opened 2016-01-16
Fixes & new network output

zuckschwerdt:
Hi! Are you interested in these changes? Batched together here, each commit is a seperate issue. I can break out the commits into individual pull-requests if that's prefered.

Mostly simple fixes to syntax and compiler warnings, please ask if something is unclear.

The most interesting part is the addition of UDP network output. I'm running a few rtl_433 instances as probes and those report metrics to a central server. It's currently just the data fields as JSON string. This could be extended to support common (metrics) formats, like e.g. SYSLOG, STATSD, COLLECTD, GRAPHITE, AMQP. Currently I'm running a simple relay to cleanup and translate the JSON-UDP to those targets.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -50,6 +50,7 @@ struct data_printer;
 extern struct data_printer data_json_printer;
 extern struct data_printer data_kv_printer;
 extern struct data_printer data_csv_printer;
+extern struct data_printer data_udp_printer;
 
 /** Constructs a structured data object.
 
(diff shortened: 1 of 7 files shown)


merbanan:
Resend all the ok'ed parts as a new pull so I can merge those first. Then you can take the rest as separate  pulls.

zuckschwerdt:
Will do. As for the server part, socat or netcat will do, additionally I just posted some examples: https://github.com/zuckschwerdt/rtl_433_relay

PR #262, opened 2016-01-17
fix: broken bit operation

zuckschwerdt:
The intent `// Last bit is always 0` is not reflected int the code `b[3] && 0x7F` -- which translates to: `b[3] != 0 && 1`, i.e. `b[3] != 0`. Quite the opposite than intended!

The test files confirm that indeed the last bit is always `1`. This might be an artifact of the decoder, 24 bit length seem reasonable.

The bit count is 25 and `b[3]` will contain only a single bit (at MSB). Thus writing `b[3]` is sufficient. Or more verbose: `(b[3] && 0x80)` or even `((b[3] && 0x80) != 0)`.

The test for `ID 0x00000 not supported` is broken too. Actually here every ID with any single byte of all zeros will be dropped.

This patch keeps the original logic and changes the comment, as opposed to: https://github.com/zuckschwerdt/rtl_433/commit/484c87c986a9e6ad6d771cae54efabf3b147380a
But please note that
* more (valid) IDs will be accepted, e.g. 0xFF00 and 0x00FF
* the (valid) "test" command (0x00) will be accepted for Chuango
* the (maybe valid) command (0x00) will *still* be blocked for generic_remote

### DIFF
--- a/src/devices/chuango.c
+++ b/src/devices/chuango.c
@@ -30,8 +30,8 @@ static int chuango_callback(bitbuffer_t *bitbuffer) {
 
 	// Validate package
 	if ((bits == 25)
-	 && (b[3] && 0x7F)	// Last bit is always 0
-	 && (b[0] != 0x00) && (b[1] != 0x00) && (b[2] != 0x00)	// Reduce false positives. ID 0x00000 not supported
+	 && (b[3] & 0x80)	// Last bit (MSB here) is always 1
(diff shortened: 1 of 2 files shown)


merbanan:
Well @vestom might be able to sort this out if this is wrong. I'll merge it for now.

vestom:
Damn, that is some sloppy coding :-/ Thanks for noticing!
Last bit should always be zero though - if I had remembered to invert it like the three other bytes that is... I guess it must have been late.
Your code works, and that is good :-)

Issue #260, opened 2016-01-16
Temperature in Celsius, Fahrenheit or even both?


zuckschwerdt:
When using or converting to the new data printer format, do we give temperature readings in Celsius, Fahrenheit or even both?
I'd like to see some easily processable output (with JSON). I don't think giving both would add any value, only make things more verbose. But that might be just me?
Are there historical requirements and what are the opinions on this?

Quick census on `.field` keys:
- temperature_F
  - ambient_weather.c
- temperature_C
  - alecto.c
  - hideki.c
  - oil_watchman.c
  - oregon_scientific.c
  - prologue.c
  - wt450.c

rct:
The decoder should output what ever units are "native" to the device/message.

If you wind up with multiple conversions, you may loose some precision due to rounding floats for output to ascii.  

You may also obscure small changes which could be jitter or could be useful data.

The LaCrosse TX 433 mhz units I've seen use celcius natively (just offseting to avoid the need for a sign) so they should output Celsius.     

(I say that even though I'm from one of the few/only countries that uses Fahrenheit, so the output will no longer be directly meaningful to me. ;-)

zuckschwerdt:
Entirely reasonable, output all data as native as possible. (I'd even avoid floats as most devices encode a decimal and dividing by 10 isn't all perfect in binary.)
My current metric logging (graphite) doesn't honor units, it would be great to get a standardized output if at all possible.
Perhaps an option to convert values (native, C, F) in the data printer would be ok?

merbanan:
A normalizing step to select Celsius or Fahrenheit output, would be welcome.

rct:
A normalizing step would be welcome only if raw output was still accessible.

zuckschwerdt:
I was thinking about a command line option, e.g. `--c-to-f` & `--f-to-c` which would trigger code in `data_acquired_handler`:
* if `--c-to-f ` and the type is DATA_DOUBLE and the key is temperature_C then convert
* if `--f-to-c ` and the type is DATA_DOUBLE and the key is temperature_F then convert

And maybe, if there is a trailing `" C"` (`" F"`) in the format string switch that.
Also some sensors are opinionated and have build-in conversion. I'd remove that.

Current practise:
```
alecto.c:
  "temperature_C", "Temperature", DATA_FORMAT, "%.02f C", DATA_DOUBLE, (float) temp / 10.0F,
ambient_weather.c:
[...]

PR #263, opened 2016-01-17
new command line option -C native | f-to-c | c-to-f

zuckschwerdt:
Introduces new command line option `-C native | si | customary` that hooks into `data_acquired_handler` and
if `si` and type is DATA_DOUBLE and key is temperature_C then celsius2fahrenheit,
if `customary` and type is DATA_DOUBLE and key is temperature_F then fahrenheit2celsius.
Also if there is a FORMAT the last character "C" / "F" will be switched accordingly.

See discussion at https://github.com/merbanan/rtl_433/issues/260

Written for and tested with
```
alecto.c:
  "temperature_C", "Temperature", DATA_FORMAT, "%.02f C", DATA_DOUBLE, (float) temp / 10.0F,
ambient_weather.c:
  "temperature_F", "", DATA_FORMAT, "%.1f", DATA_DOUBLE, temperature,
hideki.c:
  "temperature_C", "Temperature",   DATA_FORMAT, "%.02f C", DATA_DOUBLE, temp/10.f,
oil_watchman.c:
  "temperature_C", "", DATA_DOUBLE, temperature,
oregon_scientific.c:
  "temperature_C", "Temperature", DATA_FORMAT, "%.02f C", DATA_DOUBLE, get_os_temperature(msg, sensor_id),
prologue.c:
  "temperature_C", "Temperature", DATA_FORMAT, "%.02f C", DATA_DOUBLE, temp,
wt450.c:
  "temperature_C", "Temperature",DATA_FORMAT, "%.02f C", DATA_DOUBLE, temp,
```

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -28,6 +28,7 @@
 #include "pulse_detect.h"
 #include "pulse_demod.h"
 #include "data.h"
+#include "util.h"
 
 
 static int do_exit = 0;
@@ -45,6 +46,12 @@ int debug_output = 0;
 int quiet_mode = 0;
 int overwrite_mode = 0;
 
+typedef enum  {
+    CONVERT_NATIVE,
+    CONVERT_SI,
+    CONVERT_CUSTOMARY
+} conversion_mode_t;
+static conversion_mode_t conversion_mode = CONVERT_NATIVE;
 
 int num_r_devices = 0;
 
@@ -113,6 +120,7 @@ void usage(r_device *devices) {
             "\t\t 3 = Raw I/Q samples (cf32, 2 channel)\n"
             "\t\t Note: If output file is specified, input will always be I/Q\n"
             "\t[-F] kv|json|csv Produce decoded output in given format. Not yet supported by all drivers.\n"
+            "\t[-C] native|si|customary Convert units in decoded output.\n"
             "\t[<filename>] Save data stream to output file (a '-' dumps samples to stdout)\n\n", 
             DEFAULT_FREQUENCY, DEFAULT_SAMPLE_RATE, DEFAULT_LEVEL_LIMIT);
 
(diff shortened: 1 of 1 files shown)


rct:
Perhaps thinking ahead for the weather stations, the command line args should be native|metric|english?   

We might want to handle KmPH <-> MPH, MM <-> IN., etc.

rct:
(at this point only suggesting a change to the command line arguments, since I don't know if there are any weather stations using the new data format routines, I know the Acurite doesn't yet.)

zuckschwerdt:
I thought of using a bit field later on for extensibility, but your suggestion is even better!
(also it might be worthwhile to consider introducing even more structure into the data fields, e.g. explicit units, significant digits. Or maybe to encourage a common structure for the FORMAT string.
I could make a table on the current practice and all current units if something in this direction is wanted.

zuckschwerdt:
Can someone check if "english" is what we want to convey here? According to Wikipedia there seems to be "imperial", "english" and "U.S. customary".
I've always though Fahrenheit and Inches to be "imperial" but I can't make sense of it anyway ;)

rct:
Nothing is ever simple.  ;-)

I'm fine wih imperial, but I'm sure there are some other opinions.   Some weather related sources I check didn't really have consensus, and add S.I. To the mix for the metric side.

zuckschwerdt:
As far as I could gather (https://en.wikipedia.org/wiki/Imperial_and_US_customary_measurement_systems) "english" is pre-1824 and "imperial" users now all converted to metric. So we are really talking about "U.S. customary" here. But "us-customary" isn't a nice option value at all.
People with a stake in using that option need to speak up :)

merbanan:
Native, si and customary.

A bit boring but somewhat correct. Imperial units or simplified English units would have been fun also but confusing.

Patch ok otherwise.

eras:
Further development ideas: with this I think it might be a good idea to go away with the DATA_FORMAT being used to format temperatures and just always format temperature_C and temperature_F values with some common format string, such as the "%.2f C" or "F" every code uses anyway..

And to go a bit further would be to add actual type to the data side, so in addition to DATA_DOUBLE there would be DATA_UNIT_C or even DATA_ABS_TEMP, DATA_UNIT_C. This would work better if there are sensors that have multiple measured values we wish to convert between and can't expect them all to have a predefined name.

zuckschwerdt:
Most likely we'd need to keep data type (and maybe precision) separate from the units. E.g. humidity is usually an integer, and then there is one sensors reporting a float...

zuckschwerdt:
I changed the code (and squashed the commit) to `native | si | customary`.

PR #264, opened 2016-01-17
Fix abs() call on unsigned int

zuckschwerdt:
Calling `abs()` on `unsigned int` is implementation (compiler) specific. As far as I can see the current code gets the intended result, but by chance only. `abs()` seems to convert unsigned values with set high bit to negative values and then removes the sign. Let's make the intention explicit and sound.

### DIFF
--- a/src/pulse_demod.c
+++ b/src/pulse_demod.c
@@ -17,6 +17,11 @@
 #include <stdlib.h>
 
 
+static unsigned int absdiff(unsigned int a, unsigned int b)
+{
+	return a > b ? a - b : b - a;
+}
+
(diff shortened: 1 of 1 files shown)


merbanan:
Imo just changing the pulses and device struct members to int should remove the need for this wrapper, does it not ?

vestom:
Yes, I also think we should just change to using signed int. I guess there are more potential bugs in using unsigned arithmetic... However the chance of not running on a 2s complement machine is probably slim ;-)

vestom:
I went ahead and changed it to int (#265). Using unsigned was always a bit nasty due to all the arithmetic going on... Thanks for noticing this!!

merbanan:
I merged the other patch. Thanks for noticing the issue.

PR #265, opened 2016-01-17
Change pulse data from unsigned to int to avoid undefined behaviour

vestom:
Thanks to zuckschwerdt for noticing this!
Tested with test corpus with no differences.


merbanan:
Jolly good.

PR #266, opened 2016-01-18
Xc0348 fixes

viraptor:
Fixed parsing issues, added CRC check and json result


merbanan:
The csv struct is missing.

PR #281, opened 2016-01-26
new feature: allow multiple outputs (WIP)

zuckschwerdt:
This is tested and works. *It's not intended/proposed for merge right now though.*

I'd like to allow multiple output handlers, i.e. allowing multiple `-F handler` arguments.
This code refactors the single `output_format` field to a chain of `output_handler`s, and allows the generalization of setup and tear down callbacks for each handler. It touches as few code as possible to do so.
It might not make immediate sense to introduce this but it clears the way to enhance the output printers:
The next step would be to add file arguments to the outputs. E.g. `-F kv:stderr -F json:/tmp/socket -F csv:log.txt` would be nice.

(TODO: aux_free() should be a general close() and there might need to be a sensible default if no `-F` is given.)

What are the thoughts on this? Is this appreciated? Is there a better way maybe?

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -316,7 +316,10 @@ void data_print(data_t* data, FILE *file, data_printer_t *printer, void *aux)
 		.aux     = aux
 	};
 	ctx.printer->print_data(&ctx, data, NULL, file);
-	fputc('\n', file);
+	if (file) {
+		fputc('\n', file);
+		fflush(file);
+	}
 }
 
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
This is now ready for merge.
- There is no change in default behaviour, i.e. if there is no `-F ` a KV output will be used.
- Using a single `-F` will act the same as before
- Using multiple `-F` will now output to all the destinations instead of just the last one

Also the change tidies up the special handling of CSV aux data setup/teardown.

merbanan:
I'll test and review this in a few days.

merbanan:
Looks good.

PR #286, opened 2016-02-01
Return correct events count from oil_watchman and emontx drivers

dwmw2:


merbanan:
Ok, but what can you do with that ?

dwmw2:
Have a happy feeling that my code is correct? Isn't that enough?

But seriously, I was working on making it store captures only for signals that *weren't* decoded, and this was a prerequisite.

merbanan:
I am just trying to understand all the code that I merge. At a brief look I didn't but I assume that you know what you where doing.

PR #288, opened 2016-02-02
Add remote control on chip HT680

igorok107:
FCC ID: [ERY919A2](https://fcc.io/ERY919A2)
ery919a2


merbanan:
Hi, have you sent test signals to the rtl_433_tests repository? And new decoders must use the data_make() api when reporting events. Look at rubicson.c as an example. And you need to change the code so that only one event per button press is generated.

PR #291, opened 2016-02-04
adding a bitbuffer_find_repeated_row()

zuckschwerdt:
this adds a function `bitbuffer_find_repeated_row()` with examples how it simplifies device code.
Just taking some 'middle' row and comparing all the fields to some other rows isn't good enough to reliably get 'good' data. Now you can explicitly get a 'strong' row that is repeated a given number of times.


merbanan:
Can you add a:

    if (!(bits == 36))
        return 0;

check also ? That way we wont calculate this in vain.

zuckschwerdt:
Actually I've prepared some updates/cleanup for both devices already :) I'll rebase and send it.

PR #292, opened 2016-02-04
code cleanup

zuckschwerdt:
This is a cleanup of the code for Prologue and Nexus sensors.
The functional changes are to accept a trailing 0 bit as artefact from the demodulation and added output of the channel number for Nexus.


merbanan:
Ok, but still no early row bit number check.

PR #298, opened 2016-02-08
Decode device id from current_cost, and support more devices

davidsheldon:
Decode the data from my currentcost device. Added the device id to the output.

The high bits of bytes 2,4 and 6 tell us if the data for channels 0,1 and 2 is valid. Is it possible to not include values in the data object based on the flags easily? Should we just always supply them?


merbanan:
Just switch on whatever fields you have and create different objects. That should work fine.

enavarro222:
@davidsheldon Thanks for the contribution ! I was not sure about the split between constant header / device ID, I only have one sensor.

Thanks also for your PR in rlt_433_test.

PR #307, opened 2016-02-21
LaCrosse WS-2310 json output

jvestman:
Changed LaCrosse WS-2310 to use data_make. However, It no longer reports customary units for wind speed and rainfall as better place for unit conversions would be data_acquired_handler().

### DIFF
--- a/src/devices/lacrossews.c
+++ b/src/devices/lacrossews.c
@@ -19,6 +19,7 @@
 
 #include "rtl_433.h"
 #include "util.h"
+#include "data.h"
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
The wind/gust speed has `wind_key = msg_type == 3 ? "wind_speed_ms":"gust_speed";` -- I guess the `_ms` suffix is missing. The `output_fields` lists this as expected.

jvestman:
@zuckschwerdt Thanks for catching that.

merbanan:
Thank you

PR #312, opened 2016-02-24
Output less decimal places to get rid of bogus digits

ygator:
This eliminates the inaccurate digits.


rct:
It is only bogus digits if it happens to be large values like temperature.  

Putting this restriction in a general routine that outputs machine readable JSON without knowing what the expected precision should be sure feels non-general.

ygator:
I don't disagree.  However, I think it is better then what was there.

Also this is why I had suggested early on that the values be represented by strings.  They would be able to be represented exactly as read.  Just the conversion functions if enabled would have to deal with them.

PR #314, opened 2016-02-25
Changed json fields name on Fine Offset WH1080

ovrheat:
Updated Fine Offset WH1080 with more correct json fields name + some cosmetic change.
Still I can't figure out how to format json data fields, in the same way of the 'normal' screen output by using DATA_FORMAT, "%02d" (i.e. '02' instead of '2' as json looks now). Can you please give me some hint?

### DIFF
--- a/src/devices/fineoffset_wh1080.c
+++ b/src/devices/fineoffset_wh1080.c
@@ -8,8 +8,8 @@
  * (aka Froggit WH1080)
  * (aka .....)
  *
- * This module is based on Stanisław Pitucha ('viraptor' https://github.com/viraptor ) earl code for the Digitech XC0348 Weather Station, 
- * which seems to be a rebranded Fine Offset WH1080 Weather Station. 
+ * This module is based on Stanisław Pitucha ('viraptor' https://github.com/viraptor ) earl code for the Digitech XC0348 
+ * Weather Station, which seems to be a rebranded Fine Offset WH1080 Weather Station. 
  *
  * Some info and code derived from Kevin Sangelee's page: 
  * http://www.susa.net/wordpress/2012/08/raspberry-pi-reading-wh1081-weather-sensors-using-an-rfm01-and-rfm12b/ .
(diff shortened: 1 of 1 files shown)


ygator:
How it is right now you don't.  Json double is hard coded to %.3f now with my change and before %f.  Json int is hard coded to %d.  However, it should not matter as json data is for programs to read and they can format the data however they want. Key value defaults to the same as json unless a format is specified.

merbanan:
Json output is not for pretty printing. Thus the issue you mention does not affect usage.

ovrheat:
Thank you to both of you. Yes it's true, json is not for pretty printing. I was only thinking that in some particular case it would have been useful, like in case of time data (that's the case). I think that nobody will log hours as '1', or minutes as '3' on their scripts, so everybody will be forced to implement a '0' before single digits, and I was just trying a way to avoid this little waste of time by giving out the already correct output format. Not a big trouble, anyway. :+1:

zuckschwerdt:
What is consuming the time data? Maybe an ISO 8601 string (e.g. `2016-02-25T11:46:42` would be better suited?

ovrheat:
@zuckschwerdt:  No, not necessarily. The WH1080 json time output is already like 'Hour '0', Minute '1', Second '0', Month '2' and so on. Every datetime data is already present so the user is already able to build a  '2016-02-25T11:46:42' string, if this is what he wants.
But it would have been a good start if json data was presented as Hour '00' instead of Hour '0', it would have been more coherent with time data format. Nobody will use hour '0' as he will always be forced to translate '0' to '00'. 
Anyway, as already said, this is not an important question, I was only wondering if there was a way, that's why I asked.  We can leave it as is.

ygator:
I agree for json the time should be output as iso 8601 string that includes UTC.  So it would be 2016-02-25T11:46:42Z

rct:
I think there are two reasonable JSON representations for time  ISO8601 (preferably in UTC), or better yet the raw milliseconds since the epoch. 

If we're doing JSON output of raw data to be used by other software milliseconds since the epoch offers a number of advantages - 1) make millisecond resolution timestamps available, 2) timezone interpretation is up to the downstream software. 3) it is reasonably compact and should be handled by most (reasonable) date/time libraries.

PR #318, opened 2016-02-27
Added data labels for KlikAanKlikUit output

jvestman:
Sorry, omitted data labels from initial patch. Apparently, the convention is that data labels should be omitted for time, model and id and the rest of the fields should have labels, right?


merbanan:
I don't really know, but they are needed for some values for correct processing.

PR #322, opened 2016-03-03
The contact sensors required a larger pulse tolerance

kajksa:
The magnetic contact sensors from Chuango required a larger tolerance.

### DIFF
--- a/src/devices/chuango.c
+++ b/src/devices/chuango.c
@@ -66,7 +66,7 @@ static int chuango_callback(bitbuffer_t *bitbuffer) {
 
 
 PWM_Precise_Parameters pwm_precise_parameters = {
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
@kajksa <b>Many sensor</b> require... What sample rate are you using? Did you raise the default of 250000 Hz? Note that the tolerance is given in samples and depended on the sample rate (i.e. tolerance 20 needs to be 80 at 1MHz, to get 80 µs in each case).
This ought to fixed some day.

kajksa:
I am using the default of 250000 Hz and did not think of that issue. Thanks for pointing it out. I am working on adding some other device and perhaps some day I will feel confident enough about the code that I will attempt to fix it.

PR #325, opened 2016-03-07
Add Quhwa doorbell support

kajksa:


src/devices/quhwa.c
```diff
@@ -0,0 +1,73 @@
...
+	    
+	b[0] = ~b[0];
+	b[1] = ~b[1];
+	b[2] = ~b[2];
+	
+	unsigned bits = bitbuffer->bits_per_row[0];
+
+	if ((bits == 18) &&
+	    ((b[1] & 1) && (b[1] & 2))  && // Last two bits of second byte are one
+	    (b[2]==0xFF)) // Last two bits are one
```
zuckschwerdt: The comment does not fit the code here.

kajksa: OK, you probably right. I am trying to express that for the 18 bits the last 4 are ones 

[01] {18} 61 bb c0 : 01100001 101110**11 11**

So how do I do that?

zuckschwerdt: I see, since you inverted the last byte the remaining unused 0 bits are set to 1 too.
Maybe this makes the intent more obvious:
((b[1] & 0x03) == 0x03) &&
((b[2] & 0xC0) == 0xC0)

kajksa: Yes, that is much clearer. Now I will just figure out if I have the git skills to fix it... :)

PR #341, opened 2016-03-24
add -T option to set run duration

panarom:
run rtl_433 for a specified amount of seconds, then quit.  Useful for discrete (rather than continuous) data gathering.

NOTE: I am unable to test in synchronous mode, because (regardless of the changes in this PR), rtl_433 segfaults on my armv7l machine when I pass in the -S option. Please review carefully, if you decide to merge (if not, this functionality can be implemented with existing options by doing some arithmetic with the sample rate & passing the appropriate sample count parameter)

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -35,6 +35,8 @@ static int do_exit = 0;
 static int do_exit_async = 0, frequencies = 0, events = 0;
 uint32_t frequency[MAX_PROTOCOLS];
 time_t rawtime_old;
+int duration = 0;
+time_t stop_time;
 int flag;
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
You need to specify an output destination, if unset -S will crash.

vestom:
Introduces build error - please fix:
projects/rtl_433/src/rtl_433.c:725:25: error: ‘rawtime’ undeclared (first use in this function)
     if (duration > 0 && rawtime >= stop_time) {
                         ^
rawtime is not in scope, as it is only locally declared in if sentence above...

merbanan:
commit reverted

PR #342, opened 2016-03-24
Prevent crash when no output file is given with `-S`

zuckschwerdt:
This prevents `fwrite()` on unintialized `demod->out_file` and gives an informative error message.


panarom:
:+1: 
@zuckschwerdt I didn't realize that's what I was doing wrong, thanks for finding/diagnosing/fixing.

zuckschwerdt:
It wasn't obvious, I needed to fire up gdb to pinpoint the fault. Thanks for reporting!

PR #343, opened 2016-03-24
use uintptr_t instead of unsigned long - fixes build error on win64

somesortoferror:


merbanan:
Looks good. I will merge this when I can test it at home.

PR #346, opened 2016-03-25
Pulse analyzer: Add PCM detection. Minor cleanup.

vestom:


merbanan:
In vestom we trust. Looks good.

PR #349, opened 2016-03-26
Acurite tower json

panarom:


panarom:
this pr adds some quotation marks in standard output where they probably shoudln't be. will resubmit once that's fixed.

zuckschwerdt:
This introduces an inconsistency with the JSON output. I.e. with JSON I expect the ID field to be a basic integer number. This breaks that assumption for one sensor with an output of a quoted hex-string.
More importantly all e.g. temperature and humidity outputs will be converted to strings with affixed unit in the JSON output...
I guess the post processing should be left to the JSON consumer.

Also you should not convert and ouput a second temperature scale (Fahrenheit).

merbanan:
Yes, only output the native data. And int for id. We should document that.

PR #350, opened 2016-03-27
Fix build via mingw32

mnhauke:


merbanan:
I really dislike #ifdefs. We should work towards not needing them.

PR #351, opened 2016-03-27
Acurite tower json

panarom:
the acurite device is missing json functionality, so I've implemented it for the acurite device I happen to have.  I could add code for other acurite devices, but I'm unable to test those, so I refrained at present.
 
Additionally, the json printer for ints & floats had a hard-coded format string, which means e.g. a single char can't be printed as an int, nor an integer as a hexadecimal string. I'm somewhat dissatisfied with this code, but it's not immediately clear to me how to DRY it up: suggestions welcome.  Alternatively I could refactor it so that I pass a pre-computed DATA_STRING, instead of a DATA_INT with a DATA_FORMAT parameter, to get the desired char & hex strings.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -417,15 +417,24 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
 	    tempc = acurite_txr_getTemp(bb[4], bb[5]);
 	    tempf = celsius2fahrenheit(tempc);
 
-	    printf("%s Acurite tower sensor 0x%04X Ch %c: %3.1F C %3.1F F %d %% RH\n",
-		   time_str, sensor_id, channel, tempc, tempf, humidity);
+        data_acquired_handler(data_make("time","",DATA_STRING, time_str,
+                                        "model", "",DATA_STRING,"Acurite Tower Sensor",
+                                        "id", "Sensor ID", DATA_FORMAT, "%#X", DATA_INT, sensor_id,
+                                        "channel", "Channel", DATA_FORMAT, "%c", DATA_INT, channel,
+                                        "temperature_C", "Temp ℃", DATA_DOUBLE, tempc,
+                                        "temperature_F", "Temp ℉", DATA_DOUBLE, tempf,
+                                        "relative_humidity", "Relative Humidity (%)", DATA_INT, humidity, NULL));
+
 
 	    // currently 0x44 seens to be a normal status and/or type
 	    // for tower sensors.  Battery OK/Normal == 0x40
-	    if (sensor_status != 0x44)
-		printf("%s Acurite tower sensor 0x%04X Ch %c, Status %02X\n",
-		       time_str, sensor_id, channel, sensor_status);
-
+        if (sensor_status != 0x44) {
+            data_acquired_handler(data_make("time","",DATA_STRING, time_str,
+                                            "model", "",DATA_STRING,"Acurite Tower Sensor",
+                                            "id", "Sensor ID", DATA_FORMAT, "%#X", DATA_INT, sensor_id,
+                                            "channel", "Channel", DATA_FORMAT, "%c", DATA_INT, channel,
+                                            "sensor_status", "Sensor Status", DATA_INT, sensor_status, NULL));
+        }
 	}
 
 	// The 5-n-1 weather sensor messages are 8 bytes.
--- a/src/data.c
+++ b/src/data.c
@@ -394,12 +394,28 @@ static void print_json_string(data_printer_context_t *printer_ctx, const char *s
 
 static void print_json_double(data_printer_context_t *printer_ctx, double data, char *format, FILE *file)
 {
-	fprintf(file, "%.3f", data);
+    if(format == NULL) {
+	    fprintf(file, "%.3f", data);
+    } else {
+        int length = snprintf(NULL, 0, format, data) + 1; //add space for null byte
+        char *formatted_str = malloc(sizeof(char) * length);
+        sprintf(formatted_str, format, data);
+        print_json_string(printer_ctx, formatted_str, "", file);
+        free(formatted_str);
+    }
 }
 
 static void print_json_int(data_printer_context_t *printer_ctx, int data, char *format, FILE *file)
 {
-	fprintf(file, "%d", data);
+    if(format == NULL) {
+	    fprintf(file, "%d", data);
+    } else {
+        int length = snprintf(NULL, 0, format, data) + 1; //add space for null byte
+        char *formatted_str = malloc(sizeof(char) * length);
+        sprintf(formatted_str, format, data);
+        print_json_string(printer_ctx, formatted_str, "", file);
+        free(formatted_str);
+    }
 }
 
 /* Key-Value printer */


zuckschwerdt:
As noted in #349 please do not format JSON output. The pretty printing is up to the JSON consumer.
Also try to match already existing format conventions, i.e. here:
- ID as int
- channel as int
- do not compute/output derived values like Fahrenheit
- use e.g. "%.02f C" as format string for temperature
- use e.g. "%u %%" as format string for humidity

rct:
FWIW, Since you brought it up:
* Acurite channels are chars ('A', 'B', 'C' for most of the weather related devices)
* Regarding battery status, I think I forgot the submit a patch once I got the device on an adjustable power supply to confirm the battery bit.   
* Note however, I choose not to redundantly print "battery OK" on 99% of the messages since that seems pointless.  Instead I think it makes much more sense to send a message when you've got a non-typical status byte indicating battery low, or some status that hasn't been seen yet. 
* Alternatively, I could just see outputting the status bits as a value, especially when most of the bits haven't been seen/decoded yet.

Also, since we are doing JSON:
* Why are numerical values being output with extra type information in the value that needs to be parsed out instead of just having a type for temp_c and temp_f?   I know that's the way other devices are doing it, but it sure seems wrong.
* If there are cases where humidity isn't a percentage, then perhaps it should get a different type?

src/devices/acurite.c
```diff
@@ -417,15 +417,24 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
...
 	    tempf = celsius2fahrenheit(tempc);
 
-	    printf("%s Acurite tower sensor 0x%04X Ch %c: %3.1F C %3.1F F %d %% RH\n",
-		   time_str, sensor_id, channel, tempc, tempf, humidity);
+        data_acquired_handler(data_make("time","",DATA_STRING, time_str,
+                                        "model", "",DATA_STRING,"Acurite Tower Sensor",
+                                        "id", "Sensor ID", DATA_FORMAT, "%#X", DATA_INT, sensor_id,
+                                        "channel", "Channel", DATA_FORMAT, "%c", DATA_INT, channel,
+                                        "temperature_C", "Temp ℃", DATA_DOUBLE, tempc,
+                                        "temperature_F", "Temp ℉", DATA_DOUBLE, tempf,
```
merbanan: Only output the native type, that looks like to be Celsius. Sensor id's should be int.

src/devices/acurite.c
```diff
@@ -417,15 +417,24 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
...
 	    // currently 0x44 seens to be a normal status and/or type
 	    // for tower sensors.  Battery OK/Normal == 0x40
-	    if (sensor_status != 0x44)
-		printf("%s Acurite tower sensor 0x%04X Ch %c, Status %02X\n",
-		       time_str, sensor_id, channel, sensor_status);
-
+        if (sensor_status != 0x44) {
+            data_acquired_handler(data_make("time","",DATA_STRING, time_str,
+                                            "model", "",DATA_STRING,"Acurite Tower Sensor",
+                                            "id", "Sensor ID", DATA_FORMAT, "%#X", DATA_INT, sensor_id,
```
merbanan: Sensor Id as int.

merbanan:
Regarding the json changes. Can you elaborate when you need the code that you added?

src/devices/acurite.c
panarom: @merbanan do you agree with @zuckschwerdt that the channel should also be an int?
If both id & channel become ints, then there's no need for my refactoring of the json printing in data.c

panarom:
@merbanan assuming I understand your question correctly, I wanted to print out an integer (id) as a hexidecimal, and a single char (channel) as a character, by way of casting to an int.  I noticed that the json_print_int had a format parameter tha was unused, so I implemented it a.)to get the formatted printing I wanted and b.)to make it consistent with the string printing function, that honors the format string.
But given that I need to actually generated the formatted string anyway, so that I can pass it to print_json_string so that it can be quoted/potentially esacped, then it's probably better to revert the changes I made, generate the strings elsewhere, and then use DATA_STRING instead of hacking DATA_INT.
However, even if we use decimal integers for device ids, I'm still not certain we should force a floating point number to have exactly 3 digits past the decimal point: I'd suggest that should be left up to the sensor data's precision, which implies the print_json_float should honor a formatting string, even if there's no present need for print_json_int to do so.

panarom:
@zuckschwerdt sorry, I wasn't deliberately ignoring your comment in #349 : I didn't see it before I reopened this PR

panarom:
also, what's the preference for updating a PR? add additional modifying commits? squash them/force push? or close this one and open a new one?

panarom:
I agree with @rct that the output string for temperature should not have a 'C' character in it: it should be strictly numeric; although I suppose if I really support that position I should go refactor it everywhere else…

ygator:
@panarom See #312 for the reason the json float is currently %.3f.
I agree with you that the json float should handle a format since kv already does and for that matter all routines should handle formats to be made consistent.  Unit conversions might have to be looked at though.

zuckschwerdt:
Each issue should most likely be a single commit. Squash and force push seems the way to go.

It seems we have different expectations/requirements of the outputs. Perhaps we could write down the requirements and goals for each and work on that before changing anything.

merbanan:
@panarom id and channel should be ints. Regarding PR's do what is most easy for you. But one commit per "feature" is the way to go. And we need to formalize the output.

merbanan:
And I think accepting a format parameter for float is the way forward. That way we can add decimal rounding for decoder where we know it. And that means partial revert of #312 . Does that sound like a fair trade-off ?

My idea with rtl_433 is to make it the middle layer for radio devices and software that process the sensor data. It should just lightly normalize the data so processing the data is easy. So if a device has channel A/B/C/D then translate that to 1,2,3,4. The device could add another member with the native output just for bookkeeping. But this all comes back to an initial formalization of the json structure syntax.

panarom:
@ygator I know I just said that the temperature should be strictly numeric; but the reason I refactored print_json_float the way I did was to handle the case of a more elaborate format string, and still print it in json.  E.g. imagine you want the json value of rainfall to be ```1.34 "```, with a quotation mark for inches: that quotation mark would need to be escaped for proper json rendering, which is why the data and format string are passed to sprintf, so that the fully formatted string may be passed on to print_json_string for escaping.

ygator:
@merbanan I agree and If I would of noticed the format I would have just output my data as %.1f after allowing for format to be used for json doubles.  Then all other decoders could be updated to use the format as well which would solve the bogus digits.

ygator:
I see now that the format is only used for kv because json would have to use the same one.  For temperature all of the decoders are using a format with the unit embedded in it.

panarom:
One more question before I start a new branch: according to the acurite documentation(http://www.acurite.com/accuracy), the sensor's only accurate to within ~1 ℃: can I round the temperature to the nearest integer, or should I leave it as a floating point?

merbanan:
Output whatever is on the receiver display.

merbanan:
It should be quite linear which useful information. Ie 10.5 is 0.5 more then 10.

panarom:
@rct  any idea how many significant figures are on an acurite receiver display?  I only have the thermometers (because they were inexpensive): the only receiver I've got is my rtlsdr dongle.
@merbanan : the picture on the website looks like it displays to the nearest degree.
Otherwise I'll just leave it as what's currently in acurite.c; but I'll feel guilty about it because I distinctly recall the lectures on significant figures in science class ;-)

ygator:
It looks to me like all the decoder functions are one decimal, so you should output what the decoder functions decode which is one decimal.

rct:
@panarom - Acurite has multiple displays including an internet bridge which just does 433 mhz -> acu-link internet service.  More information is available in the IP messsages sent by the link than is displayed on the consoles.  I wouldn't trust their documents/specifications that I've seen but would go by what appears to be the range specified by the according to the encoding.

The encoding for the tower (596) only covers tenths of a degree Celsius. 

Note: the encoding/native formats for Acurite devices varies amount devices even with the devices that share the same or a very similar message format like the tower and the 5-n-1.  HOWEVER, the 5-n-1 reports natively in Fahrenheit to a tenth of a degree.

Note: The txr callback handles both the 5-n-1 and tower sensors.  If you make any changes outside of the printing, make sure the 5-n-1 still works.   They should probably both be converted to the new printing system at the same time.

rct:
@merbanan - Why do you feel the Acurite Channel should be in int instead 'A', 'B' or 'C' as is marked on the physical devices?

And which int do think should be output?  

The encoding is 0b00: 'C', 0b01: Not used, 0b10: 'B', 0b11: 'A'

If you xor that it's close, but still doesn't map {0,1,2} -> {'A','B','C'{ because of the one unused value.

merbanan:
@rct Guess we need to change to string then. It would have been nice to make it a number. But I am undecided. The principle is to use model, channel and sensor id to uniquely identify sensors. It's not easy to know the effect of changing the ground structure.

rct:
FWIW, I don't know which specs are authoritative, but:
* it doesn't look like JSON has a char type so it would need to be a string.  
* It doesn't look like bases/radix are supported in JSON numbers either. 

Though what the most common parsers support is a different matter.

ygator:
I think we should fall back to what @merbanan has said in the past.  Output what comes out of the decoder with no conversions.  It is up to the receiver to do what it wants with the output.  So channel would be a number.  Most everything would be a number including battery.  I don't like the string OK or LOW.

merbanan:
low_battery=1 instead ?

zuckschwerdt:
`low_battery=1` sounds reasonable. I retain all values with a time-series db and actually use `battery=0` with percentage as unit (the other value being `1`). This yields nice battery level gauges that can also be stacked.

ygator:
I would prefer battery be before low.  So battery_low or since I don't like underscored field names then BatteryLow/batteryLow/etc...
Plus it would be consistent as we do use C_temperature

vestom:
FWIW, I agree that the goal is to have the JSON output be normalized and SW centric, such that higher layers (e.g. OpenHAB/Domoticz like software) can easily treat different sensors models of same type in the exact same way. 

This means each sensor should not have different ID types, battery status types etc. Furthermore we should prefer numeric values instead of strings: E.g. 1 instead of "OK" or "true" - strings are in no way more boolean than integers. I do not think it matters whether IDs are nicely numbered (0,1,2...), as long as the values are unique. Some sensors (e.g. FineOffset) will output a random 8 bit value anyway.

However for "pretty printing" (kv) it could/should be converted to something more sensor specific (e.g. A, B, C), that looks humanely familiar to the sensor nomenclature and precision.

We should probably define some sensor classes or at least some pre-defined keys, such that higher level software will not have to look at the sensor model to infer whether it is a temperature/humidity/rain/wind/door/smoke/button/... sensor.

Just my 2 cents :-)

zuckschwerdt:
So "A", "B", "C" will translate to channel numbers 65...67 in JSON. Good enough, i guess.
If I get around to it I'll compile a table of current practice on keys and values and we can smooth out the edge cases and turn that into a guide.

ch0mik:
Hello,

Im looking for rtl433 modifications for write WX data to JSON file lika as  https://github.com/asdil12/pymultimonaprs/blob/master/pymultimonaprs.json

Vy 73! de SQ7MRU, op. Pawel

rct:
@ch0mik - rtl_433 currently has support to output a number of sensors as JSON, (or CSV, or key-value).    What kind of sensors do you have?   The conversion to structured data for all devices has been a little stalled until there is better consensus on how things should be represented. 

I'm not familar with pymultimonaprs.  I'll have to take a look at it.  If you want weather data to send out via APRS, it should be doable though I expect you'd have to right a bit of a wrapper to parse, ranslate, and keep the required state.  That shouldn't be too hard to do in Python.

I suggest moving any discussion about this out of this GitHub issue, which is specific to 2 Acurite devices, and take it to the rtl_433 Google Group.

73 de W2RCT, Rob.

rct:
I'm somewhat interested in trying to move the structured data (aka JSON) conversion forward for Acurite devices.  I've been doing small amounts of acurite clean up on the code as I've worked on other devices this past summer. 

There was a lot of info and opinions in this thread.   It will probably be most useful in the long term if we took a bit of a survey of the devices that are currently supported and summarized the current output, and the available data in the raw messages.

Also there are two ideas which can conflict with each other:

1 - Generic data representations easy to consume by clients without regard for the specific model/manufacturer:  (@vestom wrote)

> I agree that the goal is to have the JSON output be normalized and SW centric, such that higher layers (e.g. OpenHAB/Domoticz like software) can easily treat different sensors models of same type in the exact same way.
> 
> This means each sensor should not have different ID types, battery status types etc. Furthermore we should prefer numeric values instead of strings: E.g. 1 instead of "OK" or "true" - strings are in no way more boolean than integers. I do not think it matters whether IDs are nicely numbered (0,1,2...), as long as the values are unique. Some sensors (e.g. FineOffset) will output a random 8 bit value anyway.

2 - Provide access to a somewhat raw representation of the data, so that no information or precision is lost by conversion to a generic representation.

Acurite channel numbers are a good example.  For software consumption, the Acurite Channel number can be dropped for the 5-n-1, Tower sensor, and perhaps some others, because there is a fixed ID of at least 16 bits that is unique.   However, for people dealing with the physical sensors, and display consoles, the A, B, C, channel indication is more meaning than the ID.   (The ID is not marked on the devices.)     Adding the channel number as an extra data item probably wouldn't confuse the majority of consumers that would just ignore fields they don't understand.

A number of devices have a few bits or a byte that contain some status info, like battery OK vs. LOW.   The other bits may be completely unused or reserved for future use.   There may be status information in there which the person writing the decoder hasn't figured out yet.   As above this can probably be handled by having both a generic battery OK/Low (boolean) data element, and a "status" or "model specific raw status" element.   (This would need to be revisited if any devices report battery voltage.)

Or perhaps there should be some mode flags to output cooked, generic data, or output raw data.

kpreid:
@rct A suggestion to avoid the conflict you describe:

* Both normalized and model-specific information is produced by the decoders.
* In the final JSON output, they are distinguishable from each other (e.g. every model-specific field is prefixed with the model name and a separator `{"channel": ..., "acurite$channel": ...}`, or they could be in a JSON object inside a specific field `{"channel": ..., "raw": {"channel": ...}}`).
* (Optional) There is a flag to omit the model-specific information entirely, so the output is simple and non-redundant.

This allows applications processing the data to do so generically, but also do special processing if they can do better than rtl_433 does with some particular model-specific information, because both sets are available.

merbanan:
@rct adding a raw tag to the json output would be fine by me. The goal is still to have a strict sane super set of json tags the different decoder modules can implement. Right now it is somewhat ok with regards to temperature sensors but there is documentation of it. But any work towards making things better is always welcome.

zuckschwerdt:
There are some good points about formatting in here. Let's keep this open until I get around to pick this up. note: this isn't for merge anymore as #436 already went in.

Issue #353, opened 2016-03-28
AcuRite Rain Gauge False Positives


ygator:
Can someone with an AcuRite Rain Gauge do more processing to filter out it triggering on Oregon Scientific Protocol V1 Sensors.

I am getting:
AcuRite Rain Gauge Total Rain is 1.0mm
Raw Message: 4a 90 02 00 00
{"time" : "2016-03-28 14:54:33", "model" : "OSv1 Temperature Sensor", "sid" : 4, "channel" : 2, "battery" : "OK", "temperature_C" : 19.300}

merbanan:
Doesn't look like we have any samples, feel free to disable it.

merbanan:
I re thought it. I think what we need is instead a method of disabling a decoder by default. So when we register it the default state is disabled. But it can be enabled with -R34 or what number i got registered at. That way if you have one of these you just enable it explicitly. Does it make sense ?

zuckschwerdt:
If I understand correctly the `disabled` parameter in the device struct works that way, i.e. it disables a noisy or development device by default.

merbanan:
@zuckschwerdt can you enable it also via -Rx ?

zuckschwerdt:
I assumed so and tested it too. Yes, using `-Rx` on a disabled device works as expected.

rct:
There are a number of devices that are likely to produce false positives as they have no CRC or checksum.  A number of the generic" remotes" that have been added recently have no error checking. 

IIRC, waveman also generates a lot of false positives, but it's not a big problem for me because I enable the 4 devices I need, but that also goes back to before the performance was vastly improved by @vestom

Should devices without error checking be disabled by default?

merbanan:
Devices that are are causing false positives should be disabled by default. Some signals are unique in their own way so they don't need a crc.

rct:
If I wasn't running with -R x -R y -R z, I would be of the opinion that Waveman is causing problems.

merbanan:
Patch welcome.

rct:
I agree that the implementation of the Acurite rain gauge isn't very complete and lacks samples.  

I'm not going to bother to submit a patch for Waveman since I use -R ... -R ...

I'm just trying to note looking forward as more "generic remotes" get implemented using the same chips (EV1527, RT1527, FP1527, HS1527 and HT12E) you might want to consider a class of devices/timings that need to enabled.

homeautomaton:
[...]

PR #354, opened 2016-03-28
Disable Acurite Rain Guage as it produces false positivies of Oregon …

ygator:
…Scientific v1
See #353


rct:
Clearly `return 0` is not the right thing to do.

PR #358, opened 2016-04-01
add -T option to set run duration

panarom:
run rtl_433 for a specified amount of seconds (minimum), then quit.  Useful for discrete (rather than continuous) data gathering

This should fix the build error that @vestom found
I also just tested the build (but only the autoconf version, not the cmake) before opening the PR, just to check


merbanan:
I tested the cmake build system.

PR #360, opened 2016-04-08
missing backtick

0xDA221E:


zuckschwerdt:
Don't change all the line endings -- you switched to DOS/CRLF. The file has UNIX/LF line endings. There are 2 bad lines though, i.e. CRLF. Maybe fix those.

0xDA221E:
Strange, I edited the line endings in the GitHub editor directly...

PR #361, opened 2016-04-08
missing backtick in readme.md, now in LF

0xDA221E:


merbanan:
You have to squash the patch. It looks good otherwise.

PR #368, opened 2016-04-19
introduce functionality to decode the Oregon Scientific SL109H Remote Thermal Hygro Sensor

panarom:


src/devices/oregon_scientific_sl109h.c
```diff
@@ -0,0 +1,158 @@
...
+    "temperature_C",
+    "humidity",
+    NULL
+};
+
+r_device oregon_scientific_sl109h = {
+    .name           = "Oregon Scientific SL109H Remote Thermal Hygro Sensor",
+    .modulation     = OOK_PULSE_PPM_RAW,
+    .short_limit    = 2800/*760*/,
+    .long_limit     = 4400/*1050*/,
```
panarom: when I ran rtl_433 with the analyze flags, I got back the commented values as the estimated limits; but using them failed to produce any resutls. However, I noticed that the esperanza_ews was picking up the sl109h signals, so  I switched the limits to those (not sure if this is just a coïncidence, but they're approximately quadruple the analysis-estimated values), and it started working.  Not sure if there's a bug, or I just don't understand how the analysis works.

merbanan:
Do we have test signals for this transmitter ?

panarom:
I do; should I create a PR on rtl_433_tests?

panarom:
https://github.com/merbanan/rtl_433_tests/pull/91

panarom:
@merbanan I've rebased the PR; anything else I should do before it's ready for merging?

src/devices/oregon_scientific_sl109h.c
ygator: I just ran into that as well.  Looking at rtl_433.c it shows the values are divided by ((float) 1000000 / (float) samp_rate.  Where the sample_rate defaults to DEFAULT_SAMPLE_RATE which is 250000.  So it basically divides the values by 4.  I found a commit for this back in January that says "Switch protocol decoder limits to µs instead of "samples at 250000Hz"".  I wasted a few hours working on a decoder to find out the values were divided by 4.  I ended up just multiplying the values by 4 and moved on.  They should probably change all output of samples to µs.

merbanan:
@panarom no, thanks for the contribution.

merbanan:
@ygator you are correct regarding the samples vs us stuff.

merbanan:
@ygator patch welcome

roger-:
@panarom Sorry to resurrect this thread but I'm looking for a good outdoor humidity sensor -  is the SL109H any good (in terms of accuracy, battery life, etc.)?

Issue #375, opened 2016-05-14
Ambient Weather F007TH interprets negative temperatures incorrectly


meadowface:
Data capture has been added in merbanan/rtl_433_tests#92.

The current code uses a uint32_t when processing instead of an int32_t and also incorrectly assumes a sign bit.

A fix in a pull request to follow.

PR #376, opened 2016-05-14
Fix incorrect handling of negative temperatures in ambient_weather.c

meadowface:
As described in issue #375


zuckschwerdt:
You could use `int`-type as the code is not dependent on a specific width.

meadowface:
Good point.  Thanks.  It was a misguided attempt at keeping the style.  Corrected in the PR.

PR #379, opened 2016-05-19
Add 'battery low' decoding

ovrheat:
- Add 'battery low' decoding on FineOffset WH1080.
- Some cosmetics on source.


merbanan:
You mix a functional change with a cosmetical change. Can you split this into 2 patches? Looks good otherwise.

ovrheat:
That's right, I'll close this...

PR #380, opened 2016-05-23
Added data output (for json/csv/etc.) for Esperanza EWS

jonas-koeritz:
Implemented data output format for Esperanza EWS.

### DIFF
--- a/src/devices/esperanza_ews.c
+++ b/src/devices/esperanza_ews.c
@@ -1,4 +1,6 @@
 #include <bitbuffer.h>
+#include "util.h"
+#include "data.h"
 #include "rtl_433.h"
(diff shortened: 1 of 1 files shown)


merbanan:
Are there tests samples for this sensor ?

jonas-koeritz:
I don't own this sensor. There seems to be one in my neighbourhood and i noticed the ugly output. I can not verify the values.
The humidity seems kind of wrong on the one I receive, it shows 5% when its actually about 95%.
There is also a Prologue sensor in range that show humidity as 10 %.

merbanan:
Ok, well the prologue protocol might have a humidity sensor or not.

PR #381, opened 2016-05-23
Add the ability to execute a program for each decoded message

aabaker:
It is helpful for my use case to be able to execute a program every time a valid message is received, passing it as parameters the values in the message. The program can, for example, add the values received to a database.

This code implements this feature for Posix hosts, I don't have a Dev/test environment to attempt a WIN32 version.

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -21,6 +21,8 @@
  */
 
 #include <stdbool.h>
+#include <sys/types.h>
+#include <sys/wait.h>
 
 #include "rtl-sdr.h"
 #include "rtl_433.h"
@@ -30,6 +32,7 @@
 #include "data.h"
 #include "util.h"
 
+#define ARG_MAX_LEN 256
 
 static int do_exit = 0;
 static int do_exit_async = 0, frequencies = 0, events = 0;
@@ -48,6 +51,7 @@ int debug_output = 0;
 int quiet_mode = 0;
 int utc_mode = 0;
 int overwrite_mode = 0;
+char *child_name=NULL;
 
 typedef enum  {
     CONVERT_NATIVE,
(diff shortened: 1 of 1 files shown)


merbanan:
Can you show how a command line would look and the output of that ?

aabaker:
If I launch rtl_433 with
rtl_433 -q -e echo
then when I get a message it will execute, for example
echo --time "2016-05-23 23:51:06" --model "Weather Sensor THGR122N" --id 81 --channel 1 --battery OK --temperature_C "11.70 C" --humidity "68 %"
I've included the quotes as if this was a command typed at the shell, those values get passed as a single argument with the quotes removed just as a quoted argument does from the shell

Of course running echo isn't very useful, the script I actually call is

#!/usr/bin/env python

import argparse
import sys
import rrdtool

parser = argparse.ArgumentParser(description='Add values to rrdtool databases')
parser.add_argument('--time')
parser.add_argument('--model' )
parser.add_argument('--id', required=True )
parser.add_argument('--channel', required=True )
parser.add_argument('--battery' )
parser.add_argument('--temperature_C' )
parser.add_argument('--humidity' )

args=parser.parse_args()

if (args.id == '230'):
  rrdtool.update('temp2.rrd', 'N:'+args.temperature_C.split()[0])
elif (args.id == '81'):
  rrdtool.update('temp1.rrd', 'N:'+args.temperature_C.split()[0])
  rrdtool.update('humidity.rrd', 'N:'+args.humidity.split()[0])

panarom:
@aabaker one other possibility to consider, because it may be cleaner: use the JSON output format to standard out, then pipe to your program, which will parse the JSON on standard in
(One of the flaws with this suggestion, though, is that JSON output isn't available for every device).

merbanan:
@aabaker I have to go with @panarom on this, use a pipe and parse the json.

rct:
@aabaker once you are already (Python, Perl, whatever) it's pretty much straightforward to read from stdin (unbuffered / linebuffered) and do what ever you want with the data.

Currently I'm updating about a dozen RRDtool archives from a bunch of different sensors and creating daily log files that can be replayed if necessary.  I'm using Perl since I started before the JSON stuff and the regular expression bits were easier for me (old habits).   

If I rewrite in Python, I'll post it.  

There should probably be a "contrib" section of this repo for consumer scripts and stuff (or at least a markdown doc with a link to other repos.)

aabaker:
Once I looked at the JSON output and discovered 

- it is being output as one record per line so there is no need to try to find record boundaries and
- there is an fflush() after each record is output so using pipes won't cause buffering issues

decoding it in python is trivial

``` python
#!/usr/bin/env python
import json
import sys
import rrdtool

while(True):
  line=sys.stdin.readline()
  data=json.loads(line)
  if data['id'] == 81:
    rrdtool.update('temp1.rrd', 'N:'+str(data['temperature_C']))
    rrdtool.update('humidity.rrd', 'N:'+str(data['humidity']))
  elif data['id'] == 230:
    rrdtool.update('temp2.rrd', 'N:'+str(data['temperature_C']))
```

PR #385, opened 2016-05-26
add TFA pool thermometer support

kissandr:


zuckschwerdt:
You also need to change `include/rtl_433.h` (MAX_PROTOCOLS), `include/rtl_433_devices.h` (DECL), and  `README.md`.
And fold the code into a single commit.

PR #389, opened 2016-06-06
Add support for Kedsum Temperature/Humidity sensor

nebulous:
Adds support for http://amzn.to/1Y3DHb6, which was the cheapest temp/humidity sensor I could find on Amazon. Very similar to a couple of other sensors, with frustrating differences in endianness.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -58,7 +58,8 @@
 		DECL(springfield) \
 		DECL(oregon_scientific_sl109h) \
 		DECL(acurite_606) \
-		DECL(tfa_pool_thermometer)
+		DECL(tfa_pool_thermometer) \
+		DECL(kedsum)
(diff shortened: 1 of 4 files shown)


merbanan:
Can you upload some sample signals also? Patch is ok, will merge when I get test signals.

nebulous:
Added in tests PR 98
https://github.com/merbanan/rtl_433_tests/pull/98

mastadook:
Is it possible somehow to get the Temperature in Celsius instead of Fahrenheit?
Any Idea?

zuckschwerdt:
If you just need the output to ber Celsius use the `-C si` parameter. https://github.com/merbanan/rtl_433/blob/master/src/rtl_433.c#L131

mastadook:
Thank you, that worked fine.

PR #395, opened 2016-07-06
Added Steelmate TPMS support

johnjore:


src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+		//Battery voltage is stored as half the mV
+		uint8_t tmpbattery_mV = ~reverse8(bb[i][7]);
+		uint16_t battery_mV = tmpbattery_mV * 2;
+
+		//Checksum is a sum of all the other values
+		uint8_t payload_checksum = ~reverse8(bb[i][8]);
+		uint8_t calculated_checksum = preAmble + ID1 + ID2 + p1 + tempFahrenheit + tmpbattery_mV;
+		if (payload_checksum != calculated_checksum)
+			continue;
+
```
merbanan: The payload calculation must be done first. So we can abort early.

src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+		sprintf(sensorIDhex, "0x%04x", sensorID);
+		
+		//Pressure is stored as twice the PSI
+		uint8_t p1 = ~reverse8(bb[i][5]);
+		double pressurePSI = (double)p1 / 2;
+
+		//Temperature is stored in Fahrenheit. Note that the datasheet claims operational to -40'C, but can only express values from -17.8'C
+		uint8_t tempFahrenheit = ~reverse8(bb[i][6]);
+		double tempCelcius = (double)(tempFahrenheit - 32) * 5 / 9;
+
```
merbanan: If the native representation is Fahrenheit you must use that. There is a command line flag that will auto convert.

src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+		uint8_t ID1 = ~reverse8(bb[i][3]);
+		uint8_t ID2 = ~reverse8(bb[i][4]);
+		uint16_t sensorID = (ID1 << 8) + ID2;
+		char sensorIDhex[7];
+		sprintf(sensorIDhex, "0x%04x", sensorID);
+		
+		//Pressure is stored as twice the PSI
+		uint8_t p1 = ~reverse8(bb[i][5]);
+		double pressurePSI = (double)p1 / 2;
+
```
merbanan: Use float, double is not needed.

src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+		//Payload is inverted Manchester encoded, and reversed MSB/LSB order
+
+		//Preamble
+		uint8_t preAmble = ~reverse8(bb[i][2]);
+
+		//Sensor ID
+		uint8_t ID1 = ~reverse8(bb[i][3]);
+		uint8_t ID2 = ~reverse8(bb[i][4]);
+		uint16_t sensorID = (ID1 << 8) + ID2;
```
merbanan: Declare all needed variables at the start.

src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+
+		//Valid preamble? (Note, the data is still wrong order at this point. Correct pre-amble: 0x00 0x00 0x01)
+		if (bb[i][0] != 0x00 || bb[i][1] != 0x00 || bb[i][2] != 0x7f)
+			continue;
+
+		//Payload is inverted Manchester encoded, and reversed MSB/LSB order
+
+		//Preamble
+		uint8_t preAmble = ~reverse8(bb[i][2]);
+
```
merbanan: The preAmble looks unused, so comment this code so it doesn't get compiled.

src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+#include "pulse_demod.h"
+#include "util.h"
+
+static int steelmate_callback(bitbuffer_t *bitbuffer) {
+	if (debug_output >= 1) {
+		fprintf(stdout, "Steelmate TPMS decoder\n");
+		bitbuffer_print(bitbuffer);
+		fprintf(stdout, "\n");
+	}
+
```
merbanan: Move this block to the other debug section.

src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+		//Checksum is a sum of all the other values
+		uint8_t payload_checksum = ~reverse8(bb[i][8]);
+		uint8_t calculated_checksum = preAmble + ID1 + ID2 + p1 + tempFahrenheit + tmpbattery_mV;
+		if (payload_checksum != calculated_checksum)
+			continue;
+		
+		if (debug_output >= 1) {
+			fprintf(stdout, "Sensor ID: %04x. Pressure: %f PSI. Temperature: %f 'C. Battery %i mV\n", sensorID, pressurePSI, tempCelcius, battery_mV);
+		}
+
```
merbanan: Move this so we only get one debug block.

src/devices/steelmate.c
johnjore: We abort early when checking the first 3 bytes. Payload is calculated after all bytes are inverted and MSB/LSB flipped. Along the way, the values are extracted to calculate the checksum.

src/devices/steelmate.c
johnjore: Ok. I can change that.

src/devices/steelmate.c
johnjore: Ok. I can change that.

src/devices/steelmate.c
johnjore: Ok, I can change that, but prefer to do after checking preamble to avoid declaring variables that are never used?

src/devices/steelmate.c
johnjore: Hm... I would prefer to have an output before checking preamble and checksum as there are still issues capturing packets with the existing settings?

src/devices/steelmate.c
johnjore: Hm... Maybe drop it instead as the output should show the same data.

src/devices/steelmate.c
merbanan: When we try to decode 1 packet, all decoders run. If all decoders have this under debug you would be spammed totally in the logs when enabling debug. Add it under a debug define or something. But it can't active by default. But fix as much as you can and submit a new version and I'll review again.

src/devices/steelmate.c
```diff
@@ -1,48 +1,116 @@
...
+ *
+ * Copyright © 2016 Benjamin Larsson
+ * Copyright © 2016 John Jore
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ */
+
```
merbanan: Can you add links to the specification here?

johnjore: There are no specifications available. It is is an Infinion then the IC is programmable and each vendor decides how to implement. There are 4 ID bits, but only 2 are transmitted. IC can transmit the force excerted on it as wheel spinns, but is not transmitted. IC nativly supports CRC8 but a addative checksum is used instead. I'm guessing temp, pressure and voltage as they "appear" to corrolate wth observed behavour. I dont have any documents that states the chipset outputs in fahrenheit, double PSI's or half mV.

merbanan: Ok, can you document it in the format like I did for the other TPMS ?

johnjore: Can do that. I had it documented as the values were extracted, but can add a blob at the top as well.

johnjore:
Will resubmitt after making a few changes.

PR #396, opened 2016-07-06
Added Steelmate TPMS support

johnjore:
Left the debug session at the top, but commented out. Variables are declared after data packet length and first 3 bytes are validated. Each byte is extracted and stored to calculate the checksum. There are only a couple of additional calculations, spreading them out makes it harder to read / understand in the future. This pull request could be considered an update from the existing stub, but tweaks may be required over the next weeks/months to ensure each sensor is picked-up and more experience is gained.

### DIFF
--- a/src/devices/steelmate.c
+++ b/src/devices/steelmate.c
@@ -1,38 +1,115 @@
 /* Steelmate TPMS FSK protocol
  *
  * Copyright © 2016 Benjamin Larsson
+ * Copyright © 2016 John Jore
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation; either version 2 of the License, or
  * (at your option) any later version.
- */
+ * 
+ * Packet payload: 9 bytes.
+ * Bytes 2 to 9 are inverted Manchester with swaped MSB/LSB:
+ * 
+ *                               0  1  2  3  4  5  6  7  8 
+ *                    [00] {72} 00 00 7f 3c f0 d7 ad 8e fa
+ * After translating            00 00 01 c3 f0 14 4a 8e a0
+ *                              SS SS AA II II PP TT BB CC
+ * S = sync, (0x00)
+ * A = preamble, (0x01)
+ * I = id, 0xc3f0
+ * P = Pressure as double the PSI, 0x14 = 10 PSI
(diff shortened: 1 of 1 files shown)


src/devices/steelmate.c
```diff
@@ -1,48 +1,125 @@
...
+		//Valid preamble? (Note, the data is still wrong order at this point. Correct pre-amble: 0x00 0x00 0x01)
+		if (bb[i][0] != 0x00 || bb[i][1] != 0x00 || bb[i][2] != 0x7f)
+			continue;
+
+		//Payload is inverted Manchester encoded, and reversed MSB/LSB order
+		uint8_t preAmble, ID1, ID2, p1, tempFahrenheit, tmpbattery_mV, payload_checksum, calculated_checksum;
+		uint16_t sensorID, battery_mV;
+		float pressurePSI;
+		char sensorIDhex[7];
+
```
merbanan: Move the declaration of the variables to the start of the block.

src/devices/steelmate.c
```diff
@@ -1,48 +1,125 @@
...
+		//Battery voltage is stored as half the mV
+		tmpbattery_mV = ~reverse8(bb[i][7]);
+		battery_mV = tmpbattery_mV * 2;
+
+		//Checksum is a sum of all the other values
+		payload_checksum = ~reverse8(bb[i][8]);
+		calculated_checksum = preAmble + ID1 + ID2 + p1 + tempFahrenheit + tmpbattery_mV;
+		if (payload_checksum != calculated_checksum)
+			continue;
+
```
merbanan: First extract only the needed data to validate the checksum. Then do the rest of the calculations. That way if the checksum fails we do a minimum of operations.

merbanan:
After the fixes I will merge the code.

src/devices/steelmate.c
johnjore: I really, really don't like splitting them as the flow is lost and makes is much harder to read so will change to something like this instead:
                for (int j=0; j < 7; j++)
                {
                        c_bb[j] = ~reverse8(bb[i][j+2]);
                        calculated_checksum += c_bb[j];
                }
I'll revist in a couple of weeks, as I've run out of time for now.

johnjore:
Closed. Will revist in a few weeks.

Issue #397, opened 2016-07-07
Support for Elro Wireless Chime DB286A


fabian-z:
Device is a simple (and cheap) doorbell, sold as a sender/receiver bundle.
It is unknown if the devices are factory paired, however no pairing is necessary during setup.
Signal log captured a single door bell signal, the sending unit does only send one signal per button press.

Test data merged in: https://github.com/merbanan/rtl_433_tests/pull/107

I plan on trying to add support myself, but I have not yet dived into the rtl_433 codebase. :smile:

fabian-z:
Closed with https://github.com/merbanan/rtl_433/commit/48312e0424442620a1646142a0ef880f648d7bf0

Issue #23, opened 2014-01-05
Help with values for short_limit, long_limit and reset_limit


barrie0482:
Hi Benjamin, 

I am trying to add decoding of data for the Oregon Scientific RMR382 (Sender THN132N) to rtl_433. I am having difficulty understanding how to work out the values for **short_limit**, **long_limit** and **reset_limit**. I have had a go, but I don't think I have the correct values set. I have detailed my values below.

I have also included rtl_433 -a output for an **Oregon Scientific THN132N sender** and rtl_433 -a output for a **Holman Industries iWeather WS5029 sender**.

Any assistance in working out these values would be appreciated.

Thanks.

Cheers, Barrie

```
r_device iweather = {
    /* .id             = */ 9,
    /* .name           = */ "iWeather Weather Station",
    /* .modulation     = */ OOK_PWM_D,
    /* .short_limit    = */ 126,
    /* .long_limit     = */ 248,
    /* .reset_limit    = */ 3000,
    /* .json_callback  = */ &iweather_callback,
};

```

```
Registering protocol[01] Rubicson Temperature Sensor
Registering protocol[02] Prologue Temperature Sensor
Registering protocol[03] Silvercrest Remote Control
Registering protocol[04] ELV EM 1000
Registering protocol[05] ELV WS 2000
Registering protocol[06] Waveman Switch Transmitter
Found 1 device(s):
  0:  Realtek, RTL2838UHIDIR, SN: 00000001

Using device 0: Generic RTL2832U OEM
Found Rafael Micro R820T tuner
Exact sample rate is: 250000.000414 Hz
Sample rate set to 250000.
Sample rate decimation set to 0. 250000->250000
Bit detection level set to 10000.
Tuner gain set to Auto.
Reading samples in async mode...
Tuned to 433920000 Hz.
```

**rtl_433 -a output for an Oregon Scientific THN132N sender**

```
*** signal_start = 5476938, signal_end = 5581042
signal_len = 104104,  pulses = 196
Iteration 1. t: 174    min: 112 (56)    max: 236 (140)    delta 117
Iteration 2. t: 174    min: 112 (56)    max: 236 (140)    delta 0
Pulse coding: Short pulse length 112 - Long pulse length 236

Short distance: 131, long distance: 252, packet distance: 2324

p_limit: 174

[...]

PR #398, opened 2016-07-12
Add initial Elro DB286A support

fabian-z:
Some aspects of my implementation:

* ID is only given as a binary string representation, I currently doubt something more sensible can be extracted from the pattern
* ID string alternates between two values for each sending unit
* Looking at the signal, there are both long/short pulses and long/short pauses afterwards. I chose OOK_PULSE_PWM_PRECISE here.
* The current values for short_limit / long_limit / reset_limit are taken from measuring short pulse length, long pulse length and the length of the pause between data patterns in Audacity. These values allow to receive all 15 repetitions of the data pattern.

Tracking issue: #397 (also referencing signal captures in [merbanan/rtl_433_tests](https://github.com/merbanan/rtl_433_tests))

Looking forward to your feedback.

### DIFF
--- a/include/bitbuffer.h
+++ b/include/bitbuffer.h
@@ -72,6 +72,7 @@ unsigned count_repeats(bitbuffer_t *bits, unsigned row);
 /// Return the row index or -1.
 int bitbuffer_find_repeated_row(bitbuffer_t *bits, unsigned min_repeats, unsigned min_bits);
 
+
 /// Return a single bit from a bitrow at bit_idx position
 static inline uint8_t bitrow_get_bit(const bitrow_t bitrow, unsigned bit_idx)
 {
--- a/include/bitbuffer.h
+++ b/include/bitbuffer.h
@@ -64,10 +64,13 @@ unsigned bitbuffer_search(bitbuffer_t *bitbuffer, unsigned row, unsigned start,
 unsigned bitbuffer_manchester_decode(bitbuffer_t *inbuf, unsigned row, unsigned start,
 				     bitbuffer_t *outbuf, unsigned max);
 
+// Function to compare bitbuffer rows and count repetitions
+int compare_rows(bitbuffer_t *bits, unsigned row_a, unsigned row_b);
+unsigned count_repeats(bitbuffer_t *bits, unsigned row);
+
 /// Find a repeated row that has a minimum count of bits.
 /// Return the row index or -1.
 int bitbuffer_find_repeated_row(bitbuffer_t *bits, unsigned min_repeats, unsigned min_bits);
-
 
 /// Return a single bit from a bitrow at bit_idx position
 static inline uint8_t bitrow_get_bit(const bitrow_t bitrow, unsigned bit_idx)
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -67,6 +67,7 @@
 		DECL(elro_db286a) \
 		DECL(template)
 
+
 typedef struct {
 	char name[256];
 	unsigned int modulation;
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -64,8 +64,8 @@
 		DECL(steelmate) \
 		DECL(schraeder) \
 		DECL(lightwave_rf) \
+		DECL(elro_db286a) \
 		DECL(template)
-
 
 typedef struct {
 	char name[256];
--- a/src/devices/elro_db286a.c
+++ b/src/devices/elro_db286a.c
@@ -1,4 +1,4 @@
-/* Initial doorbell implementation for Elro DB286A devices
+/* Doorbell implementation for Elro DB286A devices
  * 
  * Note that each device seems to have two codes, which alternate
  * for every other button press.
@@ -44,15 +44,15 @@ static int doorbell_db286a_callback(bitbuffer_t *bitbuffer) {
 		return 0;
 	}
 	
+	if (count_repeats(bitbuffer, 1) < DB286A_MINPATTERN) {
+		return 0;
+	}
+	
 	//Get hex string representation of code pattern
 	for (i = 0; i <= DB286A_CODEBYTES; i++) {
 	    idsp += sprintf(idsp, "%02x", b[i]);	
 	}
 	id_string[DB286A_CODECHARS] = '\0';
-
-	if (count_repeats(bitbuffer, 1) < DB286A_MINPATTERN) {
-		return 0;
-	}
 	
 	local_time_str(0, time_str);
 	
--- a/src/devices/elro_db286a.c
+++ b/src/devices/elro_db286a.c
@@ -9,14 +9,14 @@
 #include "data.h"
 #include "util.h"
 
-static int doorbell_db286a_callback(bitbuffer_t *bitbuffer) {
+//33 pulses per data pattern
+#define	DB286A_PULSECOUNT		33
+//One leading zero pulse + 15*33 data pattern - 1 missing trailing zero pulse
+#define	DB286A_TOTALPULSES		DB286A_PULSECOUNT*15
+//Minimum data pattern repetitions (14 is maximum)
+#define	DB286A_MINPATTERN		5
 
-	//33 pulses per data pattern
-	const unsigned pulsecount = 33;
-	//One leading zero pulse + 15*33 data pattern - 1 missing trailing zero pulse 
-	const unsigned totalpulses = 33*15;
-	//Minimum data pattern repetitions (14 is maximum)
-	const unsigned minpattern = 5;
+static int doorbell_db286a_callback(bitbuffer_t *bitbuffer) {
 	
 	char time_str[LOCAL_TIME_BUFLEN];
 	data_t *data;
@@ -24,17 +24,17 @@ static int doorbell_db286a_callback(bitbuffer_t *bitbuffer) {
 	uint8_t *b = bb[0];
 	unsigned bits = bitbuffer->bits_per_row[0];
 
-	if (bits != totalpulses) {
+	if (bits != DB286A_TOTALPULSES) {
 		return 0;
 	}
 	
 	//33 pulses + trailing null byte for C string
 	//Example pattern: 001101111111011000101010011011001
 	
-	char id_string[pulsecount + 1];
+	char id_string[DB286A_PULSECOUNT + 1];
 	char *idsp = id_string;
 	
-	char bitrow_string[totalpulses + 1];
+	char bitrow_string[DB286A_TOTALPULSES + 1];
 	char *brp = bitrow_string;
 	
 	const char *brpcount = bitrow_string;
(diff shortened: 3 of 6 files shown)


zuckschwerdt:
The alternating values might be there to distinguish between individual key presses. Usually seen with IR senders it's called a "repeat bit". Is there any relation between the alternating values?

fabian-z:
I tried looking for a relation between the patterns, but I could not find one yet:

```
001101111111011000101010011011001
011001101011100001111010001011001
```

The only thing that catched my eye is that both end in ```11001```.
An additional data point comes from https://github.com/pimatic/pimatic-homeduino/issues/31, where the user gets a (too short) value of

```
10010011000011100010101001101100
```

It could be possible that all these devices produce codes ending in "11001" (which would be a nice check for signal validity), but we can't be sure without additional data - unfortunately I own only one device right now.

If you see any relation, I would love to get a hint.

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,104 @@
...
+
+static int doorbell_db286a_callback(bitbuffer_t *bitbuffer) {
+
+	//33 pulses per data pattern
+	const unsigned pulsecount = 33;
+	//One leading zero pulse + 15*33 data pattern - 1 missing trailing zero pulse 
+	const unsigned totalpulses = 33*15;
+	//Minimum data pattern repetitions (14 is maximum)
+	const unsigned minpattern = 5;
+
```
merbanan: Add these values as a define instead.

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,104 @@
...
+	}
+	
+	//33 pulses + trailing null byte for C string
+	//Example pattern: 001101111111011000101010011011001
+	
+	char id_string[pulsecount + 1];
+	char *idsp = id_string;
+	
+	char bitrow_string[totalpulses + 1];
+	char *brp = bitrow_string;
```
merbanan: These 2 variables that uses a variable reference might be vla's. Use a define and they will for sure be constants.

include/rtl_433_devices.h
```diff
@@ -64,9 +64,9 @@
 		DECL(steelmate) \
 		DECL(schraeder) \
 		DECL(lightwave_rf) \
+		DECL(elro_db286a) \
 		DECL(template)
 
-
```
merbanan: Stray line removal.

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,104 @@
...
+	
+	//Get first id pattern in transmission
+	strncpy(idsp, bitrow_string+1, pulsecount);
+	id_string[pulsecount] = '\0';
+	
+	//Check if pattern is received at least x times
+	
+	unsigned count = 0;
+	const char *tmp = bitrow_string;
+
```
merbanan: Declare all variables at the start of the block.

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,104 @@
...
+	if (count < minpattern) {
+		return 0;
+	}
+	
+	local_time_str(0, time_str);
+	
+	data = data_make(
+				"time",			"",		DATA_STRING, time_str,
+				"model",		"",		DATA_STRING, "Elro DB286A",
+				"id",			"Code",	DATA_STRING, id_string,
```
merbanan: Use code instead or button id.

merbanan:
I am not really sure why you create a binary string of the button code. Why not just use the hex value ?

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,104 @@
...
+	uint8_t *b = bb[0];
+	unsigned bits = bitbuffer->bits_per_row[0];
+
+	if (bits != DB286A_TOTALPULSES) {
+		return 0;
+	}
+	
+	//33 pulses + trailing null byte for C string
+	//Example pattern: 001101111111011000101010011011001
+
```
merbanan: Can you move this code block below the variable declarations ?

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,104 @@
...
+	
+	while((brpcount = strstr(brpcount, id_string))) {
+	   count++;
+	   brpcount++;
+	}
+	
+	if (count < DB286A_MINPATTERN) {
+		return 0;
+	}
+
```
merbanan: I still don't understand why this is needed. Do you get all the bits on one row ?

src/devices/elro_db286a.c
fabian-z: I tried to declare the processing variables only after making sure that the bitbuffer is going to be processed, but I see your point in readability - I will move it in a new commit.

src/devices/elro_db286a.c
fabian-z: Yes, this is actually the case. I tried to adjust the *_limit parameters in r_device, but I only received either 15 messages (1bit zero message, 14 x complete data pattern, 1x data pattern missing trailing zero pulse) with one row each, or 1 message with all bits on one row (as it is implemented now).

fabian-z:
I will change my implementation to generate hex codes instead of binary strings. Already tried to create decimal values (which gave no satisfying result) - hex seems to be the sensible choice in this case to shorten the id code.
Regarding the row separation issue, I 'd be thankful for any insights on how to tune the parameters.

merbanan:
rtl_433 -a -r tests/elro/01/gfile002.data gives:

Input format: uint8
*** signal_start = 40713, signal_end = 327681
signal_len = 286968,  pulses = 495
Iteration 1. t: 246    min: 122 (255)    max: 370 (240)    delta 50
Iteration 2. t: 246    min: 122 (255)    max: 370 (240)    delta 0
Pulse coding: Short pulse length 122 - Long pulse length 370

Short distance: 127, long distance: 375, packet distance: 1741

p_limit: 246
bitbuffer:: Number of rows: 16 
[00] {1} 00 : 0
[01] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[02] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[03] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[04] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[05] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[06] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[07] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[08] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[09] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[10] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[11] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[12] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[13] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[14] {33} 99 47 85 d3 00 : 10011001 01000111 10000101 11010011 0
[15] {32} 99 47 85 d3 : 10011001 01000111 10000101 11010011 
Test mode file issued 5 packets


So it should be possible to get it working.

fabian-z:
I thought so, too. Unfortunately, I was not able to replicate the results of ```rtl_433 -a```. Is there some way to get the parameters used by this analyzer?
https://github.com/merbanan/rtl_433/issues/23#issuecomment-31602828 provided some guidance, but using the values suggested in the limit variables, I get separated messages instead of separate bitrows.

Regarding using the bitbuffer directly without binary string representations, that would be much easier if the message would separate properly. Otherwise the implementation would have to deal with shifting the byte array by one bit, which would bloat it a lot.

merbanan:
.modulation		= OOK_PULSE_PWM_RAW,
	.short_limit	= 800,
	.long_limit		= 1500*4,
	.reset_limit	= 2000*4,

merbanan:
That gives the same output but inverted bits.

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,83 @@
+/* Initial doorbell implementation for Elro DB286A devices
+ * 
+ * Note that each device seems to have two id patterns, which alternate
+ * for every other button press.
+ * 
+ * Example pattern: 001101111111011000101010011011001
+ */
+
```
merbanan: This is incorrect now, and you need to add a proper copy right header.

include/bitbuffer.h
```diff
@@ -64,11 +64,14 @@ unsigned bitbuffer_search(bitbuffer_t *bitbuffer, unsigned row, unsigned start,
...
 
+// Function to compare bitbuffer rows and count repetitions
+int compare_rows(bitbuffer_t *bits, unsigned row_a, unsigned row_b);
+unsigned count_repeats(bitbuffer_t *bits, unsigned row);
+
 /// Find a repeated row that has a minimum count of bits.
 /// Return the row index or -1.
 int bitbuffer_find_repeated_row(bitbuffer_t *bits, unsigned min_repeats, unsigned min_bits);
 
-
```
merbanan: Stray line removal.

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,83 @@
...
+		return 0;
+	}
+	
+	//Get hex string representation of code pattern
+	for (i = 0; i <= DB286A_CODEBYTES; i++) {
+	    idsp += sprintf(idsp, "%02x", b[i]);	
+	}
+	id_string[DB286A_CODECHARS] = '\0';
+	
+	printf("%s", id_string);
```
merbanan: Remove this printout or add a debug block.

src/devices/elro_db286a.c
```diff
@@ -0,0 +1,81 @@
...
+		return 0;
+	}
+	
+	//Get hex string representation of code pattern
+	for (i = 0; i <= DB286A_CODEBYTES; i++) {
+	    idsp += sprintf(idsp, "%02x", b[i]);	
+	}
+	id_string[DB286A_CODECHARS] = '\0';
+
+	if (count_repeats(bitbuffer, 1) < DB286A_MINPATTERN) {
```
merbanan: Move this block before the get hex representation block, that way we break as early as possible for bad data.

merbanan:
After these last fixes I will merge the code.

src/devices/elro_db286a.c
fabian-z: Sorry, I saw it in the last moment and already force pushed the commit again. You were too fast :smile:

fabian-z:
Thanks for merging and your excellent review!
The demodulator parameters you suggested were very helpful, too.

fabian-z:
I just found a spot that slipped through my attention:
https://github.com/fabian-z/rtl_433/commit/156665e04a4d28a375d842041e5e6d7b07c2b938

It is not crucial but the defined array is longer than needed.
Could you pull in my commit or shall I open a quick PR?

merbanan:
Pr

PR #410, opened 2016-08-08
Add THN802 and PCR800a to oregonscientific.c

timmehrees:
Editted src/devices/oregonscientific.c to include the THN802 Outdoor Temperature Sensor, and the likely Australian version of the PCR800 (I called it PCR800a) with a slightly different ID.


merbanan:
Next time split the functional and cosmetic changes.

zuckschwerdt:
There should be no output of calculated values like temperature_F as there is a Fahrenheit-conversion switch available.

timmehrees:
Thanks for the heads up merbanan, will attempt to do that next time. First time contributing to anything, so still learning the ropes.... Also, i'm not a programmer, I just copied the previous entries, and edited as needed, so I didn't notice the conversion switch. Thanks for all of your work on this, great project!

Issue #406, opened 2016-08-06
Clean up fprintf() usage + auto enable


merbanan:
The code modules under devices/*.c needs a review. All fprintf() output should be to stderr and no fprintf() should be on their own. All fprintf()'s must be under debug output guards.

Some modules are not yet converted to make_data() format. Set the enable flag for those modules so they need to be explicitly enabled via the -R parameter.

Remove all printf() usage.

rct:
> _"no fprintf() should be on their own. All fprintf()'s must be under debug output guards."_ 

This seems a bit incomplete and/or heavy handed -  You're conflating debug output and error reporting.   There are exceptions that should be reported that aren't necessarily structured data and aren't debugging output.

How should exceptions be reported?

How should error reporting be done (when it's desired)?

rct:
See #419 - I think part of  this issue is to have an easy way to get cleanly parseable structured data output.  While clean up is still necessary, rather than just disable all of the unstructured devices, I think a better way is to start identifying which devices output clean structured data, in a manner that can be used by command line option processing.

ludwich66:
Please see my google groups entry.
https://groups.google.com/forum/#!topic/rtl_433/yogNIYaHZOM

merbanan:
By default there should be no printfs now. Some old decoders might use printf but they are disabled by default. Closing issue.

PR #412, opened 2016-08-09
Add the sample code 

wufengyi:
Add sample code to better understand the requirement.
https://github.com/merbanan/rtl_433/issues/406


wufengyi:
Hi,Benjamin

please review the sample code to see if I understand it correctly.
also,I think you should explain more about"All fprintf() output should be to stderr and no fprintf() should be on their own".what does this mean?

Thanks.
Jeffson.

src/devices/new_template.c
```diff
@@ -64,6 +64,6 @@ r_device template = {
 	.long_limit     = (56+33)*4,
 	.reset_limit    = (56+33)*2*4,
 	.json_callback	= &template_callback,
-	.disabled		= 1,
+	.disabled		= 0,//Set the enable flag for those modules so they need to be explicitly enabled via the -R parameter.
 	.fields			= csv_output_fields,
```
merbanan: This one is already ok. You got the meaning inverted. The text is good. Change it back to 1 and keep the text.

src/devices/acurite.c
```diff
@@ -472,10 +472,12 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
 		tempf = acurite_getTemp(bb[4], bb[5]);
 		tempc = fahrenheit2celsius(tempf);
 		humidity = acurite_getHumidity(bb[6]);
-
+//remove all printf usage
+#if 0
 		printf("%s Acurite 5n1 sensor 0x%04X Ch %c, Msg %02x, Wind %d kmph / %0.1f mph, %3.1F C %3.1F F %d %% RH\n",
 		       time_str, sensor_id, channel, message_type,
 		       wind_speed, wind_speedmph, tempc, tempf, humidity);
+#endif
```
merbanan: Change this to a if (debug_output >= 1) {} block. Change the printf() to fprintf(stderr, ...).

src/devices/acurite.c
```diff
@@ -472,10 +472,12 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
...
 	    } else {
 		printf("%s Acurite 5n1 sensor 0x%04X Ch %c, Status %02X, Unknown message type 0x%02x\n",
 			time_str, sensor_id, channel, bb[3], message_type);
```
merbanan: You missed this printf. Add it under a if (debug_output >= 1) {} block.

src/devices/acurite.c
```diff
@@ -595,8 +597,10 @@ static int acurite_986_callback(bitbuffer_t *bitbuf) {
 	}
 
 	if ((status & 1) == 1) {
+            if (debug_output) {//All fprintf()'s must be under debug output guards.
 	    fprintf(stderr, "%s Acurite 986 sensor 0x%04x - %d%c: low battery, status %02x\n",
 		    time_str, sensor_id, sensor_num, sensor_type, status);
+            }
```
merbanan: This block is ok, but use if (debug_output >= 1), and also change the indentation.

src/devices/acurite.c
rct: Where is the rest of this commit?   Why is this being hacked up?

src/devices/acurite.c
rct: Ugh.  No.   So you are going to make it so low battery messages only get seen when debug is enabled?

When and were do you feel exceptions should be reported?

Also I don't see any calls in this commit to change this to use the data printing stuff, so this is just making useful information disappear for no good reason.

rct:
Ok, why are you modifying a device file under a commit named "Add the sample code"?

merbanan:
Device modules that don't use the make_data() api should get the ".disabled = 1" change.

If a device module that uses the make_data() api uses printf, that use should be protected with a debug block.

src/devices/acurite.c
rct: Ok, same question, why is some PR about sample code, going to change the behavior of a currently working device?

I don't think there should be any changes to acurite.c in this PR.   Ideally if the behavior of a device is getting changed it should be pretty explicit.  (especially there hasn't been any releases or tags to make it easier to go back to a known working version.)

rct:
Ok, if you want to enforce that, then lets have a PR that changes all of them at once and only changes that so it can be identified, backed out, etc. 

Having a tag or a release or notice before the change is made would also be a nice courtesy to those who have written device code and are currently using it.

src/devices/acurite.c
merbanan: I didn't see that the printout was sensor data. The only change will be that the device needs to be explicitly enabled.

merbanan: I have a plan on how to go forward towards a release. This is part of it but it will take some time.

rct:
Where and how am I supposed to figure out that a commit named "Add the sample code" is disabling a working device that I am using?

merbanan:
This PR wont be merged in its current form. This is just an initial request for comments.

rct:
That's not really clear.  Some groups put a [WIP] or [DO NOT MERGE] in the PR subject line for experimental stuff.

wufengyi:
thanks for your review effort,I will give you the new code ASAP.

merbanan:
If you run

grep data_make src/devices/*

you will get a group of files, lets call it api2 files. The rest of the files we can call api1 files.

Api1 files should get .disabled =1 set in the r_device struct

Api2 files that uses *printf without a debug block should get a debug block around the fprintf. Right now you have mixed it a bit.

When this project started the device modules used printf to output data. After a while we got the data_make() functions, not all device modules have been converted yet. So now I want all Api1 modules disabled by default and all Api2 device modules to not use fprintf unless you enable debug mode.

khkremer:
This simple "grep" approach will not always work: In the acurite.c file, you will find two groups of devices: One that uses the new approach with data_make(), and all the others that use the old printf approach. When I added support for the 606tx sensor, I added the code to the acurite.c file - because that's where it belongs - and I used the data_make() method to output data. This means that your grep would flag this file as belonging to api2. We need to do this on a per device basis, and not by just looking at files.

wufengyi:
@merbanan 
well,I am a little confused by your comments.let us take a look at new_template.c file for an example.
in this file ,the data_make function was used.so,**this file should belong to api2 files.**
so,depending on your logic , **.disabled =0 should be set in the r_device struct**,right?
but in the ealier review, you said I got the meaning inverted.why was that?
Or for the api2 files,I should not modify the disabled flag,just leave it that way.

merbanan:
new_template.c is an example device module. It is meant as an example for developers to start a new modules from. That is why it should be disabled by default.

For the api2 files,you should not modify the disabled flag,just leave it that way.

wufengyi:
I understood,another question:
there are also some fprintf lines in api1 files,I think I do not need to add debug mode block before fprintf .am I right?

merbanan:
Correct.

PR #418, opened 2016-08-31
Clean up fprintf() usage + auto enable 

wufengyi:
Clean up fprintf() usage + auto enable #406 review request.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -155,7 +155,7 @@ static int acurite_getWindSpeed (uint8_t highbyte, uint8_t lowbyte) {
 	// TODO: sensor does not seem to be in kph, e.g.,
 	// a value of 49 here was registered as 41 kph on base unit
 	// value could be rpm, etc which may need (polynomial) scaling factor??
-	int highbits = ( highbyte & 0x1F) << 3;
+    int highbits = ( highbyte & 0x1F) << 3;
     int lowbits = ( lowbyte & 0x70 ) >> 4;
     int speed = highbits | lowbits;
     return speed;
@@ -176,7 +176,7 @@ static int acurite_getHumidity (uint8_t byte) {
 
 static int acurite_getRainfallCounter (uint8_t hibyte, uint8_t lobyte) {
     // range: 0 to 99.99 in, 0.01 in incr., rolling counter?
-	int raincounter = ((hibyte & 0x7f) << 7) | (lobyte & 0x7F);
+    int raincounter = ((hibyte & 0x7f) << 7) | (lobyte & 0x7F);
     return raincounter;
 }
 
@@ -471,8 +471,8 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
 		wind_speedmph = kmph2mph((float) wind_speed);
 		tempf = acurite_getTemp(bb[4], bb[5]);
 		tempc = fahrenheit2celsius(tempf);
-		humidity = acurite_getHumidity(bb[6]);
-
+		humidity = acurite_getHumidity(bb[6]);	
+		
 		printf("%s Acurite 5n1 sensor 0x%04X Ch %c, Msg %02x, Wind %d kmph / %0.1f mph, %3.1F C %3.1F F %d %% RH\n",
 		       time_str, sensor_id, channel, message_type,
 		       wind_speed, wind_speedmph, tempc, tempf, humidity);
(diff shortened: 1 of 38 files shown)


rct:
This is getting better but still needs work.    A few comments from a quick spot check:

1. You need a **much clearer description** of this pull request, so people can find it when things stop working as they expect.  For example the title should be something extremely clear like "Disable all devices that aren't using make_data()".    Then it would be good in the description to say something like, "any device that doesn't use make_data() will now have to be explicitly enabled by using -R".   
2. From a spot check, a number of your changes to make debug fprintf's require debug_level >= 1 (-D -D) don't make sense and should be deleted.   Requiring a higher debug level should only be necessary for verbose things dumping the bitbuffer.  Reporting checksum/parity errors at debug_level == 1 seems reasonable as long as it's not too verbose. 
3. As a matter of personal preference, I would split the changes into two commits or two PRs, to separate a) Disabling all devices that don't use make_data, and b) reducing debug fprintf output.
4. You've got extraneous white space changes that should not be in the commit.   If you feel there is white space clean-up that should be done, that should be a separate commit, titled appropriately.  This makes it easier for people to read the diffs and figure out what you are doing. (And Why).
5. Are you running the test files when making these changes to see what the effects are?

rct:
Note: If the change is made to disable all devices that don't use make_data() for JSON, CSV, etc.  It should be accompanied by an additional command line argument that will enable all devices when necessary.  A somewhat common use case for this is simple, you've got an unknown device that you want to see if there are any decoders for.  You want all available decoders enabled to see if there is a match.

Note: one of the things holding up the migration of some devices to the new data routines is the need to come to a consensus on some of the additional fields beyond temperature and humidity.  See #351

src/devices/acurite.c
```diff
@@ -471,8 +471,8 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
 		wind_speedmph = kmph2mph((float) wind_speed);
 		tempf = acurite_getTemp(bb[4], bb[5]);
 		tempc = fahrenheit2celsius(tempf);
-		humidity = acurite_getHumidity(bb[6]);
-
+		humidity = acurite_getHumidity(bb[6]);	
+
```
rct: this white space change shouldn't be here.  The above indenting changes (possibly tabs to/from spaces) should also not be in this commit in my opinion.

src/devices/acurite.c
```diff
@@ -674,7 +674,7 @@ static int acurite_606_callback(bitbuffer_t *bitbuf) {
 
     local_time_str(0, time_str);
 
-    if (debug_output > 1) {
+    if (debug_output >= 1) {
```
rct: This is a good use of -D -D (debug_output >= 1).

src/devices/alecto.c
```diff
@@ -89,7 +89,7 @@ static int alectov1_callback(bitbuffer_t *bitbuffer) {
         /* Quit if checksup does not work out */
         if (csum != (bb[1][4] >> 4) || csum2 != (bb[5][4] >> 4)) {
             //fprintf(stdout, "\nAlectoV1 CRC error");
-            if(debug_output) {
+            if(debug_output >= 1) {
```
rct: In my opinion this change should be deleted.   This isn't overly verbose, so it seems appropriate for debug_level == 1.

src/devices/bresser_3ch.c
```diff
@@ -51,7 +51,7 @@ static int bresser_3ch_callback(bitbuffer_t *bitbuffer) {
     b[4] = ~b[4];
 
     if (((b[0] + b[1] + b[2] + b[3] - b[4]) & 0xFF) != 0) {
-        if (debug_output) {
+        if (debug_output >= 1) {
```
rct: Same comment as above, not overly verbose, so should require a higher debug level.

src/devices/lacrosse.c
```diff
@@ -180,10 +180,10 @@ static int lacrossetx_callback(bitbuffer_t *bitbuffer) {
 			// message integrity.
 			if (msg_nybbles[5] != msg_nybbles[8] || 
 			    msg_nybbles[6] != msg_nybbles[9]) {
-				if (debug_output) {
-			    fprintf(stderr,
-				    "LaCrosse TX Sensor %02x, type: %d: message value mismatch int(%3.1f) != %d?\n",
-				    sensor_id, msg_type, msg_value, msg_value_int);
+				if (debug_output >= 1) {
```
rct: No.  shouldn't require -D -D,   you want to be able to see cases like this (where there are message types you haven't seen before), without higher levels of debugging.   Again, not overly verbose.

merbanan:
Merged manually

merbanan:
@rct I added the ability to enable all decoder via the command line.

wufengyi:
@merbanan ,no further suggestions on this pull request?

merbanan:
@wufengyi no its fine.

PR #421, opened 2016-09-03
Add RadioHead ASK device

enavarro222:
This add a device to decode [RadioHead ASK](http://www.airspayce.com/mikem/arduino/RadioHead/classRH__ASK.html#details).

I think it may be "out of the scope" of rtl_433. This is a "generic" decoder for some "home made" devices (based on RadioHead ASK) and not a decoder for a specific device... What do you think about that ?

The only practical problem I see about the bitrate (RH_ASK_SPEED) which is defined by a constant in rtl_433 but may vary from one RadioHead ASK device to an other.

Note thate some code came from RadioHead source (for CRC, and 4to6 decoding) but both project are GPL so I guess there is no issue with that. Maybe references should be more visible ?

### DIFF
--- a/src/devices/radiohead_ask.c
+++ b/src/devices/radiohead_ask.c
@@ -1,0 +1,178 @@
+#include "rtl_433.h"
+#include "pulse_demod.h"
+#include "util.h"
+
+// Note: some code here provide from RadioHead source code.
+// http://www.airspayce.com/mikem/arduino/RadioHead/index.html
+
+// Transmitter speed in bits per seconds
+#define RH_ASK_SPEED 2000
+#define RH_ASK_BIT_LEN (int)1e6/RH_ASK_SPEED
+
+// Maximum message length (including the headers, byte count and FCS) we are willing to support
+// This is pretty arbitrary
+#define RH_ASK_MAX_PAYLOAD_LEN 67
+#define RH_ASK_HEADER_LEN 4
+#ifndef RH_ASK_MAX_MESSAGE_LEN
+ #define RH_ASK_MAX_MESSAGE_LEN (RH_ASK_MAX_PAYLOAD_LEN - RH_ASK_HEADER_LEN - 3)
+#endif
+
+uint8_t payload[RH_ASK_MAX_PAYLOAD_LEN] = {0};
+int data_payload[RH_ASK_MAX_MESSAGE_LEN];
+
+// 4 bit to 6 bit symbol converter table
+// Used to convert the high and low nybbles of the transmitted data
+// into 6 bit symbols for transmission. Each 6-bit symbol has 3 1s and 3 0s 
+// with at most 3 consecutive identical bits
+static uint8_t symbols[] = {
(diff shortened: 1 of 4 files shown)


src/devices/radiohead_ask.c
```diff
@@ -0,0 +1,178 @@
...
+    return ((((uint16_t)data << 8) | hi8 (crc)) ^ (uint8_t)(data >> 4) 
+            ^ ((uint16_t)data << 3));
+}
+
+static int radiohead_ask_callback(bitbuffer_t *bitbuffer) {
+    // Get time
+    char time_str[LOCAL_TIME_BUFLEN];
+    data_t *data;
+    local_time_str(0, time_str);
+
```
merbanan: Declare all needed variables here.

src/devices/radiohead_ask.c
```diff
@@ -0,0 +1,178 @@
...
+        return 0;
+    }
+
+    start_pos += init_pattern_len;
+    // read "bytes" of 12 bit
+    unsigned int pos;
+    uint8_t rxBits[2] = {0};
+    unsigned int nb_bytes=0;
+    uint8_t msg_len = RH_ASK_MAX_MESSAGE_LEN;
+    for(pos = start_pos; pos < len && nb_bytes < msg_len; pos += 12){
```
merbanan: Indent by 4.

src/devices/radiohead_ask.c
```diff
@@ -0,0 +1,178 @@
+#include "rtl_433.h"
+#include "pulse_demod.h"
+#include "util.h"
+
+// Note: some code here provide from RadioHead source code.
+// http://www.airspayce.com/mikem/arduino/RadioHead/index.html
```
merbanan: When the crc is fixed only the 6to4 code is from the RadioHead code. So that can then this note can be changed to that.

merbanan:
The code/feature is clean enough to be accepted. Fix the requested changes and the code will be merged.

enavarro222:
Ok thanks;

For the CRC I got different results with the ones from util.c. So for simplicity I just copy paste the one from RH. I have to check further why I got different values.

enavarro222:
Here it is.

CRC was a bit tricky for me because the RadioHead ASK function it's quite not trivial implementation of CRC-16-CCITT ! (see here https://github.com/Yveaux/RadioHead/blob/master/RadioHead/RHCRC.cpp#L79)

merbanan:
I can look at the crc. There are only that many ways to do a crc.

enavarro222:
For CRC, with some "brut force" testing I get that it match with a `~crc16(payload, msg_len-2, 0x8408, 0xFFFF);`

According to https://en.wikipedia.org/wiki/Cyclic_redundancy_check#Standards_and_common_use it indeed match with a CRC-16-CCITT...

enavarro222:
@merbanan is it ok or did you see something to improve before to merge ?

merbanan:
The only thing needing improvements are my reaction skills.

enavarro222:
@merbanan thanks ! :)
and concerning your reaction skills it's hard to blame you on that, you are more than reactive most of the time !

PR #430, opened 2016-09-29
Minor fixes for auto level as default

vestom:
Making auto level (-l 0) the default broke the old analyzer (-a).
Minor documentation fixes.
Corrected documented level range to [0-16384] as maximum level for a unit phasor is 128*128. Anything above 16384 will be ripple due to saturation/overdrive (phasor outside unit circle). This is the reason 8000 has worked so well as a fixed detection threshold :-)


merbanan:
Thanks.

merbanan:
Why do you use my old code ? :)

vestom:
To record and save samples :-)

PR #431, opened 2016-10-03
Acurite 6045 lightning prelim. and 609 cleanup.

rct:
* Acurite 6045M Lightning detector preliminary decoding.  I don't have a display, just the sensor, hoping someone who has the display can assist.  Currently working:  humidity and count of lighting strikes detected.  The sensor is supposed to report temperature but I haven't  seen anything in the message that looks reasonable.

* Acurite 609TXC:
 * low battery reporting
 * report device ID, 1 byte which changes on each power up.
 * Enabled device since it now uses structured data.

* Fixed misnamed checksum function, was named CRC but the implementation is a checksum.

* Fixed minor rain counter bug if both the old and new 5-n-1 weather decoders were enabled.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -11,6 +11,7 @@
  * - 609TXC "TH" temperature and humidity sensor (609A1TX)
  * - Acurite 986 Refrigerator / Freezer Thermometer
  * - Acurite 606TX temperature sesor
+ * - Acurite 6045M Lightning Detector (preliminary)
  */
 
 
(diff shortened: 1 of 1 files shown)


src/devices/acurite.c
```diff
@@ -512,6 +534,30 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
...
+		   time_str, sensor_id, channel, message_type, temp, humidity, strike_count, strike_distance);
+
+	    // FIXME Temporarily dump message data until the decoding improves.
+	    // Include parity indicator.
+	    for (int i=0; i < browlen; i++) {
+		char pc;
+		pc = byteParity(bb[i]) == 0 ? ' ' : '*';
+		fprintf(stdout, " %02x%c", bb[i], pc);
+	    }
+	    printf("\n");
```
merbanan: Use only fprintf().

PR #434, opened 2016-10-19
add Maverick ET-73x BBQ Sensor

gismo2004:
Up to now, there is no support for Maverick ET-73x BBQ sensors. 

The implementation is based on a forum discussion:
https://forums.adafruit.com/viewtopic.php?f=8&t=25414

The reason for the bad code style: I'm no coder! Sorry

cheers

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -65,8 +65,8 @@
 		DECL(schraeder) \
 		DECL(lightwave_rf) \
 		DECL(elro_db286a) \
-		DECL(efergy_optical) \
-		DECL(hondaremote) \
+                DECL(efergy_optical) \
+                DECL(hondaremote) \
 		DECL(template) \
 		DECL(fineoffset_XC0400) \
 		DECL(radiohead_ask) \
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -65,8 +65,8 @@
 		DECL(schraeder) \
 		DECL(lightwave_rf) \
 		DECL(elro_db286a) \
-                DECL(efergy_optical) \
-                DECL(hondaremote) \
+		DECL(efergy_optical) \
+		DECL(hondaremote) \
 		DECL(template) \
 		DECL(fineoffset_XC0400) \
 		DECL(radiohead_ask) \
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -65,14 +65,15 @@
 		DECL(schraeder) \
 		DECL(lightwave_rf) \
 		DECL(elro_db286a) \
-                DECL(efergy_optical) \
-                DECL(hondaremote) \
+        DECL(efergy_optical) \
+        DECL(hondaremote) \
 		DECL(template) \
 		DECL(fineoffset_XC0400) \
 		DECL(radiohead_ask) \
-		DECL(kerui)                                                    \
-        DECL(maverick_et73x)                                           \
-		DECL(fineoffset_wh1050)
+		DECL(kerui) \
+		DECL(fineoffset_wh1050) \
+		DECL(honeywell)\
+		DECL(maverick_et73x)
 
 
 typedef struct {
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -70,9 +70,9 @@
 		DECL(template) \
 		DECL(fineoffset_XC0400) \
 		DECL(radiohead_ask) \
-		DECL(kerui) \
-		DECL(fineoffset_wh1050) \
-		DECL(honeywell)
+		DECL(kerui)                                                    \
+        DECL(maverick_et73x)                                           \
+		DECL(fineoffset_wh1050)
 
 
 typedef struct {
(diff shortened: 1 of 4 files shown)


include/rtl_433_devices.h
```diff
@@ -65,14 +65,15 @@
 		DECL(schraeder) \
 		DECL(lightwave_rf) \
 		DECL(elro_db286a) \
-                DECL(efergy_optical) \
-                DECL(hondaremote) \
+		DECL(efergy_optical) \
+		DECL(hondaremote) \
 		DECL(template) \
```
merbanan: Remove this change

src/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ add_executable(rtl_433
 	pulse_detect.c
 	rtl_433.c
 	util.c
-        devices/acurite.c
+	devices/acurite.c
 	devices/alecto.c
```
merbanan: Remove this change

src/devices/maverick_et73x.c
```diff
@@ -0,0 +1,253 @@
+//for further information please referre to https://forums.adafruit.com/viewtopic.php?f=8&t=25414
+
```
merbanan: Add proper copyright header

src/devices/maverick_et73x.c
```diff
@@ -0,0 +1,253 @@
...
+#define MAV_MESSAGE_LENGTH 104
+#define TEMPERATURE_START_POSITION_S1 8
+#define TEMPERATURE_START_POSITION_S2 13
+#define TEMPERATURE_BIT_COUNT 5
+
+static unsigned int msg_converted[26];
+static char msg_hex_combined[26];
+static time_t time_now_t, last_time_t;
+static int32_t session_id;
+
```
merbanan: Static variables are not allowed, place them where they are needed.

src/devices/maverick_et73x.c
```diff
@@ -0,0 +1,253 @@
...
+    if (param=='6')
+        retval = 1;
+    if (param=='9')
+        retval = 2;
+    if (param=='a' || param=='A')
+        retval = 3;
+
+    return retval;
+}
+
```
merbanan: Replace this with a lookup table.

src/devices/maverick_et73x.c
```diff
@@ -0,0 +1,253 @@
...
+    //check for correct header (0xAA9995)
+    if((bitbuffer->bb[0][0] != 0xAA || bitbuffer->bb[0][0] != 0xaa ) || bitbuffer->bb[0][1] != 0x99 || bitbuffer->bb[0][2] != 0x95)
+        return 0;
+    
+    //convert hex values into quardinary values
+    convert_bitbuffer(bitbuffer);
+    
+    //checksum is used to represent a session. This means, we get a new session_id if a reset or battery exchange is done. 
+    chk_xor = (calculate_checksum(checksum_data()) & 0xffff) ^ checksum_received();
+
```
merbanan: If the checksum fail you are supposed to reject the buffer.

src/devices/maverick_et73x.c
```diff
@@ -0,0 +1,253 @@
...
+    local_time_str(0, time_str);
+
+    if(debug_output)
+        fprintf(stderr, "checksum xor: %x\n", chk_xor);
+
+    data = data_make("time",           "",                      DATA_STRING,                         time_str,
+                     "id",             "Session_ID",            DATA_INT,                            chk_xor,
+                     "temperature_C1", "TemperatureSensor1",    DATA_FORMAT, "%.02f C", DATA_DOUBLE, get_temperature(TEMPERATURE_START_POSITION_S1),
+                     "temperature_C2", "TemperatureSensor2",    DATA_FORMAT, "%.02f C", DATA_DOUBLE, get_temperature(TEMPERATURE_START_POSITION_S2),
+                     "status",         "Status",                DATA_STRING,                         dev_state,
```
merbanan: Here you should add a crc member to inform that the message was received with crc.

merbanan:
Looks good. A few things to handle and you need to send some test signals.

src/devices/maverick_et73x.c
```diff
@@ -0,0 +1,253 @@
...
+        //data bit at current position is "1"
+        //do XOR with mask
+        if((data >> i) & 0x01)
+            csum ^= mask;
+        
+        mask = shiftreg(mask);
+    }
+    return csum;
+}
+
```
merbanan: All this checksum stuff looks overly complicated. What are you actually doing ?

src/devices/maverick_et73x.c
rct: What is the the point of adding a static pseudo "data" element?  Messages that are invalid (Bad CRC, Checksum, etc) shouldn't be output, so there will never be anything but CRC OK.   

The fact that the protocol uses CRC (vs checksum, parity, nothing) isn't a property of the message.  It's a property of the protocol/device.  

If you want consumers to be able to know/make decisions based on device/protocol properties, then there should probably be an initial message that contains the device/protocol properties.  In addition to the error detection type, it could emit, resolution, min/max, that could be used in filtering bad data, etc.   Also, it would be useful to know which devices have a fixed ID, and which are ephemeral (new at every battery change).

src/devices/maverick_et73x.c
merbanan: Add a table where the indexes are the values you need.

quart_tab[10] = {0,0,0,0,0,1,0,0,2,3} should do it.

merbanan:
Merged.

PR #436, opened 2016-10-28
added kv/json capabilities to acurite_txr

akhenakh:
also handling battery state

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -362,7 +362,7 @@ static int acurite_th_callback(bitbuffer_t *bitbuf) {
 		     "battery",		"",		DATA_STRING,	battery_low ? "LOW" : "OK",
 		     "status",		"",		DATA_INT,	status,
 		     "temperature_C", 	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, tempc,
-		     "humidity",	"Humidity",	DATA_INT,	humidity,
+		     "humidity",        "Humidity",	DATA_INT,	humidity,
 		     NULL);
 
 	data_acquired_handler(data);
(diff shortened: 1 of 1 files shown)


akhenakh:
just realized it's sort of a duplicate of https://github.com/merbanan/rtl_433/pull/351

rct:
Yes, there was a lot of discussion about data representaiton in #351.   I think some consensus was on some of the points.

src/devices/acurite.c
```diff
@@ -468,17 +470,23 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
...
+                "channel",  		"",     	DATA_STRING, 	&channel_str,
+                "temperature_C", 	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, tempc,
+                "humidity",		"Humidity",	DATA_INT,	humidity,
+                "battery",        	"Battery",    	DATA_STRING, 	battery_low ? "LOW" : "OK",
+
+                NULL);
+
+        data_acquired_handler(data);
 
+        return 1;
```
rct: No, this is wrong.   It shouldn't just return here.   

1. There are possibly still unprocessed rows in the the bitbuffer.
2. IIRC the return value is supposed to be the number of events processed, though I don't remember if anything is doing anything with this. 

Also, the indentation is now wrong.

src/devices/acurite.c
```diff
@@ -468,17 +470,23 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
...
+            battery_low = sensor_status >>7;
+		
+        data = data_make(
+                "time",			"",		DATA_STRING,	time_str,
+                "model",		"",		DATA_STRING,	"Acurite tower sensor",
+                "id",			"",		DATA_INT,	sensor_id,
+                "channel",  		"",     	DATA_STRING, 	&channel_str,
+                "temperature_C", 	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, tempc,
+                "humidity",		"Humidity",	DATA_INT,	humidity,
+                "battery",        	"Battery",    	DATA_STRING, 	battery_low ? "LOW" : "OK",
```
rct: See @vestom's comment in #351, this should probably be an int, or something more boolean than another string to parse.

rct: Also, per some of the discussion we had, I'd would output the raw status byte (in addition to battery) in case there are other status bits/message type indications that aren't being parsed yet.   This allows a consumer with more acurite specific implementation to do something with it.

akhenakh: changed `battery` for an int, (note it was copied from acurite_th_callback)
and added `status` raw value as an int

src/devices/acurite.c
akhenakh: fixed

akhenakh:
the json is now outputted 3 times for the same event (not every time)
relates to late return change, is it supposed to be that way ?

src/devices/acurite.c
rct: Ok, I'll fix  _th_ (609) to normalize battery next time I submit changes for acurite.c

merbanan:
Looks ok.

panarom:
@rct thanks for doing this. I never managed to get around to 351 ( @merbanan should I go ahead and close it?)

merbanan:
@panarom is the issue fixed? Then we can close if not lets keep it up.

PR #441, opened 2016-11-08
Correct negative temperatures

tvoverbeek:
Negative temperatures were reported incorrectly. In a signed short
(16-bit) the 12-bit temperature is always positive.
This sensor is also used in weather stations.

### DIFF
--- a/src/devices/tfa_pool_thermometer.c
+++ b/src/devices/tfa_pool_thermometer.c
@@ -18,6 +18,7 @@ static int pool_temperature_sensor_callback(bitbuffer_t *bitbuffer) {
 	char time_str[LOCAL_TIME_BUFLEN];
     local_time_str(0, time_str);
 	int i,device,channel;
+	int iTemp;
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Using signed short in the intermediary step uTemp won't do anything. The compiler will turn the calculation into int anyway. You might consider using int right away. Otherwise you could shift by 12 use signed short to recognise the sign-bit and shift left by 4 afterwards. But that will confuse a casual reader for sure.

tvoverbeek:
Agree with zuckschwerdt. Updated wit using int for the intermediate temperature.

zuckschwerdt:
It is preferred that you squash your commits. It's easier to pull if each topic has a single commit only. Force pushing a branch with an open pull request works fine on Github.

tvoverbeek:
Commits squashed into 1. OK?

merbanan:
I trust this to give the correct values.

PR #447, opened 2016-11-18
Improved hideki.c

Matthies:
... to detect data from the wind/temperature and the rain sensor. Changed pulse_demod_clock_bits to allow a missing short gap at the end of the message which is needed for the new sensor types.


Matthies:
Do you need more data files? Ludewich66 uploaded several in the test repository.

PR #450, opened 2016-12-02
Update pulse_detect.c

gustep12:
Counting down in the gaps histogram, from longest to shorter detected gaps, should help to make the PPM_RAW pulse detection more robust. If there are accidentally some too-short gaps in the transmission, they will now be ignored rather than used for parsing the transmission.

A more detailed description and sample transmissions which get fixed by this change are [here.](https://groups.google.com/forum/#!topic/rtl_433/4ICr3h6ZE-M)


merbanan:
Hopefully this makes things better.

gustep12:
Thank you. Another strategy would be to sort the gap bins number of counts, then set the .short_limit as the average gap length of the two highest counting bins, and .long_limit just above the combined maximum of these two highest counting bins. This should be even more robust.

PR #451, opened 2016-12-02
Update acurite.c

gustep12:
This sanity check was too stringent and kept failing on my Acurite 606TX sensor. Full description of the issue and sample data [here.](https://groups.google.com/forum/#!topic/rtl_433/4ICr3h6ZE-M)

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -771,7 +771,7 @@ static int acurite_606_callback(bitbuffer_t *bitbuf) {
       return 0;
 
     // do some basic checking to make sure we have a valid data record
-    if ((bb[0][0] == 0) && (bb[1][4] == 0) && (bb[7][0] == 0x00) && ((bb[1][1] & 0x70) == 0)) {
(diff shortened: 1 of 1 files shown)


merbanan:
Hi, as there is a proper checksum check after the sanity check the change is fine. But there is a bug in the code I would like you to fix. After the data_acquired_handler(data); line there should be a return 1; line. Can you add this as a second commit and I will merge the code.

rct:
Yes it should return after a successful decode.

FWIW, while we are talking about the acurite 606 handler, only the first row of the bitbuffer is processed.  I don't know if the message repeats fall in the same bitbuffer or not given the timing changes.  If they do fall in the same bitbuffer, and the first iteration of the message is invalid, the whole message will be discarded.  There are a couple of handlers that have this somewhat lazy (in my opinion) processing that should be fixed to make things more robust.

(I don't own an 606. I was just checking what changed, since I own a  bunch of the other acurite devices.)

PR #458, opened 2016-12-11
Added support for RF-tech 433 remote temperature sensor.

hackaroni:
Support for RF-tech/INFRA remote temperature sensor.  See pull request in rtl_433_tests for image of sensor and some test data.


hackaroni:
Ok, I'll make a new commit/pull request with the two changes suggested above.

PR #462, opened 2017-01-03
Added support for LaCrosse temperature/humidity sensor TX141TH-Bv2

aromring:
The LaCrosse sensor TX141TH-Bv2 is used by a quite popular new Color
Forecast Station (model C85845) among other products. It uses a novel
(for LaCrosse) protocol that is unlike other protocols created by this
company. Hence, it deserves a callback function in rtl_433,, doesn’t
it? :)


include/rtl_433_devices.h
```diff
@@ -65,16 +65,17 @@
 		DECL(schraeder) \
 		DECL(lightwave_rf) \
 		DECL(elro_db286a) \
-                DECL(efergy_optical) \
-                DECL(hondaremote) \
+        DECL(efergy_optical) \
+        DECL(hondaremote) \
 		DECL(template) \
```
merbanan: Stray change.

include/rtl_433_devices.h
```diff
@@ -65,16 +65,17 @@
...
 		DECL(fineoffset_XC0400) \
 		DECL(radiohead_ask) \
 		DECL(kerui) \
 		DECL(fineoffset_wh1050) \
 		DECL(honeywell) \
 		DECL(maverick_et73x) \
-		DECL(rftech)
+		DECL(rftech) \
+        DECL(lacrosse_TX141TH_Bv2)
```
merbanan: Use tabs.

src/devices/lacrosse_TX141TH_Bv2.c
```diff
@@ -0,0 +1,230 @@
...
+    if (0==id || 0==humidity || humidity > 100 || temp_c < -40.0 || temp_c > 60.0) {
+        if (debug_output) {
+            fprintf(stderr, "LaCrosse TX141TH-Bv2 data error\n");
+        }
+        return 0;
+    }
+
+    data = data_make("time",    "Date and time", DATA_STRING,    time_str,
+                     "temperature_C", "Temperature in deg C", DATA_FORMAT, "%.2f C", DATA_DOUBLE, temp_c,
+                     "temperature_F", "Temperature in deg F", DATA_FORMAT, "%.2f F", DATA_DOUBLE, 9.0*temp_c/5.0+32.0,
```
merbanan: Only report the hardware native temperature.

src/devices/lacrosse_TX141TH_Bv2.c
```diff
@@ -0,0 +1,230 @@
...
+static int lacrosse_tx141th_bv2_callback(bitbuffer_t *bitbuffer) {
+	bitrow_t *bb = bitbuffer->bb;
+	data_t *data;
+	char time_str[LOCAL_TIME_BUFLEN];
+    local_time_str(0, time_str);
+	int i,j,k,nbytes,npacket,kmax;
+    uint8_t id=0,status=0,battery_low=0,test=0,humidity=0,maxcount;
+    uint16_t temp_raw=0;
+    float temp_c=0.0;
+    data_and_count dnc[LACROSSE_TX141TH_PACKETCOUNT];
```
merbanan: data_and_count dnc[LACROSSE_TX141TH_PACKETCOUNT] = {0}; should remove the need for a memset.

src/devices/lacrosse_TX141TH_Bv2.c
```diff
@@ -0,0 +1,230 @@
...
+    kmax=0;
+    maxcount=0;
+    if(npacket>1) {
+        for(k=0;k<npacket;++k) {
+            if(dnc[k].count>maxcount) {
+                maxcount=dnc[k].count;
+                kmax=k;
+            }
+        }
+    }
```
merbanan: Row 130 to 171 should be removable if we can figure out the crc. Can you post a few bitstream dumps ?

merbanan:
Looks good, makefile.am needs an amendment also. And sample signals also.

src/devices/lacrosse_TX141TH_Bv2.c
aromring: There are a few dumps reported in the comments:
0x87 0x02 0x3c 0x3b 0xe1
0x87 0x02 0x7d 0x37 0xe1
0x87 0x01 0xc3 0x31 0xd8
0x87 0x02 0x28 0x37 0xd8
but I'll post more after coming back home tonight. 
Even with CRC figure out, I'd still leave this code in. First, I've done experiments with the sensor into which I've purposely put spent batteries. This made the signal very weak, which also models large distances, obstacles like reinforced concrete walls, etc. The worst burst I've got had 7 unique packets: 6 counts of the correct data and 6 bogus ones with a count of 1 each. That's signal quality of 50%. Hence, there is a reason why LaCrosse sends 12 packets in a single burst and it's a good idea to take full advantage of this. Besides, in the very first 15 packets I've ever recorded there were as many as 2 (!) instances of identical CRCs, which I've listed above.  It's a pretty unreliable CRC, if you ask me, if the last byte is a CRC at all. 
Second, the impact  of mode calculations on performance is quite minimal and it's good to see the measure of signal quality anyway. One might want to move the sensor into a better spot, for example.

aromring:
All right, here are the promised examples of packets (4 data bytes plus 5th as a presumed checksum):
87	01	C3	31	D8
87	01	F2	25	AE
87	02	20	35	53
87	02	28	37	D8
87	02	32	39	CF
87	02	3C	3B	E1
87	02	47	3C	F8
87	02	50	3C	BF
87	02	59	3C	AB
87	02	60	3B	CC
87	02	67	3A	DB
87	02	6D	3A	05
87	02	73	39	7B
87	02	78	38	17
87	02	7D	37	E1
87	02	C7	1E	28
87	02	CA	1E	15
87	02	D1	1E	29
43	02	50	49	9e
43	42	50	49	7e
43	02	51	4a	f5
43	02	50	4a	b3
43	02	51	49	d8

Sample signals are attached.

[RawSignalData.zip](https://github.com/merbanan/rtl_433/files/684010/RawSignalData.zip)

src/devices/lacrosse_TX141TH_Bv2.c
```diff
@@ -0,0 +1,227 @@
...
+        if (debug_output) {
+            fprintf(stderr, "LaCrosse TX141TH-Bv2 data error\n");
+        }
+        return 0;
+    }
+
+    data = data_make("time",    "Date and time", DATA_STRING,    time_str,
+                     "temperature", "Temperature in deg F", DATA_FORMAT, "%.2f F", DATA_DOUBLE, temp_f,
+                     "humidity",    "Humidity", DATA_FORMAT, "%u %%", DATA_INT, humidity,
+                     "quality", "Signal quality",   DATA_FORMAT, "%.1f %%", DATA_DOUBLE, 100.0*((float)dnc[kmax].count)/LACROSSE_TX141TH_PACKETCOUNT,
```
merbanan: I missed this also. The output from the decoder is only to have parameters from the bitstream and static info data about the model. I like the idea though but I don't really know how to add it properly.

merbanan:
43	02	51	4a	f5
43	02	50	4a	b3

That one is important. As it shows that it is not a checksum but a crc.

aromring:
I've just noticed that among 23 packets the "D8" CRC repeats as many as three (!) times:

87 01 C3 31 D8
87 02 28 37 D8
43 02 51 49 d8

A well designed CRC should not have repetition rate so high. It makes me wonder why LaCrosse developed this, I think, custom method instead of using one of the established algorithms.

src/devices/lacrosse_TX141TH_Bv2.c
merbanan: The quality property is not part of the bitstream, thus by the current api it can not be there.

aromring: Hmm... In my humble opinion, the quality factor is a part of the bitstream - it has been directly derived from it. And it's very useful to the end user.
But - let me know if you want me to remove it.

merbanan: Well what unit has the quality factor? Right now it is usable for you with this sensor. How do you apply the same to all the others and make sure the quality factor has the same meaning? These are not questions that can be answered easily thus I'd prefer you to remove it for the initial commit. If we can come up with a quality metric that can be used for all decoder then we can add it outside the decoders. Like the timestamp, but right now every decoder reimplements it because I'm to lazy to fix it.

zuckschwerdt:
It's a reverse Galois LFSR with bytes reflected, gen 31 key f4 ;) This is used in at least three other decoders. Could be a hardware feature.

zuckschwerdt:
Checksum added with 8f4e817

PR #464, opened 2017-01-07
akhan-100F14 catches things it doesn't recognise fully

klattimer:
This prevents the code from hitting a dead end for devices which are similar, the signal is left up for grabs when it doesn't fully understand what's been received.

### DIFF
--- a/src/devices/akhan_100F14.c
+++ b/src/devices/akhan_100F14.c
@@ -48,19 +48,10 @@ static int akhan_rke_callback(bitbuffer_t *bitbuffer) {
 									"data",		"Data (4bit)",	DATA_STRING,	CMD,
 									NULL);
 
-		} else {
-			data = data_make(	"time",		"",				DATA_STRING,	time_str,
-									"device",	"",				DATA_STRING,	"Akhan 100F14 remote keyless entry",
-									"id",			"ID (20bit)",	DATA_FORMAT, 	"0x%x", 	DATA_INT, ID,
-									"data",		"Data (4bit)",	DATA_FORMAT, 	"0x%x", 	DATA_INT, dataBits,
-									"other",		"Attention",	DATA_STRING,	"The data received is not used by the akham keyfob. This might be another device using a HS1527 OTP encoder with the same timing.",
-									NULL);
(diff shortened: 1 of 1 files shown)


merbanan:
I will just assume this is correct. Can you send samples for this decoder, there seem to be missing.

klattimer:
I'm still working on that bit - I've got a cheap doorbell I'm currently working on writing decoders for and this was getting in the way. I'm still trying to work out precisely how to create new devices from the recorded data - I'm recording a short and long pulse and bits with rtl_433 -a and can see the number of rows and the bits per row - but building a new device around just that isn't as easy as I first thought... Picking the modulation and setting the parameters there seems tricky, I think more hints in the analysis output for which modulation is possible... 

My main aim here is eventually to be able to take a JSON package from this and re-send the signal with an arduino and OOK radio attached.

merbanan:
Use -A it is a better analyzer.

And for arduino:

https://github.com/merbanan/OOK_gen

merbanan:
Basically you need another tool and reimplement everything in reverse. You need to extend the json with signal modulation parameters and bit packing transforms. Generalizing might not be worth it.

klattimer:
The arduino code there will be very useful, I also have a bit of code for doing oregon scientific sensors which is extremely useful too... I'm also using a HopeRF RFM12B to send FSK signals on 868Mhz (I need to get rtl433 working on this frequency too).

I was intending to extend the json to include a brand and model, allowing us to better catalogue devices then simply re-use the structs from the supported devices, maybe even a device catalogue online where the data can be keyed into a database and shared before it's integrated in git?

I suppose I'm proposing somekind of online library of reverse engineered RF signals. We'd collect from users a device photo, signal samples, a link to a manufacturer page or distributor page, brand and model numbers etc... From there we can process the signals determine the necessary values for the demodulator/decoder and generate device templates in C from there. Obviously multiple signals would be required but that shouldn't be too hard if model and brand is included.

There's already 70 devices supported here, I think another thousand must be out there easily within our reach with some epic crowd sourcing... I have at least another 4 to add and I've hardly got started yet!

PR #467, opened 2017-01-16
Support for Acurite 00275rm and 00276rm sensors

jonrifkin:
I've added support for the Acurite 00275rm and 00276rm sensors.  If you need test data, I can upload that as well.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -751,102 +751,102 @@ static int acurite_606_callback(bitbuffer_t *bitbuf) {
 
 
 static int acurite_00275rm_callback(bitbuffer_t *bitbuf) {
-	int browlen, battery_low, id, model, valid = 0;
-	uint8_t *bb;
-	data_t *data;
-	char *model1 = "00275rm", *model2 = "00276rm";
-	float tempc, ptempc, tempf, ptempf;
-	uint8_t probe, humidity, phumidity, water;
+    int browlen, battery_low, id, model, valid = 0;
+    uint8_t *bb;
+    data_t *data;
+    char *model1 = "00275rm", *model2 = "00276rm";
+    float tempc, ptempc, tempf, ptempf;
+    uint8_t probe, humidity, phumidity, water;
 
-	local_time_str(0, time_str);
+    local_time_str(0, time_str);
 
-	if (debug_output > 1) {
-		fprintf(stderr,"acurite_00275rm\n");
-		bitbuffer_print(bitbuf);
-	}
+    if (debug_output > 1) {
+        fprintf(stderr,"acurite_00275rm\n");
+        bitbuffer_print(bitbuf);
+    }
 
-	for (uint16_t brow = 0; brow < bitbuf->num_rows; ++brow) {
-		if (bitbuf->bits_per_row[brow] != 88) continue;
-		bb = bitbuf->bb[brow];
-		if (debug_output) {
-		    fprintf(stderr,"acurite_00275rm: ");
-		    for (int i=0; i<11; i++) fprintf(stderr," %02x",bb[i]);
-		    fprintf(stderr,"\n");
+    for (uint16_t brow = 0; brow < bitbuf->num_rows; ++brow) {
+        if (bitbuf->bits_per_row[brow] != 88) continue;
+        bb = bitbuf->bb[brow];
+        if (debug_output) {
+            fprintf(stderr,"acurite_00275rm: ");
+            for (int i=0; i<11; i++) fprintf(stderr," %02x",bb[i]);
+            fprintf(stderr,"\n");
                 }
-		id = (bb[0]<<16) | (bb[1]<<8) | bb[3];
-		battery_low = (bb[2] & 0x40)==0;
-		model       = (bb[2] & 1);
-		tempc       = 0.1 * ( (bb[4]<<4) | (bb[5]>>4) ) - 100;
-		probe       = bb[5] & 3;
-		tempf       = 1.8*tempc + 32;
-		humidity    = ((bb[6] & 0x1f) << 2) | (bb[7] >> 6);
-		//  No probe
-		if (probe==0) {
-			data = data_make(
-				"time",            "",             DATA_STRING,    time_str,
-				"model",           "",             DATA_STRING,    model ? model1 : model2,
-				"probe",           "",             DATA_INT,       probe,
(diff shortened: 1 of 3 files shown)


rct:
Thanks for doing this.   Please add samples to rtl_433_tests. 

These look close to the 592TXR's minus the extra probe.  What happens if both the new protocol and -R 39 are enabled?   Also did you try any 592TXR samples to see if there are any false positives?

Also, what indent are you using, it looks like 7 or 8.   The reset of acurite I think is a 4 space indent.

khkremer:
It looks like the sensor is reporting temperature in C. You are however outputting both C and F (calculated from C). When you look at all other sensors, they only output the temperature in the unit native to the sensor. This avoids problems with rounding and values that are potentially converted twice - once in rtl_433 and once again in the software that receives the data.

jonrifkin:
You're welcome.  And thanks for rtl_433, it has saved me a huge amount of effort.  I know because I've previously developed code to decode the 915mhz signal from a Lacrosse sensor, starting with the output from rtl_fm, and it took forever.  I should have just abandoned that and used a 433mhz sensor and rtl_433 from the start.

I've added to the rtl_433_tests repository under the tests/acurite/08 directory.

I've tried other samples and have found no false positives.  In fact, while I developed the code I also had an Acurite TXR that was sending live signals.  I had no issue with false positives.  Furthermore, I wouldn't expect a problem, because two callback functions require different bit counts in the input signal.  The txr callback (acurite_txt_callback) expects two separate ranges of bit counts; either exactly 72 bits, or between 56 and 64 bit inclusive.  The new 00275rm callback expects exactly 88 bits, so there's no overlap.

Sorry about the tab and indent conflict in the source code.  I've just pushed a correction to my repository/branch; I've replaced the tabs in my callback with spaces.

I also just removed the F output from the code, following what the other sensors do.

rct:
@jonrifkin - 

> ...  I've previously developed code to decode the 915mhz signal from a Lacrosse sensor, starting with the output from rtl_fm, and it took forever. I should have just abandoned that and used a 433mhz sensor and rtl_433 from the start.

FWIW, I started doing some LaCrosse 915 Mhz decoding in rtl_433, but I gave up because I only had one sensor, and found the frequency hopping a little annoying.  I think someone else added support for some of the LaCrosse 915 mhz sensors not too long ago.

> The txr callback (acurite_txt_callback) expects two separate ranges of bit counts; either exactly 72 bits, or between 56 and 64 bit inclusive. The new 00275rm callback expects exactly 88 bits, so there's no overlap.

Good to know about the callbacks and thanks for being thorough.  I was wondering because I thought I saw some of the new devices had some degree of interchangeability.   The TXR callback could be tightened up, I think, because it will only be either 56 or 64, depending on whether an extra byte got inserted.  I seem to recall some of that may have been due to the older demodulator code.

Do the new sensors also transmit every 18 seconds?

src/devices/acurite.c
```diff
@@ -852,3 +943,14 @@ r_device acurite_606 = {
...
 };
+
+r_device acurite_00275rm = {
+    .name           = "Acurite 00275rm,00276rm Temp/Humidity with optional probe",
+    .modulation     = OOK_PULSE_PWM_TERNARY,
+    .short_limit    = 320,  // = 4* 80,  80  is reported by -G option
+    .long_limit     = 520,  // = 4*130, 130  "
+    .reset_limit    = 608,  // = 4*152, 152  "
+    .json_callback  = &acurite_00275rm_callback,
+    .disabled       = 1,
```
rct: Since this outputs structured data and doesn't look experimental, I believe this should be enabled by default (disabled = 0).

jonrifkin:
> Do the new sensors also transmit every 18 seconds?
Yes, that's what I'm seeing for both the internal and external.

merbanan:
The indentation is a little off. Please fix that and add a crc check assuming the crc calculation is correct. If the crc description is correct then add a crc member to the structured output. Look in other code to see how it is done.

Add the needed change to build this with autotools also.

src/devices/acurite.c
```diff
@@ -125,6 +125,19 @@ static int acurite_checksum(uint8_t row[BITBUF_COLS], int cols) {
...
+        uint8_t extract = *bytes++;
+        for ( uint8_t tempI = 8; tempI; tempI-- ) {
+            uint8_t sum = (crc ^ extract) & 0x01;
+            crc >>= 1;
+            if (sum) crc ^= poly;
+            extract >>= 1;
+        }
+    }
+    return crc;
+}
```
merbanan: If you look in rtl_433/src/util.c we have crc code, reuse that.

merbanan:
After the crc fix everything is ok and I will merge the code.

jonrifkin:
I've 1) incorporated and verified the crc code (thanks for the code), 2) added the crc member to the structured output, 3) cleaned up the indentation, and 4) unset the disabled flag (i.e. set .disabled=0,).

I also modified the decoding algorithm.  The device sends three redundant signals, and I was printing each signal as it came (following the design of the Acurite TXR callback).  Now, I combine the three signals into one, so that any bit position with non-agreeing values uses the majority value.  The crc is then done on the combined signal.  The advantages, as I see them, are that 1) we have more robust signal decoding and 2) we've removed redundant output.  I hope this is acceptable. 

All these changes are in my acurite-00275rm branch.

> Add the needed change to build this with autotools also.
I don't know what changes this requires.  If you point me in the right direction maybe I can take care of that.

merbanan:
The autotools comment was wrong on my part. Nothing to bother with.

If there is a crc present I only bother to check the first row. If one where to do it properly one should do some kind of divide and conquer over the rows and only include the ones with proper crc.

Anyway as I have not figured out the truly optimal way how to reduce this I leave it to the submitting author to decide.

src/devices/acurite.c
jonrifkin: OK, done.  Found crc16() in util.c, and removed the my unnecessary crc code.

jonrifkin:
OK, cleaned up crc code and re-pushed it into my acurite-00275rm branch.

merbanan:
Are you really really sure it is supposed to be crc16 and not crc8 ? Anyway I'll merge the patch now. Thanks for your submission.

jonrifkin:
I tested both on historical data.  Crc16 gave 0 for each unless I manually perturbed the data.  Crc8 didn't.  And, the crc8 algorithm didn't match the one presented above. Crc16 did.

PR #469, opened 2017-01-17
Add protocol definition for oregon scientific THN129 temp only sensor.

sbsrouteur:
Added THN129 sensor. Sensor follows the v2.1 protocol, and only has 1 temp measurement.

### DIFF
--- a/src/devices/oregon_scientific.c
+++ b/src/devices/oregon_scientific.c
@@ -19,6 +19,7 @@
 #define ID_WGR800   0x1984
 #define ID_WGR968   0x3d00
 #define ID_UV800    0xd874
+#define ID_THN129   0xcc43  // THN129 Temp only
 
 float get_os_temperature(unsigned char *message, unsigned int sensor_id) {
   // sensor ID included  to support sensors with temp in different position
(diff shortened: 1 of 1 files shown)


src/devices/oregon_scientific.c
```diff
@@ -352,6 +353,23 @@ static int oregon_scientific_v2_1_parser(bitbuffer_t *bitbuffer) {
...
+        float temp_c = get_os_temperature(msg, sensor_id);
+        data = data_make(
+            "time",          "",            DATA_STRING, time_str,
+            "brand",         "",            DATA_STRING, "OS",
+            "model",         "",            DATA_STRING, "THN129",
+            "id",            "House Code",  DATA_INT,    get_os_rollingcode(msg, sensor_id),
+            "channel",       "Channel",     DATA_INT,    get_os_channel(msg, sensor_id), // 1 to 5
+            "battery",       "Battery",     DATA_STRING, get_os_battery(msg, sensor_id) ? "LOW" : "OK",
+            "temperature_C",  "Celcius",    DATA_FORMAT, "%.02f C", DATA_DOUBLE, temp_c,
+            "temperature_F",  "Fahrenheit", DATA_FORMAT, "%.02f F", DATA_DOUBLE, ((temp_c*9)/5)+32,
```
merbanan: Remove this. Only the signal native representation can be there.

merbanan:
Some signal samples would be nice also.

src/devices/oregon_scientific.c
sbsrouteur: no sure to understand what you mean, the sensor transmits, chan & bat status (in message begin after Id) then temp in degree_C. Which os these lines should I remove?

merbanan: temperature_F

sbsrouteur:
Added a sample signal file.
https://github.com/merbanan/rtl_433_tests/pull/136

merbanan:
Thanks

PR #475, opened 2017-01-22
devices: mebus: Use data API for results

dholm:
Use the data API to post results from the Mebus device so that it can
generate standard format output.


merbanan:
This change results in that the decoder can be enabled by default. Change .disabled = 1 to 0 and the patch is ready to be merged.

dholm:
@merbanan Fixed!

Issue #477, opened 2017-01-25
Add support for LaCrosse TX29 and TX35 (this post include .data and protocol spec)


psa-jforestier:
Hello, and thanks for all this work. I have two different sensors from LaCrosse Europe (868,2 MHz), and a clone from Conrad.
The current version of RTL_433 can't decode it, but the protocol is documented, and in -a mode I can see the bytes, and I saved a valid .data file. I dont know how to contribute to this project. Do I have to create a fork of the rtl_433_tests project, add all necessary information, and create a Pull Request ?

data files :
[gfile.zip](https://github.com/merbanan/rtl_433/files/729991/gfile.zip)

TX29-IT & TX35
--------------
Captured signals for LaCrosse compatible TX29-IT, made by Conrad 868 Mhz temperature. The TX9U-IT version is 915MHz, dont know if it use the same protocol.

This should also work for LaCrosse TX35 and  which use the same protocol, with humidty.

Device Info
-----------
* Buy it online : http://www.conrad.fr/ce/fr/product/646282/
* Owner's Manual : FR : http://www.lacrossetechnology.fr/donnees/documents/produits/TX29IT+/PDF-fr-TX29IT+.pdf
* EAN: 4016138439489

Device Notes
------------

* Random Device ID - When the device is powered up (batteries are
  changed) it will generate a random ID byte that it will send with
  all transmissions.
* Messages are sent once.
* There is temperature and humidty in the same message. If (humidity & 0x7F == 0x6A) it meens there is no humidity sensor.

Protocol Documentation
----------------------
A nice page documenting how to decode the protocol is available here http://fredboboss.free.fr/articles/tx29.php/

Basically : 

* carrier frequency : 868.3 MHz
* modulation OOK
* baudate 17.24 kpbs
* 64-bit messages (8 bytes, including preamble and crc)
* CRC is a CRC-8 with the polynom x8 + x5 + x4 + 1 (poly=0x31, int=0, final xor=0) of bytes 4 to 7

```
Nibble num :
---1 ---2 ---3 ---4 ---5 ---6 ---7 ---8 ---9 --10 --11 --12 --13 --14 --15 --16
Example Data bytes(gfile-tx29.data) : 
   a    a    2    d    d    4    9    2    8    4    4    8    6    a    e    c
Example Data bits (gfile-tx29.data) : 
1010 1010 0010 1101 1101 0100 1001 0010 1000 0100 0100 1000 0110 1010 1110 1100
Bytes num :
[...]

PR #479, opened 2017-01-30
LaCrosse TX29-IT / TX35DTH-IT weather sensor

psa-jforestier:
Here is a decoder for the LaCrosse TX29-IT / TX35DTH-IT weather sensor, also compatible with sensor made by Conrad and by StarMétéo. This sensors transmit on 868,2Mhz, i successfully decode 4 sensors at the same time with this decoder, tuning at 868240000Hz.
Theses devices use the same bit-level protocol, but RF signal is different (pulse length is 55us or 105us depending of the device), this pull-request add two decoders in rtl_433_devices.h and only one decoder in devices/lacrosse_tx35.c .

See issue https://github.com/merbanan/rtl_433/issues/477

I hope you will like it ;) !

Jerome.

### DIFF
--- a/src/devices/lacrosse_tx35.c
+++ b/src/devices/lacrosse_tx35.c
@@ -41,7 +41,8 @@ one byte as the preamble. I own 3 sensors TX29, and two of them send a long prea
 So this decoder synchronize on the 0xaa 0x2d 0xd4 preamble, so many 0xaa can occurs before.
 Also, I added 0x9 in the preamble (the weather data length), because this decoder only handle 
 this type of message.
-TX29 and TX35 sahre the same protocol, but pulse are different length.
+TX29 and TX35 sahre the same protocol, but pulse are different length, thus this decoder
+handle the two signal and we use two r_device struct (only differing by the pulse width)
 
 How to make a decoder : https://enavarro.me/ajouter-un-decodeur-ask-a-rtl_433.html
 */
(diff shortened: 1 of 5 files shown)


src/devices/lacrosse_tx35.c
```diff
@@ -0,0 +1,193 @@
...
+			continue;
+		}
+
+		/*
+		 * Now that message "envelope" has been validated,
+		 * start parsing data.
+		 */
+		uint8_t sensor_id, newbatt, battery_low;
+		uint8_t humidity; // in %. If > 100 : no humidity sensor
+		float temp_c; // in °C
```
merbanan: Declare all variable at start of a block.

src/devices/lacrosse_tx35.c
```diff
@@ -0,0 +1,193 @@
...
+
+Developer's comments
+====================
+I have noticed that depending of the device, the message received has different length.
+It seems some sensor send a long preamble (33 bits, 0 / 1 alternated), and some send only
+one byte as the preamble. I own 3 sensors TX29, and two of them send a long preamble.
+So this decoder synchronize on the 0xaa 0x2d 0xd4 preamble, so many 0xaa can occurs before.
+Also, I added 0x9 in the preamble (the weather data length), because this decoder only handle 
+this type of message.
+TX29 and TX35 sahre the same protocol, but pulse are different length, thus this decoder
```
merbanan: *share

src/devices/lacrosse_tx35.c
```diff
@@ -0,0 +1,193 @@
...
+	return lacrosse_it(bitbuffer, 29);
+}
+
+/**
+ ** Wrapper for the TX35 device
+ **/
+static int lacrossetx35_callback(bitbuffer_t *bitbuffer) {
+	return lacrosse_it(bitbuffer, 35);
+}
+
```
merbanan: Add defines for the different types. Then send the type define as parameter here instead.

src/devices/lacrosse_tx35.c
```diff
@@ -0,0 +1,193 @@
...
+		uint8_t humidity; // in %. If > 100 : no humidity sensor
+		float temp_c; // in °C
+		sensor_id = ((out[3] & 0x0f) << 2) | ((out[4]>>6) & 0x03);
+		temp_c = 10.0 * (out[4] & 0x0f) +  1.0 *((out[5]>>4) & 0x0f) + 0.1 * (out[5] & 0x0f) - 40.0;
+		newbatt = (out[4] >> 5) & 1;
+		battery_low = (out[6]>>7) & 1;
+		humidity = 1 * (out[6] & 0x7F); // Bit 1 to 7 of byte 6
+		//if (humidity == LACROSSE_TX29_NOHUMIDSENSOR) 
+		//	humidity = -1; // The TX29-IT sensor do not have humidity. It is replaced by a special value
+
```
merbanan: Here check the parameter type and don't send the humidity for a sensor that does not have a humidity value.

src/devices/lacrosse_tx35.c
```diff
@@ -0,0 +1,193 @@
...
+	.name           = "LaCrosse TX29IT Temperature snsor",
+	.modulation     = FSK_PULSE_PCM,
+	.short_limit    = 55,
+	.long_limit     = 55,
+	.reset_limit    = 4000,
+	.json_callback  = &lacrossetx29_callback,
+	.disabled       = 0,
+	.demod_arg      = 0,
+	.fields         = output_fields,
+};
```
merbanan: Define another output_fields list that doesn't have humidity.

merbanan:
Looks good, some needed changes and the code is good to be merged. And send samples if there are none yet in the test repository.

PR #490, opened 2017-03-11
Fix compile error on ARM with GCC with lack of parens.

nicheath:
I'm using rtl_433 on a BeagleBone Black and gcc does not compile `battery = bb[1][1] & 0x8f >> 7;` correctly.
It executes as if it was written `battery = bb[1][1] & (0x8f >> 7);` Resulting in the application always printing "Low".

Solution: add explicit parens.

The change to the earlier statement was just added for explicit behavior even though it was compiling correctly.

Linux arm 4.4.9-ti-r25 #1 SMP Thu May 5 23:08:13 UTC 2016 armv7l GNU/Linux
$ gcc --version
gcc (Debian 4.9.2-10) 4.9.2

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -727,12 +727,12 @@ static int acurite_606_callback(bitbuffer_t *bitbuf) {
 	    // Processing the temperature: 
             // Upper 4 bits are stored in nibble 1, lower 8 bits are stored in nibble 2
             // upper 4 bits of nibble 1 are reserved for other usages (e.g. battery status)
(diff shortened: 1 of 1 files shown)


nicheath:
original rtl_433 is compiled as such
```
1c4f2: 785b               ldrb  r3, [r3, #1]
1c4f4: f003 0301       and.w   r3, r3, #1
1c4f8: 60fb                str r3, [r7, #12]
```

modified is compiled correctly
```
1c4f2: 785b            ldrb  r3, [r3, #1]
1c4f4: 09db            lsrs    r3, r3, #7
1c4f6: b2db            uxtb    r3, r3
1c4f8: 60fb             str r3, [r7, #12]
```

PR #500, opened 2017-03-19
Acurite 06045 lightning sensor temperature fix.

rct:
* correctly decode temperature as 12 bit value
* correct status/message type to 2 remaining "flag" bits.  (Meaning TBD)
* correct distance to head of storm to just lower 5 bits
* Add l_status flag bits from distance byte as separate field.  (Meaning TBD)
* Add documentation of my current understanding of message format

Decoding of Acurite 06045 lightning sensor is still a bit of a work in progress, but lightning strike count, temperature and humidity should be fairly stable and usable.  

Distance to head of storm seems more reasonable, but is likely not the value shown by the display in Miles or KM.  Will need help from someone who has an Acurite display for this sensor to figure out remaining decoding.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -328,6 +328,107 @@ static float acurite_6045_getTemp (uint8_t highbyte, uint8_t lowbyte) {
     return temp;
 }
 
+/*
+ * Acurite 06045m Lightning Sensor decoding.
+ *
+ * Specs:
+ * - lightning strike count
+ * - extimated distance to front of storm, up to 25 miles / 40 km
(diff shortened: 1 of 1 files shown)


ggowins:
I have one of the Acurite displays that shows the lightning strike data from this sensor.  Let me know what you'd like me to check, and I can get the info for you.  Unfortunately the weather looks good for the several days, no lightning in the forecast.

rct:
@ggowins Thanks for your offer to help.  Can you contact me directly or via WXForum -  http://www.wxforum.net/index.php?topic=30139.50

(my email address is in the AUTHORS file for this repo.)

PR #504, opened 2017-03-22
Fix #503

### LINKED ISSUE #503: rtl_433/src/util.c:117:4: error: non-void function 'local_time_str' should return a value [-Wreturn-type]
I get the following error during the compilation of this program using `Apple LLVM version 8.0.0 (clang-800.0.42.1)`.

```
[  1%] Building C object src/CMakeFiles/rtl_433.dir/util.c.o
/Users/casa/rtl_433/src/util.c:117:4: error: non-void function 'local_time_str' should return a value [-Wreturn-type]
                        return;
                        ^
1 error generated.
```

crushedice2000:
It compiles well returning `0` but I don't know if this is right.

# Edit:
 
I didn't read the function previously. Sorry. I only done a diagonal read of the code and thought that it was only to exit prematurely from the function in case of error.
Now it returns `buf`as pointed by @zuckschwerdt and @rct.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -114,7 +114,7 @@ char* local_time_str(time_t time_secs, char *buf) {
 		extern float sample_file_pos;
 		if (sample_file_pos != -1.0) {
 			snprintf(buf, LOCAL_TIME_BUFLEN, "@%fs", sample_file_pos);
-			return 0;
+			return buf;
 		}
 		time(&etime);
 	} else {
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
The function prototype for `local_time_str()` in `util.h` is to return `void`. It's also how the function is used in all devices.
Best to replace the `char*` with `void` and remove the `buf` from the final return.

zuckschwerdt:
This change was introduced with fc560a9 (https://github.com/merbanan/rtl_433/commit/fc560a927ce37a69bb95284e7b3ae8f2ecbfb8df). Return `buf` instead of `0` as noted in https://github.com/merbanan/rtl_433/blob/master/include/util.h#L85

rct:
This changed in commit #497  It should return buf.   I think that other return was missed.

PR #505, opened 2017-03-23
add support for FineOffset WH3080 weather station

ovrheat:
Reworked '**Fine Offset WH1080 weather station**' module to integrate support for '**Fine Offset WH3080 weather station**', which in substance uses the same WH1080 protocol plus UV/Lux sensors addition.
So: weather/datetime protocol is identical for both devices, while UV/Lux decoding is only triggered by WH3080.


merbanan:
Keep the name or use git move first. Right now the diff is huge and it is hard to see what is changed.

ovrheat:
Ok, I'll close this and I'll submit a new pull request.

PR #511, opened 2017-04-03
Add support for Fine Offset WH3080 weather station

ovrheat:
- Reworked 'Fine Offset WH1080' module to add support to Fine Offset WH3080 UV/Light decoding. 
WH3080 looks like a WH1080 with added UV/Light sensors onboard. UV/Light data are sent separately from weather data in a 'once-every-60-seconds' cycle, while weather data remains identical for both devices with their 48 seconds cycle.
- Added 'Signal type' field on datetime output to display which kind of datetime signal the device is synchronized to (DCF77 or WWVB/MSF)
- Corrected fields name on datetime output.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,3 +19,4 @@ Baruch Even <baruch@ev-en.org>
 Andrea <andrea@andreaaizza.com>
 Helge Weissig <helgew@grajagan.org>
 Robert Fraczkiewicz <aromring@gmail.com>
+Nicola Quiriti <nik@wifi4all.it>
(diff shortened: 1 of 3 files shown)


merbanan:
You remove lots of static. Why?

ovrheat:
I thought it should be better... If you think it could make troubles we can put them back again

zuckschwerdt:
If you mean "better" as code style, then no. OO languages might discourage static things. But in C static is more like "private" in other popular languages: access control. every non-static is exported.

ovrheat:
Fine!... Exactly the opposite I wanted!...   Let's get back to static, I'll submit a pull request in no time.

PR #521, opened 2017-04-29
honeywell.c: fix pointer overrun, sprintf hex buffer needs extra byte…

emard:
… to hold NULL char

### DIFF
--- a/src/devices/honeywell.c
+++ b/src/devices/honeywell.c
@@ -23,7 +23,7 @@ static int honeywell_callback(bitbuffer_t *bitbuffer) {
   char time_str[LOCAL_TIME_BUFLEN];
   const uint8_t *bb;
   char device_id[6];
-  char hex[2];
+  char hex[3];
   char binary [65];
   int heartbeat = 0;
   uint16_t crc_calculated = 0;


zuckschwerdt:
this is a fix, indeed. The code is still ugly though. It's meant to extract 20 bits of ID, right? Can't simple bit shifting give us a plain int? Are you interested in cleaning this up? Otherwise I'll give it a go :)

emard:
Maybe better you make it if you have real hardware
I can jump in too but my transmitter is custom :)

I'm using it to decode a signal from this my project

https://github.com/emard/low-power-counter

On the 20-bit field this hardware encodes the state
of the counter, on the 4-bit "unknown" field encodes
channel number (0-3). and 8-bit "data" field might be used
instead of "serial" to receive from several different transmitters

I liked the honeywell protocol becuase it sends 64-bits fast and
has standard 16-bit CRC

zuckschwerdt:
I don't have the hardware as the 345 MHz band is not legal in europe ;)
There are however some test signals in the rtl_433_tests repository. My approach would be to simply shift-and-add the nibbles instead of mucking around with hex-printing nibbles.
I'll prepare a patch and verify it with the test signals.

zuckschwerdt:
Looking closer there is actually a lot wrong with this device. Why is there a time_unix field and a binary dump of the whole packet? This smells like left over debugging code. And the crc flag will always show "ok" as a crc error will return early. I'll send a cleanup patch for you to test then.

emard:
hi!

I'd like parts of packet maybe hex sprintf'd to some large string and isolated
the parts and printed hex/dec e.g. what do you thing

reserved: 0x8=8, serial: 0x12345=74565, data: 0xAB=171 (open/close, on/off)

Super, there's a plenty of useful test signals

zuckschwerdt:
I've just cleaned what's there and added the channel (the nibble / 4bits before the 20bit id). Can you lay out the other interesting parts, preferably like this:
```
PP PP C DDDDD EE SS SS
P: 16bit Preamble and sync bit (ff fe)
C: 4bit Channel
D: 20bit Device serial number
E: 8bit Event Information, where 0x80 = Open/Close, 0x04 = Heartbeat
S: 16bit CRC
```

emard:
HI
that packet interpretation looks great!
PPPP C DDDDD EE SSSS
I guess that not much more all we know about original packet

My device sends packets with this meaning:
P: 16bit preamble always FFFE
C: 4bit channel 0-3
DDDDD: 20bit counter value
EE: 8bit device number (allows for multiple devices in proximity)
SSSS: 16bit crc

rct:
only partially following this, just curious,  how wide spread will your device be?   Are you using a different frequency so you don't provide false positives to nearby alarm systems as the counter value cycles through device IDs that might be in use?

emard:
I don't expect any significant spread of the device, maybe few on Earth
at most :)... I only know of my installation. I use it to monitor pulses of
water flow meter

this "Channel" nibble can distinguish my device from honeywell door sensor.
Honeywell has value 8 in nibble "C", my device can have there values 0,1,2,3

But of course I can slightly change crc, or the bit timing so my device
will not mix with honeywell.

My device (as described on git) needs external transmitter so user can
choose 315,433,868 or any other

I use 433.92 transmitter SYN115 cheap ebay for 1$

zuckschwerdt:
I've updated #522 -- `event` will give you the device numer and `id` will be the counter for your device. It's somewhat named wrong for your case but compatible to the original devices this way.

emard:
That's fine for me, fields will be mis-named but I know what they
mean.
Where is it uploaded, can I try your code to see how it looks like?

emard:
Yes yes I compiled it, that's it, looks beautiful now :)

zuckschwerdt:
#522 is merged and this pr can be closed

PR #530, opened 2017-05-10
adds a watchdog timer to async callback, abort if async events stop arriving

zuckschwerdt:
In async mode rtlsdr might just stop sending events, rtl_433 won't notice and idle forever.
To test just unplug the receiver, nothing will happen. (One of my receivers has thermal problems and will lock up at random intervals.)
This timer will ensure events keep coming to the callback and trigger the existing signal handler to abort as cleanly as possible otherwise.

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -599,6 +599,10 @@ static void rtlsdr_callback(unsigned char *iq_buf, uint32_t len, void *ctx) {
     int i;
     char time_str[LOCAL_TIME_BUFLEN];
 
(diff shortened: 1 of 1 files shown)


merbanan:
Looked harmless enough.

zuckschwerdt:
With the current buffer size (128k) the callback should be called every 500ms at 250k sample rate (lowest possible setting for receivers if I'm not mistaken).
I though about making the margin bigger like 2 or 3 seconds but I guess we'll see if problems or corner cases arise :)

PR #535, opened 2017-05-13
harmonize device data output

zuckschwerdt:
Harmonize device data output to `model` instead of `device` or `make`, and `crc` instead of `checksum`. All devices should now use a common set of basic data output: `time`, `model`, `id`, and sometimes `type` as well as `crc`: `OK` for checksummed transmissions. Mostly it's just a single out of ~79 devices that isn't in line with common output naming. This will make post-processing easier and give new device authors a fixed starting point no matter which example code they are looking at.

### DIFF
--- a/src/devices/steelmate.c
+++ b/src/devices/steelmate.c
@@ -87,7 +87,7 @@ static int steelmate_callback(bitbuffer_t *bitbuffer) {
 
 		data = data_make("time", "", DATA_STRING, time_str,
 			"type", "", DATA_STRING, "TPMS",
-			"make", "", DATA_STRING, "Steelmate",
+			"model", "", DATA_STRING, "Steelmate",
 			"id", "", DATA_STRING, sensorIDhex,
 			"pressure_PSI", "", DATA_DOUBLE, pressurePSI,
 			"temperature_F", "", DATA_DOUBLE, (float)tempFahrenheit,
@@ -106,7 +106,7 @@ static int steelmate_callback(bitbuffer_t *bitbuffer) {
 static char *output_fields[] = {
 	"time",
 	"type",
-	"make",
+	"model",
 	"id",
 	"pressure_PSI",
 	"temperature_F",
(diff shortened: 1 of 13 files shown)


zuckschwerdt:
My hope is, that this change isn't too inconvenient to existing users and a benefit to general post-processing/logging. Comments otherwise very welcome.

src/devices/steelmate.c
```diff
@@ -87,12 +87,12 @@ static int steelmate_callback(bitbuffer_t *bitbuffer) {
...
 		data = data_make("time", "", DATA_STRING, time_str,
 			"type", "", DATA_STRING, "TPMS",
-			"make", "", DATA_STRING, "Steelmate",
+			"model", "", DATA_STRING, "Steelmate",
 			"id", "", DATA_STRING, sensorIDhex,
 			"pressure_PSI", "", DATA_DOUBLE, pressurePSI,
 			"temperature_F", "", DATA_DOUBLE, (float)tempFahrenheit,
 			"battery_mV", "", DATA_INT, battery_mV,
-			"checksum", "", DATA_STRING, "OK",
+			"crc", "", DATA_STRING, "OK",
```
merbanan: This really is a checksum. Naming it crc is wrong. Maybe change all instances to message integrity check instead (mic) ?

merbanan:
Looks ok besides the crc thingy. Maybe add mic = crc | checksum | none ?

zuckschwerdt:
Very good idea! There is no reason to output `CRC=BAD` (and no device currently does that). How about `integrity` =`CRC`, `CHECKSUM`... to make it easier to comprehend at a glance?
I'll convert all instances and update the PR.

merbanan:
Mic instead of integrity and I'm happy. Mic is an actual term.

zuckschwerdt:
done. All `crc` and `checksum` data fields are converted. I'll do a thorough review and add some more `mic` keys for other devices later.
Would you be ok with a whitespace cleanup PR? There are quite some trailing whitespaces and mixed indentations I'd like to sort out.

merbanan:
White space changes would be nice. Feel free to send a PR

PR #536, opened 2017-05-14
Fix whitespace

zuckschwerdt:
First commit deletes trailing whitespace, second commit fixes indentation for each file to either 4 spaces or one tab.
It surprised me how many files had mixed tabs and spaces, and this effort might need a second pass to thoroughly beautify the device code.
Also if the commit is too big / too intrusive I'm okay with dropping this idea ;)


merbanan:
Tnx

PR #537, opened 2017-05-15
Update rtl_433.h

strowger:
MAX_PROTOCOLS incremented to 79. 

Without this change, starting with the -G option fails with an error.


zuckschwerdt:
The problem originated with 2acd8f9253f80ecfac6ee55cd7d4b8cd2e2a695e

PR #539, opened 2017-05-17
fix Company name, output flags and data code

zuckschwerdt:
The company is named "Schrader" not "Schraeder". Adds output of flags, pressure, and temperature in undecoded format. Include the high nibble of the ID. The ID is given on the device with 8 hex digits not just 7, but that might be omitted in the protocol.

### DIFF
--- a/src/devices/schraeder.c
+++ b/src/devices/schraeder.c
@@ -1,6 +1,7 @@
-/* Schraeder TPMS protocol
+/* Schrader TPMS protocol
  *
- * Copyright © 2016 Benjamin Larsson
+ * Copyright (C) 2016 Benjamin Larsson
+ * and 2017 Christian W. Zuckschwerdt <zany@triq.net>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
I'm confident I nailed the decoding, see https://zuckschwerdt.org/schrader-readings.txt for a log of over 1000 unique readings taken near a busy road. Analysis of the ID shows 0x01944919 as the biggest (newest) issued serial ID so far. Without the first nibble the distribution of IDs is just too random. The extended ID is also in line with https://github.com/merbanan/rtl_433_tests/tree/master/tests/Schraeder. For the company name see https://www.schraderinternational.com/Connect-with-Schrader

zuckschwerdt:
Updated the PR with decoded pressure and temperature readings. The FCC description gave it away. Statistical analysis on the decocded values confirms the (straight forward) mapping.

merbanan:
Impressive I must say. We need a unit for the pressure though.

merbanan:
And now it is time to start vehicle tracking :)

zuckschwerdt:
The unit here is bar. I didn't want to go with mbar/hPa as the value is not absolute but differential to the ambient pressure (i.e. the plain values your car manual tells you).
Maybe the data field should be named `pressure_BAR` (`_bar`, `_Bar`?). That would be a clear scheme with minimal change. And, like temperature, this could be detected and converted easily. Thoughts?

zuckschwerdt:
Contrary to Schrader's statement, http://www.tpmsmadesimple.com/pdf/Schrader%20Technical%20Response-RF%20Security-2010.pdf, "while  it  is  possible  to  track  someone  by  their  tire  IDs, the  feasibility  of  doing so  would  be  quite  low", I get readings from fast passing vehicles with ~50m LOS. I'd surely consider this a privacy problem. Also think about a set of sensors with known distance to derive the vehicles average speed…

PR #540, opened 2017-05-17
Added data struct for efergye2

DavidBel86:
Added standard formated output capability
Also corrected the calculation method and added some information (id, battery status, learn mode...).

### DIFF
--- a/src/devices/efergy_e2_classic.c
+++ b/src/devices/efergy_e2_classic.c
@@ -4,7 +4,9 @@
  * on frequency ~433.55 MHz. The data that is transmitted consists of 8
  * bytes:
  *
- * Byte 1-4: Start bits (0000), then static data (probably device id)
+ * Byte 1: Start bits (00)
+ * Byte 2-3: Device id
+ * Byte 4: Learn mode, sending interval and battery status
  * Byte 5-7: Current power consumption
  * Byte 8: Checksum
  *
--- a/src/devices/efergy_e2_classic.c
+++ b/src/devices/efergy_e2_classic.c
@@ -78,7 +78,7 @@ static int efergy_e2_classic_callback(bitbuffer_t *bitbuffer) {
 	uint8_t learn = (bytes[3] & 0x80)>>7;
 	uint8_t interval = (((bytes[3] & 0x30)>>4)+1)*6;
 	uint8_t battery = (bytes[3] & 0x40)>>6;
-	float current_adc = (float)((bytes[4] << 8 | bytes[5]) << fact) / 0x8000;
+	float current_adc = (float)((bytes[4] << 8 | bytes[5]) * pow(2,fact)) / 0x8000;
 	uint16_t address = bytes[2] << 8 | bytes[1];
 	//fprintf(stdout, "Addr: %02X%02X,Amps: %.2f A, inter: %d s, bat: %d, learn: %d\n",bytes[1],bytes[2],current_adc,interval,battery,learn);
 	//fprintf(stdout, "%04X,%.2f,%d,%d,%d\n",address,current_adc,interval,battery,learn);
--- a/src/devices/efergy_e2_classic.c
+++ b/src/devices/efergy_e2_classic.c
@@ -28,8 +28,6 @@ static int efergy_e2_classic_callback(bitbuffer_t *bitbuffer) {
 	signed char fact = 0;
 	data_t *data;
 	char time_str[LOCAL_TIME_BUFLEN];
-	
-	local_time_str(0, time_str);
 
 	if (num_bits < 64 || num_bits > 80) {
 		return 0;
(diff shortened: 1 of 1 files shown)


src/devices/efergy_e2_classic.c
```diff
@@ -18,10 +18,18 @@
...
+#include "math.h"
 
 static int efergy_e2_classic_callback(bitbuffer_t *bitbuffer) {
 	unsigned num_bits = bitbuffer->bits_per_row[0];
 	uint8_t *bytes = bitbuffer->bb[0];
+	signed char fact = 0;
+	data_t *data;
+	char time_str[LOCAL_TIME_BUFLEN];
+	
+	local_time_str(0, time_str);
```
zuckschwerdt: move `local_time_str` below all checks, ideally just above `data_make`

src/devices/efergy_e2_classic.c
```diff
@@ -60,24 +68,59 @@ static int efergy_e2_classic_callback(bitbuffer_t *bitbuffer) {
...
+	else
+	{
+	    fact = bytes[6];
 	}
 
+	uint8_t learn = (bytes[3] & 0x80)>>7;
+	uint8_t interval = (((bytes[3] & 0x30)>>4)+1)*6;
+	uint8_t battery = (bytes[3] & 0x40)>>6;
+	float current_adc = ((float)(256 * bytes[4] + bytes[5]) * pow(2,fact))/ 32768;
+	uint16_t address = bytes[1] + (bytes[2] << 8);
```
zuckschwerdt: use bitwise OR

src/devices/efergy_e2_classic.c
```diff
@@ -60,24 +68,59 @@ static int efergy_e2_classic_callback(bitbuffer_t *bitbuffer) {
...
+	}
+	else
+	{
+	    fact = bytes[6];
 	}
 
+	uint8_t learn = (bytes[3] & 0x80)>>7;
+	uint8_t interval = (((bytes[3] & 0x30)>>4)+1)*6;
+	uint8_t battery = (bytes[3] & 0x40)>>6;
+	float current_adc = ((float)(256 * bytes[4] + bytes[5]) * pow(2,fact))/ 32768;
```
zuckschwerdt: use `bytes[4] << 8 | bytes[5]` and `1 << fact` and `0x7fff`

src/devices/efergy_e2_classic.c
```diff
@@ -60,24 +68,59 @@ static int efergy_e2_classic_callback(bitbuffer_t *bitbuffer) {
 		return 0;
 	}
 
-	const unsigned VOLTAGES[] = {110, 115, 120, 220, 230, 240, 0};
-
-	double current_adc = 256 * bytes[4] + bytes[5];
-	for (unsigned i = 0; VOLTAGES[i] != 0; ++i) {
-		double power  = (VOLTAGES[i] * current_adc * (1 << bytes[6])) / 32768;
-		fprintf(stderr, "Power consumption at %u volts: %.2f watts\n", VOLTAGES[i], power);
+	if (bytes[6] > 0x80)
```
zuckschwerdt: shouldn't this be `>=` (or `&0x80`) as 0x80 would already be a negative value? But I'm not entirely following, isn't bytes[6] 2s-complement? Wouldn't a cast to `signed char`do?

src/devices/efergy_e2_classic.c
DavidBel86: Do you mean, instead of '+' in current_adc and address calculation?

src/devices/efergy_e2_classic.c
DavidBel86: Ok

src/devices/efergy_e2_classic.c
DavidBel86: Yes, it actually should, don't know why I went in that wierd if then else...

src/devices/efergy_e2_classic.c
zuckschwerdt: I was thinking about `uint16_t address = bytes[2] << 8 | bytes[1];`

src/devices/efergy_e2_classic.c
zuckschwerdt: I.e. `float current_adc = (float)((bytes[4] << 8 | bytes[5]) << fact) / 0x8000;`

src/devices/efergy_e2_classic.c
zuckschwerdt: :) Then go for `(int8_t)bytes[6]`

DavidBel86: Mmmh, will not that convert the value of bytes[6] (uint8_t) to an int8_t (keeping the same positive value then)? Shouldn't I reinterpret bytes[6] as a signed number? (ex: using reinterpret_cast<int8_t>(bytes[6])). For memo, fact can be negative to shift the comma to the left.
I haven't been developing in C for a while to be honest... ;)

DavidBel86: BTW: I reverted back to 'pow(2,fact)' to allow both right and left shifting. Do you have other solution?

zuckschwerdt: Good thing you caught that. That looked clean but was entirely stupid of me. Sorry. Convert to float (it will be double internally, there should be no floats in the code base… discussion for another day, though) apply scaling factor. Just like you originally had.

Dividing by 0x8000 is essentially a shift-right by 15. This results in a factor of pow(2, fact-15). I take it this is because the "mantissa" is 16-bit and needs to be normalized to 0<M<1. I think I get it now!

There is integer promotion, arithmetic conversion, and the cast operator in C. The cast `(int8_t)` will reinterpret here (and the result will actually be `int` …integer promotion with sign extension!). Other places like `(float) someint` actually trigger arithmetic conversion. C is misleading at best :)

Also, your `pow()` call will promote the signed char to integer and then convert to double. The const `2` will be converted to double at compile time. To skip some of this, it's best to declare temporary variables, like `fact` as integer -- promotion isn't that expensive but will likely lead to wrong assumptions.

zuckschwerdt: Though I know a fair bit about C, and when in doubt am not afraid to look at the assembler output, I am wrong on occasion or will at least not explain clearly enough :)
As for other peoples device code here it's best to assume that it is tweaked to work, but mostly with bad assumptions… C isn't the "native" language for most, and even when it is, C is designed to bite you…

DavidBel86:
I've rewritten the calculation method, tested, it's working fine.

gtortone:
Hi,
variable 'fact' is defined two times (line 28 and line 73)

DavidBel86:
Indeed, just saw it, it's clean now, just dubble checked.
Sorry about that

PR #545, opened 2017-05-20
Add comment to explain unit conversion

sbsrouteur:


sbsrouteur:
Is this any cleaner, it looks the same from here?

zuckschwerdt:
nah. not  good, based on an old head. Try `git pull upstream master` `git rebase master` `git push -f` before opening a branch.
Also a note of `1 inHg = 3386.389 Pa` and a multiplication with `33.86389` would be better. I.e. `…/100.0*33.86389` (the F on 100.0 is not useful as the calculation will be done in double anyway). Thanks for picking up my query so promptly!

merbanan:
You still need to redo this. The PR does not apply cleanly.

sbsrouteur:
closing reset everything and did a new one (#546)

PR #550, opened 2017-05-22
hop interval param added (multiple frequencies)

gtortone:
I added a parameter (-H) to specify hop interval in case of multiple frequencies.

Example:
./rtl_433 -f 433535000 -f 434019000 -R 36 -R 19 -H 15

in current master branch default value (600 seconds) was a fixed values,
and in some case is not suitable...


rct:
If you are changing command line parameters, README.md should generally get updated too.  It contains the usage output so it's mostly a copy-and-paste to keep it in sync.

PR #552, opened 2017-05-24
adding Citroen TPMS device

zuckschwerdt:
Another TPMS protocol. Similar to Schrader and Steelmate in FSK, Manchester and packet layout.


merbanan:
Thanks, don't forget samples if there are none.

PR #561, opened 2017-06-06
Fix hex output padding for unit ID field for lightwaverf.

darrenbuse:
When printing the unit ID hex value for lightwaverf devices there is no padding to handle any leading '0's of the individual bytes that make up the ID. If any of the bytes have a value less than 0x0F only a single character is printed, as these are then concatenated together an incorrect value is produced.

Example:
Correct Unit ID: F10F6F is incorrectly printed as: F1F6F (missing '0')


merbanan:
Thanks.

PR #562, opened 2017-06-09
Thermopro TP-11.

pekon:
I tested this with my device. I'll send a separate pull request with the test files.

### DIFF
--- a/src/devices/thermopro_tp11.c
+++ b/src/devices/thermopro_tp11.c
@@ -39,21 +39,18 @@ static int valid(unsigned data, unsigned check) {
 }
 
 static int thermopro_tp11_sensor_callback(bitbuffer_t *bitbuffer) {
+    if (bitbuffer->num_rows < 3) return 0;
     bitrow_t *bb = bitbuffer->bb;
-    data_t *data;
-    char time_str[LOCAL_TIME_BUFLEN];
-    local_time_str(0, time_str);
     int good = -1;
 
-    for (int i = 0; good < 0 && i < 4; i++) {
+    // Compare first four bytes of rows that have 32 or 33 bits.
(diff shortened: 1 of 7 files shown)


zuckschwerdt:
Add new devices at the end, otherwise existing IDs are invalidated (`-R` flag).
Also `bitbuffer_find_repeated_row()` might be useful.

pekon:
Thanks! bitbuffer_find_repeated_row assumes that all rows have exactly the same number of bits. For this device the first three rows have 33 bits and the last has 32 bits. This code only compares the first 32 bits.

zuckschwerdt:
good point, that needs to be fixed someday. `compare_rows()` should return a prefix length which can be matched up with `min_bits`.

zuckschwerdt:
One more thing: move the `local_time_str()` call down to just before `data_make()`.

pekon:
Done. Also made sure the code exits as soon as it detects that no more than two rows can be same.

src/devices/thermopro_tp11.c
```diff
@@ -0,0 +1,100 @@
...
+                && bb[i][1] == bb[j][1]
+                && bb[i][2] == bb[j][2]
+                && bb[i][3] == bb[j][3]
+                && ++equal_rows >= 2) good = i;
+        }
+    }
+    if (good < 0) return 0;
+    // bb[good] is equal to at least two other rows, decode.
+    unsigned value = (bb[good][0] << 16) + (bb[good][1] << 8) + bb[good][2];
+    unsigned device = value >> 12;
```
merbanan: Move the declaration to the top, and I prefer unsigned int over just unsigned.

merbanan:
Looks good. One tiny change and samples before I merge.

pekon:
Done. Thanks!

PR #572, opened 2017-07-05
remove false positives from Esperanza/Kedsum

zuckschwerdt:
This removes false positive from Esperanza EWS vs Kedsum. Kedsum starts with 15 sync pulses and Esperanza with only 2. Also improves reception with Kedsum and moves the `local_time_str` after all checks.

### DIFF
--- a/src/devices/esperanza_ews.c
+++ b/src/devices/esperanza_ews.c
@@ -1,9 +1,3 @@
-#include <bitbuffer.h>
-#include "util.h"
-#include "data.h"
-#include "rtl_433.h"
-
-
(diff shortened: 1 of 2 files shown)


nebulous:
The addition of the block between lines 32 and 40 causes a regression for my Kedsum sensors which no longer work after this patch.

zuckschwerdt:
That block, i.e. checking for leading sync pulses, is tied to the change `reset_limit`. Maybe your device timing differs from the recorded test samples? Can you check that or upload one sample from your device?

PR #573, opened 2017-07-07
replace binary literals with hex

zuckschwerdt:
Since we are targeting C99 (or rather gnu99) we can not use binary literals. It will work with recent GCC (and Clang) but I still have older (Clang) releases that do not support this extension.

Perhaps we'd need to target the C++14 standard which includes binary literals, but that seems to harsh for the few odd places where binary numbers are used and easily replaced with hex.

### DIFF
--- a/src/devices/current_cost.c
+++ b/src/devices/current_cost.c
@@ -13,12 +13,12 @@ static int current_cost_callback(bitbuffer_t *bitbuffer) {
     local_time_str(0, time_str);
 
     uint8_t init_pattern[] = {
-        0b11001100, //8
-        0b11001100, //16
-        0b11001100, //24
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
I.e. this fixes
```
src/devices/current_cost.c:16:9: warning: binary integer literals are a GNU extension [-Wgnu-binary-literal]
…
src/devices/intertechno.c:18:60: warning: binary integer literals are a GNU extension [-Wgnu-binary-literal]
…
src/devices/mebus.c:16:54: warning: binary integer literals are a GNU extension [-Wgnu-binary-literal]
…
```

merbanan:
C99 is fine. C++14 is not.

PR #577, opened 2017-07-08
fix compiler warnings; possible undef bug

zuckschwerdt:
fix a compiler warning about a missing if-branch leading to undefined variables and a possible undefined variable bug. I.e.
```
src/devices/acurite.c:950:24: warning:
variable 'data' is used uninitialized whenever 'if' condition is false
```
and
```
src/devices/fineoffset_wh1080.c:319:16: warning:
variable 'msg_type' is used uninitialized whenever 'if' condition is false
```


zuckschwerdt:
…also a signed/unsigned warning

PR #590, opened 2017-07-20
Switched to new data API

Lefinnois:
Hi, 
I've switched my code to the new data API and so re-enable the device.
Regards

### DIFF
--- a/src/devices/cardin.c
+++ b/src/devices/cardin.c
@@ -1,5 +1,4 @@
 #include "rtl_433.h"
-#include "pulse_demod.h"
 #include "util.h"
 
 /*
@@ -21,8 +20,6 @@ static int cardin_callback(bitbuffer_t *bitbuffer) {
 	char rbuttonJ2[64] = { 0 };
 	data_t *data;
 	char time_str[LOCAL_TIME_BUFLEN];
-
-	local_time_str(0, time_str);
 
 	// validate message as we can
 	if((bb[0][2] & 48) == 0 && bitbuffer->bits_per_row[0] == 24 && (
@@ -120,6 +117,7 @@ static int cardin_callback(bitbuffer_t *bitbuffer) {
 			strcpy(rbuttonJ1, "--o");
 		}
 
+		local_time_str(0, time_str);
 		data = data_make(
 			"time",       "",                 DATA_STRING, time_str,
 			"model",      "",                 DATA_STRING, "Cardin S466",
--- a/src/devices/cardin.c
+++ b/src/devices/cardin.c
@@ -22,10 +22,10 @@ static int cardin_callback(bitbuffer_t *bitbuffer) {
 
 	// validate message as we can
 	if((bb[0][2] & 48) == 0 && bitbuffer->bits_per_row[0] == 24 && (
-				(bb[0][2] & 3) == 3 ||
-				(bb[0][2] & 9) == 9 ||
-				(bb[0][2] & 12) == 12 ||
-				(bb[0][2] & 6) == 6) ) {
+				(bb[0][2] & 0x0f) == 3 ||
+				(bb[0][2] & 0x0f) == 9 ||
+				(bb[0][2] & 0x0f) == 12 ||
+				(bb[0][2] & 0x0f) == 6) ) {
 
 /*
 		fprintf(stdout, "------------------------------\n");
--- a/src/devices/cardin.c
+++ b/src/devices/cardin.c
@@ -1,4 +1,6 @@
 #include "rtl_433.h"
+#include "pulse_demod.h"
+#include "util.h"
 
 /*
  * Cardin S466-TX2 generic garage door remote control on 27.195 Mhz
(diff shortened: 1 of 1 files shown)


src/devices/cardin.c
```diff
@@ -1,4 +1,6 @@
 #include "rtl_433.h"
+#include "pulse_demod.h"
```
zuckschwerdt: Is there a reason to include pulse_demod headers?

src/devices/cardin.c
```diff
@@ -15,6 +17,12 @@ static int cardin_callback(bitbuffer_t *bitbuffer) {
 	bitrow_t *bb = bitbuffer->bb;
 	int i, j, k;
 	unsigned char dip[10] = {'-','-','-','-','-','-','-','-','-', '\0'};
+	char rbuttonJ1[64] = { 0 };
+	char rbuttonJ2[64] = { 0 };
+	data_t *data;
+	char time_str[LOCAL_TIME_BUFLEN];
+
+	local_time_str(0, time_str);
```
zuckschwerdt: only run this line after all checks, usually just before data_make()

src/devices/cardin.c
Lefinnois: This came from new_template, sorry.

src/devices/cardin.c
Lefinnois: done

src/devices/cardin.c
```diff
@@ -92,40 +102,56 @@ static int cardin_callback(bitbuffer_t *bitbuffer) {
...
 		if((bb[0][2] & 9) == 9) {
-			fprintf(stdout,                  "2 --o (this is right button)\n");
-			fprintf(stdout, "                 1 o--\n");
+			strcpy(rbuttonJ2, "--o (this is right button)");
+			strcpy(rbuttonJ1, "o--");
 		}
 		if((bb[0][2] & 12) == 12) {
-			fprintf(stdout,                  "2 o-- (this is left button or two buttons on same channel)\n");
-			fprintf(stdout, "                 1 o--\n");
+			strcpy(rbuttonJ2, "o-- (this is left button or two buttons on same channel)");
```
zuckschwerdt: this looks really verbose. It's a 4-bit value, the output should be something like a single string with about 4 characters -- not 2 lines of (static) text. Generally a device should decode the values and leave presentation to some other software. Compact output suitable for a database storage would be ideal.

src/devices/cardin.c
```diff
@@ -92,40 +102,56 @@ static int cardin_callback(bitbuffer_t *bitbuffer) {
...
+			NULL);
+
+		data_acquired_handler(data);
+
 		return 1;
 	}
 	return 0;
 }
 
+static char *csv_output_fields[] = {
```
zuckschwerdt: usually this can be named output_fields, as it's used for json too.

src/devices/cardin.c
zuckschwerdt: maybe just using bb[0][2] as index to an array with possible output codes would be best here. strcpy really isn't needed.

Lefinnois: The remote control has two buttons.

The second group of switches (DIP 2 in the picture) can only be guessed if the right button is used.

One of the 4 possible combinations may mean that the left button has been used or that the right button does the same as the left button.

One of the 4 conditions is therefore not 100% reliable regarding the position of the switches.

I do not know how to mean this in 4 characters: the button used, the position of the switches and the doubt about the use of the left button.

s466

zuckschwerdt: currently your output is either of
```
"--o (this is right button)" "--o"
"--o (this is right button)" "o--"
"o-- (this is left button or two buttons on same channel)" "o--"
"o-- (this is right button)" "--o"
```
it could just be ("dip2" not "rbuttonJ2" and "rbuttonJ1")
```
"--o RB --o"
"--o RB o--"
"o-- LB/2B o--"
"o-- RB --o"
```
Anyway, I don't get that "--o" symbol. I thought those DIPs are tristate with -=off +=on o=float?
But it seems to be ascii-art for a single DIP slider? If so, that's a good picture for debug output but structured output then should be:
```
"00"
"01"
"11?"
"10"
```
If those DIPs are not tristate then a bitfield (`010101010`) would be much cleaner that the -o+ symbols for the "dipswitch".

You can keep the ascii-art output for debug mode by the way.

Lefinnois: DIP1 is tristate
DIP2 is not

I did so : 
keeping "+-+-+-+-o" style for the main dipswitch (DIP1)
using "11R", "10R", "01R" and "00L?"
getting ride of strcpy() and the tests on bb[0][2]

src/devices/cardin.c
Lefinnois: done

src/devices/cardin.c
zuckschwerdt: great solution! On last nit-picking: `rbutton[((bb[0][2] & 15) / 3)-1]` will go out-of-bounds for bb[0][2] < 3 or >= 15. Maybe the intent is clearer with just 16 strings of:
```
char *rbutton[4] = { "?", "?", "?", "11R", "?", "?", "10R", "?", "?", "01R", "?", "?", "00L?", "?", "?", "?" };
```

Lefinnois: This should not happen because bb[0][2] test is part of the message validation line 24.

zuckschwerdt: You are right, I didn't catch that. But doesn't it only ensure that at least two bits are set? I can't read decimal very well (with hex I can visualize the bit pattern easy). Maybe that test is intended to be like: (bb[0][2] & 0xf) == 3 ...? (i.e bb[0][2] = 15 matches all four conditions!)

Lefinnois: Damn ! You're absolutely right ! Thanks !

Issue #593, opened 2017-08-01
Command just quitting on Hop?


coogle:
Hello,

I have a series of devices that I need to listen for on different frequencies. Specifically a series of Honeywell sensors and then an Acurite 592TXR..

I thought I could use the -H command to potentially listen to both frequencies but this is what happens:

```
$ rtl_433 -f 344975000 -f 433920000 -R 40 -R 70 -H 5 -F json
Registering protocol [1] "Acurite 592TXR Temp/Humidity, 5n1 Weather Station, 6045 Lightning"
Registering protocol [2] "Honeywell Door/Window Sensor"
Registered 2 out of 90 device decoding protocols
Found 1 device(s):
  0:  Realtek, RTL2838UHIDIR, SN: 00000001

Using device 0: Generic RTL2832U OEM
Found Rafael Micro R820T tuner
Exact sample rate is: 250000.000414 Hz
[R82XX] PLL not locked!
Sample rate set to 250000.
Bit detection level set to 0 (Auto).
Tuner gain set to Auto.
Reading samples in async mode...
Tuned to 344975000 Hz.
Signal caught, exiting!

User cancel, exiting...
```

Here's the trouble -- I didn't ^C at all... after 5 seconds (the length of the -H time, if I change it to 1 it happens immediately) the command just dies instead of switching frequencies.

Am I doing something wrong?

zuckschwerdt:
This might be related to #85c4e76faa4ba226ccabb97a1a91cd3bde28e7a2
I'll need to have a look.

coogle:
I can confirm that by rolling back #530 this bug goes away:

A binary without the changes in #530:

```
$ ./rtl_433 -f 344975000 -f 433920000 -R 40 -R 70 -H 5 -F json
Registering protocol [1] "Acurite 592TXR Temp/Humidity, 5n1 Weather Station, 6045 Lightning"
Registering protocol [2] "Honeywell Door/Window Sensor"
Registered 2 out of 90 device decoding protocols
Found 1 device(s):
  0:  Realtek, RTL2838UHIDIR, SN: 00000001

Using device 0: Generic RTL2832U OEM
Found Rafael Micro R820T tuner
Exact sample rate is: 250000.000414 Hz
[R82XX] PLL not locked!
Sample rate set to 250000.
Bit detection level set to 0 (Auto).
Tuner gain set to Auto.
Reading samples in async mode...
Tuned to 344975000 Hz.
Tuned to 433920000 Hz.
[...]

PR #596, opened 2017-08-02
improve rtlsdr callback watchdog timer

zuckschwerdt:
don't restart the timer if we are aborting anyway; cancel timer on rtlsdr_cancel_async() which might be too slow to exit the async loop before the timeout. (also minior whitespace to align on 4 spaces)


zuckschwerdt:
s.a. #593, thanks to @coogle for finding this.

PR #597, opened 2017-08-12
Use default number of buffers to fix crash of two rtl_433 instances

rpeyron:
Hi,
Thanks for wonderful rtl_433 !  
I wanted to use it simultaneously two usb dongles, but the second keeps failing with "Failed to submit transfer 31!  WARNING: async read failed.".
I noticed that it works ok with the default number of USB buffers instead of 32 (keeps crashing with 31). This may be something due to my hardware, and seems related to bandwidth, as if I lower the size of buffers I can grow the number of buffers accordingly. But as I cannot see the use of a higher number of buffers as the default one and that other users could have the same issue, I suggest you this ridiculously mall change.
Thanks again.

### DIFF
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -30,7 +30,7 @@
 #define DEFAULT_FREQUENCY       433920000
 #define DEFAULT_HOP_TIME        (60*10)
 #define DEFAULT_HOP_EVENTS      2
-#define DEFAULT_ASYNC_BUF_NUMBER    32
+#define DEFAULT_ASYNC_BUF_NUMBER    0 // Force use of default value (was : 32)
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
Interesting find and sound fix. Can you report what hardware this occured on? I use as much as 4 dongles at the same time without noticing buffer related issues.

rpeyron:
I use two dongles : one is a low cost chineese RTL2832U with tuner Fitipower FC0012, and the other one is the one sold by rtl-sdr.com  RTL2832U+R820T2 1PPM TXCO.  The problem occurs with both of them, whatever is the first plugged and used. Motherboard : Z77-DS3H. The ports used are USB2 only (with no other USB devices). Hope this helps.

PR #602, opened 2017-09-12
Add Beurer BG64 Bathroom Scale

johnjore:


merbanan:
The patch is broken. Some files are completely changed.

zuckschwerdt:
Your commit touches whole files with no actual change. Did you maybe convert the line nedings by accident?
Also you can return "valid" as is, i.e. counting how many packets were decoded.

johnjore:
@zuckschwerdt: All files in the PR are changed, if you look closely. No idea why Git didn't do a proper diff, but happy to close the PR if its not of interest.

src/devices/beurer_bg64.c
```diff
@@ -0,0 +1,126 @@
...
+                //Checksum, add each nibble, modulus 16 is highest nibble of column 9
+                if (( (bb[i][0] >> 4) + (bb[i][0] & 0xf) + (bb[i][1] >> 4) + (bb[i][1] & 0xf) + 
+                      (bb[i][2] >> 4) + (bb[i][2] & 0xf) + (bb[i][3] >> 4) + (bb[i][3] & 0xf) + 
+                      (bb[i][4] >> 4) + (bb[i][4] & 0xf) + (bb[i][5] >> 4) + (bb[i][5] & 0xf) + 
+                      (bb[i][6] >> 4) + (bb[i][6] & 0xf) + (bb[i][7] >> 4) + (bb[i][7] & 0xf) + 
+                      (bb[i][8] >> 4) + (bb[i][8] & 0xf) ) % 16 != (bb[i][9] >> 4) )
+		  continue;
+
+		//bitbuffer_print(bitbuffer);
+		//fprintf(stdout, "\n");
```
merbanan: Remove the 2 lines above. Except that and the changes for the unrelated lines the code is ready the be merged.

zuckschwerdt:
You can always add commits to a PR. Or squash-commit /amend which keeps the merge history lean and is preferred.
Getting comments and reviews is a sure sign of interest if anything.
Maybe you edited the source on a Windows system? Without careful usage of some git settings (core.autocrlf ...) Git will treat line endings as significant.

src/devices/beurer_bg64.c
johnjore: Would rather keep the two comments as one of the nibbles, possibly battery, is not properly understood yet, and should be revisit in 3-4 months when the batteries have been utilized a little. As they are commented out, they have no adverse impact on performance.

PR #603, opened 2017-09-15
Add support for the Thermopro TP-12.  This is based heavily on the code

tomfelker:
for the Thermopro TP-11, and the modulation is compatible, but the bit
layout is fairly different and there are more sensors, so I've made this a
separate device.


panarom:
Can you add a couple things?
1. update README.md with new device
2. update the device count in rtl_433.h
3. add the device in Makefile.am as you did in CMakeLists.txt

merbanan:
Hi, patch looks ok but sample signals are needed.

PR #604, opened 2017-09-30
Acurite Rain Gague emit structured data

dpavlin:
I'm collecting my sensors and some others from my environment so I noticed that  Acurite Rain Gague doesn't emit JSON. I structured it based on similar rain sensors which already exist in code.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -216,8 +216,25 @@ static int acurite_rain_gauge_callback(bitbuffer_t *bitbuffer) {
     if ((bb[0][0] != 0) && (bb[0][1] != 0) && (bb[0][2]!=0) && (bb[0][3] == 0) && (bb[0][4] == 0)) {
 	    float total_rain = ((bb[0][1]&0xf)<<8)+ bb[0][2];
 		total_rain /= 2; // Sensor reports number of bucket tips.  Each bucket tip is .5mm
-        fprintf(stdout, "AcuRite Rain Gauge Total Rain is %2.1fmm\n", total_rain);
-		fprintf(stdout, "Raw Message: %02x %02x %02x %02x %02x\n",bb[0][0],bb[0][1],bb[0][2],bb[0][3],bb[0][4]);
+
+		if (debug_output > 1) {
+			fprintf(stdout, "AcuRite Rain Gauge Total Rain is %2.1fmm\n", total_rain);
+			fprintf(stdout, "Raw Message: %02x %02x %02x %02x %02x\n",bb[0][0],bb[0][1],bb[0][2],bb[0][3],bb[0][4]);
+		}
+
+		uint8_t id = bb[0][0];
+		data_t *data;
+		local_time_str(0, time_str);
+
+		data = data_make(
+			"time",	"",		DATA_STRING,	time_str,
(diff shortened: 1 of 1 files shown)


src/devices/acurite.c
```diff
@@ -216,8 +216,23 @@ static int acurite_rain_gauge_callback(bitbuffer_t *bitbuffer) {
     if ((bb[0][0] != 0) && (bb[0][1] != 0) && (bb[0][2]!=0) && (bb[0][3] == 0) && (bb[0][4] == 0)) {
 	    float total_rain = ((bb[0][1]&0xf)<<8)+ bb[0][2];
 		total_rain /= 2; // Sensor reports number of bucket tips.  Each bucket tip is .5mm
+/*
```
rct: Instead of commenting these out, wrap these in an if debug > 1, or just delete.

src/devices/acurite.c
```diff
@@ -216,8 +216,23 @@ static int acurite_rain_gauge_callback(bitbuffer_t *bitbuffer) {
...
         fprintf(stdout, "AcuRite Rain Gauge Total Rain is %2.1fmm\n", total_rain);
 		fprintf(stdout, "Raw Message: %02x %02x %02x %02x %02x\n",bb[0][0],bb[0][1],bb[0][2],bb[0][3],bb[0][4]);
+*/
+		uint8_t id = bb[0][0];
```
rct: please fix indentation.

src/devices/acurite.c
```diff
@@ -236,7 +251,7 @@ static float acurite_th_temperature(uint8_t *s){
 }
 
 // Acurite 609 Temperature and Humidity Sensor
-// 5 byte messages
+// 6 byte messages
```
rct: Why did this change?   If accidental please revert.

src/devices/acurite.c
```diff
@@ -216,8 +216,23 @@ static int acurite_rain_gauge_callback(bitbuffer_t *bitbuffer) {
...
+*/
+		uint8_t id = bb[0][0];
+    	data_t *data;
+    	local_time_str(0, time_str);
+
+		data = data_make(
+			"time",	"",		DATA_STRING,	time_str,
+			"model",	"",		DATA_STRING,	"Acurite Rain Gague",
+			"id",		"",		DATA_INT,	id,
+			"rain_total", 	"Total Rain",	DATA_FORMAT,	"%.1f mm", DATA_DOUBLE, total_rain,
```
rct: I'm not sure if there is any convention for the other rain gauges, but I *think* the units should be part of the label, not part of the data.  That way the data items shouldn't require extra string parsing to turn them back into usable values.

dpavlin: `git grep '"rain'` is so much less precise than that :-) which form would you prefer?

dpavlin:
I do not own sensor itself, since I'm listening to my sensors, and this one is just in range, so I can't really verify protocol. I'm collecting it, so I will have some idea about false detection in site time, I hope.

src/devices/acurite.c
dpavlin: what is prefered indentation? I use tabsize=4, but I also see spaces in file so I'm confused.

PR #606, opened 2017-10-03
Make LaCrosse TX141TH sensors return temperature_C so that they are more consistent with other temperature sensors

nahall:
Most other temperature sensors in RTL_433 return temperature_C, but this LaCrosse TX141TH sensor was only returning the Fahrenheit temperature in the "temperature" field, so this adds the Celsius temperature in the temperature_C field.

This is helpful for other programs like Domoticz which requests a CSV output from RTL_433 and only looks for data in the temperature_C field.


merbanan:
Can't really say what the native representation is so patch ok.

zuckschwerdt:
Actually the °C output is native. Scaling the raw value with 9/5 is the conversion. https://github.com/nahall/rtl_433/blob/master/src/devices/lacrosse_TX141TH_Bv2.c#L184
The output in °F should be removed.

PR #608, opened 2017-10-10
Added decoding of X10 CR12A ON/OFF remote commands

PalantirWielder:
If a command packet from the CR12A remote is recognized then the House [A-F], Unit [1-16] and Command [ON,OFF] is decoded and printed after the decoded bytes.
Note: Decoding of the remote's "SCAN" button was not implemented.


PalantirWielder:
The code change is relatively small. I've tested this code on my own system and found it to work without issue.
Let me know if there are any changes that need to be made such that the code is more compatible with this code base.

merbanan:
Json output support would be nice. But this looks like an improvement and is thus approved.

PR #609, opened 2017-10-11
Add Fuzhou Emax WEC-1502 Temperature sensor; Taylor, AccuTemp, ...

pjkundert:
This is the underlying sensor in many inexpensive temperature sensor/display units, such as the AccuTemp, Taylor, etc.  See: https://fccid.io/WEC-1502


merbanan:
Big patch, it will take me a while to review this.

pjkundert:
Yes, sorry -- it grew as I realized there was a major bug in bitbuffer_add_row.  I think it's worth taking the plunge, though.  Future decoders will be quite a bit simpler and more dependable, using bitbuffer_find_repeated_rows instead of just guessing that their decoded signal is on row 0 or row 1...

Also, added support for another very inexpensive and popular temperature sensor, the La Crosse TX019, Board version 0.  This is a very, very common sensor like the WEC-1502.  It uses mostly the same decoding as the La Cross TX141, but without humidity.

pjkundert:
Brought up-to-date with master; fixed a defect in another protocol, schraeder.c that prevented building...

zuckschwerdt:
This looks really promising. I'll have to check if some device relied on the (broken) assumption that sync bits produce empty rows.
Can you revert the change to `schraeder.c` as it's already merged. Thanks for the heads up there!

pjkundert:
I've corrected a bunch of decoders that were based on the assumption that the first row was always empty, but there might be more.  I ran the rtl_433_tests, and I think that it was an improvement.  The Mebus test was clearly incorrect, so it's been disabled by default 'til someone fixes it.

zuckschwerdt:
quick update: I'm adding a few more json check files to rtl_433_tests to spot regressions more easily. then this PR can be integrated.

zuckschwerdt:
There are a few device which only differ in the number of sync pulses (Esperanza, Kedsum, S3318p).
How about adding 	`uint16_t	syncbits_before_row[BITBUF_ROWS]` to `bitbuffer_t` and keep tally?

zuckschwerdt:
Some changes are already added incrementally. And the TX019-Bv0 went in, unreleated, as TX141-Bv2.
Can you reduce this to the Solight TE44 changes and split out the add_row and repeated_row changes in two more PRs? Otherwise I'll slowly work my way through this -- but I'd rather give you the attribution :)

Mindavi:
Is there anything from this still left that requires it to stay open?

zuckschwerdt:
I've been incrementally applying some of this. Still need to get around to check all suggestions are implemented -- perhaps you can take a look?

merbanan:
I went through some of the things. I can take care of the new_template.c parts.

merbanan:
x10_rf.c changes look good.

merbanan:
calibeur.c changes looks good.

merbanan:
mebus and solight needs some investigation

merbanan:
@zuckschwerdt can you look if the bitbuffer changes are sane?

zuckschwerdt:
The bitbuffer_add_row collapsing of empty rows is nice, but needs extensive testing if some decoders rely on empty rows for sync information. We do have an extra field for sync but it's not used by decoders so far I think.

Mindavi:
I did a check of the bitbuffer stuff, but can't see any obvious problems with the current implementation. Possibly needs a better look.

Mindavi:
I've applied the patch here: https://github.com/merbanan/rtl_433/pull/609/files#diff-8326cbd8def5b1e6c70234312e3e4d55a4816c7b92d0fcf3121e6bc56ef92af7R58 (line 58-60) and found the following decoders don't work as before anymore:

```
rfraw file Kedsum-TH_kedsum-60.0f-66h.rfraw does not generate any output
rfraw file Esperanza-EWS_gfile078.rfraw does not generate any output
rfraw file Ford-CarRemote_ford-unlock001.rfraw does not generate any output
rfraw file AlectoV1-Wind_gfile001_H13726A_Silvercrest_Weather_2008-Auriol_2011-Wind-event_0.2S_0.6G_315DEG.rfraw does not generate any output
rfraw file GT-WT03_g001_434.101M_250k.rfraw does not generate any output
rfraw file AlectoV1-Temperature_gfile005.rfraw does not generate any output
rfraw file RF-tech_gfile002.rfraw does not generate any output
rfraw file Emos-TTX201_g003_433.92M_250k.rfraw does not generate any output
rfraw file AlectoV1-Rain_gfile004.rfraw does not generate any output
```

This list is probably non-exhaustive. rtl_433_tests might help here.

I've also got some local work where I've generated rfraw files for most decoders. Maybe we could add something like that to the rtl_433 repository for testing?

zuckschwerdt:
> I've also got some local work where I've generated rfraw files for most decoders. Maybe we could add something like that to the rtl_433 repository for testing?

We could add the rfraw samples to the r_devices perhaps. Like #1263 but with real rf data (instead of bitbuffers) for more robust testing.

merbanan:
The patch was never updated and some of the valuable parts have been merged in other commits.

PR #612, opened 2017-10-12
brennenstuhl sensor output

ukshady:
Fixed output formatting for JSON and CSV


merbanan:
The whole code is a bit messy and should be cleaned up but this is a step towards removing fixed output and is thus approved.

PR #619, opened 2017-11-01
Add support for Philips outdoor temperature sensor

ccoff:
This sensor is used with various Philips clock radios. 

I've also submitted a pull request for the signal files over at https://github.com/merbanan/rtl_433_tests/pull/189

### DIFF
--- a/src/devices/philips.c
+++ b/src/devices/philips.c
@@ -1,0 +1,231 @@
+/* 
+ * Philips outdoor temperature sensor -- used with various Philips clock 
+ * radios (tested on AJ3650)
+ *
+ * Not tested, but these should also work: AJ7010, AJ260 ... maybe others?
+ *
+ * A complete message is 112 bits:
+ *      4-bit initial preamble, always 0
+ *      4-bit packet separator, always 0, followed by 32-bit data packet. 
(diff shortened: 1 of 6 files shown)


src/devices/philips.c
```diff
@@ -0,0 +1,231 @@
...
+    uint8_t a, b, c;
+    uint8_t packet[PHILIPS_PACKETLEN];
+    uint8_t r_crc, c_crc;
+    uint8_t channel, battery_status;
+    int tmp;
+    float temperature;
+    data_t *data;
+
+    /* Get the time */
+    local_time_str(0, time_str);
```
zuckschwerdt: only call local_time_str after the last early return, preferably just before data_make

ccoff: The debug statements in the validation checks use the time from local_time_str() as well -- do you not want the time in those? Or, I could call local_time_str() twice, once before the early returns, and again just before data_make()

zuckschwerdt: Since it's merged already consider it for future work where applicable. The reasoning is that we generally don't want all the decoders to do expensive work where it's likely not a valid packet.

ccoff: Gotcha, thanks for the review.

PR #626, opened 2017-11-10
Honeywell improvements

panarom:


src/devices/honeywell.c
```diff
@@ -46,20 +64,28 @@ static int honeywell_callback(bitbuffer_t *bitbuffer) {
 
     channel = bb[2] >> 4;
     device_id = ((bb[2] & 0xf) << 16) | (bb[3] << 8)| bb[4];
+
     event = bb[5];
-    state = (event & 0x80) >> 7;
-    heartbeat = (event & 0x04) >> 2;
+
+    if(event == 0x20) add_motion_detector(device_id);
```
mcolyer: It seems as though the logic from the [source](https://github.com/jhaines0/HoneywellSecurity/blob/18a4ac82e79cca532b1eadb33327d49a73ce7df1/rpi/digitalDecoder.cpp#L73) is slightly different, `event > 0x20`, is this intentional?

panarom: the logic from the source is different; but I have both motion sensors and door/window sensors at my house, and this is the code that appears to work for me.  OTOH, if you read through the issues in the referenced project (that cover the sensor encoding), the code here is definitely not successfully handling all potential cases: https://github.com/jhaines0/HoneywellSecurity/issues/2 .  However, since the previous honeywell.c code didn't appear to handle motion sensors at all; and since it works for me with both, I opened this PR.  I was actually gonna see if I could get the original honeywell.c contributor (adam1010) to try out my branch; but he seem's to've dropped off of github (or at least, prevented himself from being @-ed in commentary).  I can keep the honeywell changes local to my forked copy of rtl_433 if you like; but I'd still really like to get the json-formatting changes committed: so I can open a different PR for that if necessary.

mcolyer: > the code here is definitely not successfully handling all potential cases: jhaines0/HoneywellSecurity#2

Yep, as it turns out I've been using this `securityguy`'s fork of HoneywellSecurity in my own house (and it forked it [here](https://github.com/mcolyer/Honeywell5800)) as it seems to work with more of my sensors. 

My motion sensor seems to work well with rtl_433's Honeywell protocol but it doesn't seem to see any messages from my 5816 door sensors. Even if I disable the checksum messages it doesn't seem to correctly decode the 5816 byte stream. When I turn on debug mode in rtl_433 it says something about thinking it's a frequency shift encoding but I don't think that's right. I could try to debug further but I'd need some pointers about what to look for next.

> I can keep the honeywell changes local to my forked copy of rtl_433 if you like; but I'd still really like to get the json-formatting changes committed: so I can open a different PR for that if necessary.

No need to keep them out, as this change doesn't negatively impact my house as it didn't recognize my 5816's before this change either - just wanted to make sure this was intentional.

panarom: @mcolyer do you want to collaborate on further fixing up honeywell.c so that it works for both of us/incorporates more functionality referred to in jhaines0's project?

mcolyer:
> do you want to collaborate on further fixing up honeywell.c so that it works for both of us/incorporates more functionality referred to in jhaines0's project?

I don't have a ton of time at the moment, but if there's something specific you'd want me to test I could probably try it. Alternatively, these sensors are really cheap on [Amazon](https://smile.amazon.com/Ademco-5816-Honeywell-Wireless-Transmitter/dp/B0006BCCRM/ref=sr_1_1?ie=UTF8&qid=1511269075&sr=8-1&keywords=5816) if you wanted more sensors to try it out on.

panarom:
@mcolyer huh, I've got the exact same model, and it's working for me (  this is the same model you've got, correct? http://a.co/3kAFvgx ).  However, I have noticed that signal reception/decoding can be very finicky: highly dependant on antenna placement/orientation.  Do you have any loose sensors with which you can test; or are they all already mounted on doors/windows? Alternatively, is there a laptop you can use to test in closer physical proximity to a sensor?  I've also seen the case where the 'open/triggered' signal is successfully received/decoded, but the corresponding 'close/triggered' isn't; or vice versa.  If you get a chance to test them again, let me know how it turns out; and thanks for your feedback on this PR

mcolyer:
> However, I have noticed that signal reception/decoding can be very finicky: highly dependant on antenna placement/orientation.

Yeah I was wondering if that was the case. It's odd that the same hardware can decode the signals with https://github.com/mcolyer/Honeywell5800 but not rtl_433. My one theory is that Honeywell5800 sets the gain to `350` instead of using the auto value but I haven't had a chance to try setting the gain manually in rtl_433 to see if that has an effect.

> Do you have any loose sensors with which you can test; or are they all already mounted on doors/windows? Alternatively, is there a laptop you can use to test in closer physical proximity to a sensor?

I don't have any loose sensors but I do have a laptop and had the laptop right next to the sensor and it still didn't detect the signal.

> I've also seen the case where the 'open/triggered' signal is successfully received/decoded, but the corresponding 'close/triggered' isn't

I've actually got this exact case with my front door at the moment in Honeywell5800 - so I wonder if there is something else going on in that particular case.

> If you get a chance to test them again, let me know how it turns out; and thanks for your feedback on this PR

I'm traveling currently but I'll investigate further when I return and let you know.

PR #629, opened 2017-11-14
added unit (Fahrenheit) to output of Ambient Weather Temperature Sensor

exploide:
This improves consistency with most other sensor outputs and decreases confusion when wondering why it has 40 degrees in November ;)

### DIFF
--- a/src/devices/ambient_weather.c
+++ b/src/devices/ambient_weather.c
@@ -154,7 +154,7 @@ ambient_weather_parser (bitbuffer_t *bitbuffer)
             "device", "House Code", DATA_INT, deviceID,
             "channel", "Channel", DATA_INT, channel,
             "battery", "Battery", DATA_STRING, isBatteryLow ? "Low" : "Ok",
-            "temperature_F", "Temperature", DATA_FORMAT, "%.1f", DATA_DOUBLE, temperature,
+            "temperature_F", "Temperature", DATA_FORMAT, "%.1f F", DATA_DOUBLE, temperature,
             "humidity", "Humidity", DATA_FORMAT, "%u %%", DATA_INT, humidity,
             NULL);
         data_acquired_handler(data);


zuckschwerdt:
Someday we will drop units from the raw values, but for now your patch is indeed consistent.

exploide:
Agree, would be cleaner when the printing function would dispatch `temperature_C` and `temperature_F` and add the appropriate unit as necessary. I guess this requires broader changes to the code and all the device callbacks.

zuckschwerdt:
It is unclear what JSON output would be best:
- `{"temperature_C" : 12.3}`
- `{"temperature" : {"unit" : "C", "value" : 12.3}}`
- `{"temperature" : 12.3, "temperature_unit" : "C"}`

There are merrits and drawbacks either way. I somewhat favor (1) for it's brevity. But it would require string manipulation in consumers (not necessarily regex, `strrchr(x, '_')` should do).

exploide:
When I imagine writing a JSON consumer, having to deal with unpredictable key names feels dirty. On the other hand, as I saw in multiple device callbacks, there can be arbitrary (?) additional key names for a sensor, which are also kind of unpredictable.

From a JSON perspective, I like the second approach, but regarding consistency with other export formats like CSV, nested key-value pairs are not always applicable.

The third approach is very simple and straight forward but loosens the coupling between the two values.

Now, I rethought what you probably already imagined :P

Given that key names seem to be arbitrary anyhow, having different keys for different temperature units doesn't sound so bad for me, too.

zuckschwerdt:
Yes, I was reasoning something like that. Never hurts to get a fresh look though. I guess that consumers need to know keys and units beforehand in any case if they want to use the values (and if someone really just wants to log everything and not proccess it further the keys don't need to be sensible ;) )

merbanan:
patch welcome

PR #631, opened 2017-11-14
Update rtl_433.h

w1wra:
Change the size of MAX_PROTOCOLS. 

If you try to run rtl_433 -G you get the following error: 

Max number of protocols reached 94
Increase MAX_PROTOCOLS and recompile

Correct value should be 95 now that new protocol has been added.

### DIFF
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -42,7 +42,7 @@
 
 #define MINIMAL_BUF_LENGTH      512
 #define MAXIMAL_BUF_LENGTH      (256 * 16384)
-#define MAX_PROTOCOLS           94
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Some of the latest new devices must have missed this. `95` ist the right value for now.

w1wra:
I agree 95 is correct, until the next device is added. I was suggesting padding so this would not crop up again so fast.

w1wra:
I can adjust the request to 95 if it is deemed more proper.

zuckschwerdt:
Yes, ideally use `--amend` and the `push --force` to change the commit.

w1wra:
I think I did the amend/force correctly.

PR #636, opened 2017-11-23
Added support for Telldus (312623, F007TPH) and Proove (311433)

Spx1p:
thermo-/hygrometers.

### DIFF
--- a/src/devices/telldus.c
+++ b/src/devices/telldus.c
@@ -6,7 +6,7 @@
  *      1-bit starter bit
  *      48-bit data packet 
  *
- * 48-bit data packet format:
+ * 48-bit data packet format (non-inverted):
  *
  * 00000000 AAAAABBB CCCCDDDD DDDDDDDD EEEEEEEE FFFFFFFF 
  *
@@ -45,7 +45,6 @@ static int telldus_callback(bitbuffer_t *bitbuffer)
 	uint8_t r_crc, c_crc;
 	uint8_t channel;
 	uint8_t humidity;
-	uint8_t message[PACKETLEN-1];
 	int tmp;
 	float temperature;
 	data_t *data;
@@ -87,14 +86,9 @@ static int telldus_callback(bitbuffer_t *bitbuffer)
 		fprintf(stderr, "\n");
 	}
 
-	/* Read message for crc calculation */
-	for (i = 0; i < PACKETLEN-1; i++) {
-		message[i] = bb[i];
-	}
-
 	/* Correct CRC? */
 	r_crc = bb[PACKETLEN - 1]; /* Last byte in the data */
-	c_crc = crc8(message, PACKETLEN-1, 0x31, INIT_CRC);
+	c_crc = crc8(bb, PACKETLEN-1, 0x31, INIT_CRC);
 
 	if (r_crc != c_crc) {
 		if (debug_output) {
@@ -112,10 +106,10 @@ static int telldus_callback(bitbuffer_t *bitbuffer)
 
 	/* Sensor id */
 	sensor_id = (bb[2] >> 4) & 0x0f;
-	/* Let's rename sensor 0111 to 1, and 1000 to 2 */
-	if(sensor_id == 0x07)
+	/* Let's rename sensor 0111 to 1, and 1000 to 2. */
+	if(sensor_id == 0x07) /* Devices that have one sensor, it is always 0111. */
 		sensor_id = 0x01;
-	else
+	else /* Devices that have two sensors, the second one is 1000. */
 		sensor_id = 0x02;
 
 	/* Channel */
@@ -151,7 +145,6 @@ static int telldus_callback(bitbuffer_t *bitbuffer)
 						 "channel",			"Channel",		DATA_INT,		channel,
 						 "sensor", 			"Sensor id",	DATA_INT,		sensor_id,
 						 "temperature_C",	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, temperature,
-						 "humidity",		"Humidity",		DATA_STRING,	"N/A",
 						 NULL);
 	}
 	data_acquired_handler(data);
--- a/src/devices/telldus.c
+++ b/src/devices/telldus.c
@@ -31,134 +31,113 @@
 #include "rtl_433.h"
 #include "util.h"
 
-#define BITLEN 				48
-#define PACKETLEN    		6
-#define STARTBYTE	  		0x00
-#define INIT_CRC			0x2e
-
-/** telldus_crc8_bit():
- * Compute CRC-8 for a byte sequence, one bit at a time.
- *
- * This function was generated using crcany (https://github.com/madler/crcany)
- * License information: "This code is under the zlib license, permitting free
- * commercial use."
- */
-static unsigned telldus_crc8_bit(unsigned crc, void const *mem, size_t len) {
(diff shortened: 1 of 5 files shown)


src/devices/telldus.c
```diff
@@ -0,0 +1,195 @@
...
+#define INIT_CRC			0x2e
+
+/** telldus_crc8_bit():
+ * Compute CRC-8 for a byte sequence, one bit at a time.
+ *
+ * This function was generated using crcany (https://github.com/madler/crcany)
+ * License information: "This code is under the zlib license, permitting free
+ * commercial use."
+ */
+static unsigned telldus_crc8_bit(unsigned crc, void const *mem, size_t len) {
```
zuckschwerdt: Best to use thecrc8 from util.h https://github.com/merbanan/rtl_433/blob/master/include/util.h#L49

src/devices/telldus.c
```diff
@@ -0,0 +1,195 @@
...
+    return crc;
+}
+
+
+static int telldus_callback(bitbuffer_t *bitbuffer) 
+{
+    char time_str[LOCAL_TIME_BUFLEN];
+    uint8_t *bb;
+    unsigned int i;
+	uint8_t sensor_id;
```
zuckschwerdt: use 4 spaces or one tab as indent for the whole file.

src/devices/telldus.c
```diff
@@ -0,0 +1,195 @@
...
+    uint8_t r_crc, c_crc;
+    uint8_t channel;
+	uint8_t humidity;
+    int tmp;
+    float temperature;
+    data_t *data;
+
+    /* Get the time */
+    local_time_str(0, time_str);
```
zuckschwerdt: place local_time_str() after all early returns, just before data_make()

src/devices/telldus.c
Spx1p: Hum, I only use 1 tab and on vim my tabstop setting is 4, so it should be just like you described.

src/devices/telldus.c
```diff
@@ -0,0 +1,195 @@
...
+	if (humidity == 0x00) { /* if there is no humidity sensor */
+		data = data_make("time",			"",				DATA_STRING,	time_str,
+						 "model",			"",				DATA_STRING,	"Telldus/Proove thermometer",
+						 "channel",			"Channel",		DATA_INT,		channel,
+						 "sensor", 			"Sensor id",	DATA_INT,		sensor_id,
+						 "temperature_C",	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, temperature,
+						 "humidity",		"Humidity",		DATA_STRING,	"N/A",
+						 NULL);
+	} else {
+		humidity = ~humidity;
```
zuckschwerdt: looks like your whole buffer is inverted? Use bitbuffer_invert() instead. https://github.com/merbanan/rtl_433/blob/master/include/bitbuffer.h#L47

src/devices/telldus.c
zuckschwerdt: there are just a few stray tabs in this file, use only 4 spaces then. don't mix :)

Spx1p: Oh, I think what's the problem. I used one of the previous devices as a base and it differs because of that.

src/devices/telldus.c
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	/* Read message for crc calculation */
+	for (i = 0; i < PACKETLEN-1; i++) {
+		message[i] = bb[i];
+	}
+
+	/* Correct CRC? */
+	r_crc = bb[PACKETLEN - 1]; /* Last byte in the data */
+	c_crc = crc8(message, PACKETLEN-1, 0x31, INIT_CRC);
```
zuckschwerdt: you don't actually need `message` (and the copying), `bb` will suffice, or not?

src/devices/telldus.c
```diff
@@ -0,0 +1,182 @@
...
+	bitbuffer_invert(bitbuffer);
+	bb = bitbuffer->bb[1];
+
+	/* Sensor id */
+	sensor_id = (bb[2] >> 4) & 0x0f;
+	/* Let's rename sensor 0111 to 1, and 1000 to 2 */
+	if(sensor_id == 0x07)
+		sensor_id = 0x01;
+	else
+		sensor_id = 0x02;
```
zuckschwerdt: is this a catch-all for `!=0x07`? might be worth a comment that other ids never appear or something like that,

Spx1p: Usually (non inverted) message gives a sensor id 1000, but if there are two sensors on a device, the other sensor is 0111. There is a comment about it at the very top where the data packet is explained.

src/devices/telldus.c
Spx1p: You are correct, I'll clean up.

src/devices/telldus.c
```diff
@@ -0,0 +1,182 @@
...
+						 "temperature_C",	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, temperature,
+						 "humidity",		"Humidity",		DATA_FORMAT,	"%i%%", DATA_INT, humidity,
+						 NULL);
+	} else {
+		data = data_make("time",			"",				DATA_STRING,	time_str,
+						 "model",			"",				DATA_STRING,	"Telldus/Proove thermometer",
+						 "channel",			"Channel",		DATA_INT,		channel,
+						 "sensor", 			"Sensor id",	DATA_INT,		sensor_id,
+						 "temperature_C",	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, temperature,
+						 "humidity",		"Humidity",		DATA_STRING,	"N/A",
```
zuckschwerdt: best to remove this line "N/A" is not what parser for humidity data would expect. No sensor value: no output :)

src/devices/telldus.c
zuckschwerdt: ok, got it. It might be best to describe everything after the invert. that's probably how the data is intended to be. the decoder just made the wrong assumption (long/short vs 0/1).

Spx1p: That might be my fault too. But I got this working with OOK_PULSE_PWM_TERNARY modulation, so I stick with it :)

zuckschwerdt:
Interesting note in #639 that this seems to be the same as Fineoffset WH2?

Spx1p:
Fineoffset WH2 looks very similar to telldus/proove. The pulse analyzer gives two different kind of results with the test data. Some files are guessed as pulse pwm with fixed gap (pulse_demod_pwm)
`[00] {49} ff a4 38 6d ff d0 80 : 11111111 10100100 00111000 01101101 11111111 11010000 1`
and some as pulse pwm with starter bit (pulse_demod_pwm_ternary)
`[00] {0} : `
`[01] {48} 00 b1 87 b6 00 91 : 00000000 10110001 10000111 10110110 00000000 10010001`

I went for two-row version here, so I did not have to realign bytes (and calculate new hex values, for example to check crc with reveng).
Now, I did recalculate manually couple crc8 for those 49 bit versions, and WH2 actually matches here with that. Remove the first bit, and also discard the first byte (0xFF) for crc8 calculation, and it matches when using same parameters that WH2 uses.
The data is grouped almost same way as WH2, but there are differences too.

zuckschwerdt:
this was merged with #639 and the model is output too now. Please verify and open a new PR if there are things left to fix. Thanks!

PR #637, opened 2017-11-24
Stop after outputting successful event(s)

jmpesp:
If p_events > 0 and this flag is enabled, cause rtl_433 to exit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Running:
 ```
 Usage:	= Tuner options =
 	[-d <RTL-SDR USB device index>] (default: 0)
-    [-i <RTL-SDR USB device serial number (can be set with rtl_eeprom -s)>]
+	[-i <RTL-SDR USB device serial number (can be set with rtl_eeprom -s)>]
 	[-g <gain>] (default: 0 for auto)
 	[-f <frequency>] [-f...] Receive frequency(s) (default: 433920000 Hz)
 	[-H <seconds>] Hop interval for polling of multiple frequencies (default: 600 seconds)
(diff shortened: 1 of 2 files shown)


jmpesp:
diff ignoring whitespace: https://github.com/merbanan/rtl_433/pull/637/files?w=1

panarom:
@jmpesp out of curiosity, what's your usecase for this feature?  I thought about implementing this for an automated response to a wireless doorbell press; but ended up doing something different.

jmpesp:
@panarom I wrote a small golang program that execs rtl_433, takes the output (listening to a wireless temperature sensor) and writes it into InfluxDB:
```
cmd := exec.Command("rtl_433", "-R", "13", "-F", "json", "-E", "-T", "50")
```
 I'm using cron to wake up the program every minute. With 254fd2b, rtl_433 exits gracefully after outputting one "point", with the alternative being the golang program needing to kill it after successfully receiving that point. This also means that rtl_433 will return 0 on exit, meaning that any non-zero exit is a problem that I have to look at further.

zuckschwerdt:
I thought this might for testing or debug. Wouldn't running rtl_433 continously be a more robust way? Or is it to reduce load? If you feel (like me) that parsing stdout and keeping rtl_433 running is to fragile my patch for UDP output might be worth a look: https://github.com/zuckschwerdt/rtl_433/commits/develop -- I run rtl_433 from supervisord and consume the UDP in a simple bridge (to MQTT, Graphite, ...) https://github.com/zuckschwerdt/rtl_433_relay

jmpesp:
@zuckschwerdt Running rtl_433 in this way consumes about 5% of the tiny Atom CPU, it's not load :) I initially toyed with the idea of pushing json lines from a continuously running rtl_433 onto a go channel but liked the idea of cron more. I also like the idea of MQTT (thanks for the suggestion, will be looking into it), but keeping dependencies down is also a goal for this project.

PR #639, opened 2017-11-26
Add support for Telldus variants of fineoffset

peahonen:


peahonen:
Not sure if this is a duplicate of #636, only noticed that one afterwards.
Anyway, Telldus sensors looked like variants of fineoffset. Main difference being only the start bit.
Fineoffset WH2 and Telldus callbacks in this patch are pretty much the same and common parts could be refactored into one.

zuckschwerdt:
Can you detail the difference to WH2? I remember the start bit being the difference between WH2/WH5 I think.

peahonen:
WH2/5 difference is the LSB of the first byte. WH2/Telldus difference is the extra bit buffer row in the front.

To decode Telldus, you just need to change fineoffset demod_arg to 1 to make pulse_demod_pwm() to skip the extra start bit.

However, this will result in incorrect decodes with WH5 logic. So this is why I had to make a Telldus specific copy of WH2/5 decoder and remove the WH5 case there.

Should you accept the patch, the fineoffset_WH2_callback() and fineoffset_telldus_callback() could possibly refactored together to eliminate the duplication of common WH2-like structure.

zuckschwerdt:
Thanks for the implementation and explanation! I'll have a look at the sample data to see if the difference can be abstracted away without too much extra logic. Otherwise this PR is a more elegant way.

peahonen:
I could help with the refactoring as well. I just did not spend time on that after realising I might have a duplicate here.
I also now uploaded the recordings I used when working on the decoding. Sorry for no notes, but they were decoded fine with this patch.

zuckschwerdt:
I'd say it's best to roll this into the Fineoffset WH2/WH5 case.
Can you confirm, that adding at https://github.com/merbanan/rtl_433/blob/master/src/devices/fineoffset.c#L54
```
    else if (bitbuffer->bits_per_row[0] == 49 &&
            bb[0][0] == 0xFF) // Telldus
        bitbuffer_extract_bytes(bitbuffer, 0, 9, b, 40);
```
Gives the desired results? I think adding a model var to that case statement would furhter improve the code. And perhaps checking `bb[0][1]&0x80 == 0x80` would be good.

peahonen:
Yes. This actually had the same decode results. Just even more elegant way.

PR #642, opened 2017-12-07
change grab file name to gNNN_FFFM_RRRk.cu8

zuckschwerdt:
This changes the grab file name from "gfile001.data" to a recognizable pattern like "g001_433.92M_1024k.cu8":
- "g001" grab no. 1
- "433.92M" base freq. always suffixed with "M", should work well for the whole range of e.g. 0.1M to 2400M. (maybe "MHz" so other SI-prefixes could be used?)
- "1024k" sample rate. always suffixed with "k", should work well for the whole range of e.g. 50k to 3200k. (maybe "ks" (kilosamples) so other SI-prefixes could be used?)
- "cu8" universallay agreed to signify complex-unsigned-8

Did I miss something? Any other ideas? Comments very welcome!

(rtl-sdr samplerates are 250k, 1000k, 1024k, 1536k, 1792k, 1920k, 2048k, 2160k, 2560k, 2880k, 3200k -- and look all nice and recognizable, I don't know about other hardware though.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Examples:
 | `rtl_433 -p NN -R 1 -R 9 -R 36 -R 40` | Typical usage: Enable device decoders for desired devices. Correct rtl-sdr tuning error (ppm offset).
 | `rtl_433 -a` | Will run in analyze mode and you will get a text description of the received signal.
 | `rtl_433 -A` | Enable pulse analyzer. Summarizes the timings of pulses, gaps, and periods. Can be used in either the normal decode mode, or analyze mode.
-| `rtl_433 -a -t` | Will run in analyze mode and save a test file per detected signal (gfile###.data). Format is uint8, 2 channels.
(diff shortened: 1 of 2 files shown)


merbanan:
Magnificent, commit whenever.

merbanan:
Only comment, make it easy to parse, ie always the same length of the fields.

zuckschwerdt:
Yes, I was pondering a strict fixed format. But then I guess people will edit this by hand, there will be other software, who knows. A parser should split on `-`, `_`, ` ` and then see if it's a float plus suffix (`M`, `k `). Should be easy enough (I already have working support in my spectogram testbed). I just don't known if M, k will be universal enough. I *feel* that it will cover all cases (as e.g. 1M sample rate is not good, is it 1024k or 1000k? Also even kHz and GHz frequencies can be recognized plain when written in M).
I think I'll let this sit for a week for anyone to chime in with things I missed ;)

zuckschwerdt:
I have some example C parser code I'll link to this later. Basically detecting separators and `strtod(3)`.

Issue #641, opened 2017-12-07
Stack-fault using MSVC


gvanem:
After some troubles building with *clang-cl*, I get a stack-fault on a simple command like `rtl_433.exe -h`.
Since MSVC's *cl* is not supported it seems. But I do use MSVC' *link*. 

The culprit is the huge ` r_device devices[]` array on stack in `main()`.  
Linking with `-stack:20000000` fixes it (<a href="https://www.codecogs.com/eqnedit.php?latex=\fn_jvn&space;\tiny&space;20*10^6" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\fn_jvn&space;\tiny&space;20*10^6" title="\tiny 20*10^6" /></a> bytes!!).  The default stack-size is 2MByte AFAIK.

So, can this be done differently?

zuckschwerdt:
`r_device` should be 304 bytes on 64-bits (tested with clang) and there are 98 devices right now. 30k should do (I see 29792 bytes). There is something else wrong here. Can you debug where the memory is going? Maybe pepper it with fprintf...sizeof?

gvanem:
You're right. The size of `devices[]` is only 28224 bytes. I'm on 32-bit (although 64-bit Win-10).
But linking with `-stack:0x200000` fixes the crash (which happens before `main()`). So it must be something else that triggers it. I fail to see why. No constructors in the code AFAICS.

Will try MinGW next.

*PS*. You can change the `__MINGW32__` to `_WIN32` around `setenv()`.

zuckschwerdt:
Does `_WIN32` catch MinGW as well as MSVC?
Other than `sentenv()` https://github.com/merbanan/rtl_433/blob/master/src/rtl_433.c#L1035 there is also `#include "getopt/getopt.h"` in https://github.com/merbanan/rtl_433/blob/master/include/rtl_433.h#L22 -- should we update that too?

zuckschwerdt:
If you are using cmake, can you take a look at https://github.com/merbanan/rtl_433/blob/master/CMakeLists.txt#L32 and see if we can enable some compiler flags?

gvanem:
> Does _WIN32 catch MinGW as well as MSVC?

Catch? Yes, `_WIN32` is a built-in on both. If that is what you mean.

And the `rtl_433.h` WIN32-part, should IMHO be written a bit clearer. Like this:
```c
#ifdef _WIN32
  #include <windows.h>
  #include <io.h>
[...]

PR #643, opened 2017-12-07
Clean up WIN32 include/ifdef

zuckschwerdt:
By @gvanem in #641
use _WIN32 to address both MinGW and MSVC.
Cleaned up WIN32 includes in rtl_433.h.
MinGW's <unistd.h> includes <getopt.h>.

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -1032,9 +1032,13 @@ int main(int argc, char **argv) {
                 }
                 break;
             case 'U':
-#if !defined(__MINGW32__)
+#ifdef _WIN32
+                putenv("TZ=UTC+0");
+                _tzset();
+#else
                 utc_mode = setenv("TZ", "UTC", 1);
-                if(utc_mode != 0) fprintf(stderr, "Unable to set TZ to UTC; error code: %d\n", utc_mode);
+                if(utc_mode != 0)
+                    fprintf(stderr, "Unable to set TZ to UTC; error code: %d\n", utc_mode);
 #endif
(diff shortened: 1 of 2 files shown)


src/rtl_433.c
```diff
@@ -1032,7 +1032,7 @@ int main(int argc, char **argv) {
                 }
                 break;
             case 'U':
-#if !defined(__MINGW32__)
+#if !defined(_WIN32)
```
gvanem: Although  there is no `setenv()` on MinGW/MSVC, one can use `putenv()`:
```diff
--- a/src/rtl_433.c 2017-12-06 22:08:24
+++ b/src/rtl_433.c 2017-12-07 13:38:45
@@ -1032,9 +1032,13 @@
                 }
                 break;
             case 'U':
-#if !defined(__MINGW32__)
+#ifdef _WIN32
+                putenv ("TZ=UTC+0");
+                _tzset();
+#else
                 utc_mode = setenv("TZ", "UTC", 1);
-                if(utc_mode != 0) fprintf(stderr, "Unable to set TZ to UTC; error code: %d\n", utc_mode);
+                if(utc_mode != 0)
+                  fprintf(stderr, "Unable to set TZ to UTC; error code: %d\n", utc_mode);
 #endif
                 break;
             case 'W':
```

Ref: https://docs.microsoft.com/en-gb/cpp/c-runtime-library/reference/tzset

zuckschwerdt: thanks! updated.

PR #646, opened 2017-12-08
RTGN129 author name

siddolo:
Code in PR #634


zuckschwerdt:
Can you also add a header to `oregon_scientific.c`? Look at e.g. `fineoffset.c` for a template. It would be great if you can find other authors in git blame.

siddolo:
I don't know who originally wrote oregon_scientific.c so i don't know how to write the copyright row. I only added the RTGN129 support.

zuckschwerdt:
Looks like this dates back to the first days. A glances at https://github.com/merbanan/rtl_433/commits/master/src/devices/oregon_scientific.c shows maybe 8 authors. I need to do this on a rainy day then ;)

siddolo:
done

zuckschwerdt:
Thanks. The first line in `oregon_scientific.c` should be like "Various Oregon Scientific protocols" not "Fineoffset...", right?
Sorry, about the authors: I think it's best if every author opts-in to have his name and address in the AUTHORS file. Just put all the names in the oregon_scientific.c header in the Copyright line.

siddolo:
do you think that this is the right format?

zuckschwerdt:
thanks!

PR #648, opened 2017-12-09
replace -i <serial> with -d :<serial>

zuckschwerdt:
This unifies the device selection:
- no `-d` or `-d -1` : select the first available rtlsdr device
- `-d <index>` : select rtlsdr device by number
- `-d :<serial>` : select rtlsdr device by serial (replaces -i serial)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Running:
 ```
 Usage:	= Tuner options =
 	[-d <RTL-SDR USB device index>] (default: 0)
-	[-i <RTL-SDR USB device serial number (can be set with rtl_eeprom -s)>]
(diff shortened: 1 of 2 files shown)


panarom:
@zuckschwerdt so this isn't a huge deal, but given that sdr-dongle serial numbers are customizeable (e.g. could be user-set to 0, 1, 2), is there any cause for concern regarding collisions of SN/USB-device-IDs?

zuckschwerdt:
The serial is always prefixed with a colon `:` (i.e. `-d :00000001` for the factory default), the device number just a plain int.
I thought about using the more idiomatic hash`#` sign, but that likely needs to be esacaped in the shell.
The idea is to have a flexible format for `-d` that can be extended without adding ever more option switches ;)

PR #649, opened 2017-12-13
Adding interlogix devices driver

baileybrent:
Really enjoyed building this and reading/working with the rtl_433 code.  Really well done.  Thank you!

### DIFF
--- a/src/devices/interlogix.c
+++ b/src/devices/interlogix.c
@@ -1,0 +1,404 @@
+/* Interlogix/GE Wireless Device Decoder
+ *
+ * Copyright © 2017 Brent Bailey - bailey.brent@gmail.com
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ */
+
+#include "rtl_433.h"
+#include "pulse_demod.h"
+#include "util.h"
+
+
+/*
+ * Interlogix/GE Wireless 315.508 mhz Devices
+ *
+ * Found actual frequency to be -f 319511500
+ *
+ * Deocding done per us patent #5761206
+ * https://www.google.com/patents/US5761206
+ *
+ * Protocol Bits
+ * ________________________________
+ * 00-02 976 uS RF front porch pulse
+ * 03-14 12 sync pulses, logical zeros
+ * 15 start pulse, logical one
+ * 16-35 20 bit sensor identification code (ID bits 0-19)
+ * 36-39 4 bit device type code (DT bits 0-3)
+ * 40-42 3 bit trigger count (TC bit 0-2)
+ * 43 low battery bit
+ * 44 F1 latch bit NOTE: F1 latch bit and debounce are reversed.  Typo or endianess issue? TODO: check other latch bits
+ * 45 F1 debounced level
+ * 46 F2 latch bit
+ * 47 F2 debounced level
+ * 48 F3 latch bit (cover latch for contact sensors)
+ * 49 F3 debounced level
+ * 50 F4 latch bit
+ * 51 F4 debounced level
+ * 52 F5 positive latch bit
+ * 53 F5 debounced level
+ * 54 F5 negative latch bit
+ * 55 even parity over odd bits 15-55
+ * 56 odd parity over even bits 16-56
+ * 57 zero/one, programmable
+ * 58 RF on for 366 uS (old stop bit)
+ * 59 one
+ * 60-62 modulus 8 count of number of ones in bits 15-54
+ * 63 zero (new stop bit)
+ *
+ * Protocol Description
+ * ________________________________
+ * Bits 00 to 02 are a 976 ms RF front porch pulse, providing a wake up period that allows the
+ *      system controller receiver to synchronize with the incoming packet.
+ * Bits 3 to 14 include 12 sync pulses, e.g., logical 0's, to synchronize the receiver.
+ * Bit 15 is a start pulse, e.g., a logical 1, that tells the receiver that data is to follow.
+ * Bits 16-58 provide information regarding the transmitter and associated sensor. In other
+ *      embodiments, bits 16-58 may be replaced by an analog signal.
+ * Bits 16 to 35 provide a 20-bit sensor identification code that uniquely identifies the particular
+ *      sensor sending the message. Bits 36 to 39 provide a 4 bit device-type code that identifies the
+ *      specific-type of sensor, e.g., smoke, PIR, door, window, etc. The combination of the sensor
+ *      bits and device bits provide a set of data bits.
+ * Bits 40 through 42 provide a 3-bit trigger count that is incremented for each group of message
+ *      packets. The trigger count is a simple but effective way for preventing a third party from
+ *      recording a message packet transmission and then re-transmitting that message packet
+ *      transmission to make the system controller think that a valid message packet is being transmitted.
+ * Bit 43 provides the low battery bit.
+ * Bits 44 through 53 provide the latch bit value and the debounced value for each of the five inputs
+ *      associated with the transmitter. For the F5 input, both a positive and negative latch bit are provided.
+ * Bit 55 provides even parity over odd bits 15 to 55.
+ * Bit 56 provides odd parity over even bits 16 to 56.
+ * Bit 57 is a programmable bit that can be used for a variety of applications, including providing an
+ *      additional bit that could be used for the sensor identification code or device type code.
+ * Bit 58 is a 366 ms RF on signal that functions as the "old" stop bit. This bit provides compatibility with
+ *      prior system controllers that may be programmed to receive a 58-bit message.
+ * Bit 59 is a logical 1.
+ * Bits 60 to 62 are a modulus eight count of the number of 1 bits in bits 15 through 54, providing enhanced
+ *      error detection information to be used by the system controller. Finally, bit 63 is the "new" stop bit,
+ *      e.g., a logical 0, that tells the system controller that it is the end of the message packet.
+ *
+ *
+ *
+ */
+
+
+#define INTERLOGIX_CRC_POLY    0x07
+#define INTERLOGIX_CRC_INIT    0x00
+
+//preamble (and single sync bit) should be 0000 0000 0000 1
+//TODO: only searching for 0000 0001 and should be searching for all 13 bits of preamble
+static unsigned char preamble[1] = { 0x01 };
+
+static uint8_t reverse4(uint8_t  nib)
+{
+        int revnib = 0;
+
+        revnib += (nib >> 3) & 0x1;
+        revnib += (nib >> 1) & 0x2;
+        revnib += (nib << 1) & 0x4;
+        revnib += (nib << 3) & 0x8;
+
+        return(revnib);
+}
+
+//assumes little endian - find if bit is set to 1 using position 0-7 left to right
+int isNthBitSet (uint8_t c, int n) {
+        static uint8_t  mask[] = {128, 64, 32, 16, 8, 4, 2, 1};
+        return ((c & mask[n]) != 0);
+}
+
+//assumes little endian - usage printBits(sizeof(i), &i);
+void printBits(size_t const size, void const * const ptr)
+{
+        unsigned char *b = (unsigned char*) ptr;
+        unsigned char byte;
+        int i, j;
+
+        for (i=size-1;i>=0;i--)
+        {
+                for (j=7;j>=0;j--)
+                {
+                    byte = (b[i] >> j) & 1;
+                    fprintf(stderr, "%u", byte);
+                }
+        }
+        fprintf(stderr, " ");
+        //puts("");
+}
+
+char * int2bin(uint8_t i)
+{
+        size_t bits = sizeof(uint8_t) * 8;
+
+        char * str = malloc(bits + 1);
+        if(!str) return NULL;
+        str[bits] = 0;
+
+        // type punning because signed shift is implementation-defined
+        unsigned u = *(unsigned *)&i;
+        for(; bits--; u >>= 1)
+                str[bits] = u & 1 ? '1' : '0';
+
+        return str;
+}
+
+static int interlogix_callback(bitbuffer_t *bitbuffer) {
+        char time_str[LOCAL_TIME_BUFLEN];
+        uint8_t *bb;
+        data_t *data;
+        int row = 0;
+        local_time_str(0, time_str);
+
+        char device_type[1];
+        char *device_type_name = "";
+        char device_serial[6];
+        char device_raw_message[256];
+        char *low_battery = "";
+        char *f1_latch_state = "";
+        char *f2_latch_state = "";
+        char *f3_latch_state = "";
+        char *f4_latch_state = "";
+        char *f5_latch_state = "";
+
+        if (debug_output >= 1) {
+                fprintf(stderr,"%s GE/Interlogix Wireless Devices Template Callback\n", time_str);
+        }
+        if (debug_output >= 2 ) {
+
+                fprintf(stderr, "bitbuffer:: Number of rows: %d \n", bitbuffer->num_rows);
+                fprintf(stderr, "bitbuffer:: Length of 1st row: %d \n", bitbuffer->bits_per_row[row]);
+                fprintf(stderr, "bitbuffer:: Printing 1st row: \n");
+
+                //this is typically done in the bitbuffer_print function but depends upon length so doing it explicitly here
+                for (uint16_t bit = 0; bit < bitbuffer->bits_per_row[row]; ++bit) {
+                                if (bitbuffer->bb[row][bit/8] & (0x80 >> (bit % 8))) {
+                                                fprintf(stderr, "1");
+                                } else {
+                                                fprintf(stderr, "0");
+                                }
+                                //if ((bit % 8) == 7) { fprintf(stderr, " "); }   // Add byte separators
+                }
+                fprintf(stderr, "\n");
+                if (debug_output >=3)
+                        //print all rows
+                        bitbuffer_print(bitbuffer);
+        }
+
+        unsigned pos;
+
+        unsigned bit_offset = bitbuffer_search(bitbuffer, row, 0, preamble, sizeof(preamble)*8);
+
+        if (bitbuffer->bits_per_row[row] - bit_offset < 45) { //message should be at least 45 bits not including preamble bits
+                if (debug_output >= 3)
+                        fprintf(stderr, "Found valid preamble but message size too small, exiting! \n");
+                return 0;
+        }
+        else {
+               if (debug_output >= 3)
+                        fprintf(stderr, "Found valid preamble and valid message size, begin parsing \n");
+
+                //set message starting postion (just past preamble and sync bit)
+                bit_offset = sizeof(preamble)*8 + bit_offset;
+
+                uint8_t message[6];
+                unsigned msgLength = bitbuffer->bits_per_row[row]-bit_offset+2; //TODO: what happens if a different message size arrives?
+                if (debug_output >= 2)
+                        fprintf(stderr, "Message length: %d \n", msgLength);
+
+                bitbuffer_extract_bytes(bitbuffer, row, bit_offset, message, msgLength);
+
+
+                //TODO: Check message integrity (CRC/Checksum/parity) using crc8
+                //      message does not include sync bit that needs to be included in CRC calculation
+                //      missing 2 trailing bits in protocol that include CRC and stop bit?
+                if (crc8(message, 5, INTERLOGIX_CRC_POLY, INTERLOGIX_CRC_INIT) != message[5]) {
+                        if (debug_output >= 1)
+                                fprintf(stderr, "CRC Check for message payload failed! \n");
+                        //return 0;
+                }
+
+                if (debug_output >= 1) {
+
+                        //grab 6, 4 bit nibbles from bit buffer. need to reverse them to get serial number
+                        fprintf(stderr, "\nDevice Serial Number: ");
+                        fprintf(stderr, "%02x", reverse8(message[2]));
+                        fprintf(stderr, "%02x", reverse8(message[1]));
+                        fprintf(stderr, "%02x", reverse8(message[0]));
+                        fprintf(stderr, "\n");
+
+                        fprintf(stderr, "Device Message: ");
+                        printBits(sizeof(message[0]), &message[0]);
+                        printBits(sizeof(message[1]), &message[1]);
+                        printBits(sizeof(message[2]), &message[2]);
+                        printBits(sizeof(message[3]), &message[3]);
+                        printBits(sizeof(message[4]), &message[4]);
+                        printBits(sizeof(message[5]), &message[5]);
+                        fprintf(stderr, "\n");
+
+                }
+
+                sprintf(device_type, "%01x", (reverse8(message[2])>>4));
+
+                //TODO: look at conditionally parsing message for specific device types - raw message
+                //      will be passed so parsing for specific device can be done at higher layer
+                switch((reverse8(message[2])>>4)) {
+                        case 0xa: device_type_name = "contact sensor"; break;
+                        case 0xf: device_type_name = "keyfob"; break;
+                        case 0x4: device_type_name = "motion sensor"; break;
+                        case 0x6: device_type_name = "heat sensor"; break;
+
+                        default: device_type_name = "unknown"; break;
+                }
+
+
+
+
+                low_battery = (message[3] & 0x10) ? "on" : "off";
+                sprintf(device_serial, "%02x%02x%02x", reverse8(message[2]), reverse8(message[1]), reverse8(message[0]));
+                sprintf(device_raw_message, "%s%s%s", int2bin(message[3]), int2bin(message[4]), int2bin(message[5]));
+
+                f1_latch_state = (message[3] & 0x04) ? "open" : "close";
+                f2_latch_state = (message[3] & 0x01) ? "open" : "close";
+                f3_latch_state = (message[4] & 0x40) ? "open" : "close";
+                f4_latch_state = (message[4] & 0x10) ? "open" : "close";
+                f5_latch_state = (message[4] & 0x04) ? "open" : "close";
+
+
+                if (debug_output >= 1) {
+
+                        //10110110 00001000 11100101 TTT0DA01 DC01DE00 00110000
+                        if (message[3] & 0x10) { //tested
+                                fprintf(stderr, "Low Battery ");
+                        }
+                        else {
+                                fprintf(stderr, "Battery OK ");
+                        }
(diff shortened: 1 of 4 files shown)


merbanan:
Automake change are missing.

src/devices/interlogix.c
```diff
@@ -0,0 +1,404 @@
...
+
+        return str;
+}
+
+static int interlogix_callback(bitbuffer_t *bitbuffer) {
+        char time_str[LOCAL_TIME_BUFLEN];
+        uint8_t *bb;
+        data_t *data;
+        int row = 0;
+        local_time_str(0, time_str);
```
zuckschwerdt: can you move the local_time_str call down after the last early return, preferably just before data_make?

src/devices/interlogix.c
```diff
@@ -0,0 +1,404 @@
...
+                switch((reverse8(message[2])>>4)) {
+                        case 0xa: device_type_name = "contact sensor"; break;
+                        case 0xf: device_type_name = "keyfob"; break;
+                        case 0x4: device_type_name = "motion sensor"; break;
+                        case 0x6: device_type_name = "heat sensor"; break;
+
+                        default: device_type_name = "unknown"; break;
+                }
+
+
```
zuckschwerdt: can you reduce empty lines to at most 1 consecutive?

src/devices/interlogix.c
```diff
@@ -0,0 +1,404 @@
...
+}
+
+static int interlogix_callback(bitbuffer_t *bitbuffer) {
+        char time_str[LOCAL_TIME_BUFLEN];
+        uint8_t *bb;
+        data_t *data;
+        int row = 0;
+        local_time_str(0, time_str);
+
+        char device_type[1];
```
zuckschwerdt: the buffer is too short. remember the terminating \0 byte :)

src/devices/interlogix.c
```diff
@@ -0,0 +1,404 @@
...
+                        case 0x4: device_type_name = "motion sensor"; break;
+                        case 0x6: device_type_name = "heat sensor"; break;
+
+                        default: device_type_name = "unknown"; break;
+                }
+
+
+
+
+                low_battery = (message[3] & 0x10) ? "on" : "off";
```
zuckschwerdt: Look at the other decoders for usual keys. `"battery",       "Battery",     DATA_STRING, battery_low ? "LOW" : "OK",`

zuckschwerdt:
well done and documented. just a few style updates.

baileybrent:
Parsing through the changes now.  If you have a minute, take a look at the crc logic... not sure what I am missing there.

src/devices/interlogix.c
baileybrent: I wasn't sure this was actually needed and the json/csv output seemed to handle it without the terminating byte so I had left it out... adding it back now.

zuckschwerdt: you are using `sprintf` on this, which always adds a terminating byte (no proper strings without the closing 0). Maybe the actual stack memory here is aligned to 64 bits, …or you a silently trashing another vars value. OOB access is too easy in C, bad nonetheless :)

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+ *
+ *
+ *
+ */
+
+#define INTERLOGIX_CRC_POLY    0x07
+#define INTERLOGIX_CRC_INIT    0x00
+
+//preamble (and single sync bit) should be 0000 0000 0000 1
+//TODO: only searching for 0000 0001 and should be searching for all 13 bits of preamble
```
zuckschwerdt: a preamble isn't meant to be matched completly, the sender or receiver could be "warming up" and the signal distorted. You can safely write `only searching for bottom 8 bits of the 13 bits preamble`

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+static unsigned char preamble[1] = { 0x01 };
+
+static uint8_t reverse4(uint8_t  nib)
```
zuckschwerdt: is this used?

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+        revnib += (nib >> 3) & 0x1;
+        revnib += (nib >> 1) & 0x2;
+        revnib += (nib << 1) & 0x4;
+        revnib += (nib << 3) & 0x8;
+
+        return(revnib);
+}
+
+//assumes little endian - find if bit is set to 1 using position 0-7 left to right
+int isNthBitSet (uint8_t c, int n) {
```
zuckschwerdt: same, is this used?

baileybrent: nope. not at the moment

src/devices/interlogix.c
baileybrent: nope. not at the moment

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+        }
+        fprintf(stderr, " ");
+        //puts("");
+}
+
+char * int2bin(uint8_t i)
+{
+        size_t bits = sizeof(uint8_t) * 8;
+
+        char * str = malloc(bits + 1);
```
zuckschwerdt: ok, this is rather bad. malloc and then just leaking the memory (in the sprintf's below).
Just remove the whole int2bin business and sprintf("%x... hex.

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+                return 0;
+        }
+        else {
+               if (debug_output >= 3)
+                        fprintf(stderr, "Found valid preamble and valid message size, begin parsing \n");
+
+                //set message starting postion (just past preamble and sync bit)
+                bit_offset = sizeof(preamble)*8 + bit_offset;
+
+                uint8_t message[6];
```
zuckschwerdt: this buffer might be too short, you copy `msgLength` bytes below.

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+                f1_latch_state = (message[3] & 0x04) ? "open" : "close";
+                f2_latch_state = (message[3] & 0x01) ? "open" : "close";
+                f3_latch_state = (message[4] & 0x40) ? "open" : "close";
+                f4_latch_state = (message[4] & 0x10) ? "open" : "close";
+                f5_latch_state = (message[4] & 0x04) ? "open" : "close";
+
+                if (debug_output >= 1) {
+
+                        //10110110 00001000 11100101 TTT0DA01 DC01DE00 00110000
+                        if (message[3] & 0x10) { //tested
```
zuckschwerdt: all these are best condensed like: `fprintf(stderr, message[3] & 0x10 ? "Low Battery " : "Battery OK ");`

src/devices/interlogix.c
baileybrent: that would print hex.. how do I print binary?

zuckschwerdt: there is no easy way to print binary;) you'll get used to see hex as binary in your mind after a while, i guess. The decoder is meant to decode the known pieces of the signal. For development I'd just dump the whole thing (as hex) and fiddle with it in python or something.

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+
+r_device interlogix = {
+    .name          = "Interlogix GE Security Device Decoder",
+    .modulation    = OOK_PULSE_PPM_RAW,
+    .short_limit   = ((27+57)/2)*4,
+    .long_limit    = (245)*4,
+    .reset_limit   = (121)*4,
+    .json_callback = &interlogix_callback,
+    .disabled      = 0,
+    .demod_arg     = 1,
```
zuckschwerdt: PPM does not have a demod arg.

src/devices/interlogix.c
```diff
@@ -0,0 +1,392 @@
...
+ * See:
+ * - pulse_demod.h for descriptions
+ * - rtL_433.h for the list of defined names
+ *
+ */
+
+r_device interlogix = {
+    .name          = "Interlogix GE Security Device Decoder",
+    .modulation    = OOK_PULSE_PPM_RAW,
+    .short_limit   = ((27+57)/2)*4,
```
zuckschwerdt: The timing in your samples look a bit longer. Best use:
```
.short_limit = (120+240)/2, // Threshold between 120µs short and 240µs long gap
.long_limit = 360, //  Maximum gap size before new row of bits
.reset_limit = 360, // Maximum gap size before End Of Message
```

zuckschwerdt:
On the question of viewing a binary representation:
Maybe run the general purpose decoder (from the `feat-gdecode` branch)

```
rtl_433 -q -F json -X interlogix:OOK_PPM_RAW:180:360:360:41 -r interlogix/04/gfile001.cu8
```

E.g. to get all samples

```
for f in interlogix/*/*.cu8 ; do rtl_433 -q -F json -X interlogix:OOK_PPM_RAW:180:360:360:41 -r "$f" ; done
```

This gives you these unique messages:

```
{59} 00 0f 07 2a 92 0e a5 40
{59} 00 08 6c a9 90 08 a5 60
{59} 00 08 6c a9 94 0e ad 80
{59} 00 0f 07 2a 90 4a ad 80
{59} 00 0d b0 47 2e 68 ad c0
{59} 00 0d b0 47 29 08 a9 00
```

A quick way to get the binary representation:
```
$ bpython
from bitstring import BitArray
BitArray("0x 00 0f 07 2a 92 0e a5 40").bin
```
or as a 1-liner:
`python2.7 -c $'from bitstring import BitArray\nprint(BitArray("0x 00 0f 07 2a 92 0e a5 40").bin)'`

Also you might want to try:
- Jupyter
- Atom + Hydrogen
- VSCode + Jupyter

This then gives you (truncated to 59 bits):
```
00000000000011110000011100101010100100100000111010100101010
00000000000010000110110010101001100100000000100010100101011
00000000000010000110110010101001100101000000111010101101100
00000000000011110000011100101010100100000100101010101101100
00000000000011011011000001000111001011100110100010101101110
00000000000011011011000001000111001010010000100010101001000
```

Your protocol description does not mention a CRC-8, just some parity bits (aka CRC-1). How did you notice there should be a CRC?

baileybrent:
I will give the general purpose decoder a shot on a few other devices as I think it would be a good place to start for a new device.  Regarding the conversion to binary, I am essentially doing that already with some python code but would have rather have had it done in the interlogix device decoder but I am moving on for now.

My assumption was that the modulus 8 bits were in fact the CRC-8 check but I guess I stand corrected.

baileybrent:
I have question if you a minute to take a look.. I was just adding the logic to add the parity check but I see some anomalies with the output of the bitbuffer_extract_bytes function.  
`bitbuffer_extract_bytes(bits, 0, bit_offset, message, msgLength);`
The last byte in the returned message array is sometimes not correct.  To resolve this I hard coded an additional 2 bits so that the total message length is 48 (fits evenly into the byte array).
`unsigned msgLength = bits->bits_per_row[0]-bit_offset+2;`
Is there a better way to deal with this?  Should I hard code the msgLength to 48 as others do?

zuckschwerdt:
Your messages array will contain random data, it's not initialized to 0's. When you copy less than the full buffer length there will be random bits at the end.
The bitbuffer will always have 0's at unused positions. That why copying more bits zeros out your messages array.
Best practice would be to fix messages to a expected length (i.e. not use msgLength) and copy a fixed amount of data. It will make the code easier to follow and more robust. In any case your code can't handle an arbitrary length messages array. It expects a fixed amount of data. I.e. abort if there are too few bits in the bitbuffer otherwise ignore extra bits.

zuckschwerdt:
I just noticed that the indentation seems odd. Use 4 spaces per indent level and don't mix with tabs.
Also use `int` or `unsigned int` when no other type is needed, e.g. on `calcParity`, `msgParity`, `bitc`. The compiler will compute using int anyway (integer promotion, maybe read http://www.idryman.org/blog/2012/11/21/integer-promotion/ -- C will be happy to bite you any day).

baileybrent:
Interesting read and good to know...  Changes made.
Hey two other things to note. 
I would love to hear how you are doing your timing calculations.  I had to parse through audacity and manually measure the timing (and cover the samples via the bit rate).  I assume you have a better way to get these timings?
I also noticed that if you run rtl_433 without any arguments the list of devices is not in the correct order (78 is shown as interlogix devices instead of 98) but when I run with the -h argument the ordering is correct.  I haven't looked at the logic yet but wanted to note it as it could cause confusion.

zuckschwerdt:
It's Audacity and measuring with the ruler by count of samples for me too ;)

That apparent bug with the device list is due to disabled devices. Try `rtl_433 -R 52 -R 82` and you'll see what I mean.

Sorry, `uint16_t` is too short also. Broadly speaking a plain `int` will be the size of the architecture (64 bits mostly). And C will use that type to do calculations.

> Some data types like char , short int take less number of bytes than int, these data types are automatically promoted to int or unsigned int when an operation is performed on them. This is called integer promotion. For example no arithmetic calculation happens on smaller types like char, short and enum. They are first converted to int or unsigned int, and then arithmetic is done on them. If an int can represent all values of the original type, the value is converted to an int . Otherwise, it is converted to an unsigned int.

zuckschwerdt:
I hope my comments don't bother you as to nit-picky. The next contributor might just pick your code as a template, best to polish it now.
That said… the dominant and prefered style here is K&R. Opening braces should be on the same line as the statement, except for functions. https://en.wikipedia.org/wiki/Indentation_style#K&R

baileybrent:
No problem. I am just trying to pay it forward but I would note the K&R style request somewhere so the back and fourth can be reduced.  I also used uint16_t specifically as it was use the rtl_433 code base elsewhere (see `bitbuffer_print`)  Anyway.. hope my contributions helped.

baileybrent:
I was just re-reading the logic in the bitbuffer_print and I see it is limited by BITBUF_MAX_PRINT_BITS.  I see why uint16_t is better used there and not elsewhere.

zuckschwerdt:
Your guess is likely what the original author had in mind also. But modern compilers will do very interesting things to code, even if it looks deceptively simple.
Best to leave the types as general as possible and allow the compiler to optimize freely. If you force a fixed size on storage chances are the optimization will miss.
I guess `bitbuffer_print` isn't the only place that needs attention :)
Write a short C function and compile with `gcc -S snippet.c` vs `gcc -O2 -S snippet.c`. Compare ;)
E.g. this and then with `int` for `uint16_t`
```
#include <stdint.h>
int main(int argc, char **argv)
{
    for (uint16_t i = 0; i < argc; ++i) argv[i] = 0;
}
```

src/devices/interlogix.c
```diff
@@ -0,0 +1,345 @@
...
+            }
+        }
+        fprintf(stderr, "\n");
+        if (debug_output >=3)
+            //print all rows
+            bitbuffer_print(bitbuffer);
+    }
+
+    unsigned int bit_offset = bitbuffer_search(bitbuffer, row, 0, preamble, sizeof(preamble)*8);
+
```
zuckschwerdt: insert ` bit_offset += sizeof(preamble)*8;` here

src/devices/interlogix.c
```diff
@@ -0,0 +1,345 @@
...
+    if (bitbuffer->bits_per_row[row] - bit_offset < 45) //message should be at least 45 bits not including preamble bits
+    { 
+        if (debug_output >= 1)
+            fprintf(stderr, "Found valid preamble but message size too small, exiting! \n");
+        return 0;
+    }
+    else 
+    {
+        //set message starting postion (just past preamble and sync bit)
+        bit_offset = sizeof(preamble)*8 + bit_offset;
```
zuckschwerdt: remove

src/devices/interlogix.c
```diff
@@ -0,0 +1,345 @@
...
+
+    unsigned int bit_offset = bitbuffer_search(bitbuffer, row, 0, preamble, sizeof(preamble)*8);
+
+    if (bitbuffer->bits_per_row[row] - bit_offset < 45) //message should be at least 45 bits not including preamble bits
+    { 
+        if (debug_output >= 1)
+            fprintf(stderr, "Found valid preamble but message size too small, exiting! \n");
+        return 0;
+    }
+    else
```
zuckschwerdt: no need to put the code here in an `else` block

src/devices/interlogix.c
```diff
@@ -0,0 +1,345 @@
...
+    {
+        //set message starting postion (just past preamble and sync bit)
+        bit_offset = sizeof(preamble)*8 + bit_offset;
+
+        //even parity check (for bits 15-54) against even parity bit at bit 55
+        unsigned int calcParity = 0;
+        unsigned int msgParity = 0;
+        for (unsigned int bitc = bit_offset; bitc < bitbuffer->bits_per_row[row]; bitc++ )
+        {
+            if (bitc == (bit_offset+40)) { //15 bits for preamble and sync so bit 41 is actually documented bit 55
```
zuckschwerdt: wouldn't it be easier to just calculate the parity and then check the result after the loop?

src/devices/interlogix.c
```diff
@@ -0,0 +1,345 @@
...
+ * Bit 59 is a logical 1.
+ * Bits 60 to 62 are a modulus eight count of the number of 1 bits in bits 15 through 54, providing enhanced
+ *      error detection information to be used by the system controller. Finally, bit 63 is the "new" stop bit,
+ *      e.g., a logical 0, that tells the system controller that it is the end of the message packet.
+ *
+ *
+ *
+ */
+
+#define INTERLOGIX_CRC_POLY    0x07
```
zuckschwerdt: the CRC defs are not needed

src/devices/interlogix.c
```diff
@@ -0,0 +1,345 @@
...
+        {
+            byte = (b[i] >> j) & 1;
+            fprintf(stderr, "%u", byte);
+        }
+    }
+    fprintf(stderr, " ");
+}
+
+char * int2bin(uint8_t i)
+{
```
zuckschwerdt: this func does not work as intended and should be removed

baileybrent:
Let me know if you see any changes that might be needed.  The logic works great and I have been running it for a while now without any major issue.  I do see about 2-5% of the time a message is read and the message is not decoded properly (serial number usually slightly off).

baileybrent:
The patent documentation states that "Bits 60 to 62 are a modulus eight count of the number of 1 bits in bits 15 through 54, providing enhanced error detection.." but those bits are never received so it appears I am left to deal with the even/odd parity check.  Could there be another checksum that I am missing?  I ran some sample messages through RevEng but haven't been able to find anything yet.

baileybrent:
OK I reverted back to my original timings just for the heck of it and I am not seeing any problems with the decoding now.  If you have a minute, can you pull your ruler back out to see if your timing measurements could have been off?  My original timings:
`    .short_limit   = ((27+57)/2)*4,
    .long_limit    = (245)*4,
    .reset_limit   = (121)*4,`

zuckschwerdt:
Please only use µs, other values (samples/s at some samplerate) might be confusing.
True timing should be 120 us and 240 us (with 120 us pulses, i.e. 1:1 and 1:2 ratios). But the detection might lag a bit and give slightly shorter values for gaps.
`long_limt > reset_limit` means you only get one row. Which fits your signal. I'd use (120+240)/2 for short, 500 us for reset and 1000 for long too. But the values 180:360:360 will work the same.

zuckschwerdt:
Notes on output:
`"model"` `"GE Interlogix Device"`: shorten then name, drop the "Device".
The "model" should be unique and short (thinking of MQTT here), a descriptive name can be given in `r_device`.

`"device_type"`, `"device_type_name"`: choose just one. Don't output the same data multiple ways. I'd use e.g. `"sensor": "heat"` (i.e. drop the ".. sensor" suffix too). Other decoder should be able to use the same output style, so we get a mostly uniform output across devices.

`"device_serial"`: this is commonly `"id"`.
`"low_battery"`: this should mimic other devices, i.e. `"battery",       "Battery",     DATA_STRING, battery_low ? "LOW": "OK"` (this might change in the future to be a gauge 1..0 perhaps).

zuckschwerdt:
Those "latches" are DIP switches /contact pads on the board (F = fuse)? Like a house code to set a common "address"? Or is this a volatile input like a push button?

baileybrent:
Those timings are what I had to remove to resolve the misreads that were occurring from 2-5% of the time.  Specifically it looks like if I decrease the short_limit, it resolves the issue.  

The 'latches' represent switch states for the devices. Latch 3 is generally the 'tamper' switch (momentary switch on the pcb), the other latches are either reed switches or a terminal block for external switches.  From what I can tell, not all 'latches' are used on all devices GE/ITI makes.

I will work on making the formatting changes but I am hoping you can give me any insight on the checksum validation and/or what other applications due to make sure that the data is accurate.  The simple parity check doesn't provide much integrity and it looks like there isn't any other data to assist with the validation (mod 8 count bits are not present in any message).  I can always attempt to filter in my higher level application (ie; filter data that hasn't been received > 1/2/3 times with the exact same message) but my guess is that the device manufacturer is not doing this (but I could be wrong).

zuckschwerdt:
Just document the nominal timings and use whatever works best. I'll rework the PPM decoder similar to the PWM_PRECISE later.

Ok, then the prefered keys are `button1`, … to get an interface common with other devices.

I'll look at the checksums, maybe later tonight.

zuckschwerdt:
if you want to get an optimum separation between short and long gaps: run the analyzer (`-A`) on all files and in `Gap width distribution:` note the overall biggest sample count (in square brackets) for the shorter gap and the smallest sample count for the longer gap. Then calculate a middle value. Should be something like 170 µs.

baileybrent:
As a future reference, the FCC document has some decent information on the timings: [https://fccid.io/document.php?id=566521](url)

> The sensor employs amplitude modulation and transmits 61 bit or 80 bit ITI protocol.
The duty cycle correction factor will be derived for both 61 bit and 80 bit ITI protocol.
A 61 bit ITI protocol packet begins with an "ON" time of 854 mS, each bit has an "ON"
time of 122 mS, and one bit has an additional 244 mS. The total "ON" time of a single
packet is:
854 mS + 61 * 122 mS + 244 mS = 8.54 mS.
Only one packet is sent in any given 100 mS window for a duty cycle correction factor of:
20*LOG(8.54/100) = -21.4 dB
In 80 bit ITI protocol packet each bit has an "ON" time of 122 mS. The total "ON" time of
a single packet is:
80 * 122 mS = 9.76 mS.
Only one packet is sent in any given 100 mS window for a duty cycle correction factor of:
20*LOG(8.54/100) = -20.2 dB
The maximum allowed correction factor is -20.0 dB.

zuckschwerdt:
That should be µs (or us) not mS (also in your source, don't use uS)

zuckschwerdt:
For your parity checks: pull up the bit extraction, then:
```
    uint8_t message[INTERLOGIX_MSG_BIT_LEN / 8];
    bitbuffer_extract_bytes(bitbuffer, row, bit_offset, message, INTERLOGIX_MSG_BIT_LEN);

    // even data bits from message[0 .. 40] and odd data bits from message[1 .. 41]
    // i.e. 5 bytes and two (top-most) bits.
    int parity = message[0] ^ message[1] ^ message[2] ^ message[3] ^ message[4]; // parity as byte
    parity = (parity >> 4) ^ (parity & 0xF); // fold to nibble
    parity = (parity >> 2) ^ (parity & 0x3); // fold to 2 bits
    parity ^= message[5] >> 6; // add check bits
    int parity_error = parity ^ 0x3; // both parities are odd, i.e. 1 on success

    if (parity_error) {
        if (debug_output >= 1)
            fprintf(stderr, "Parity check failed (%d %d)\n", parity >> 1, parity & 1);
        return 0;
    }
```

baileybrent:
I have been testing with this logic and I have yet to see a single misread.  Thank you.  I also noticed that my ebay rtl2832u was acting a bit flaky but the parity check appears to be handling the flakiness.  My SMArt is rock solid (with or without a tuned antenna).

zuckschwerdt:
Same here, the SMArt's are solid, but the NESDR Mini Knockoffs crash when getting hot. It's better since I installed a tiny heatsink with thermalpads to the R820T2 (the chip near the antenna port that's too hot to touch…).

zuckschwerdt:
Brent, another device just took the spot of protocol 99. You get to be protocol 100🎈. Please rebase and update the PR: `git fetch upstream master; git rebase upstream/master`resolve conflicts, add, `git rebase --continue; git push -f` should work.

baileybrent:
Woot! Lucky me.  You see anything blocking a merge?  I finished my testing and have tested against 6-7 different device types that use this protocol (against a total of 43 or so devices).

zuckschwerdt:
43 devices? Your house must be a tricked out home automation geek dream.
Works and looks ok. I'm probably going to streamline it a bit (cutting down on the debugs), though. The code should be easy to glace over -- I need to maintain 100 devices, they need to be short and uniform.
The serial should be named `id`. The message[2] is both id and type? Perhaps those 4 bit should be removed? Or is the id given on the device in this way? `device_type_name` needs a catchier key. And I'd like to see the latch states as `button1`… to fit the scheme with other somewhat similar devices. Since all bits are decoded, the `device_raw_message` is a duplication, isn't it?`
(Harmonizing keys is a goal for the upcoming MQTT etc. output.)
Lastly make sure to add the `devices/interlogix.c`line last in the Makefiles etc.

baileybrent:
The devices are mostly window/door contacts so nothing that fun.

I can remove the `debug > 1` output and slim things down further but a few other items to note:
- You prefer naming 'device_type_name' to 'device_type'? 
- message[2] does contain both id and type as that is how the device is labeled on the pcb so I would prefer to leave it as it is.
- Serial is already sent as id in the output.  Are you asking to also name the internal variable as id?
- raw_message still has value as the protocol should be decoded as I have it but I have seen some minor differences between devices (some of which I do not own).  In theory if the data that is decoded, isn't correct, the person consuming the data can use their own parsing logic using the raw_message data.
- I can change to label to output as a button for these reed contact switches but it does pain me a bit.   I would prefer to leave the output as a 'switch' OPEN/CLOSED as that passes the hall test for me and anyone consuming the data can probably easily understand it (it also lines up with home automation applications like HA).  Button output, depending upon the type (momentary/latching/etc) would probably need a negative output like PRESSED/UNPRESSED or DOWN/UP if we opt to go down that route.

zuckschwerdt:
* ok
* ok 
* ah ok, csv_output_fields needs the update too ("device_serial", "low_battery")
* the person consuming wrong data should get in contact or send a patch ;)
* good point. then perhaps something like `switch1` or `contact1`?

baileybrent:
This should be complete and pretty well tested (received quite a bit of feedback already) but let me know if there are any other changes/comments.

zuckschwerdt:
sorry for the delay and thanks for the device code. You missed increasing `MAX_PROTOCOLS` and the addition to the `README.md`. I changed `device_raw_message` to just `raw_message` and updated the style somewhat -- which then uncovered a (now fixed) nasty bug in the bitbuffer function: d4dd87fdc196cd8c00d4948e959a02b8afa3c358

Issue #471, opened 2017-01-18
need a better mechansim for identifying units


matthewwall:
there is no way to know what units a decoder is sending.  the rtl_433 command has an option for units:

    -C (native | si | customary)

but what are the units of 'native'?  that is not clear from the decoder output, nor is it clear where one would go to figure it out.

what are the units of 'customary'?

also, does 'si' really mean 'si', as in meter instead of kilometer for every measure of length?  it would help both the writers of the decoders and the users of their output to know exactly which units will be returned, e.g., centimeter vs meter vs millimeter vs kilometer.  does 'si' mean that rain will be reported in millimeter but lightning strike distance will be reported in kilometers?

the decoder output should be self-identifying wrt units and unit system - the consumer of the decoder output might have no idea what was specified to the rtl_433 program.

one approach is to include a 'unit_system' in the json output.  this parameter would have a value of METRIC, METRICWX, or US.  each of these unit systems then has a clear definition of which units will be used for each type of measurement.

as things are now in rtl_433, ensuring consistent units is a nightmare.

zuckschwerdt:
Quick note:
- **native** is whatever the device delivers to the decoder
- **si** is a currently a conversion from °F to °C
- **customary** (you might know this as US, Imperial, …) is a currently a conversion from °C to °F
It would be nice to convert e.g. rain measurements, we'd need a way for decoders to report the unit first.
The JSON output will show the unit of temperature readings on the key as "_F" or "_C" postfix. This isn't perfect but other options like nested structures "{value: x, unit: y}" or unrelated keys "temperature:x, temperature_unit: y" are not much better.

matthewwall:
proposal:

start by enumerating the actual units to be used for 'si' and 'customary'.  do this in the 'how to write a decoder' guide for rtl_433 so that writers of the decoders (or those who want to fix them) know what to write.  the unit groups and their associated units would be:
```
temperature - C or F
depth (e.g., snow, rain) - mm or inch
distance (e.g., distance to lightning strike) - km or mile
altitude - meter or foot
pressure - mbar or inHg
humidity - percent
rain rate - mm/period or inch/period
```
then there are (at least) two options for communicating the units.

option 1: use the append-units-to-label pattern in every decoder.  for example:

    {"time": xxxx, "temperature_C": z}
    {"time": xxxx, "temperature_F": y} # never just 'temperature'
    {"time": xxxx, "rain_mm"}
[...]

PR #651, opened 2017-12-17
Fix unit conversions for Acurite 5n1

claws:
I recently purchased an Acurite 5n1 device and was using rtl_433 to decode the transmitted signal. I wanted units reported in metric (e.g. SI style) units. When I used the ``-C si`` command line argument I observed that only the temperature field was being modifed to metric leaving some other fields reporting in imperial (e.g. the wind speed reading in miles per hour and rain accumulation in inches). This was unexpected and annoying, so I have tried to improve rtl_433 a little to fix this problem.

The fixes in this merge request may partially address the problem decribed in #471.

These changes solve the problem I had with rtl_433.

Change description:

- In ``util.c`` and ``util.h`` add:
  - additional converter functions for inch2mm and mm2inch.
  - A string pattern replacement function (found at StackOverflow) to support field
    name modification. This function is called by the ``data_aquired_handler`` in
    ``rtl_433.c`` when performing unit conversions.
  
- In ``rtl_433.c`` update the ``data_aquired_handler`` function performing unit
  conversion to support additional conversions (mph to kph, kph to mph, inch to
  mm and mm to inch).
  No doubt other conversion routines may be useful here in future but the code as is 
  solves the problem highlighted in this issue. The updates may improve the existing
  conversion support capabilities (if fields are named appropriately) and act as
  an example for others wanting to add more in the future as required.
  
- In ``acurite.c`` modify the 5n1 message parsing function in ``acurite.c`` to adopt
  data field names that can be used within the units conversion function to swap
  fields between customary and metric (i.e. si) units.
  - Use the pretty key data field to retain the existing output labels.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -1,6 +1,6 @@
 /**
  * Various utility functions for use by device drivers
- * 
+ *
  * Copyright (C) 2015 Tommy Vestermark
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -119,5 +119,29 @@ float kmph2mph(float kph);
 /// @return speed in kilometers per hour
 float mph2kmph(float kph);
 
+
+/// Convert millimeters (mm) to inches (inch)
+///
+/// @param mm: measurement in millimeters
+/// @return measurement in inches
+float mm2inch(float mm);
+
+/// Convert inches (inch) to millimeters (mm)
+///
+/// @param inch: measurement in inches
+/// @return measurement in millimeters
+float inch2mm(float inch);
+
+
+
+
+/// Replace a pattern in a string. This utility function is
+/// useful when converting native units to si or customary.
+///
+/// @param orig: string to search for patterns
+/// @param rep: the pattern to replace
+/// @param with: the replacement pattern
+/// @return a new string that has rep replaced with with
+char *str_replace(char *orig, char *rep, char *with);
 
 #endif /* INCLUDE_UTIL_H_ */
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
very nice! some one need to tackle this :)
If kmph is native (or the closest thereof) for acurite you can drop the 2 instances of `kmph2mph() in acurite` -- you are changing the key anyway.
Can you verify if `d->format` is really static? Or do we need to `free` the old value?

claws:
With regard to the 2 ``kmph2mph`` calls, I agree that it would make sense to clean up that area of the code.  From the ``acurite.c`` file itself I can't tell what the actual native format emitted by the Acurite device is. In this case it seems dependent on what the original author chose to emit from the low level functions that read the raw values and turn them into something useful. For example, temperature is converted from raw to Fahrenheit whereas wind speed is read as cup turns and converted to kmph (probably because the conversion function was obtained from somewhere else that was using kmph).

I wanted to make a minimal change that focused on the problem I specifically had. This change leaves the default (native) output as imperial - even though one of the low level function returns metric. This meant I needed to leave the ``kmph2mph`` callls in place. I can follow up later with a separate refactor pull request to address these issues. Or I can make those changes in this merge request if that is the direction from maintainers.

I'll look into the whether d->format is static and what the native format for the 5n1 device is.

claws:
The ``data_free`` call made at the end of ``data_acquired_handler`` function in ``rtl_433.c`` does appear to free the ``d->format`` string.

claws:
I have updated the pull request with a commit that properly calls free on the ``d->format`` string that gets replaced with a new string when a conversion is applied.

zuckschwerdt:
thanks!

rct:
@zuckschwerdt there are a number of existing users of this, including those using weewx_sdr so it would have been nice to let this get some review.

zuckschwerdt:
Ok, sorry. On the frontend it's two changed labels:
- `wind_speed` -> `wind_speed_mph`
- `rainfall_accumulation` -> `rainfall_accumulation_inch`

PR #655, opened 2017-12-18
Adds a boolean data type for structured output

tecknicaltom:
This adds a DATA_BOOL type for structured output. The value should be a
C99 bool type (through <stdbool.h>) and the JSON/KV/CSV output will be
true or false. This should be helpful for fields such as a low battery
status or other binary sensor fields.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -26,12 +26,12 @@
 typedef enum {
 	DATA_DATA,		/* pointer to data is stored */
 	DATA_INT,		/* pointer to integer is stored */
+	DATA_BOOL,		/* pointer to a bool is stored */
 	DATA_DOUBLE,		/* pointer to a double is stored */
 	DATA_STRING,		/* pointer to a string is stored */
 	DATA_ARRAY,		/* pointer to an array of values is stored */
 	DATA_COUNT,		/* invalid */
 	DATA_FORMAT,		/* indicates the following value is formatted */
-	DATA_BOOL,		/* pointer to a bool is stored */
 } data_type_t;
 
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
looks very good. I'll have to think about the whole `battery`, `battery_low`, `battery_ok` / `YES`, `NO`, `OK`, `LOW` example for a few days.

tecknicaltom:
I plan to finish the implementation for the DSC sensors, which will involve (in addition to including the raw status byte) decoding the status byte to expose whether the sensor is indicating open (e.g. door/window opened), low battery, tamper sensor, heartbeat... Additionally it might be useful to expose whether the CRC check passed. A boolean data type seemed appropriate for these types of things.

rct:
A boolean would be useful as there is a range of 0/1, True/False. 

@tecknicaltom - I've got the DSC bits mapped out (I'm the original author) though I haven't been decided on naming.  I'd like to compare notes via a different issue or direct. My email is in the Authors file.

zuckschwerdt:
Yes, using boolean values totally makes sense. Though I pondering whether this will work well with consumers. I take it that time series loggers will store and aggregate numerical types well. Thus using `0` and `1` will work by default. Is there support for true booleans in e.g. Graphite, Influx, Sqlite and such? Is `battery_low: 1` or `battery_low: true` easier to process?
How about popular home automation tools, do they parse bools or ints as flags?

zuckschwerdt:
Let's maybe come back to this PR when #827 and other discussion around formatting has settled.

PR #666, opened 2018-01-14
wh1080: Fix warning about uninitialized use

frej:
Initialize 'br' to NULL in order to safely detect when an unrecognized
packet is received.


zuckschwerdt:
note though that this code is not broken (there is no path where br is used before initialized), just somewhat ugly. This fix is about the compiler warning only.
Could you actually remove l. 263-266, and add else-return0 at l.290, thanks!

frej:
Good catch, serves me right for not fully analysing the control flow. I have updated the patch to your suggestions.

Issue #30, opened 2014-04-21
Adding X10-Security protocol


wherzig:
Hi,

rtl_433 is really great. This evening I was able to add reception of the X10 security protocol  to rtl_433. I was able to receive a DS10A sensor with this (changed at 433 Mhz). It should also work with sensors for the Marmitec safeguard alarm system at 433 Mhz. See http://www.marmitek.com/de/produkt-details/home-automation-security/x-10-security/safeguard.php. I have this also at home.

It just took a few hours today. Your hints for short_limit, long_limit and reset_limit are really good. I did not think that I would be so easy!

You may add the following code to rtl_433 for this to rtl_433.c if you want to.
I am looking forward to add other protocols. Unfortunately I did only take a X10 DS10A security sensor to the hotel at vacations....

Have fun!

Regards

Willi

---

static int x10sec_callback(uint8_t bb[BITBUF_ROWS][BITBUF_COLS]) {
    /\* validate */
    if (bb[1][0]^0x0f == bb[1][1] &&
        bb[1][2]^0xff == bb[1][3]) {
        fprintf(stderr, "X10SEC: id = %02x%02x code=%02x",bb[1][0],bb[1][4],bb[1][2]);
        if (debug_output)
            debug_callback(bb);
        return 1;
    }
    return 0;
}

r_device x10sec = {
    .id             = 6,
    .name           = "X10-Security",
    .modulation     = OOK_PWM_D,
    .short_limit    = 1100,
    .long_limit     = 2200,
    .reset_limit    = 10000,
    .json_callback  = &x10sec_callback,
};

// add this to the init protocol section
    register_protocol(demod, &x10sec);

zuckschwerdt:
Implemented with #671

PR #671, opened 2018-01-25
Adds x10_sec device for decoding X10 Security RF signals

anthonykava:
Please be kind.  This addresses issue #30 and adapts code provided by the OP.  Adds an x10_sec device that decodes X10 Security signals.  I'm also providing a test to rtl_433_tests.

### DIFF
--- a/src/devices/x10_sec.c
+++ b/src/devices/x10_sec.c
@@ -25,106 +25,113 @@
 #include "data.h"
 #include "util.h"
 
-static int x10_sec_callback(bitbuffer_t *bitbuffer)
-{
-	bitrow_t *bb = bitbuffer->bb;		
+static int x10_sec_callback(bitbuffer_t *bitbuffer) {
+	bitrow_t *bb = bitbuffer->bb;
 	data_t *data;
-
-	char *remarks_str="UNKNOWN";		/* human-readable remarks */
+    uint16_t r;							/* a row index				*/
+    uint8_t *b;							/* bits of a row			*/
+	uint8_t test_a;						/* test 1/2 for message		*/
+	uint8_t test_b;						/* test 2/2 for message		*/
+	char *event_str = "UNKNOWN";		/* human-readable event		*/
+	char x10_id_str[5] = "\0";			/* string showing hex value	*/
+	char x10_code_str[5] = "\0";		/* string showing hex value	*/
 	char time_str[LOCAL_TIME_BUFLEN];
-	char x10_id_str[5]="\0";			/* string showing hex value */
-	char x10_code_str[5]="\0";			/* string showing hex value */
 
-	/* validate what we received ...
-	 * used intermediate variables test_a and test_b to suppress
-	 * warnings generated when compiling using:
-	 *
-	 *   if((bb[1][0]^0x0f)==bb[1][1] && (bb[1][2]^0xff)==bb[1][3])
-	 */
-	uint8_t test_a=bb[1][0]^0x0f;
-	uint8_t test_b=bb[1][2]^0xff;
-	if(test_a==bb[1][1] && test_b==bb[1][3])
-	{
-		/* set remarks based on code received */
-        switch(bb[1][2])
-		{
-			case 0x04:
-				remarks_str="DS10A DOOR/WINDOW OPEN";
-                break;
-			case 0x06:
-				remarks_str="KR10A KEY-FOB ARM";
-                break;
-			case 0x0c:
-				remarks_str="MS10A MOTION TRIPPED";
-                break;
-			case 0x46:
-				remarks_str="KR10A KEY-FOB LIGHTS-ON";
-                break;
-			case 0x82:
-				remarks_str="SH624 SEC-REMOTE DISARM";
-                break;
-            case 0x84:
-                remarks_str="DS10A DOOR/WINDOW CLOSED";
-                break;
-			case 0x86:
-				remarks_str="KR10A KEY-FOB DISARM";
-                break;
-			case 0x88:
-				remarks_str="KR15A PANIC";
-                break;
-			case 0x8c:
-				remarks_str="MS10A MOTION READY";
-                break;
-			case 0x98:
-				remarks_str="KR15A PANIC-3SECOND";
-                break;
-			case 0xc6:
-				remarks_str="KR10A KEY-FOB LIGHTS-OFF";
-                break;
+    for ( r=0; r<bitbuffer->num_rows; ++r) {
+        b = bitbuffer->bb[r];
+
+		/* looking for five bytes */
+        if (bitbuffer->bits_per_row[r] < 40) {
+			continue;
 		}
 
-		/* get x10_id_str, x10_code_str, and time_str ready for output */
-		sprintf(x10_id_str,"%02x%02x",bb[1][0],bb[1][4]);
-		sprintf(x10_code_str,"%02x",bb[1][2]);
-		local_time_str(0, time_str);
+		/* validate what we received ...
+		 * used intermediate variables test_a and test_b to suppress
+		 * warnings generated when compiling using:
+		 *
+		 *   if((b[0]^0x0f)==b[1] && (b[2]^0xff)==b[3])
+		 */
+		test_a = b[0]^0x0f;
+		test_b = b[2]^0xff;
+		if (test_a != b[1] || test_b != b[3]) {
+			continue;
+		} else {
+			/* set event_str based on code received */
+	        switch(b[2]) {
+				case 0x04:
+					event_str = "DS10A DOOR/WINDOW OPEN";
+					break;
+				case 0x06:
+					event_str = "KR10A KEY-FOB ARM";
+					break;
+				case 0x0c:
+					event_str = "MS10A MOTION TRIPPED";
+					break;
+				case 0x46:
+					event_str = "KR10A KEY-FOB LIGHTS-ON";
+					break;
+				case 0x82:
+					event_str = "SH624 SEC-REMOTE DISARM";
+					break;
+				case 0x84:
+					event_str = "DS10A DOOR/WINDOW CLOSED";
+					break;
+				case 0x86:
+					event_str = "KR10A KEY-FOB DISARM";
+					break;
+				case 0x88:
+					event_str = "KR15A PANIC";
+					break;
+				case 0x8c:
+					event_str = "MS10A MOTION READY";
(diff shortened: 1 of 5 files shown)


src/devices/x10_sec.c
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		/* build and handle data set for normal output */
+		data=data_make
+		(
+			"time",		"",				DATA_STRING, time_str,
+			"model",	"",				DATA_STRING, "X10 Security",
+			"id",		"Device ID",	DATA_STRING, x10_id_str,
+			"code",		"Code",			DATA_STRING, x10_code_str,
+			"remarks",	"Remarks",		DATA_STRING, remarks_str,
```
zuckschwerdt: `remarks` should probably be named `event`

src/devices/x10_sec.c
```diff
@@ -0,0 +1,137 @@
...
+			bitbuffer_print(bitbuffer);
+		}
+
+		/* build and handle data set for normal output */
+		data=data_make
+		(
+			"time",		"",				DATA_STRING, time_str,
+			"model",	"",				DATA_STRING, "X10 Security",
+			"id",		"Device ID",	DATA_STRING, x10_id_str,
+			"code",		"Code",			DATA_STRING, x10_code_str,
```
zuckschwerdt: don't output data twice. use either code or remarks.

src/devices/x10_sec.c
```diff
@@ -0,0 +1,137 @@
...
+ *
+ */
+
+#include "rtl_433.h"
+#include "data.h"
+#include "util.h"
+
+static int x10_sec_callback(bitbuffer_t *bitbuffer)
+{
+	bitrow_t *bb = bitbuffer->bb;
```
zuckschwerdt: use `uint8_t *b;`

src/devices/x10_sec.c
```diff
@@ -0,0 +1,137 @@
...
+{
+	bitrow_t *bb = bitbuffer->bb;		
+	data_t *data;
+
+	char *remarks_str="UNKNOWN";		/* human-readable remarks */
+	char time_str[LOCAL_TIME_BUFLEN];
+	char x10_id_str[5]="\0";			/* string showing hex value */
+	char x10_code_str[5]="\0";			/* string showing hex value */
+
+	/* validate what we received ...
```
zuckschwerdt: loop all rows, see new_template.c, then `b = bitbuffer->bb[r];`

src/devices/x10_sec.c
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		/* get x10_id_str, x10_code_str, and time_str ready for output */
+		sprintf(x10_id_str,"%02x%02x",bb[1][0],bb[1][4]);
+		sprintf(x10_code_str,"%02x",bb[1][2]);
+		local_time_str(0, time_str);
+
+		/* debug output */
+		if(debug_output)
+		{
```
zuckschwerdt: join all braces on the previous line.

src/devices/x10_sec.c
```diff
@@ -0,0 +1,137 @@
...
+	 * warnings generated when compiling using:
+	 *
+	 *   if((bb[1][0]^0x0f)==bb[1][1] && (bb[1][2]^0xff)==bb[1][3])
+	 */
+	uint8_t test_a=bb[1][0]^0x0f;
+	uint8_t test_b=bb[1][2]^0xff;
+	if(test_a==bb[1][1] && test_b==bb[1][3])
+	{
+		/* set remarks based on code received */
+        switch(bb[1][2])
```
zuckschwerdt: add a space between keywords and parens `if (…`

src/devices/x10_sec.c
anthonykava: Renaming to 'event'

src/devices/x10_sec.c
anthonykava: I think there's merit to having the hex code (code) and the human-readable event (will be 'event', 'remarks' in this version) as separate fields, e.g., for CSV or JSON output.  The hex code could be in the human-readable 'event' but then you have to parse for it if you're using the output.

src/devices/x10_sec.c
anthonykava: Changing to loop per new_template.c

src/devices/x10_sec.c
anthonykava: Moving braces

src/devices/x10_sec.c
anthonykava: Adding spaces

zuckschwerdt:
just add more commits to this same branch.

src/devices/x10_sec.c
```diff
@@ -0,0 +1,144 @@
...
+    uint16_t r;							/* a row index				*/
+    uint8_t *b;							/* bits of a row			*/
+	uint8_t test_a;						/* test 1/2 for message		*/
+	uint8_t test_b;						/* test 2/2 for message		*/
+	char *event_str = "UNKNOWN";		/* human-readable event		*/
+	char x10_id_str[5] = "\0";			/* string showing hex value	*/
+	char x10_code_str[5] = "\0";		/* string showing hex value	*/
+	char time_str[LOCAL_TIME_BUFLEN];
+
+    for ( r=0; r<bitbuffer->num_rows; ++r) {
```
zuckschwerdt: don't use space after opening parens

src/devices/x10_sec.c
```diff
@@ -0,0 +1,144 @@
...
+				case 0x98:
+					event_str = "KR15A PANIC-3SECOND";
+					break;
+				case 0xc6:
+					event_str = "KR10A KEY-FOB LIGHTS-OFF";
+					break;
+			}
+
+			/* get x10_id_str, x10_code_str, and time_str ready for output */
+			sprintf(x10_id_str,"%02x%02x",b[0],b[4]);
```
zuckschwerdt: insert space after every comma

zuckschwerdt:
looks good. The indentation appears funny in some lines. I'd prefer 4 spaces instead of tabs.

src/devices/x10_sec.c
```diff
@@ -0,0 +1,146 @@
...
+         * used intermediate variables test_a and test_b to suppress
+         * warnings generated when compiling using:
+         *
+         *   if((b[0]^0x0f)==b[1] && (b[2]^0xff)==b[3])
+         */
+        test_a = b[0] ^ 0x0f;
+        test_b = b[2] ^ 0xff;
+        if (test_a != b[1] || test_b != b[3]) {
+            continue;
+        } else {
```
zuckschwerdt: you don't need this `else`, which reduces the indent for everything below. Otherwise it looks perfect now!

src/devices/x10_sec.c
```diff
@@ -0,0 +1,143 @@
...
+
+        /* validate what we received ...
+         * used intermediate variables test_a and test_b to suppress
+         * warnings generated when compiling using:
+         *
+         *   if((b[0]^0x0f)==b[1] && (b[2]^0xff)==b[3])
+         */
+        test_a = b[0] ^ 0x0f;
+        test_b = b[2] ^ 0xff;
+        if (test_a != b[1] || test_b != b[3]) continue;
```
zuckschwerdt: this warning is due to integer promotion and possible sign extend on 0xff. You can get the same result without warning by:
`if ((b[0] ^ 0x0f) != b[1] || (b[2] ^ b[3]) != 0xff) continue;` or maybe
`if ((b[0] ^ b[1]) != 0x0f || (b[2] ^ b[3]) != 0xff) continue;` for consistency.

src/devices/x10_sec.c
```diff
@@ -0,0 +1,132 @@
...
+ * (at your option) any later version.
+ *
+ */
+
+#include "rtl_433.h"
+#include "data.h"
+#include "util.h"
+
+static int x10_sec_callback(bitbuffer_t *bitbuffer) {
+    bitrow_t *bb = bitbuffer->bb;
```
zuckschwerdt: this var is also unused. everything else looks perfect.

zuckschwerdt:
Sorry, didn't see it earlier: you need to increment `#define MAX_PROTOCOLS 98` in https://github.com/merbanan/rtl_433/blob/master/include/rtl_433.h#L46

zuckschwerdt:
anytime ;)
With the new looping, did you notice there are 10 packets, 5x window open, then 5x window close?
This is an interesting case as the events are not separated but in a single "stream". I might add some utility logic to condense this into just to events.
Also did you notice this signal in fact has 41 bits? The last one might be the parity on the last byte (or just a stop bit).
Can you record more samples with different conditions?

anthonykava:
Yes, even before the looping, I was seeing 5x CLOSE and 5x OPEN in that sample, and I see that every time.  I believe the the X10 sensors always transmit events multiple times in case of reception errors.  I've seen this on the W800 RF receiver too.

I did not notice the 41-bits.  I'm going to try recording more samples this weekend with a more appropriate antenna (although the stock RTL-SDR antenna did just fine at short range).

zuckschwerdt:
It is common to see the same packet repeated for reliability. I wonder however why open and close are in the same run of packets? Did you quickly open and close?
The reception is fine btw, it seems rather strong. If you want to picture the signal quickly, just drop it into http://triq.net/iqs to get a spectrogram view.

anthonykava:
The sample is based on pressing the 'TEST' button on a DS10A sensor.  The test button generates a CLOSE event immediately followed by an OPEN event.  I can easily capture samples of the reed switch itself in operation with a delay between events.  Reception should be rather strong at a distance of 20 cm! I'll upload more samples.

anthonykava:
Some quick samples with separated OPEN and CLOSE events:

[x10_sec.ds10.tests.2018-01-26.2015Z.tar.gz](https://github.com/merbanan/rtl_433/files/1669074/x10_sec.ds10.tests.2018-01-26.2015Z.tar.gz)

zuckschwerdt:
So a normal signal is repeated 5 times, which could be used to increase reliability. And a close/open sequence indicates the test function (and is not useful for normal operation).
The protocol mentions that the additional bit could be a parity on the last byte `8e`. If you happen to change the address, be sure to look if that bit ever changes. (Use `rtl_433 -q -F json -R 0 -X x10sec:OOK_PPM_RAW:1100:2000:10000 -r FILE` to quickly look at the bytes.

PR #672, opened 2018-01-26
Spelling fix: gague -> gauge

brycied00d:


zuckschwerdt:
Thanks. There is apparently no test data for this device in rtl_433_tests (otherwise that would need changing too).
Do you have this device and can you supply perhaps 3-5 samples?

brycied00d:
> Do you have this device and can you supply perhaps 3-5 samples?

Sorry! I only noticed the spelling mistake while poking around.

PR #674, opened 2018-02-03
Fix on_connect signature and topic subscriptions in MQTT examples.

linville:
Paho changed the signature of `on_connect` to require 4 arguments as of [v1.3.0](https://github.com/eclipse/paho.mqtt.python/issues/197) (breaking change in Paho, previously it was optional).

Remove the `/#` from the subscription since it wouldn't match the example topic which didn't include anything after rtl_433.

### DIFF
--- a/tests/mqtt_rtl_433_test.py
+++ b/tests/mqtt_rtl_433_test.py
@@ -48,14 +48,14 @@ def print_sensor_state():
     sys.stdout.flush()  # Print in real-time
 
 
-def on_connect(client, userdata, rc):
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks. Are you adventurous enough to try the `feat-mqtt` branch? It works with either Paho or Mosquitto. The (re)connect needs work and it's not fully async right now, but that fixed it will be merged in the next weeks.

linville:
Thanks, I am checking it out now!

PR #676, opened 2018-02-09
Fix finding Paho and Mosquitto cmake modules on case sensitive file systems

linville:
find_package(..) is case sensitive. Change case to match `cmake/Modules/FindMosquitto.cmake` and `cmake/Modules/FindPaho.cmake`.

Original code worked on macOS (APFS, case insensitive) but failed on Ubuntu (ext4).


zuckschwerdt:
Thanks!

PR #683, opened 2018-02-13
New devices: Euroster 3000TX, Elro DB270

rp3tya:


zuckschwerdt:
The current idea is to offer "doorbell"-style devices where no meaningful data can be extracted as simple "recipes". I.e.
`-X "ELRO_DB270_Wireless_Doorbell:OOK_PWM:300:930:11000,repeats>=4,bits>=25"`

... which reminds me: gap_limit is missing from the flex decoder.

zuckschwerdt:
both now work with the flex decoder
`-X "Euroster_3000TX:OOK_MC_ZEROBIT:1000:0:4800,bits=32"`

Would you be ok to start a file `HOWTO.md` (in the root folder) with a short description, the flex examples and links to the test files?

rp3tya:
Not sure i understand what you mean with the howto file...

The thermostat works great with the general decoder, but the doorbell picks up all kind of devices and doesn't output the right content.

zuckschwerdt:
Yes, the code was broken. I just fixed it, there should be no more random output.
For the file: make a section for each device, write someting about the device, the decoder line and some sample output -- if you want :)

rp3tya:
Great!

For the file: something like [this](https://github.com/rp3tya/rtl_433/blob/master/decoders.cfg)?

zuckschwerdt:
I was thinking of a simple README-like file. This is even better! We can add a parser and automate things (later this year maybe). Very nice.

zuckschwerdt:
@rp3tya can you remove the other files from the PR so I can merge `decoders.cfg`?

PR #687, opened 2018-02-16
change array separator in csv to ";"

zuckschwerdt:
otherwise RadioHead ASK output will break csv parsers

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -513,6 +513,15 @@ static void print_csv_data(data_output_t *output, data_t *data, char *format)
     --csv->data_recursion;
 }
 
+static void print_csv_array(data_output_t *output, data_array_t *array, char *format)
+{
+    for (int c = 0; c < array->num_values; ++c) {
+        if (c)
+            fprintf(output->file, ";");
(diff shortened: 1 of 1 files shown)


merbanan:
This might upset some people.

merbanan:
Maybe let it be configurable ?

zuckschwerdt:
ah, no. i worded that badly. CSV is still Comma separated ;) This is about the nested separator in data arrays. No device except RadioHead ASK uses this: https://github.com/merbanan/rtl_433/blob/master/src/devices/radiohead_ask.c#L145

zuckschwerdt:
I.e.
```
time,model,id,nested_array,foo
@1,test_no_array,42,,bar
@2,test,42,[a,b,c,d,e],bar
```
now is properly
```
time,model,id,nested_array,foo
@1,test_no_array,42,,bar
@2,test_with_array,42,a;b;c;d;e,bar
```
Sorry if I gave anyone a heart attack ;)

merbanan:
All good then.

PR #700, opened 2018-03-29
Added decoder for Dish Network UHF Remote 6.3

dtiller:
Please pull this decoder into your codebase. I also uploaded test recordings and data as well as photos in the rtl_433_tests project.

### DIFF
--- a/src/devices/dish_remote_6_3.c
+++ b/src/devices/dish_remote_6_3.c
@@ -1,0 +1,156 @@
+/* Decoder for UHF Dish Remote Control 6.3, tested with genuine Dish remote.
+ *
+ * Copyright (C) 2018 David E. Tiller
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * The device uses PPM encoding,
+ * 0 is encoded as 400 us pulse and 1692 uS gap,
+ * 1 is encoded as 400 us pulse and 2812 uS gap.
+ * The device sends 7 transmissions per button press approx 6000 uS apart.
+ * A transmission starts with a 400 uS start bit and a 6000 uS gap.
+ *
+ * Each packet is 16 bits in length.
+ * Packet bits: BBBBBB10 101X1XXX
+ * B = Button pressed, big-endian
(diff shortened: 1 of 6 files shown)


src/devices/dish_remote_6_3.c
```diff
@@ -0,0 +1,156 @@
...
+static char *output_fields[] = {
+    "time",
+    "button",
+    NULL
+};
+
+r_device dish_remote_6_3 = {
+    .name          = "Dish remote 6.3 decoder",
+    .modulation    = OOK_PULSE_PPM_RAW,
+    .short_limit   = (SHORT_GAP + LONG_GAP) / 2,
```
zuckschwerdt: Please put the values here and remove the defines

src/devices/dish_remote_6_3.c
```diff
@@ -0,0 +1,156 @@
...
+        return 0;
+    }
+
+    button = b[0] >> 2;
+    button_string = button_map[button];
+	
+    local_time_str(0, time_str);
+
+    data = data_make(
+            "time",  "", DATA_STRING, time_str,
```
zuckschwerdt: add at least a model string, look at other devices for common output

dtiller:
I made the requested changes and pushed the code and test data changes. 

Thanks for reviewing my pull request.

dtiller:
Requested changes made and pushed up.

dtiller:
Requested changes made and pushed up.

zuckschwerdt:
Increase https://github.com/merbanan/rtl_433/blob/master/include/rtl_433.h#L46
and add the new device from `rtl_433 -h` to the README.txt.
My advice on `model` wasn't too useful, I guess. It should be a concise form of the r_device name. "Dish remote 6.3" should be ok, but dots, slashes an alike are not great with e.g. MQTT. Can't really say what the most fitting string would be. Don't use brand, it will be replaced with something like "class" someday.
Last thing: there is no message integrity (MIC), so `.disabled` should be `= 1`. Thanks!

dtiller:
Christian,

Changes made - code and related tests pushed. 

Thanks!

PS - I'm working on another decoder, but it's driving me nuts! I can't seem to get the correct parameters. Of course 'rtl_433 -a' gives me perfect data, but I can't seem to replicate that with a -X flex or the actual driver params. I'll keep at it.

PR #703, opened 2018-04-11
eusahues' changes to add currentcost counter devices added as per htt…

strowger:
I didn't write this patch, I got it from eusahues (at https://github.com/EusaHues/rtl_433/commit/a4188a36f0a38e8e84ea99117bd51e7c6bc8584b ) and updated it slightly.

It extends the existing currentcost device support, to support "counter" type devices. These are currentcost devices which count pulses from a light, or magnet on a meter, such as the flashing light on an electricity meter or pulse output on a gas meter.

Please would you add it?

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -48,6 +48,7 @@
 		DECL(hideki_ts04) \
 		DECL(oil_watchman) \
 		DECL(current_cost) \
+		DECL(current_cost_counter) \
 		DECL(emontx) \
 		DECL(ht680) \
 		DECL(s3318p) \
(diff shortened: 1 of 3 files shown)


EusaHues:
#blameeusa not sure why I didn’t submit it myself, sorry!

include/rtl_433_devices.h
```diff
@@ -48,6 +48,7 @@
 		DECL(hideki_ts04) \
 		DECL(oil_watchman) \
 		DECL(current_cost) \
+		DECL(current_cost_counter) \
```
zuckschwerdt: Always add devices at the end to keep the current numbering

merbanan:
Do we have sample signals for this ?

EusaHues:
Suggest this one be closed, and I'll deal with any further fixes required on https://github.com/merbanan/rtl_433/pull/751

PR #704, opened 2018-04-16
Update acurite.c to hide lightning sensor diagnostic messages

mlfreeman2:
only print lightning sensor debug output when a debug flag is specified.
otherwise these messages would get in the way of my efforts to use the json output mode

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -431,17 +431,19 @@ static int acurite_6045_decode (bitrow_t bb, int browlen) {
     strike_distance = bb[7] & 0x1f;
     l_status = (bb[7] & 0x60) >> 5;
 
-    printf("%s Acurite lightning 0x%04X Ch %c Msg Type 0x%02x: %.1f F %d %% RH Strikes %d Distance %d L_status 0x%02x -",
-	   time_str, sensor_id, channel, message_type, tempf, humidity, strike_count, strike_distance, l_status);
+    if (debug_output) {
+        printf("%s Acurite lightning 0x%04X Ch %c Msg Type 0x%02x: %.1f F %d %% RH Strikes %d Distance %d L_status 0x%02x -",
+	    time_str, sensor_id, channel, message_type, tempf, humidity, strike_count, strike_distance, l_status);
 
-    // FIXME Temporarily dump raw message data until the
-    // decoding improves.  Includes parity indicator(*).
-    for (int i=0; i < browlen; i++) {
-	char pc;
-	pc = byteParity(bb[i]) == 0 ? ' ' : '*';
-	fprintf(stdout, " %02x%c", bb[i], pc);
+	// FIXME Temporarily dump raw message data until the
+        // decoding improves.  Includes parity indicator(*).
+	for (int i=0; i < browlen; i++) {
+	    char pc;
+	    pc = byteParity(bb[i]) == 0 ? ' ' : '*';
+	    fprintf(stdout, " %02x%c", bb[i], pc);
+	}
+	printf("\n");
     }
-    printf("\n");
 
     valid++;
     return(valid);


zuckschwerdt:
Are you interested in converting that to `data_make()` code?

mlfreeman2:
I thought about it, but then i noticed it didn’t match up with what my Acurite HD display showed for the lightning stats.

The HD display said 59 strikes, with the last one 5 miles away and this said 69 strikes with the last one 7 miles away.

Not sure how to go about resolving an apparent decoding error.

 

On Apr 16, 2018, at 2:09 AM, Christian W. Zuckschwerdt <notifications@github.com> wrote:

Are you interested in converting that to data_make() code?

zuckschwerdt:
Ok, no problem. Feel free to send an updated decoding if you figure it out. Thanks for this patch.

rct:
Reading it quickly, but doesn't that patch effectively disable all lightning detector output?    This person has labeled it as debugging output, but it is the only data output.

Yes, it's not in structured data format.  But there are people using it as it is.   Besides myself, who wrote it,  I believe weewx is using the output in the text form.

The device has a cumulative 7 bit strike counter (so it wraps at 127.)   That part of the decoding I'm pretty confident in.   The distance calc and 2 of the 4 are still somewhat of a mystery.

mlfreeman2:
I figured since there was a //FIXME in the code and it wasn’t structured that someone had forgotten to add the if (debug_output) wrapper.

If you want to undo my change and restore the output to always being present, I would suggest clarifying the docs for the -F option.

The docs say this: "[-F] kv|json|csv Produce decoded output in given format. Not yet supported by all drivers.
		 append output to file with :<filename> (e.g. -F csv:log.csv), defaults to stdout.”

I interpreted “Not yet supported by all drivers” as “Nobody has added formatted support to them, so their output won’t be there at all”, not “You may encounter plain text strings of random format in between legal KV/JSON/CSV, so code extra defensively when consuming the output”.





On Apr 16, 2018, at 11:26 AM, rct <notifications@github.com> wrote:

Reading it quickly, but doesn't that patch effectively disable all lightning detector output? This person has labeled it as debugging output, but it is the only data output.

Yes, it's not in structured data format. But there are people using it as it is. Besides myself, who wrote it, I believe weewx is using the output in the text form.

The device has a cumulative 7 bit strike counter (so it wraps at 127.) That part of the decoding I'm pretty confident in. The distance calc and 2 of the 4 are still somewhat of a mystery.

mlfreeman2:
In regards to your other comment about the strike counter rolling over, is it possible that the station tracks when it received the lightning reports (hence why it sees 59 in the last 48 hours but the unit has seen 69 since it last rolled over)?

I do believe the temperature and humidity readings are probably accurately decoded based on my other sensors out near it.
 
I do wonder the mystery of the distance though. I can upload packet captures of protocol 40 if you want, but you'd have to wade through reports from my 5-in-1 and a couple indoor Acurite sensors unless you have a way I can tell devices apart when i run with the -a -t options. It didn't seem apparent to me.

zuckschwerdt:
Sudden breaking changes are not intended and not wanted, sorry for that. However the `acurite_6045_decode()` needs work. It's disabled by default and not a major problem but random text to stdout along with `data_make()` is rather unexpected.
Can someone with access to this device supply a `data_make()` PR?

mlfreeman2:
I got data_make working locally with my lightning detector and submitted a PR for that.
I didn’t mess with the decoding of the raw data though, so the last strike distance still does not match with my display.

I live in Florida, so unless we have an absurdly dry summer, I’ll have plenty of chances to debug the lightning detector further.


On Apr 16, 2018, at 1:11 PM, Christian W. Zuckschwerdt <notifications@github.com> wrote:

Sudden breaking changes are not intended and not wanted, sorry for that. However the acurite_6045_decode() needs work. It's disabled by default and not a major problem but random text to stdout along with data_make() is rather unexpected.
Can someone with access to this device supply a data_make() PR?

PR #705, opened 2018-04-16
got data_make working with the 6045m

mlfreeman2:
I decided to try to get data_make working.
I didn't change the decoding process at all. 

The original plain text message is available with debugging turned on.

If i have the time I'll try to figure out why the last strike distance differs from my Acurite HD Display and mess with the decoding in another PR.

Here's a sample of what `-F json` outputs now.
`{"time" : "2018-04-16 19:00:01", "model" : "Acurite Lightning 6045M", "sensor_id" : 28271, "channel" : "B", "temperature_F" : 64.200, "humidity" : 39, "strike_count" : 69, "strike_dist" : 7}`

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -1046,7 +1046,7 @@ r_device acurite_txr = {
     .gap_limit      = 500,  // longest data gap is 392 us, sync gap is 596 us
     .reset_limit    = 4000, // packet gap is 2192 us
     .json_callback  = &acurite_txr_callback,
-    .disabled       = 1,
+    .disabled       = 0,
     .demod_arg      = 0,    // not used
 };
 
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -421,8 +421,10 @@ static int acurite_6045_decode (bitrow_t bb, int browlen) {
     char channel_str[2];
     uint16_t sensor_id;
     uint8_t strike_count, strike_distance;
+    data_t *data;
 
     channel = acurite_getChannel(bb[0]);  // same as TXR
+    sprintf(channel_str, "%c", channel);
     sensor_id = (bb[1] << 8) | bb[2];     // TBD 16 bits or 20?
     humidity = acurite_getHumidity(bb[3]);  // same as TXR
     message_type = (bb[4] & 0x60) >> 5;  // status bits: 0x2 8 second xmit, 0x1 - TBD batttery?
@@ -445,7 +447,20 @@ static int acurite_6045_decode (bitrow_t bb, int browlen) {
 	printf("\n");
     }
 
+    data = data_make(
+       "time",			"",			DATA_STRING,	time_str,
+       "model",	        	"",			DATA_STRING,	"Acurite Lightning 6045M",
+       "sensor_id",		"",			DATA_INT,	sensor_id,
+       "channel",  		"",     		DATA_STRING, 	&channel_str,
+       "temperature_F", 	"temperature",		DATA_FORMAT,	"%.1f F", 	DATA_DOUBLE, 	tempf,
+       "humidity",         	"humidity",		DATA_INT,	humidity,
+       "strike_count",          "lightning_count",	DATA_INT, 	strike_count,
+       "strike_dist",        	"last_distance",	DATA_INT, 	strike_distance,
+     NULL);
+
+    data_acquired_handler(data);
     valid++;
+
     return(valid);
 }
 
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -450,8 +450,8 @@ static int acurite_6045_decode (bitrow_t bb, int browlen) {
     data = data_make(
        "time",			"",			DATA_STRING,	time_str,
        "model",	        	"",			DATA_STRING,	"Acurite Lightning 6045M",
-       "sensor_id",		"",			DATA_INT,	sensor_id,
-       "channel",  		"",     		DATA_STRING, 	&channel_str,
+       "sensor_id",    		NULL,  			DATA_FORMAT,    "0x%02X",   DATA_INT,       sensor_id,
+       "channel",  		NULL,     		DATA_STRING, 	&channel_str,
        "temperature_F", 	"temperature",		DATA_FORMAT,	"%.1f F", 	DATA_DOUBLE, 	tempf,
        "humidity",         	"humidity",		DATA_INT,	humidity,
        "strike_count",          "lightning_count",	DATA_INT, 	strike_count,
@@ -548,11 +548,12 @@ static int acurite_txr_callback(bitbuffer_t *bitbuf) {
                     "time",			"",		DATA_STRING,	time_str,
                     "model",	        	"",		DATA_STRING,	"Acurite tower sensor",
                     "id",			"",		DATA_INT,	sensor_id,
-                    "channel",  		"",     	DATA_STRING, 	&channel_str,
-                    "temperature_C", 	"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, tempc,
-                    "humidity",         "Humidity",	DATA_INT,	humidity,
-                    "battery",          "Battery",    	DATA_INT, 	battery_low,
-                    "status",		"",		DATA_INT,	sensor_status,
+                    "status",			"",		DATA_INT,	sensor_status,
+		    "sensor_id",    		NULL,  		DATA_FORMAT,    "0x%02X",   DATA_INT,       sensor_id,
+                    "channel",  		NULL,     	DATA_STRING, 	&channel_str,
+                    "temperature_C", 		"Temperature",	DATA_FORMAT,	"%.1f C", DATA_DOUBLE, tempc,
+                    "humidity",         	"Humidity",	DATA_INT,	humidity,
+                    "battery",          	"Battery",    	DATA_INT, 	battery_low,
                     NULL);
 
             data_acquired_handler(data);


zuckschwerdt:
@rct does this fit your use-case, can weewx-sdr be changed to handle this?
I'd say `sensor_id` should be `id` and `strike_dist` should be postfixed with a unit, not sure though.

mlfreeman2:
I went back and forth between “id” and “sensor_id” and ended up choosing “sensor_id” based on what I saw elsewhere in the file. 

I’m not sure whether the strike distance is miles or kilometers. How could we find out?

mlfreeman2:
Ok, I enabled the Acurite protocol by default. It took me a little while to figure out what you meant (I got sick earlier this week).

I also added a PR in weewx-sdr to support this change there: https://github.com/matthewwall/weewx-sdr/pull/26

I fixed up the acurite tower sensor handling too - weewx-sdr likes the presence of a sensor_id field.

zuckschwerdt:
Great! We need to check for regressions with the rtl_433_tests repo, otherwise looks good.
(Every other sensor use `id` where here `sensor_id` is used. This needs some thought.)

mlfreeman2:
weewx-sdr wants sensor_id to exist and be a 4-character hex string. It uses the JSON field name plus that plus its own internal name for the data packet to essentially name each sensor reading it detects (e.g. `humidity.6E6F.AcuriteLightningPacket`, `temperature.1140.AcuriteTowerPacket`). You then put those names into weewx.conf to tell weewx-sdr which sensor values to feed upward.

I'm not sure what requirements `id` needs to follow but that's what I learned about `sensor_id` (and thus why I kept it).

rct:
I need to make some changes.   I'll try to get them done this weekend.  IF you want to merge this, go ahead.    What needs to be addressed:

* My understanding of what constitutes the sensor ID has changed.   I want to get this fixed as close as possible to the JSON change because it will affect how the sensor is identified and how data is stored.  People will need to do some sort of mapping/renaming to get the data from before and after to match if they are doing any historical storage. 
* The battery flag is wrong
* All of the flag/status bits are now lost, so they aren't available without turning on debug and having to separately correlate. 
* Strike Distance is wrong name and value. First according to the sensor datasheet, that is the distance to what the sensor estimates as the leading edge of the storm.   As far as what units that is in, I don't know yet.  I don't have an Acurite console.  I don't have enough data from people with Acurite consoles to come up with a mapping. The few data points users have related didn't seem to have an obvious pattern.

Other notes: 
* Strike count, is just the number of strike events the sensor has counted.  7 bit counter with wraparound.  It is non-volatite.  @merbanan  I don't think there is a unit put put there.

rct:
I think all of the sensor_id's should be changed to just "id" for consistency, but I don't have all of the output in front of me.    I think the tower sensor is correct in this regard, but the 5n1 is wrong.  Note this will be a non-upward compatible change that will affect 5n1 users.

One thing I've thought would be handy is to have some place in the test repo, the output of running all of the test samples for the current version to be used as a resource until we have some better data dictionary.

rct:
Adding (mostly for my own reference) the current state of Acurite JSON output:

```
{"time" : "@0.000000s", "model" : "Acurite 5n1 sensor", "sensor_id" : 839, "channel" : "A", "sequence_num" : 0, "battery" : "OK", "message_type" : 49, "wind_speed_mph" : 2.679, "wind_dir_deg" : 337.500, "wind_dir" : "NNW", "rainfall_accumulation_inch" : 0.000, "raincounter_raw" : 66}
{"time" : "@0.000000s", "model" : "Acurite tower sensor", "id" : 12053, "channel" : "C", "temperature_C" : 26.700, "humidity" : 74, "battery" : 0, "status" : 68}
{"time" : "@0.000000s", "model" : "Acurite 609TXC Sensor", "id" : 202, "battery" : "OK", "status" : 2, "temperature_C" : 26.200, "humidity" : 76}
{"time" : "@0.000000s", "model" : "Acurite Rain Gauge", "id" : 162, "rain" : 513.500}
{"time" : "@0.524288s", "model" : "Acurite 606TX Sensor", "id" : -93, "battery" : "OK", "temperature_C" : -12.800}
{"time" : "@0.000000s", "model" : "00275rm", "probe" : 0, "id" : 3253, "battery" : "OK", "temperature_C" : 16.700, "humidity" : 23, "mic" : "CRC"}
```

rct:
This doesn't really belong here, but also for reference, I pulled a sample of the JSON output from the tests repo of any sensor that reported humidity just do get an idea of whether there was any consensus

```
{"time" : "@0.000000s", "model" : "LaCrosse TX Sensor", "id" : 48, "humidity" : 31.000}
{"time" : "@0.000000s", "model" : "Ambient Weather F007TH Thermo-Hygrometer", "device" : 37, "channel" : 5, "battery" : "Low", "temperature_F" : 67.800, "humidity" : 35}
{"time" : "@0.000000s", "model" : "LaCrosse TX141TH-Bv2 sensor", "id" : 67, "temperature_C" : 9.300, "humidity" : 73, "battery" : "OK", "test" : "No"}
{"time" : "@0.000000s", "brand" : "OS", "model" : "THGR122N", "id" : 106, "channel" : 2, "battery" : "OK", "temperature_C" : 28.300, "humidity" : 26}
{"time" : "@0.524288s", "model" : "AlectoV1 Temperature Sensor", "id" : 44, "channel" : 1, "battery" : "OK", "temperature_C" : 28.800, "humidity" : 36, "mic" : "CHECKSUM"}
{"time" : "@1.048576s", "model" : "Bresser 3CH sensor", "id" : 230, "channel" : 1, "battery" : "OK", "temperature_F" : 68.600, "humidity" : 65, "mic" : "CHECKSUM"}
{"time" : "@0.524288s", "model" : "Esperanza EWS", "id" : 8, "channel" : 1, "temperature_F" : 39.400, "humidity" : 1}
{"time" : "@0.000000s", "model" : "HIDEKI TS04 sensor", "rc" : 4, "channel" : 1, "battery" : "OK", "temperature_C" : 24.100, "humidity" : 65}
{"time" : "@0.000000s", "model" : "Fine Offset Electronics, WH2 Temperature/Humidity sensor", "id" : 209, "temperature_C" : 24.600, "humidity" : 33, "mic" : "CRC"}
{"time" : "@0.000000s", "model" : "Telldus/Proove thermometer", "id" : 167, "temperature_C" : 24.200, "humidity" : 53, "mic" : "CRC"}
{"time" : "@0.000000s", "model" : "Fine Offset Electronics WH1080/WH3080 Weather Station", "msg_type" : 0, "id" : 133, "temperature_C" : 5.600, "humidity" : 97, "direction_str" : "NE", "direction_deg" : "45", "speed" : 0.000, "gust" : 0.000, "rain" : 335.100, "battery" : "OK"}
{"time" : "@1.048576s", "model" : "inFactory sensor", "id" : 198, "temperature_F" : 66.500, "humidity" : 77}
{"time" : "@0.000000s", "model" : "Fine Offset WH5 sensor", "id" : 3, "temperature_C" : 25.700, "humidity" : 55, "mic" : "CRC"}
{"time" : "@0.524288s", "model" : "TFA-Twin-Plus-30.3049", "id" : 4, "channel" : 0, "battery" : "OK", "temperature_C" : 14.100, "humidity" : 6}
{"time" : "@0.524288s", "model" : "Oregon Scientific SL109H", "id" : 188, "channel" : 3, "temperature_C" : 20.800, "humidity" : 64, "status" : 4, "mic" : "CHECKSUM"}
{"time" : "@0.524288s", "model" : "Prologue sensor", "id" : 9, "rid" : 78, "channel" : 1, "battery" : "OK", "button" : 0, "temperature_C" : 1.000, "humidity" : 204}
{"time" : "@1.048576s", "model" : "Mebus/433", "id" : 24, "battery" : "LOW", "channel" : 4, "unknown1" : 1, "unknown2" : 11, "temperature_C" : -150.300, "humidity" : -79}
{"time" : "@0.000000s", "model" : "WT450 sensor", "id" : 13, "channel" : 1, "battery" : "OK", "temperature_C" : 25.640, "humidity" : 14}
```

src/devices/acurite.c
```diff
@@ -450,8 +450,8 @@ static int acurite_6045_decode (bitrow_t bb, int browlen) {
     data = data_make(
        "time",			"",			DATA_STRING,	time_str,
        "model",	        	"",			DATA_STRING,	"Acurite Lightning 6045M",
-       "sensor_id",		"",			DATA_INT,	sensor_id,
-       "channel",  		"",     		DATA_STRING, 	&channel_str,
+       "sensor_id",    		NULL,  			DATA_FORMAT,    "0x%02X",   DATA_INT,       sensor_id,
```
rct: Note, two issues with this, first, the size is greater than 8 bits.  since these are int values, might as well output them as decimal like the others, instead of a hex string.

mlfreeman2: Ok, understood. Will you change that in the PR you said you'll work on or should I change it?

mlfreeman2:
I do have an Acurite HD console, which appears to be the only official interpreter of the 6045m’s output that I’ve seen on acurite.com

As I mentioned earlier, during my last round of thunderstorms, my Acurite HD console was displaying 5 miles as the “distance” report from the lightning sensor but rtl_433 was picking up the “distance” field as 7 (units unknown).
This was before I started doing anything.

I *thought* it might be 7 KM but that doesn’t convert into 5 miles - closer to 4. 
There’d have to be a floating point value being transmitted and some aggressive rounding going on in the console after conversion to make that possibly work I figured.
(According to Google, 7.24220893 km is 4.5001 miles, which could be rounded up to 5 by the Acurite console).

Also, the sensor data sheet may say it’s an “estimate to the leading edge of the storm” but the Acurite HD console distinctly calls it the “last strike distance” so that’s why I named it what I did.




On Apr 21, 2018, at 11:09 AM, rct <notifications@github.com> wrote:

I need to make some changes. I'll try to get them done this weekend. IF you want to merge this, go ahead. What needs to be addressed:

My understanding of what constitutes the sensor ID has changed. I want to get this fixed as close as possible to the JSON change because it will affect how the sensor is identified and how data is stored. People will need to do some sort of mapping/renaming to get the data from before and after to match if they are doing any historical storage.
The battery flag is wrong
All of the flag/status bits are now lost, so they aren't available without turning on debug and having to separately correlate.
Strike Distance is wrong name and value. First according to the sensor datasheet, that is the distance to what the sensor estimates as the leading edge of the storm. As far as what units that is in, I don't know yet. I don't have an Acurite console. I don't have enough data from people with Acurite consoles to come up with a mapping. The few data points users have related didn't seem to have an obvious pattern.
Other notes:

Strike count, is just the number of strike events the sensor has counted. 7 bit counter with wraparound. It is non-volatite. @merbanan <https://github.com/merbanan> I don't think there is a unit put put there.

rct:
@merbanan, @zuckschwerdt - What's the preference for battery Ok/Low going forward?

Most of the Acurite output uses "battery" : "OK", but I think the intended direction is that this should really be an int.  I thought there was some talk of adding boolean type to data_make, but I don't seen anything there yet.

src/devices/acurite.c
rct: I've got it.

rct:
@mlfreeman2 - See PR #707.  I still want to add more but wanted to get you something.

src/devices/acurite.c
```diff
@@ -445,7 +447,20 @@ static int acurite_6045_decode (bitrow_t bb, int browlen) {
 	printf("\n");
     }
 
+    data = data_make(
+       "time",			"",			DATA_STRING,	time_str,
+       "model",	        	"",			DATA_STRING,	"Acurite Lightning 6045M",
+       "sensor_id",    		NULL,  			DATA_FORMAT,    "0x%02X",   DATA_INT,       sensor_id,
+       "channel",  		NULL,     		DATA_STRING, 	&channel_str,
```
rct: FYI, that's a bug.  Looks like you might have tried, &channel, but that won't work because there is DATA_CHAR, so you need a null terminated string.  so you changed back to channel_str, but left the address of operator.   I just happened to notice I was getting the wrong channel.

mlfreeman2:
Yeah, I see that and it works for me. 
Closing this one.

Issue #610, opened 2017-10-11
rtl_433 only works once. Need to pull and re-insert dongle to run again


james-fry:
Hi,

I'm using the latest rtl_433 that I built today (ubuntu 16.04)
I am seeing a strange problem where rtl_433 works and runs the first time it executes after the RTL dongle is inserted, however any subsequent attempt to execute I get this:

```
Registered 91 out of 91 device decoding protocols
Found 1 device(s):
  0:  Realtek, RTL2838UHIDIR, SN: 00000001

Using device 0: Generic RTL2832U OEM
Found Fitipower FC0012 tuner
Exact sample rate is: 250000.000414 Hz
Sample rate set to 250000.
Bit detection level set to 0 (Auto).
Tuner gain set to Auto.
Reading samples in async mode...
Tuned to 433920000 Hz.
Signal caught, exiting!

User cancel, exiting...
```

i.e. immediate exit of rtl_433.
Pulling out and re-inserting the DVB dongle clears this issue - maybe because the usb device number increments when I do this?

from syslog:
Oct 11 22:20:44 hassio kernel: [86329.258326] usb 1-1: USB disconnect, device number **5**
Oct 11 22:20:50 hassio kernel: [86335.326443] usb 1-1: new high-speed USB device number **6** using ehci-pci
Oct 11 22:20:50 hassio kernel: [86335.755231] usb 1-1: New USB device found, idVendor=0bda, idProduct=2838
Oct 11 22:20:50 hassio kernel: [86335.755234] usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
Oct 11 22:20:50 hassio kernel: [86335.755236] usb 1-1: Product: RTL2838UHIDIR
Oct 11 22:20:50 hassio kernel: [86335.755237] usb 1-1: Manufacturer: Realtek
Oct 11 22:20:50 hassio kernel: [86335.755238] usb 1-1: SerialNumber: 00000001

I think its a software issue in either rtl_433 or in rtl_sdr because an older version that I found in this docker image works fine: https://hub.docker.com/r/sysrun/rpi-rtl_433/
(It was built a year ago according to docker hub - you can see support for 67 devices vs the current 91 from above)

```
Registered 67 out of 67 device decoding protcols
Found 1 device(s):
  0:  Realtek, RTL2838UHIDIR, SN: 00000001

Using device 0: Generic RTL2832U OEM
Found Fitipower FC0012 tuner
[...]

PR #708, opened 2018-04-22
Adjust alarm from 1 second to 3 second

iceman73:
Fixes issue 610
https://github.com/merbanan/rtl_433/issues/610

Fixes crash when in docker that causes need to reinsert dongle.

Tested for 48hrs and no crashes or resets are seen.


merbanan:
win32 is untested by me so patch accepted.

PR #711, opened 2018-04-23
Add conversion hPA/inHG, kPa/PSI

zuckschwerdt:
- add automatic conversion hPA/inHG, kPa/PSI
- extends automatic conversion to all C/F fields
- minor style updates

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -12,6 +12,7 @@
 #define INCLUDE_UTIL_H_
 
 #include <stdint.h>
+#include <stdbool.h>
 #include <time.h>
 
 // Helper macros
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
I can already do it this time ;)

include/util.h
```diff
@@ -12,6 +12,7 @@
 #define INCLUDE_UTIL_H_
 
 #include <stdint.h>
+#include <stdbool.h>
```
merbanan: Is this needed for this or leftovers from the bool patch ?

zuckschwerdt:
It's needed for `bool str_endswith()` (util.h/.c) -- fits well and C99 supports it, but could also return an int.

merbanan:
Ok, all good then.

merbanan:
Can we merge this ?

PR #713, opened 2018-04-26
Add device: Kerui Door Sensor

jpbarraca:
Added support for Kerui Door Sensors such as the ones available here: http://www.keruistore.com/kerui-g18-app-control-wireless-gsm-home-security-burglar-alarm-system-outdoor-siren.php

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ Supported device protocols:
     [99]  X10 Security
     [100]  Interlogix GE UTC Security Devices
     [101]* Dish remote 6.3
+    [102]  Kerui Door Sensor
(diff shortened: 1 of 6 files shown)


zuckschwerdt:
sorry, we don't add any doorbell-style sensors as standalone anymore. Please use the general-purpose decoder and submit your device to the https://github.com/merbanan/rtl_433/blob/master/decoders.cfg -- it's a single line not 5 files touched ;)

jpbarraca:
I agree. I didn't knew about that functionality. 
After some testing I will commit a new PR.

PR #716, opened 2018-05-01
Fix typos/misspellings

Cabalist:
Mostly just common typos.  Just comments and strings.

This includes a cleanup to the sole Python script in the project.


zuckschwerdt:
Amazing how many typos there are. Thank you for being so thorough. "Celcius" vs "Celsius" will change the output column labels for CSV. I doubt anyone is using those literal, though.

merbanan:
Looks good.

Cabalist:
Thanks for the fast turnaround on this!

PR #717, opened 2018-05-01
Add Pacific PMV-C210 TMPS support

zuckschwerdt:


zuckschwerdt:
DMC works well enough for now. Will be merged in a few days.

PR #720, opened 2018-05-14
Add AcuRite 3-n-1 support

zuckschwerdt:
WIP, don't merge. The decoding doesn't seem right. Needs testing with controlled conditions.


zuckschwerdt:
Sufficiently analyzed in #801.

PR #722, opened 2018-05-18
convert dsc output to JSON

dmcwhinnie:
Not the prettiest way to create JSON, but it was only a single printf statement so escaped the quotes and added the {}

### DIFF
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -147,7 +147,7 @@ static int DSC_callback(bitbuffer_t *bitbuffer) {
         local_time_str(0, time_str);
 
         if (crc8le(bytes, DSC_CT_MSGLEN, DSC_CT_CRC_POLY, DSC_CT_CRC_INIT) == 0) {
-            printf("%s DSC Contact ESN: %06X, Status: %02X, CRC: %02X\n",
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
It would be grerat if you could convert this device to the `data_make()` format. JSON is for free then :)
(just push more commits on top in this PR.)

zuckschwerdt:
Can you test if #725 (branch [`feat-dsc`](https://github.com/zuckschwerdt/rtl_433/tree/feat-dsc)) works for you?

dmcwhinnie:
Sorry for the delay... Yes, I will test, but may not be until next week.

rct:
FWIW, I have a version that does data_make and I've broken out the bits in a way that I think is sensible (to me as the original author), but I was waiting on the boolean issue, since the output is basically boolean.

rct:
with PR #730 the output looks like the following.  This is somewhat temporary aid for the transition between the old unstructured data and structured.  The status bits are much easier to interpret when output as hex.   I will break out the bits in a new PR within a few days.

```
{"time" : "2018-05-23 09:50:16", "model" : "DSC Contact", "id" : 3520320, "esn" : "35b740", "status" : 161, "status_hex" : "a1", "mic" : "CRC"}
```

PR #725, opened 2018-05-19
Convert DSC to structured data

zuckschwerdt:
s.a. #722

### DIFF
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -73,11 +73,17 @@
 #define DSC_CT_CRC_POLY        0xf5
 #define DSC_CT_CRC_INIT        0x3d
 
-
-static int DSC_callback(bitbuffer_t *bitbuffer) {
-    bitrow_t *bb = bitbuffer->bb;
-    int valid_cnt = 0;
+static int dsc_callback(bitbuffer_t *bitbuffer)
+{
     char time_str[LOCAL_TIME_BUFLEN];
+    data_t *data;
+    uint8_t *b;
+    int valid_cnt = 0;
+    uint8_t bytes[5];
+    uint8_t status, crc;
+    uint32_t esn;
+    char status_str[3];
+    char esn_str[7];
 
     if (debug_output > 1) {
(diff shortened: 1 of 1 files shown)


src/devices/dsc.c
```diff
@@ -144,17 +146,25 @@ static int DSC_callback(bitbuffer_t *bitbuffer) {
...
+            continue;
         }
+
+        sprintf(status_str, "%02x", status);
+        sprintf(esn_str, "%06x", esn);
+        data = data_make(
+                "time", "", DATA_STRING, time_str,
+                "model", "", DATA_STRING, "DSC Contact",
+                "id", "", DATA_STRING, esn_str,
+                "status", "", DATA_STRING, status_str,
```
rct: While it is easier to read as hex, status would be easier to process as an int.

Same thing with ID.    The "natural" way for the system to identify devices is with the hex string, but in reality it is a 24 bit unsigned int.   So if I wanted to just read the output the hex string is what I'd want to see.  But for processing it would be easier for it to just be an int.

src/devices/dsc.c
```diff
@@ -144,17 +146,25 @@ static int DSC_callback(bitbuffer_t *bitbuffer) {
         esn = (bytes[1] << 16) | (bytes[2] << 8) | bytes[3];
         crc = bytes[4];
 
-        local_time_str(0, time_str);
```
rct: this got deleted so time_str is never set.

rct:
I'll take this as a start and add fixes to it.

src/devices/dsc.c
zuckschwerdt: Totally agree. It's a number an should be readable as such. But some devices give the ID in a special format and users expect that. TPMS come to mind, they print the hex ID on the device.
We can get away with just treating it as a "label" for most applications, but the same does not apply for the status -- we really want to parse that. Complicated issue to solve in general way.

rct:
See PR #730 - which I think represents a better "transitional" structure for the move to structured data.

PR #728, opened 2018-05-23
Regularize and extend Acurite 6045M comments

gdt:
Regularize the comemnts about the Acurite 6045M, in both formatting
and more importantly in listing bitmasks for every field, including
unknown fields.  While there is almost no semantic change, hoist the
current understanding of the "uknown strike status bit" from code to
also be in the comments.


zuckschwerdt:
@rct can comment and review this.

rct:
@gdt @zuckschwerdt - This looks good.  LGTM.  Thanks!

(I do have some updates, but I will be submitting a new PR to widen the ID field from 12 bits to 14 bits based on the data provided by @gdt and Rich T.)

PR #730, opened 2018-05-23
DSC data_make additions.

rct:
* Merge branch 'master' of https://github.com/merbanan/rtl_433 into feat-dsc (@zuckschwerdt's initial data_make conversion)
* Output ID and Status byte as INTs for easy parsing.
* "transitional" data fields as an aid:
  * Add ESN back as a hex ID for readers/consumers that want to see a hex string.
  * Temporarily outputting status as hex until the bits are broken out.

### DIFF
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -159,12 +159,12 @@ static int dsc_callback(bitbuffer_t *bitbuffer)
         local_time_str(0, time_str);
 
         data = data_make(
-		"time", "", DATA_STRING, time_str,
-		"model", "", DATA_STRING, "DSC Contact",
-		"id", "", DATA_INT, esn,
-		"esn", "", DATA_STRING, esn_str,
-		"status", "", DATA_INT, status,
-		"status_hex", "", DATA_STRING, status_str, // TEMP DELETE once bits are output
+                "time", "", DATA_STRING, time_str,
+                "model", "", DATA_STRING, "DSC Contact",
+                "id", "", DATA_INT, esn,
+                "esn", "", DATA_STRING, esn_str, // to be removed - transitional
+                "status", "", DATA_INT, status,
+                "status_hex", "", DATA_STRING, status_str, // to be removed - once bits are output
                 "mic", "", DATA_STRING, "CRC",
                 NULL);
         data_acquired_handler(data);
--- a/src/devices/intertechno.c
+++ b/src/devices/intertechno.c
@@ -3,8 +3,7 @@
 static int intertechno_callback(bitbuffer_t *bitbuffer) {
     bitrow_t *bb = bitbuffer->bb;
 
-      //if (bb[1][1] == 0 && bb[1][0] != 0 && bb[1][3]==bb[2][3]){
-      if(bb[0][0]==0 && bb[0][0] == 0 && bb[1][0] == 0x56){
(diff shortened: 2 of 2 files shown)


rct:
Note: the actual alarm system uses ESN and hex digits to identify sensors (aka contacts).   While the new output has redundancy between INT and hex strings, I think it's a reasonable temporary compromise.

zuckschwerdt:
Duplicate output (id/esn, status/status_hex) should okay for a transitional period. But please mark it `// to be removed` -- it would set a bad example for other decoders otherwise. There should only be one generally accepted/most useful output.
Since `id` isn't used in an "arithmetic" way I guess `esn_str` is good there, if thats also marked on the devices. For `status` either format is okay.

For consistency indentation should be 4 spaces in this file.

rct:
@zuckschwerdt - 
* converted tabs back to spaces, so hopefully at least that block should be consistent.  
* Added the requested comment to the redundant field. 
* I don't see the point of adding _str to ESN.  We don't add type qualifiers to other fields, only units.  So adding a data type designation here seems like it would be arbitrary.

rct:
circling back to my previous comment.  the ID is an int (24 bit unsigned int).  so I think it should be left as an int that should be a natural data type and easy to consume.  I think most of the other devices I've touched have an ID that is an integer value. 

Having the ESN as a hex string is a crutch for people (like myself) who might be looking at the output rather than parsing it.

As requested I marked that field as `// to be removed - transitional`

What do I have to do to get this merged?

I'm not happy with your change to output the ID as a hex string and I would prefer others don't wind up using it.

zuckschwerdt:
I don't have strong opinions about that. Like I said int is fine, most devices use that, yes. I just meant to note that string is also ok since we don't expect consumers to do arithmetic on that information.
Let this sit for a day so others can get in on the conversation otherwise I'm good with merging.

Also we finally have tags for somewhat stable versions, currently "18.05". The next version will only be tagged when all new changes are solid and tested.

rct:
@zuckschwerdt  - 

> Let this sit for a day so others can get in on the conversation otherwise I'm good with merging.

I would be fine with that if you waited before committing your own changes that I as the author of this and user of this code don't agree with.

If you want to let it sit for review than revert your change please.

zuckschwerdt:
Something went wrong in the communication there. I took your "I'll take this as a start and add fixes to it." as the go ahead.
Let's just do the coding without the calling out.

rct:
I'm sorry, I should have been more explicit about "don't merge" because I'm going to provide additional fixes.

But where we currently stand is:

* you made a change that breaks the output for any existing user of this decoder. 
* I expressed reservations about the identifier, which is fairly fundamental, but you merged anyway.
* I've provided some fixes that seem like they are a reasonable compromise for upward compatibility - it restores ESN as a hex string, and provides id as an int for new usage.
* I did this with some urgency because I don't agree with your treatment of the identifier.  

Now I'm respectfully requesting that you please either merge my changes or revert your own to fix the identifier.  I feel this should be done before anyone starts using your new identifier.

I'm sorry if you feel this is "calling you out", but I feel like I've got a vested interest in this as the original author of the DSC code and someone who has been using this for over 2 years now.  (I'm pretty sure I submitted the DSC stuff in 2015.)

dmcwhinnie:
I have tested this latest merge and everything looks good to me.

Issue #727, opened 2018-05-23
Gather Statistics


zuckschwerdt:
Picking up the idea from @rct, capturing statistics for the decoders:

rtl_433 should keep statistics on the decoder operations.
Rough outline how to do that:

For an easy integration the return code from the decoders can be extended with the following semantics:

- each invocation will be accounted
- the number of successful decodes will be counted (return value != 0)
- the number of processed messages will be counted (sum return value if > 0)
- the number of failed decodes will be counted (return value <= 0)
- decode failed with bad message length (return FAIL_LENGTH_ = -1)
- decode failed on early checks, maybe a prefix (return FAIL_EARLY = -2)
- decode failed because of the message integrity check (return FAIL_MIC = -3)
- decode failed because of sanity check  (return FAIL_SANITY = -4)

The statistics could be shown at exit, periodically, at exit or e.g. with SIGUSR1.

This is just a quick idea to get this going. All comments and suggestions welcome. I'll turn this into some experimental code soon.

gdt:
For stats, I see this as being like SNMP and counting errors in network packets.   Your list seems like a good starting point and I think we should expect it to change as a few decoders are enhanced.  Two suggestions:
 - there are sometimes fixups to data to make things decode when something is off, or at least comments indicate an extra zero that shouldn't be there, and this is notable
 - the situation of cleanly recognizing a packet as not for us is different from thinking it is and finding a problem.  Specifically I am thinking of length/checksum and then parity/value-range in Acurite, just because that's the code I have read.   Your list captures this, but I wanted to emphasize it.

gdt:
Also, there probably should be stats for the demodulators (vs. decoders), and these will be quite different, even if they are integrated into the same stats output method.

zuckschwerdt:
Good points. A difference between "decoder not applicable" and "packet broken /decoder failed" should be made clear.
If in some decoders "fixups" are needed that is likely a kludge and best handled with debug printf's -- the root cause needs to be found and solved.

rct:
So what I have in mind as the potential use cases:
1. For running (static) tests on recorded data after changes, we should be able to identify and evaluate which decoders are called and the results.  I think this will add another, dimension besides just comparing data output to help evaluate whether the matching is working correctly.
1. For long term runs (i.e. normal usage) - be able to collect data that would let us evaluate whether there are any improvements to be made.  Are tolerances too tight or too loose?  Are decoders not getting called for some buffers where they should?  
[...]

PR #733, opened 2018-05-24
Add stats reporting

zuckschwerdt:
early WIP, not for merge. s.a. #727

Adds an experimental undocumented option `-L` for stats reporting:
- 0: do not report (default)
- 1: report successful devices
- 2: report active devices
- 3: report all

Test with e.g. `rtl_433 -q -F json -L 1 -T 60`

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Usage:		= General options =
   [-F kv | json | csv | syslog | null | help] Produce decoded output in given format.
        Append output to file with :<filename> (e.g. -F csv:log.csv), defaults to stdout.
        Specify host/port for syslog with e.g. -F syslog:127.0.0.1:1514
-  [-M time | reltime | notime | hires | utc | protocol | level | bits | help] Add various meta data to each output.
+  [-M time | reltime | notime | hires | utc | protocol | level | stats | bits | help] Add various meta data to each output.
   [-K FILE | PATH | <tag>] Add an expanded token or fixed tag to every output line.
   [-C native | si | customary] Convert units in decoded output.
   [-T <seconds>] Specify number of seconds to run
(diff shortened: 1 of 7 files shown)


merbanan:
One thing somewhat related to this is that I would like to get the SNR of the signals. That way one could see if the signal from the device is good or not. I though of using the energy in the carrier and the silence as components for this. Together with the rejected statistics one could get limits to adjust ones setup against.

merbanan:
In the OOK case, FSK should be easier as the signal is always there.

zuckschwerdt:
You are right, it's an incompatible extension. I went with `dd`s interface there, for testing Ctrl-T is just so handy ;)

rct:
It looks like output doesn't occur using test mode when reading back samples.

For Linux the SIGINFO should probably need to be turned into SIGUSR1,  29 is for SIGIO.

rct:
It would be handy if the stats output had a timestamp (to make it easier to process later).

zuckschwerdt:
This feature is now ready. With `-M stats` the default is to report stats every 10 minutes. You will see e.g.
```
{"time": "2019-03-26 12:01:54.180993", "enabled": 97,
 "frames": {"count": 198, "fsk": 23, "events": 63},
 "stats": [
  {"device":  8, "name": "LaCrosse-TX",   "events": 198, "ok": 22, "messages": 22, "fail_other": 176},
  {"device": 19, "name": "Nexus-TH",      "events":  90, "ok":  4, "messages":  4, "fail_other": 86},
  {"device": 52, "name": "Bresser-3CH",   "events": 841, "ok": 23, "messages": 23, "abort_length": 818},
  {"device": 60, "name": "Schrader-TPMS", "events":4964, "ok":  8, "messages":  8, "fail_other": 4956},
  {"device": 88, "name": "Toyota-TPMS",   "events": 118, "ok":  6, "messages":  6, "fail_other": 112}]}
```
Together with levels reporting this really allows good insight into operations and reception.
(e.g. the Schrader line shows that the demod tried to call the decoder 4964 times with different bitbuffers -- something to look into…)

Devices need to be updated to report something other than the general "fail_other" and the format isn't final -- there could be more interesting stats to add.

Good enough for now though. I'll merge in a day or two.

PR #738, opened 2018-05-24
Add support for Oregon Scientific UVR128 UV sensor

ovrheat:
Add support for Oregon Scientific UVR128 UV sensor.

### DIFF
--- a/src/devices/oregon_scientific.c
+++ b/src/devices/oregon_scientific.c
@@ -31,6 +31,7 @@
 #define ID_UV800    0xd874
 #define ID_THN129   0xcc43  // THN129 Temp only
 #define ID_BTHGN129 0x5d53  // Baro, Temp, Hygro sensor
+#define ID_UVR128   0xec70
 
 float get_os_temperature(unsigned char *message, unsigned int sensor_id) {
   // sensor ID included  to support sensors with temp in different position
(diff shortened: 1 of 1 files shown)


src/devices/oregon_scientific.c
```diff
@@ -435,6 +436,20 @@ static int oregon_scientific_v2_1_parser(bitbuffer_t *bitbuffer) {
...
       return 1;
+	} else if (sensor_id == ID_UVR128 && num_valid_v2_bits==297) {
+		if ((validate_os_v2_message(msg, 297, num_valid_v2_bits, 12) == 0)) {
+		int uvidx = get_os_uv(msg, sensor_id);
+		data = data_make(
+		  "time",           "",           DATA_STRING, time_str,
+		  "model",          "",     	  DATA_STRING, "Oregon Scientific UVR128",
+		  "id",             "House Code", DATA_INT,    get_os_rollingcode(msg, sensor_id),
+		  "uv",             "UV Index",   DATA_FORMAT, "%u", DATA_INT, uvidx,
+		  "battery",        "Battery",    DATA_STRING, get_os_battery(msg, sensor_id)?"LOW":"OK",
```
merbanan: I think we concluded to use battery_low as generic name.

merbanan:
Looks good.

src/devices/oregon_scientific.c
zuckschwerdt: yes, but it looks like this just a plain copy of all the other cases in the file. I'm not happy with "%u" either (uvidx is 0 to 165). I guess we should later change all of those "battery" instances with one commit.

ovrheat:
Sorry, I'm just asking... It's not clear to me if it's ok, or if is there something I should fix so to have the pull merged (battery? "%u" ?)... 
My next pull request will be date and time decoding for the RTGN318, but I'm waiting for this pull to be closed. Thank you.

PR #739, opened 2018-05-28
DSC Sensors break out status bits

rct:
Break out the DSC sensor status bits into individual fields.

Currently using INTs but should be booleans if/when data_make supports it.

### DIFF
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -84,6 +84,8 @@ static int dsc_callback(bitbuffer_t *bitbuffer)
     uint32_t esn;
     char status_str[3];
     char esn_str[7];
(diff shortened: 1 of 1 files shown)


dmcwhinnie:
Tested this and everything looks good.  Real nice to have all the bits broken out into individual fields.

rct:
@zuckschwerdt, @merbanan - is there anything that needs to be done to get this merged?

merbanan:
Nope, looks good.

zuckschwerdt:
I'm just trying to catch up with the last week. Looks good, as @merbanan said. Just follow up with a cleanup for those temporary/experimental fields when the dust settles.

PR #742, opened 2018-06-02
Sensible Living protocol added.

Exscotticus:
Protocol for [these cool sensors](https://www.tindie.com/products/Michael_Phillips/mini-plant-moisture-sensor/?pt=ac_prod_search).


merbanan:
Can you submit some signal samples? And would it be possible to merge the protocols?

include/rtl_433_devices.h
```diff
@@ -105,7 +105,8 @@
 		DECL(x10_sec) \
 		DECL(interlogix) \
 		DECL(dish_remote_6_3) \
-		DECL(ss_sensor)
+		DECL(ss_sensor) \
```
zuckschwerdt: use the same indent (tabs)

src/CMakeLists.txt
```diff
@@ -115,6 +115,7 @@ add_executable(rtl_433
 	devices/interlogix.c
 	devices/dish_remote_6_3.c
 	devices/simplisafe.c
+        devices/sensible_living.c
```
zuckschwerdt: indent

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+
+// Transmitter speed in bits per seconds
+#define SL_ASK_SPEED 1000
+#define SL_ASK_BIT_LEN (int)1e6/SL_ASK_SPEED
+
+// Maximum message length (including the headers, byte count and FCS) we are willing to support
+// This is pretty arbitrary
+#define SL_ASK_MAX_PAYLOAD_LEN 80
+#define SL_ASK_HEADER_LEN 4
+#ifndef SL_ASK_MAX_MESSAGE_LEN
```
zuckschwerdt: why would this already be defined?

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+static uint8_t symbols[] = {
+    0xd,  0xe,  0x13, 0x15, 0x16, 0x19, 0x1a, 0x1c,
+    0x23, 0x25, 0x26, 0x29, 0x2a, 0x2c, 0x32, 0x34
+};
+
+// Convert a 6 bit encoded symbol into its 4 bit decoded equivalent
+static uint8_t symbol_6to4(uint8_t symbol)
+{
+    uint8_t i;
+    // Linear search :-( Could have a 64 byte reverse lookup table?
```
zuckschwerdt: might be worth to try a switch-case and see if the compiler can optimize it

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+    }
+    return 0xFF; // Not found
+}
+
+
+static int sensible_living_callback(bitbuffer_t *bitbuffer) {
+    // Get time
+    char time_str[LOCAL_TIME_BUFLEN];
+    data_t *data;
+    local_time_str(0, time_str);
```
zuckschwerdt: only call local_time_str just before data_make

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+    // symbols, and it is 0 for the first 8.
+    // So we only have to search half the table
+    for (i = (symbol>>2) & 8; i < 16 ; i++){
+        if (symbol == symbols[i]) return i;
+    }
+    return 0xFF; // Not found
+}
+
+
+static int sensible_living_callback(bitbuffer_t *bitbuffer) {
```
zuckschwerdt: this brace `{ ` should have it's own line

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+      0x55, // 24
+      0x51, // 32
+      0xcd, // 40
+    };
+    // The first 0 is ignored by the decoder, so we look only for 28 bits of "01"
+    // and not 32. Also "0x1CD" is 0xb38 (SL_ASK_START_SYMBOL) with LSBit first.
+    uint8_t init_pattern_len = 40;
+
+    pos = bitbuffer_search(bitbuffer, row, 0, init_pattern, init_pattern_len);
+    if(pos == len){
```
zuckschwerdt: space after keywords

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+        if(nb_bytes == 0){
+            msg_len = byte;
+        }
+        nb_bytes++;
+    }
+
+    // Get header
+    data_len = msg_len - SL_ASK_HEADER_LEN - 3;
+    house_id = payload[1];
+    module_id = payload[2] * 256 + payload[3];
```
zuckschwerdt: this should probably be a `<<8` and `|`?

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+        "sensor_type",      "Sensor Type",      DATA_INT,     sensor_type,
+        "sensor_count",     "Sensor Count",     DATA_INT,     sensor_count,
+        "alarms",           "Alarms",           DATA_INT,     alarms,
+        "sensor_value",     "Sensor Value",     DATA_INT,     sensor_value,
+        "battery_voltage",  "Battery Voltage",  DATA_INT,     battery_voltage,
+        NULL
+	);
+    data_acquired_handler(data);
+
+    return 0;
```
zuckschwerdt: return `1` on success

src/devices/sensible_living.c
```diff
@@ -0,0 +1,183 @@
...
+    "alarms",
+    "sensor_value",
+    "battery_voltage",
+    NULL
+};
+
+r_device sensible_living = {
+    .name           = "Sensible Living Mini-Plant Moisture Sensor",
+    .modulation     = OOK_PULSE_PCM_RZ,
+    .short_limit    = SL_ASK_BIT_LEN,
```
zuckschwerdt: Please use absolute values. It easier to graps when browsing through the source.

Exscotticus:
> Can you submit some signal samples?

Yes I can. Put them in the tests subfolder in the sensible-living branch?

>  And would it be possible to merge the protocols?

Good question. It certainly would be nice to have a single protocol handle all VirtualWire and RadioHead protocols. The first hurdle is that the baud rate is arbitrary. The next hurdle is that some values get split across an arbitrary number of bytes. How would the code know which bytes to combine?

I'll work on your edits. Just keep in mind that the original **radiohead_ask.c** that I modified has the same grammar issues that you want me to address—and you accepted it. I really tried to keep my file the same and only change what was necessary. This makes diffing the two files to see what I changed easier.

zuckschwerdt:
oh, hadn't noticed that [radiohead_ask.c](https://github.com/merbanan/rtl_433/blob/master/src/devices/radiohead_ask.c) is in that bad shape. I'll push some style updates.

Exscotticus:
Admittedly my code is not good. For example, I'm grabbing 10 bytes from the payload, but I don't even validate if the payload has 10 bytes. I suppose now is a good time to mention that I'm not a C developer. Point is, no offense taken. I'll make the edits.

zuckschwerdt:
I have update the code style in radiohead_ask.c somewhat. You should be able to add your device in the same file and that way share some code (the 6to4 stuff).

Add the test samples to the [rtl_433_tests](https://github.com/merbanan/rtl_433_tests/tree/master/tests/radiohead_ask) repo in tests/radiohead_ask/03

zuckschwerdt:
You can just keep adding changes to this PR freely. We'll squash the set of commits on merge.

Exscotticus:
Added test files [here](https://github.com/Exscotticus/rtl_433/tree/sensible-living/tests/radiohead_ask/03). Does it make sense for me to complete the requested edits to **sensible_living.c** if you now want to try using **radiohead_ask.c** for multiple devices? How will we get around the aforementioned baud/byte challenges? If you're trying to create an abstraction that all VirtualWire/RadioHead can use, then **radiohead_ask.c** shouldn't make any assumptions about the baud rate or the payload.

Exscotticus:
> The sample files should go to the other repo, as new PR.

Confused again. You said...

> You can just keep adding changes to this PR freely.

But a PR is associated with a branch, and so I added the test files to the branch associated with this PR. Do you want me to create a new branch and a new PR wherein I only modify **radiohead_ask.c**? If not, please by very explicit. Assume that I have about 24 hours of git training. What repo? What branch?

> See e.g. lacrosse_tx35.c for an example of multiple devices in one source file.

I see it. And I see how a distinction is being made based on different pulse characteristics. But in the case of all VirtualWire/RadioHead devices, the pulse characteristics will be the same. You could distinguish based on bit rate—until some other VirtualWire/RadioHead device comes along with the same bit rate. Then what? Because then the payloads won't be right. You can't distinguish based on payload length, because two devices could have the same length but interpret the bytes differently.

I understand that you're trying to abstract two devices under one protocol. I juts don't understand how you're going to distinguish the umpteen million potential VirtualWire/RadioHead devices in this manner.

zuckschwerdt:
The code change and the sample files are not related and placed in two separate Git repositories.

For the samples:
- Clone the rtl_433_tests repository. https://github.com/merbanan/rtl_433_tests/
- Add a branch in your copy
- Add the sample files, Commit, Push
- Pull request

For the code changes keep using this PR #742 (and branch).

If you look in the https://github.com/merbanan/rtl_433/blob/master/src/devices/lacrosse_tx35.c#L184 there are two devices, lacrosse_tx29 and lacrosse_tx35. A device is a protocol. The two protocol/devices just share the same source file, because they are somewhat related.

- Remove the samples from this PR
- Add your r_device and callback from sensible_living.c to radiohead_ask.c

Exscotticus:
> Add your r_device and callback from sensible_living.c to radiohead_ask.c

Done.

Exscotticus:
Is there anything else this pull request is waiting on?

zuckschwerdt:
I hope to get some time in the next days to integrate this. I want to layout the code so it is a bit more reusable.

zuckschwerdt:
Have a look at https://github.com/merbanan/rtl_433/blob/master/src/devices/radiohead_ask.c -- it should be easier to understand and extend in the future.

Exscotticus:
I agree. Good changes. You abstracted a large chunk of complex common code from both callbacks. Thanks!

PR #750, opened 2018-06-18
add support for TP08 remote thermometer

terhorst:
A minor modification allows the existing ThermoPro TP12 device handler to also decode the [TP-08](https://www.amazon.com/ThermoPro-TP-08-Wireless-Thermometer-Grilling/dp/B014DAVHSQ) (and probably the TP07/TP20 as well). I have tested this for several days and it seems to work fine.

It would be interesting to try and figure out the checksum that is also being transmitted...

### DIFF
--- a/src/devices/thermopro_tp12.c
+++ b/src/devices/thermopro_tp12.c
@@ -11,7 +11,7 @@
 #include "rtl_433.h"
 #include "util.h"
 
-#define MODEL "Thermopro TP08/TP12 Thermometer"
+#define MODEL "Thermopro TP12 Thermometer"
 
 /*
 A normal sequence for the TP12:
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks. Changing the "model" data_make key (via MODEL) isn't good. Chaning the r_device name is ok. You should possibly split those and remove the define.
False positives are always a consideration, at a glance I can't say wether 2 repeats are sufficient.
We'll have to let this sit a while and test for regressions.

rct:
@terhorst - FYI my TP07 has been decoding reliably with the existing TP-11 code.

zuckschwerdt:
@terhorst are you still working on this? Can you remove the MODEL macro and only change the `.name` (i.e. not in data_make).
Also how about this logic:
- if more than 5 rows then require 5 matches,
- otherwise require 2 matches.

FearrMe:
For what it's worth, with @terhorst's modification(requiring 2 matches instead of 5) my TP-20 also seems to return correct temperature values.

zuckschwerdt:
@FearrMe what model do you have? The TP08 seems to send only 3 packets, the other 16. Thus 5 is a good choice in the latter but impossible with 3 packets ;)

FearrMe:
@zuckschwerdt I have the ThermoPro TP20, this is my first time using the software so not sure what I'm doing.

terhorst:
@zuckschwerdt I made the requested changes. I am not currently able to test these though.

zuckschwerdt:
@terhorst Looks good. @FearrMe can you test and confirm?

FearrMe:
@zuckschwerdt @terhorst yup, can confirm TP20 works with the new changes! Shows up as "Thermopro TP08/TP12 thermometer" under help command and "Thermopro TP12 Thermometer" when running and outputting temperatures.

PR #751, opened 2018-06-27
Enhance currentcost device with OptiSmart (Counter) support

EusaHues:
Hopefully addresses the feedback in https://github.com/merbanan/rtl_433/pull/703

adds to existing  currentcost device (model: "CurrentCost TX") with model: "CurrentCost Counter")

The csv output was already "broken" for multiple devices as device_id was not included
Rather than add fields to the output format I've replaced "rc" (which was never populated) with device_id (144 and 3572 below)

2018-06-27 22:17:29,CurrentCost Counter,144,16409099,,
2018-06-27 22:17:34,CurrentCost TX,3572,589,534,57

The change might still break users of csv/json output format that don't check the "model" string, as I've overloaded the counter into "power0" for the csv (and json) output formats (any preferred solution to this problem gladly accepted!)

{"time" : "2018-06-27 22:17:01", "model" : "CurrentCost Counter", "dev_id" : 144, "sensor_type" : 2, "power0" : 16409095}
{"time" : "2018-06-27 22:17:05", "model" : "CurrentCost TX", "dev_id" : 3572, "power0" : 539, "power1" : 540, "power2" : 16}

For the kv output field I've called it Counter, and added the sensor "type" (Sensor Id).

2018-06-27 22:16:33 :   CurrentCost Counter
        Device Id:       144
        Sensor Id:       2
        Counter:         16409092
2018-06-27 22:16:34 :   CurrentCost TX
        Device Id:       3572
        Power 0:         538 W
        Power 1:         532 W
        Power 2:         16 W

Recordings in https://github.com/merbanan/rtl_433_tests/pull/222

### DIFF
--- a/src/devices/current_cost.c
+++ b/src/devices/current_cost.c
@@ -33,6 +33,7 @@ static int current_cost_callback(bitbuffer_t *bitbuffer) {
 
     uint8_t *packet = packet_bits.bb[0];
     // Read data
+    // Meter
     if(packet_bits.bits_per_row[0] >= 56 && ((packet[0] & 0xf0) == 0) ){
         uint16_t device_id = (packet[0] & 0x0f) << 8 | packet[1];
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
I'm not sure where `dev_id` is used (if consuming JSON), there might be some custom scripts out there. So this might cause some grief, but the prefered key would be just `id` – I guess you could change that. Also `sensor_type` needs to listed in the `output_fields` too.

EusaHues:
I didn't add dev_id, it was already there in the json and kv (as Device Id:) outputs for the "CurrentCost TX" model.
If I change that to "id" then certainly any consumers of the json will cry foul.
In my output above "CurrentCost TX" is unchanged on json and kv, it's only different on csv (where I include "dev_id" instead of the *never populated* "rc" so as to not change the output format (again, to avoid breaking people as best I can think of?).

Doesn't output_fields control the csv output, I avoided adding any extra fields to that to not break existing users.

If the existing current_cost.c isn't using the preferred style / format, can you point me to a style guide, or a "do it this way" device? (and I'll look at re-doing the whole device).

PR #753, opened 2018-06-30
[MSVC] patch include/data.h

gvanem:
The rationale behind these patches are pretty much in the source comments. MSVC issues with:
*) "Variable Length Arrays"
*) 'restrict` keyword. 
*) And '<strings.h>'.


zuckschwerdt:
Thanks again!

PR #755, opened 2018-06-30
No "Variable Length Arrays" for MSVC

gvanem:
Use 'alloca()' instead.


merbanan:
Please don't do it like that. Just replace it with the maximum size.

zuckschwerdt:
You are right, I missed that! Sorry.

PR #760, opened 2018-07-03
Make the frequency setting clearer.

gvanem:


zuckschwerdt:
People might have scripts to parse the current ouput. I don't consider it API but let's have this issue sit for a week so others get a change to chime in.

merbanan:
I am for committing this. Parsing that value makes no sense.

zuckschwerdt:
…also I'm working on the stats/meta-data addition where in the future the center frequency would be reported with every event :)

merbanan:
Hmm, maybe the stats should include an api version tag? Then that version can be bumped when incompatible changes are added.

gvanem:
@zuckschwerdt 
> People might have scripts to parse the current ouput. 

How about if one used `-f XX.YYM` on the cmd-line, then (and only then) print the frequency as
`XX.YYYMHz`? Then the `optparse.c` should be changed (and maybe moved to `util.c` since `optparse.c` is so small). 3 decimals should be enough, no?

zuckschwerdt:
Good idea, but really a minor corner case. Anyway we aren't afraid to change things. Just let other contributors get a fair warning. This will go in as-is in a few days. And since we tag versions and deliver a Changelog now there will be documented breaking changes every few months or so.

PR #768, opened 2018-07-13
 Add device: M-Bus (EN 13757-4) - Data Link layer

vestom:
Used for various meters (water, heat, electricity etc.).
The driver only implements up to the Data Link layer of the protocol and outputs a data string for further processing by the Application layer. The Application layer is quite complex (diverse) and may involve decryption with a user supplied decryption key, so should probably be done externally anyway. 
However, this is as far as I am at this moment :-)
Note: The protocol is for use at a non-default frequency (868.95 MHz) and is disabled by default.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -23,3 +23,4 @@ Nicola Quiriti <nik@wifi4all.it>
 Petr Konecny <pekon@google.com>
 Tom Felker <tomfelker@gmail.com>
 Pasquale 'sid' Fiorillo <me@pasqualefiorillo.it>
+Tommy Vestermark <tovsurf@vestermark.dk>
--- a/README.md
+++ b/README.md
@@ -198,6 +198,10 @@ Supported device protocols:
     [101]* Dish remote 6.3
     [102]* SimpliSafe Home Security System (May require disabling automatic gain for KeyPad decodes)
     [103]  Sensible Living Mini-Plant Moisture Sensor
+    [104]* Wireless M-Bus, Mode C&T, 100kbps (-f 868950000 -s 1200000)
+    [105]* Wireless M-Bus, Mode S, 32.768kbps (-f 868300000 -s 1000000)
+    [106]* Wireless M-Bus, Mode R, 4.8kbps (-f 868330000)
+    [107]* Wireless M-Bus, Mode F, 2.4kbps
 
 * Disabled by default, use -R n or -G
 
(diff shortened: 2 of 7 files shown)


zuckschwerdt:
We need to check if `BITBUF_COLS` can be ramped up this much, (do you really expect 2k bits?) but this looks wonderful!

vestom:
Test files here: https://github.com/merbanan/rtl_433_tests/pull/229

vestom:
According to the specification the messages can be even slightly larger than 256 bytes (with preamble, CRC etc.)! Things are going faster at 868 MHz and 1.2 M samples/s :-)
I have however only seen messages up to ~827 bits (~103 bytes) myself.

I have not had any problems during testing and 2kbit is well within the uint16's used for indices. The bitbuffer does get a bit large (7kbytes), due to the multiplication with rows, but it should be OK. I remember from testing a while back, that messages are either very "vertical" (many rows with short messages) or mostly "horizontal" (one or two rows with large messages), but I have no idea right now for how to optimize for that. But especially FSK messages seems to be longer and only have one row (and using better protocols in general).

zuckschwerdt:
I've removed some unwanted references to the `BITBUF_COLS` with https://github.com/merbanan/rtl_433/commit/d4a9e09497079359b0a8188507518420be7f5d6b
Only [oregon_scientific.c](https://github.com/merbanan/rtl_433/blob/master/src/devices/oregon_scientific.c) remains a mess. Not a problem as such, but that should be cleaned up someday.
I might have missed something but overall I'd say increasing the row buffer shouldn't matter much to the decoders. The demodulators could benefit from optimization, as we put those 7k on the stack with every call there – and that's for each enabled device and in a somewhat tight loop.

vestom:
Good with a bit of cleanup :-) If we want to change the way bitbuffer works, it is bad that the users rely too much on the inner working.

An idea for optimization could be to make rows dynamic in size by making the row indices pointers into one large bit array. Then we could limit the size to the maximum product of row_size*col_size, which is probably less than 512 bytes. But that solution may have some other cost.

Frankly I would be surprised, if it is possible to see any performance difference on the increased buffer size. The increased stack size is only a bigger constant added to the stack pointer. However, as the buffer is cleared at allocation, there is a little larger "memset" going on.

vestom:
Restructured the code a bit for better reuse and added some stubs/experimental decoders for the other modes. However, I miss test samples for those, so it doesn't make a lot of sense to go further.
I will probably leave it here for now - until more data appears.

PR #770, opened 2018-07-17
Clean up baseband, optimize FM, add baseband-test

zuckschwerdt:
Adds a baseband-test to evaluate and compare speed and output.
Optimizes FM demod by removing intermediate shifts (no loss of precision).


zuckschwerdt:
No change in functionality, FM demod speedup of 50% on x64 Clang (no speedup on ARM GCC – maybe GCC was smart enough to remove the shifts already).

PR #778, opened 2018-07-27
S3318P: Add Fahrenheit to Celsius conversion (°F to °C)

Koumes:
Adds Fahrenheit to Celsius conversion (°F to °C).

### DIFF
--- a/src/devices/s3318p.c
+++ b/src/devices/s3318p.c
@@ -84,7 +84,7 @@ static int s3318p_callback(bitbuffer_t *bitbuffer) {
     uint8_t channel;
     uint8_t sensor_id;
     uint16_t temperature_with_offset;
-    float temperature_f;
+    float temperature_f, temperature_c;
 
     /* IIIIIIII ??CCTTTT TTTTTTTT HHHHHHHH XB?????? PP */
     humidity = (uint8_t)(((b[3] & 0x0F) << 4) | ((b[3] & 0xF0) >> 4));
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Sorry, the device decoders should output all values in a native format only. There is `-C si` for temperature (and other metric) conversion.

zuckschwerdt:
Also I'll take this as a hint to reorganize the help output to be a bit friendlier ;)

merbanan:
Maybe make SI the default?

merbanan:
Well I'm undecided. Maybe guessing with the help of the locale, to really complicate things?

zuckschwerdt:
I've been thinking about adding config files for a while now. Something like an Ini-File parser to read all settings from a well documented file. There could be pseudo-defaults like conversion, choice of some supported center frequencies, a list of possible decoders with annotations and hints, examples of flex decoders, and so on. Also users can simply manage and exchange those configs with others. Not sure about the best format and if we should look into libraries for that.

merbanan:
I wanted to add a daemon mode for rtl_433. That would go hand in hand with a config file. Json might be an option for the config file format.

zuckschwerdt:
Since it's not machine-to-machine data JSON could be too complicated. If INI-file doesn't fit maybe something like YAML or TOML (where TOML is a natural extension of simple INI-files).

rct:
* +1 for the idea of adding a config file, especially with flex decoding and running out of option letters.
* I personally don't like the idea of having translated units be the default especially since the general idea is that rtl_433 should be machine consumed rather than human read.  I would probably have less objection if I was pretty sure the errors that were do to the combination of pretty printing / rounding were eliminated.
* I think the README should be split into multiple documents or at least sections in this order.
   * README - basic overview with links to the other docs.
   * Build/Install
   * Running
   * Adding new decoders

PR #779, opened 2018-07-28
Add Hyundai WS SENZOR

Koumes:
Add support for new device Hyundai WS SENZOR Remote Temperature Sensor. Mostly sold in central/eastern Europe (Czechia, Slovakia, Poland, Hungary).

### DIFF
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -43,7 +43,7 @@
 
 #define MINIMAL_BUF_LENGTH      512
 #define MAXIMAL_BUF_LENGTH      (256 * 16384)
-#define MAX_PROTOCOLS           107
+#define MAX_PROTOCOLS           108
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Looks good. You also need to add to https://github.com/merbanan/rtl_433/blob/master/src/Makefile.am

zuckschwerdt:
Can you also add a PR with samples, maybe pictures and README to rtl_433_tests, please.

I guess since there is no checksum and all 24 bits with the right timing will yield something this devices needs to be disabled by default?

Koumes:
I tested the current rtl_433_tests data without conflict or false positive. My code is mainly based on s3318p.c where it is not checksum also.

I have to read the git guide. I'm not an experienced developer.

PR #782, opened 2018-08-01
Add Emos TTX201

Koumes:
Add support for new device Emos TTX201 Remote Temperature Sensor.


zuckschwerdt:
I like that the code strieves to be text-book clean. But I looks over-engineered.
The 8 bits per byte is part of the protocol and not platform dependent or might change, just replace CHAR_BIT with 8.
checksum_calculate already does some magic stuff, no need to hide nibble access in HIGH, LOW macros.
If you need a fixed preamble then use `bitbuffer_search()` and don't count bits (though it is faster).
Is there no gap or sync between packets? The demodulator should really split into one row per packet.

Koumes:
Preamble may not be received complete and counting of bits is simple way to strip it. There's a little longer gap between packets, but I have no idea how to force the Manchester demodulator to packet separation into rows.

zuckschwerdt:
About the preamble: if your perfect preamble is 0x00000 0b1 then just shorten that as needed. E.g.
```
// full preamble is 0000 0000 0000 0000 0000 1 = 0x0000080 (21 bits)
static const unsigned char preamble_pattern[1] = {0x01}; // 8 bits
// ...
bitpos = bitbuffer_search(bitbuffer, row, 0, (const uint8_t *)&preamble_pattern, 8);
// ...
```

Koumes:
The most common perfect preamble is 0x000000 (20 bits) immediately followed by a checksum 0x00 to 0x3f (6 bits). It's not easy to determine the boundary.

zuckschwerdt:
I had a peek at the samples. Looks like the short/long pulses/gaps are ~500/1000 µs. I.e. a half-bit width of 500 µs (the 510 µs you use are close enough). The inter-packet gaps are ~1500 µs (preamble gap also). It should be possible to detect that and split accordingly.

The [`pulse_demod_manchester_zerobit()`](https://github.com/merbanan/rtl_433/blob/master/src/pulse_demod.c#L254) needs to be enhanced to reject over-long pulses/gaps (2.5x half-bit width should be reasonable, 1250 µs here). I'll do that shortly.

If you want you could also try to use the PWM decoder and run [`bitbuffer_manchester_decode()`](https://github.com/merbanan/rtl_433/blob/master/include/bitbuffer.h#L71) yourself.

Try the feat-vcd branch and Sigrok's pulseview or GTKWave to get a better view of the timings, e.g.

ttx201

zuckschwerdt:
I have added a quick fix to the manchester demod. If you set the tolerance (`.tolerance = 250`) pulses and gaps shorter than half-bit width minus tolerance and longer than full-bit length plus tolerance get rejected and add a new row. This should help to split the packets here.

Koumes:
Thank You. I think there's a typo.

```diff
--- a/src/pulse_demod.c
+++ b/src/pulse_demod.c
@@ -264,7 +264,7 @@ int pulse_demod_manchester_zerobit(const pulse_data_t *pulses, struct protocol_s
                if (device->tolerance > 0
                                && (pulses->pulse[n] < device->short_limit - device->tolerance
                                || pulses->pulse[n] > device->short_limit * 2 + device->tolerance
-                               || pulses->gap[n] > device->short_limit - device->tolerance
+                               || pulses->gap[n] < device->short_limit - device->tolerance
                                || pulses->gap[n] > device->short_limit * 2 + device->tolerance)) {
                        // The pulse or gap is too long or too short, thus invalid
                        bitbuffer_add_row(&bits);


```

zuckschwerdt:
Good catch, thanks. The MC demod needs a complete overhaul, e.g. the first run of short pulses could be mistaken for 1 when it is actually 0.
But for now I hope this is enough to get multiple rows for this device.

Koumes:
Reworked using the new Manchester demodulator feature. But it is less fault-tolerant. The previous version was able to better read the erroneously received data due to interference.

zuckschwerdt:
Sorry for the long delay. Merged with slight update of name and model output. Thanks again!

PR #783, opened 2018-08-03
Fill id string for Bresser device.

guek:
Correct output of "Bresser 3CH" device.
Before:
`2018-08-03 15:22:03 :	Bresser 3CH sensor 	:	250`
`	Channel:	 1`
...
After:
`2018-08-03 17:35:03 :	Bresser 3CH sensor`
`	Id:	 250`
`	Channel:	 1`
...

### DIFF
--- a/src/devices/bresser_3ch.c
+++ b/src/devices/bresser_3ch.c
@@ -77,7 +77,7 @@ static int bresser_3ch_callback(bitbuffer_t *bitbuffer) {
     data = data_make(
             "time",          "",            DATA_STRING, time_str,
             "model",         "",            DATA_STRING, "Bresser 3CH sensor",
-            "id",            "",            DATA_INT,    id,
+            "id",            "Id",          DATA_INT,    id,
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Yeah, I personally never use KV (or CSV), there are more bugs like this lurking.
Should `time` and `model` also be named for KV? Anyone with a proper use-case speak up :)

guek:
Well, seems to be such a simple patch, but it is not. Should it be named "ID" or - like I did - "Id"? Both "keys" were already used. Other example is "Channel" vs. "Channel number".
Scanning my environment I have several sensors around. And it seems that the "keys" are different, which makes parsing hard. I do not know if some programmers rely already on this different keys for different devices. Consuming software has to be device sensitive, it seems.
(I remember that once in my career a naive programmer corrected a typo of a key string and the whole system crashed...)

zuckschwerdt:
Like they say, the really hard things in programming is stuff with dates and naming things…

PR #784, opened 2018-08-03
 Adding default case to suppress maybe-uninitialized Warning from GCC

andreasmarkussen:
When running make inside ~/Projects/rtl_433/build, I get the following minor warning:

    /root/Projects/rtl_433/src/rtl_433.c:1413:13: 
             warning: 'load_mode_str' may be used uninitialized in this function [-Wmaybe-uninitialized]
             fprintf(stderr, "Input format: %s\n", load_mode_str);
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

So I have reviewed the error description and added a default case.
I am aware that this will never happen, since the load_mode is checked to be an int between 0 and 4.

See 
[GCC Warning Options](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wmaybe-uninitialized).

I also considered to suppress the warning on line level, but I was not able to find out how.

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -1400,14 +1400,14 @@ int main(int argc, char **argv) {
             demod->sample_size = sizeof(int16_t); // CF32, CS16
         }
         if (!quiet_mode) {
-            char *load_mode_str;
+            char *load_mode_str = null;
             switch (demod->load_mode) {
             case 0: load_mode_str = "CU8 (2ch uint8)"; break;
             case 1: load_mode_str = "S16 AM (1ch int16)"; break;
             case 2: load_mode_str = "S16 FM (1ch int16)"; break;
             case 3: load_mode_str = "CF32 (2ch Float32)"; break;
             case 4: load_mode_str = "CS16 (2ch int16)"; break;
-            default: load_mode_str = "Unknown load_mode " + demod->load_mode ; break;
+            default: fprintf(stderr, "Unknown load_mode: %d\nMust be between 0 and 4", demod->load_mode);  break;
             }
             fprintf(stderr, "Input format: %s\n", load_mode_str);
         }
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -1407,6 +1407,7 @@ int main(int argc, char **argv) {
             case 2: load_mode_str = "S16 FM (1ch int16)"; break;
             case 3: load_mode_str = "CF32 (2ch Float32)"; break;
             case 4: load_mode_str = "CS16 (2ch int16)"; break;
+            default: load_mode_str = "Unknown load_mode " + demod->load_mode ; break;
             }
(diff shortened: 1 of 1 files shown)


andreasmarkussen:
Some information about my gcc verion on my Debian machine inside a docker.
    gcc -v
    Using built-in specs.
    COLLECT_GCC=gcc
    COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/6/lto-wrapper
    Target: x86_64-linux-gnu
    Configured with: ../src/configure -v --with-pkgversion='Debian 6.3.0-18+deb9u1' --with-bugurl=file:///usr/share/doc/gcc-6/README.Bugs --enable-languages=c,ada,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-6 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --enable-default-pie --with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-6-amd64/jre --enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-6-amd64 --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-6-amd64 --with-arch-directory=amd64 --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
    Thread model: posix
    gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1)

andreasmarkussen:
Link to Docker Setup
https://github.com/andreasmarkussen/rtl_433_docker/blob/master/Dockerfile

zuckschwerdt:
Thanks, good catch. Adding to a char* is a C++ thing though ;) In C it will offset the char*
Make it a fixed string or better just set the `load_mode_str` to NULL on declaration.

zuckschwerdt:
it's `NULL` (which is actually just `0`) not `null` :p

PR #786, opened 2018-08-05
Add flex getter, WIP

zuckschwerdt:
Adds a spec to extract specific bits in flex decoder. WIP, not for merge.

### DIFF
--- a/src/devices/flex.c
+++ b/src/devices/flex.c
@@ -14,6 +14,44 @@
 #include "pulse_demod.h"
 #include "optparse.h"
 
+/// extract a number up to 32/64 bits from given offset with given bit length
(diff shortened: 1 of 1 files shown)


roger-:
This looks good to me!

zuckschwerdt:
Thanks. With upcoming support for config files the syntax will even be less arcane ;)

roger-:
Sweet, when's the merge happening?

zuckschwerdt:
I'm still pondering things, looking for feedback and waiting for other developers to catch up. I guess both will be merged in the next weeks. I think I'll just take the liberty of changing details later -- don't expect anything about the flex decoder to be set in stone for now ;)

Issue #777, opened 2018-07-27
Windows build description


Mindavi:
I've written a small guide on how to get building in Windows working, but I'm not sure where to put it. I've understood building for Windows isn't officially supported, but I thought it still might be useful to have it somewhere.

This is what I wrote:

# Building rtl_433 for windows

you'll need librtlsdr and libusb, they are dependencies for rtl_433
libusb has prebuilt binaries for windows, 
librtlsdr needs to be built (at least for the latest version, some binaries seem to be floating around)

## librtlsdr

taken and adapted from here: https://www.onetransistor.eu/2017/03/compile-librtlsdr-windows-mingw.html

* install mingw-w64 (https://mingw-w64.org/) and cmake (https://cmake.org/)
    * it's easiest if you select the option to include cmake in your path, otherwise you'll need to do this manually
* download the libusb binaries from https://sourceforge.net/projects/libusb/files/libusb-1.0/ or from https://libusb.info/
    * take the latest release and then download the .7z file, the other file contains the sources (or 'windows binaries' on the .info website)
* extract the archive and open the extracted folder
* copy the contents of the include folder to `<mingw_installation_folder>/include`
* copy the `mingw64/dll/libusb-1.0.dll.a` file to `<mingw_installation_folder>/lib
* copy the `mingw64/dll/libusb-1.0.dll` file to `<mingw_installation_folder>/bin`
* download the source code of librtlsdr https://github.com/steve-m/librtlsdr
* go into the librtlsdr folder
* open cmakelists.txt with an editor that knows unix line endings
* go to `# Find build dependencies` (around line 65) and comment/remove the line with `find_package(Threads)`
* add the following lines instead:

```
SET(CMAKE_THREAD_LIBS_INIT "-lpthread")
SET(CMAKE_HAVE_THREADS_LIBRARY 1)
SET(THREADS_FOUND TRUE)
```

* go into the cmake/modules folder and open FindLibUSB.cmake with a text editor
* find the lines with the following text in them

```
/usr/include/libusb-1.0
/usr/include
/usr/local/include
```

* add some extra lines to point to the mingw include folder where you extracted libusb-1.0, making it look like this
    * take note of the "" around the folder names, these are needed when there are spaces in the folder name
    * you'll need to find out the exact paths for your system

```
/usr/include/libusb-1.0
/usr/include
/usr/local/include
"C:/Program Files/mingw-w64/x86_64-8.1.0-posix-seh-rt_v6-rev0/mingw64/include"
[...]

PR #788, opened 2018-08-09
Add MinGW build description for Windows, merge all build instructions in 1 BUILDING file

Mindavi:
see #777


zuckschwerdt:
Thanks!
Can you make this a general BUILDING.md with an introduction to all build systems?
I.e. cmake, autotools, VS2015, then GCC-MinGW with your text.
Also add a hint to the README below the existing options, something like "for other build options see BUILDING.md".

PR #789, opened 2018-08-09
Fixes and example project to compile using MS Visual Studio 2015

winterrace:
Pull request as requested in https://github.com/merbanan/rtl_433/issues/777

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,22 @@
-rtl_433
-=======
+rtl_433_win
+===========
+This is a fork of merbanan/rtl_433 which compiles for Win64 under Microsoft Visual Studio (verified with VS 2015). Pre-compiled binaries are included.
+To compile it yourself, you first need to extract the files from https://osmocom.org/attachments/2242/RelWithDebInfo.zip into a folder "rtl_433_win" parallel to the rtl_433_win folder.
+It has to look like this
+
+rtl_433_win
+    include
+    src
+    vs15
+    <and so on>
+rtl-sdr
+    x64
+    <and so on>
+    
+Then simply open rtl_433_win/vs15/rtl_433.sln in Visual Studio and select between debug/release build options.
+
+Original readme from merbanan/rtl_433:
+===========
 
 rtl_433 turns your Realtek RTL2832 based DVB dongle into a 433.92MHz generic data receiver
 
--- a/README.md
+++ b/README.md
@@ -1,22 +1,4 @@
-rtl_433_win
-===========
-This is a fork of merbanan/rtl_433 which compiles for Win64 under Microsoft Visual Studio (verified with VS 2015). See release section for pre-compiled binaries.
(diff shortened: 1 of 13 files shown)


zuckschwerdt:
Thanks.
- Clean up the README ("This is a fork")
- Use restrict instead of RESTRICT and `ifdef` just for `_MSC_VER` like you already do.

winterrace:
winterrace:feat-vs2015 is now reworked as requested (hopefully ;-) )

winterrace:
Sorry, I just accidentally closed the PR. Reopened...

zuckschwerdt:
Looks good, thanks. Can you check if the `rtl_433.vcxproj.filters` file is actually needed to build? Would be one less file to care about when devices are added.

winterrace:
The rtl_433.vcxproj.filters "just" provides a mapping of all the source files (from rtl_433.vcxproj) into different virtual folders in the source file browser. The main advantage is that all device implementations are displayed in a common subfolder, that can be collapsed or expanded.

We could remove rtl_433.vcxproj.filters from the repository, it is not needed to compile the project. However, the project looks quite ugly this way (all *.c and *.h files in one common root folder)

We could also keep it. If we introduce new source files it is basically enough to add them to rtl_433.vcxproj. If they are not added to rtl_433.vcxproj.filters, they will show up in the root folder and can simply be moved by drag and drop (this will auto-update rtl_433.vcxproj.filters).

zuckschwerdt:
For the sake of simplicity I'd rather not have extra files. The reasoning is: we like to support building with other environments, but we don't necessarily want to support development in other environments. I think it would be nice to have but I do see the trouble of maintaining that. (And once we commit to something it's rather hard to remove…)
I only ever use CMake and it's already a problem to keep the Automake in top shape.

Perhaps we could get more opinions on this?

merbanan:
If a script could generate the rtl_433.vcxproj.filters file that would be preferable.

zuckschwerdt:
A small python script to refresh auxiliary files, maybe also the help output in the README etc, yes.

auphofBSF:
Well done,works very well, including using the debugger.   Instructions worked well.

On windows 10, Visual Studio 2015 Community edition I had switch from Win8.1SDK to the Win10SDK that I already had installed. I already had used the 10.0.16299.0  version

Change SDK by  selecting Properties of the project
image

winterrace:
Thanks for sharing your experience. Great that it was useful to you...

winterrace:
Today I included some minor updates the the feat-vs2015 branch: In the vs2015 project files, I renamed the German filter names to English and removed a link to an autogenerated readme file not existing any more. Also, the compilation instructions have been updated with a link to the recent librtlsdr 0.6.0 release (new Win64 project at https://github.com/winterrace/librtlsdr).
Still works well with the latest rtl_433 updates...

PR #790, opened 2018-08-10
Add a config file parser

zuckschwerdt:
This adds a config parser (option `-c`) and a sample conf file.
I also toyed with Ini-files (too inflexible) and TOML-files (to complex).
The added benefit with a simple keyword parser is that it ties neatly into the short-option parsing.

Work in progress. Not for merge, yet.


merbanan:
The structure looks good

zuckschwerdt:
We could also automatically load some paths, if found, e.g.

1. `./rtl_433.conf`
2. `~/.rtl_433.conf`
3. `/etc/rtl_433.conf`

Thoughts on that?

roger-:
> We could also automatically load some paths, if found, e.g.
> 
> 1. `./rtl_433.conf`
> 2. `~/.rtl_433.conf`
> 3. `/etc/rtl_433.conf`
> 
> Thoughts on that?

Sounds good to me.

zuckschwerdt:
The config file feature is ready for merge. Everybody have a final look and test, merge shortly.

rct:
Looking over this now, looks good so far but haven't tested yet.

Two comments/questions that are not blockers:

- What's the plan if you need some hierarchy/structure -- ie multiple settings per config object, when you can have multiple objects?

- Something I've wanted but never gotten around to, explicitly disable a protocol.  My use case is there are times when I want to enable everything so I can see what gets tripped but I want to turn off a specific decoder that has too many false positives.

zuckschwerdt:
- The parser just reads and applies every setting one after the other (just like the command line options). The file doesn't constitute one final "config" but rather a series of settings applied. I.e. You can override and stack settings as well as use appropriate keys multiple times. This "SAX parser style" allows quite some flexibility as opposed to a "DOM representation".
- You could have a config file with all protocols enabled and then comment some out. But I'm planning some enable/disable function (even after startup, like changing the frequency)...

zuckschwerdt:
@rct protocols can now explicitly be disabled using negative protocol numbers, i.e. `-R -52` keeps the default set and only disables Bresser-3CH.

Config files can be nested/stacked (use multiple `-c` and `config_file = `). I anticipate this will be a great way to load custom sets of flex definitions.

If no `-c` option is given the first found of this list will be loaded:
- ./rtl_433.conf
- ~/.rtl_433.conf
- /usr/local/etc/rtl_433.conf
- /etc/rtl_433.conf

This feature is about to be merged.

PR #793, opened 2018-08-19
Fix MinGW build

mnhauke:
src/CMakeLists.txt
With MinGW you need to link against ws2_32 and mswsock

tests/baseband-test.c
sys/uio.h is not available on windows

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -121,11 +121,18 @@ add_executable(rtl_433
 
 add_library(data data.c)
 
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
I don't remember why `<sys/uio.h>` is there. It's not needed, you can remove that unconditionally.
How about adding a `NET_LIBRARIES` list to the top CMakeLists.txt with the conditional for ws2_32 and mswsock? See e.g. `SDR_LIBRARIES`?

PR #796, opened 2018-08-26
This change was forgotten in commit eb02ef4 ("Add CS16 input and outp…

winterrace:
…ut (#773)")


zuckschwerdt:
Good catch! Thanks!

PR #798, opened 2018-08-30
Improved checking of Simplisafe data and default enable decoder

pekon:
Simplisafe devices transmit each packet twice with about 2ms gap between them. This changes the decoder to detect and remove the duplicates.

### DIFF
--- a/src/devices/simplisafe.c
+++ b/src/devices/simplisafe.c
@@ -152,7 +152,7 @@ ss_sensor_callback (bitbuffer_t *bitbuffer)
 
 	uint8_t *b = bitbuffer->bb[row];
 
-	if (b[0] != 0x33 && b[1] != 0xa0) // All Messages Must start with 0x33a0
+	if (b[0] != 0x33 || b[1] != 0xa0) // All Messages Must start with 0x33a0
 		return 0;
 
 	if (b[2] == 0x88) {
--- a/src/devices/simplisafe.c
+++ b/src/devices/simplisafe.c
@@ -146,7 +146,7 @@ ss_keypad_commands (bitbuffer_t *bitbuffer, int row)
 static int
 ss_sensor_callback (bitbuffer_t *bitbuffer)
 {
-        // Simplisafe transmits data twice. Treat second copy as a checkum
+        // Simplisafe transmits data twice. Treat second copy as a checksum
         // and reject transmissions that lack it.
         int row = bitbuffer_find_repeated_row(bitbuffer, 2, 90);
         if (row < 0) return 0;
--- a/src/devices/simplisafe.c
+++ b/src/devices/simplisafe.c
@@ -42,13 +42,18 @@ ss_sensor_parser (bitbuffer_t *bitbuffer, int row)
 	if (bitbuffer->bits_per_row[row] != 92)
 		return 0;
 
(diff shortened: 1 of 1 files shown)


merbanan:
The PR has conflicts, rebase the changes.

zuckschwerdt:
Looks ok, but it default enables this decoder. Without a Checksum we don't generally recommend that.

zuckschwerdt:
Sorry, that's not a checksum;) Checksum here is not just about the signal integrity but also about the validity. With more than 100 decoders we need to minimize chances of false positives.

merbanan:
What we want is that the decoder can be enabled by default. But we can't perform an extensive check to begin with because that check is done for almost every signal we receive. In the end with enough protocol decoders it will take to much time to decode signals on slow platforms.

So the decoding process need to quickly reject based on simple conditions. After what we can add better checks when we are more sure this is probably the correct signal.

Lets looks look at the code to begin with:

	bitbuffer_invert(bitbuffer); // Invert the Bits
	uint8_t *b = bitbuffer->bb[0];

	if (b[0] != 0x33 && b[1] != 0xa0) // All Messages Must start with 0x33a0
		return 0;

First we invert the bitbuffer, imo this should be moved after the first check. The first check is a syncworkd a syncword check, imo that is enough to enable the decoder by default. But lets see what we can do more. In the signal description we have this:

SUM CMD + CS

So it looks like there might be a checksum that can be validated. Doing this way will give a 2 stage verification in the decoding step.

zuckschwerdt:
Oh, I didn't notice the 16-bit header check. That should get rid of false positives.
Perhaps "Treat second copy as a checksum" should be "Require 2 identical packets" (that's the line I tried to comment on).

pekon:
I moved the codeword check before inverting bits to make rejections even more efficient.

I looked at the checksum value and it often does not match, for example rtl_433_tests/tests/simplisafe/contact-1JV7T-open.cu8

the (inverted) row in bit buffer looks like this;
[01] {92} 33 a0 88 8c 52 6a ec 2a a3 80 63 f0                                           

0xa3 + 0x80 = 0x123, this does not match 0x63 in the "checksum" field. So perhaps this check could be added in the future if we find out what it should be.

merbanan:
Does the 6 upper bits match on all samples you have ((0xa3+0x80)∧3F)? If it does then at least add it to the protocol description. By the text it looks like this condition is true for all messages. If the check passes on all messages then you can add it to the code and add a mic=checksum object member.

pekon:
I have sensors that produce:
[01] {92} 33 a0 88 8c 8c 1c e2 2a 5f c2 bc f0

0x5f + 0xc2 is odd, while 0xbc is even. So the dependency is not that simple. I can try to analyze more sensors and add checks if I find any. IMO duplicate row detection should catch most transmission errors and rule out other types of sensors.

pekon:
Validation in ss_sensor_parser was not too hard, I can take a look at the other functions.
AFAICS reverse8(b[8]) contains:
  1. type of device in upper nibble (motion sensor, contact, smoke, CO, keychain and others)
  2. counter in lower nibble that increments with each transmission

I think it would make sense to separate those in data_make. I will need to update test cases for those.

PR #800, opened 2018-08-31
Add VCD and SR output

zuckschwerdt:
While toying with CAN bus the other day and using a mixed signal digital oscilloscope I was very pleased with some visualizations.
I think those visualizations are better suited and easier than telling people to use Audacity.

This branch isn't ready for merge, but the pictures are already exciting!

There is a native [VCD](https://en.wikipedia.org/wiki/Value_change_dump) output (e.g. for [GTKWave](http://gtkwave.sourceforge.net/) and lots of commercial tools if you have them) and a script to compile .sr files for [Sigrok's Pulseview](https://sigrok.org/).

```
rtl_433 -q -r ./rtl_433_tests/tests/lacrosse/tx141th-bv2/tx141th-bv2-0002.cu8 \
-m 5 ~/Desktop/tx141th-bv2-0002.vcd

./tests/sigrok-conv.sh ./rtl_433_tests/tests/lacrosse/tx141th-bv2/tx141th-bv2-0002.cu8 \
 ~/Desktop/tx141th-bv2-0002.sr
```

### DIFF
--- a/tests/sigrok-conv.sh
+++ b/tests/sigrok-conv.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+if [ -z "$1" ] ; then
+  echo input file missing
+  echo "Usage: $0 input.cu8 output.sr [sample rate in kHz]"
+  exit 1
+fi
+if [ ! -r "$1" ] ; then
+  echo input not found
+  echo "Usage: $0 input.cu8 output.sr [sample rate in kHz]"
+  exit 1
+fi
+file=$1
+
+if [ -z "$2" ] ; then
+  echo output file missing
+  echo "Usage: $0 input.cu8 output.sr [sample rate in kHz]"
+  exit 1
+fi
+if [ -e "$2" ] ; then
+  echo output already exists
+  echo "Usage: $0 input.cu8 output.sr [sample rate in kHz]"
+  exit 1
+fi
+out=$2
+
+if [ -z "$3" ] ; then
+  rate=250
+else
+  rate=$3
+fi
+
+if [ ! -z "$4" ] ; then
+  echo too many arguments
+  echo "Usage: $0 input.cu8 output.sr [sample rate in kHz]"
+  exit 1
+fi
+
+# create I-data channel
+rtl_433 -q -r "$file" -m 7 analog-1-4-1 >/dev/null 2>&1
+# create Q-data channel
+rtl_433 -q -r "$file" -m 8 analog-1-5-1 >/dev/null 2>&1
+# create AM-data channel
+rtl_433 -q -r "$file" -m 5 analog-1-6-1 >/dev/null 2>&1
+# create FM-data channel
+rtl_433 -q -r "$file" -m 6 analog-1-7-1 >/dev/null 2>&1
+# create state channels
+rtl_433 -q -r "$file" -m 10 logic-1-1 >/dev/null 2>&1
+# create version tag
+echo -n "2" >version
+# create meta data
+cat >metadata <<EOF
+[device 1]
+capturefile=logic-1
+total probes=3
+samplerate=$rate kHz
+total analog=4
+probe1=FRAME
+probe2=ASK
+probe3=FSK
+analog4=I
+analog5=Q
+analog6=AM
+analog7=FM
+unitsize=1
+EOF
+
+zip "$out" version metadata analog-1-4-1 analog-1-5-1 analog-1-6-1 analog-1-7-1 logic-1-1
+rm version metadata analog-1-4-1 analog-1-5-1 analog-1-6-1 analog-1-7-1 logic-1-1
--- a/include/pulse_detect.h
+++ b/include/pulse_detect.h
@@ -41,6 +41,9 @@ void pulse_data_clear(pulse_data_t *data);		// Clear the struct
 /// Print the content of a pulse_data_t structure (for debug)
 void pulse_data_print(const pulse_data_t *data);
 
+/// Dump the content of a pulse_data_t structure as raw binary
+void pulse_data_dump_raw(uint8_t *buf, unsigned len, intmax_t buf_offset, const pulse_data_t *data, uint8_t bits);
+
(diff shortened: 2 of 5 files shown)


zuckschwerdt:
sigrok-coarse

sigrok-fine

gtkwave-coarse

gtkwave-fine

merbanan:
What mixed signal digital oscilloscope did you play with?
And I have thought about adding a rtl_sdr input for sigrok. Together with realtime ask/fsk decoding. But I'm not sure there is a real use for realtime visualisation besides that it is "nice". This offline mode seems more practial.

zuckschwerdt:
Nothing fancy or worth the name ;) A clone of that ultra-cheap Saleae 24MHz logic analyzer and a Hantek 6022BE, both with Sigrok.
I briefly pondered the idea of a Sigrok module too. But like you say the short (realtime) snapshots aren't much to work with. I think it's more useful as a visualizer for rtl_433's sample files.
Maybe I'll code some simple general purpose PCM,PWM,PPM,... modules for Sigrok -- perhaps out flex decoder can be a template for that.

rct:
Very nice!    This should be very useful.

I did something similar in the very early days. I had a perl (or python) script that would parse rtl_433 debug output to get the transitions and timings.  It made more sense to me then Audacity since previously I had been using a Bus Pirate, IR Toy, and/or a home made IR widget for digging into consumer IR signal captures.   Right now I can't find the script, but I have some old .vcd and and .olp  (for the open logic sniffer SUMP app) files lying around from 2014 when I was trying to get the LaCrosse decoder working.

merbanan:
@zuckschwerdt try this plugin https://sigrok.org/wiki/Protocol_decoder:Ook. Searching for LHT00SU1 gives a very cheap MSO device. 1(2) channel oscilloscope and 8 channel logic analyzer. I've had some ideas for sniffing the data lines on rf-remotes but it is so much easier to just grab a sample of the air with the rtl_sdr device. But maybe it would be interesting to go from la trace to reading the data with rtl_433.

Anyway this toolkit will be superior to analyze ISM signals. Maybe we can get a PSK demodulator also ?

tests/sigrok-conv.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
+if [ -z "$1" ] ; then
+  echo input file missing
+  echo "Usage: $0 input.cu8 output.sr [sample rate in kHz]"
```
merbanan: Can we add a sample rate parser here if the rate is missing from the arguments? Newer recordings have this in the name.

zuckschwerdt: Definitely, yes. This is just a stub to get the pieces together. I'll also add float-format output and remove sox. I don't think I'll add any zip code though.

merbanan:
@zuckschwerdt I have my fair share of FX2 devices (everyone should have at least one, the LHT00SU1 is a good one to get).

I think that if we add support for raw pulse data then it would be real easy to support these cheap 433 ook receivers. We just need to create a bitbuffer from the input. I'll think a bit about it. My goal is that rtl_433 + sigrok should be a one stop shop for 433MHz work.

zuckschwerdt:
Something like VCD as input, right? I.e. skip the demodulator and read `pulse_data_t` from file or OOK receiver. Sounds good. I'll make VCD (or something like that) an input and then we can add an option to read from OOK hardware.

merbanan:
https://www.sigrok.org/blog/libsigrokdecode-052-released

merbanan:
The ook decoder is there now. Pulseview together with the VCD output should be a superior visualization tool.

iDoka:
VCD support might be to extend user experience by playing with `rtl_433`.
I'm not so familiar to SR file format, but for VCD (that file can really huge size!) we can use LXT2/VZT/FST container with gzip/bzip2 compression (GTKWave have native support for that format).
Please keep in mind for future improvements this great feature.

zuckschwerdt:
@iDoka thanks. Do you have documentation links for the container specs?

iDoka:
From GTKwave man page:

> Native dumpers exist in Icarus Verilog and the open source version of VeriWell for the LXT formats so conversion with `vcd2lxt` or `vcd2lxt2` is not necessary to take direct advantage of LXT with those simulators. 

Any two way:

I. On the fly convertation: https://www.veripool.org/boards/2/topics/150-Verilator-Converting-VCD-file-to-LXT-file-during-simulation

II. Use some piece of LXT2-sources from GTKwave|Verilator|Icarus Verilog, e.g. [A way to support LXT2 file format natively](https://www.veripool.org/boards/3/topics/2601?r=2611)

zuckschwerdt:
I merging this now, as-is. There will be some restructuring to make file output more flexible. For other (better) formats and conversions please open a new issue and we'll address that later. Thanks!

roger-:
I'm testing this out and signalview opens but doesn't show any pulses:

```
./sigrok-open.sh ~/Downloads/gfile001.cu8 
	zip warning: name not matched: analog-1-4-1
	zip warning: name not matched: analog-1-5-1
	zip warning: name not matched: analog-1-6-1
	zip warning: name not matched: analog-1-7-1
	zip warning: name not matched: logic-1-1
  adding: version (stored 0%)
  adding: metadata (deflated 24%)
rm: cannot remove 'analog-1-4-1': No such file or directory
rm: cannot remove 'analog-1-5-1': No such file or directory
rm: cannot remove 'analog-1-6-1': No such file or directory
rm: cannot remove 'analog-1-7-1': No such file or directory
rm: cannot remove 'logic-1-1': No such file or directory
sr: virtual-session: No capture file 'logic-1' in session file './test.sr'.
sr: virtual-session: No capture file 'logic-1' in session file '/tmp/tmp.6qdrRAWMTh/gfile001.cu8.sr'.
```

Am I missing something?

zuckschwerdt:
Looks like the `rtl_433` in your path is not the one compiled from this branch.

roger-:
Yup, thanks!

PR #805, opened 2018-09-07
Add WT0124 to VS2015 project

winterrace:
Update of VS2015 project files after WT0124 device was added

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,118 @@
+# Building rtl_433
+
+## Linux
+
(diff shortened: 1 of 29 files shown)


zuckschwerdt:
Thanks! I'll see if I can come up with a small script to keep us updated and use this as a test. May take some time.

zuckschwerdt:
There are two `<ItemGroup>` in the vcxproj (*.h, *.c), how are these distinguished? Could that be just one big ItemGroup? Or could we label the groups?

PR #807, opened 2018-09-07
Error or waste of resources in classify_signal()

winterrace:
Local array signal_distance_data is initialized with 4000 elements, only 1000 are used.
Overwriting the array with zero at the end of the routine seems unnecessary to me...


winterrace:
Oh, but please check, if this could even be a bug:

classify_signal() evaluates signal_pulse_data in several loops with 1000 iterations, while updating the signal_distance_data array . But  signal_pulse_data has indeed 4000 elements.

Should eventually all loops be 4000 iterations long instead of 1000? Then 4000 entries for signal_distance_data would be correct.

However, clearing signal_distance_data is superflucous in any case.

zuckschwerdt:
Indeed something looks wrong. But both 1000 and 4000 are just upper bounds and won't likely be reached? Please investigate further :)

merbanan:
The reason for this buffer is because I didn't want to allocate a new buffer all the time. And because I don't know how long the signal is going to be I just took a size that should be able to handle most signals. The most optimal way would be to allocate a buffer at the beginning and the do a realloc for more if a signal is larger then the current size.

zuckschwerdt:
This is addressed with c4c71ef now. More clean ups to follow ;)

PR #810, opened 2018-09-08
Add version (-V) and help (-h) option

zuckschwerdt:
adds usual -h option for help and -V for version (available if compiled from Git)


merbanan:
jolly good.

zuckschwerdt:
Trying to get the help text in my maintainer update script and the exit code crashed the whole thing. Turns out we never had an official "-h", it worked by accident. Also people were rightfully complaining about the missing "just tell me the version" option ;) I'll let this sit for a day and merge.

PR #814, opened 2018-09-28
Skylink Motion Sensor

NorthernMan54:
This is tested and working


src/devices/skylink_ha-434tl.c
```diff
@@ -0,0 +1,145 @@
+/* skylink off-brand wireless motion sensor and alarm system on 433.3MHz
+ * Skylink HA-434TL
+ *
+
+   This motion sensor is pretty primitive, but the price is good.  It only sends messages when it see's motion, and no motion clear message.  It also does not appear to send battery levels or low battery.  It also does not appear to send regular heartbeats as well.
+
```
merbanan: Split the lines, and to me the message every 2 hours seems like a heartbeat. The text says no heartbeat signal.

src/devices/skylink_ha-434tl.c
```diff
@@ -0,0 +1,145 @@
...
+   [ 20] Pulse:  132, Gap:  384, Period:  516
+   [ 21] Pulse:  133, Gap: 5061, Period: 5194
+   new_tmplate callback:
+
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ */
```
merbanan: License and copyright at the top, add your name also if you like. Signal description below.

src/devices/skylink_ha-434tl.c
```diff
@@ -0,0 +1,145 @@
...
+
+   The message that is received, appears to only be a device serial, and does not appear to change.
+
+   2018-09-26 21:25:07 :	Skylink motion sensor   :	00000
+   pulse_demod_ppm(): Skylink HA-434TL motion sensor
+   bitbuffer:: Number of rows: 2
+   [00] { 0}           :
+   [01] {17} be 3e 80  : 10111110 00111110 1
+   Pulse data: 19 pulses
+   [  0] Pulse:  507, Gap:  498, Period: 1005   Assuming signal header
```
merbanan: This long pulse is the signal preamble.

src/devices/skylink_ha-434tl.c
```diff
@@ -0,0 +1,145 @@
...
+   {"time" : "2018-09-27 06:55:50", "model" : "Skylink motion sensor", "code" : "5e3e8"}
+   pulse_demod_ppm(): Skylink HA-434TL motion sensor
+   bitbuffer:: Number of rows: 2
+   [00] { 3} e0        : 111
+   [01] {17} 5e 3e 80  : 01011110 00111110 1
+   Pulse data: 22 pulses
+   [  0] Pulse:   12, Gap:  269, Period:  281
+   [  1] Pulse:   13, Gap:  269, Period:  282
+   [  2] Pulse:   10, Gap:  235, Period:  245
+   [  3] Pulse:  506, Gap:  499, Period: 1005  Header
```
merbanan: Preamble again.

src/devices/skylink_ha-434tl.c
```diff
@@ -0,0 +1,145 @@
...
+                        return 1;
+                }
+        }
+        return 0;
+}
+
+static char *output_fields[] = {
+        "time",
+        "model",
+        "code",
```
merbanan: Add motion here, otherwise the csv output is hard to use.

src/devices/skylink_ha-434tl.c
```diff
@@ -0,0 +1,145 @@
...
+        int code;
+        int motion;
+        int raw;
+        char code_str[6];
+        char raw_str[6];
+
+        if (debug_output > 1) {
+                printf("new_tmplate callback:\n");
+                bitbuffer_print(bitbuffer);
+        }
```
merbanan: You can remove this block.

src/devices/skylink_ha-434tl.c
```diff
@@ -0,0 +1,145 @@
...
+
+        for (int i = 0; i < bitbuffer->num_rows; ++i) {
```
merbanan: How many times does the message repeat? The textual description does not say. And you should maybe add a rejection test based on that. Otherwise we might get lots of false positives with this decoder.

merbanan:
The actual PR has conflicts so do a rebase and also squash the commits to one commit with a proper commit message. What that is fixed and all the comments the code is good to be merged.

zuckschwerdt:
Before any more work goes into this: note that we support simple fixed-id doorbell-style (e.g. motion) sensors with the flex decoder and a simple text file.
I.e. please add a spec to https://github.com/merbanan/rtl_433/blob/master/decoders.cfg

merbanan:
Can that handle the heart beat functionality?

zuckschwerdt:
Yes! :) The flex getters should be quite powerful if you just want to pull bits and not do arithmetic on them. See #786

merbanan:
I see, yeah then this is a candidate for the flex decoder.

zuckschwerdt:
Something like this should work (untested):

    rtl_433 -q -F json -R 0 \
    -X 'skylink:OOK_PPM_RAW:600:1700:10000,get={4}:motion:[5:alive 11:motion],get=@4:{12}:code' \
    -r ../rtl_433_tests/tests/skylink_motion/g001_433.92M_250k.cu8

NorthernMan54:
Tks folks for the feedback, let me do further testing with the flex decoder.  And thoughts around when #786 will be merged?

zuckschwerdt:
I think I'll get some time on sunday. There's always on more thing I'd like to change but I guess I'll merge and then let it evolve.
It goes along with #790 -- then the spec can be loaded from a config file like here: https://github.com/merbanan/rtl_433/pull/790/files#diff-764908057ecc0e4bef84e9825b8c6250R310

NorthernMan54:
@zuckschwerdt I was playing with the version in the pull request #786 and found that this worked

./rtl_433 -q -F json -R 0 -X 'skylink:OOK_PPM_RAW:600:1700:10000,bits=17,get={3}:motion:[5:motion 2:alive],get=@3:{14}:code'

And as output it gave

{"time" : "2018-09-30 17:57:47", "model" : "skylink", "count" : 1, "num_rows" : 2, "rows" : [{"len" : 0, "data" : "", "motion" : 0, "code" : 0}, {"len" : 17, "data" : "be3e8", "motion" : "motion", "code" : 97405}], "codes" : ["{0}", "{17}be3e8"]}

Repeated a bunch of times as the device repeats itself 7 times for each message

A couple of comments

1) The value returned by the @ function was a bit wacky, ie from get=@3:{14}:code.  I was expecting 15485 and not 97405 - 97405 is decimal for be3e8 which was the whole message and didn't start at the 3rd entry

2) To reduce noise, it is possible to only return rows where bits matches ? And ignore others where it doesn't?

zuckschwerdt:
re 1, thanks, I'll look into it
re 2, yes, using `match=` should work (esp. if you are looking for a single fixed ID)

zuckschwerdt:
Oh, re 1, that was a marked TODO. Fixed with 1cf4e18

PR #815, opened 2018-09-30
Add file info detection

zuckschwerdt:
Replaces the `-m N` file mode option with file info parsed from the filename.
File extensions like `.cu8` or `.am.cs16` are detected and override prefixes like `F32:AM:amplitude.raw` are possible.


merbanan:
Nice and clear

PR #816, opened 2018-10-01
kerui: cleanup

roger-:
A few minor changes:

* rename to Kerui PIR / Contact Sensor
* cleanup state names (e.g. PIR -> motion)
* move state to separate field to make easier to parse

### DIFF
--- a/src/devices/kerui.c
+++ b/src/devices/kerui.c
@@ -49,7 +49,7 @@ static int kerui_callback(bitbuffer_t *bitbuffer) {
     local_time_str(0, time_str);
     data = data_make(
             "time",     "",               DATA_STRING, time_str,
-            "model",    "",               DATA_STRING, "Kerui PIR / Contact Sensor",
+            "model",    "",               DATA_STRING, "Kerui Security",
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
A needed clean-up, thanks. All those changes are breaking the current output though. I don't know if people rely on the current names and format?
Let this sit for two weeks and see if anyone complains ;)

roger-:
No complaints yet!

zuckschwerdt:
But let's change the "model" string to the shorter, agreed format of two words while we are at it. "Kerui Security" sounds good.

roger-:
Sure -- done!

zuckschwerdt:
Sorry, only the model. The protocol name can be descriptive, like you had.

roger-:
Got it, reverted part of the edit.

zuckschwerdt:
Thanks. Reasoning (if someone stumbles apon this later):
- r_device.name is a descriptive string for the help texts
- data.model is a concise key for the consumers (which might use it as machine-readable key/tag)

Issue #821, opened 2018-10-05
Assistance for (another) Toyota TPMS


gcormier:
Wasn't sure if the issue should be opened here... 0 open in the other repo so I figured it might be best here.

PR with the information I've captured : https://github.com/merbanan/rtl_433_tests/pull/236

zuckschwerdt:
Thanks for the samples. Looks like this sensor has a half-bit width of 100 µs (all other so far have 50 µs).
I'd start by modifying the existing tpms_toyota.c device and see if that works. Then report back and we'll see where the new code fits.

E.g. at https://github.com/merbanan/rtl_433/blob/master/src/devices/tpms_toyota.c#L110

    .short_limit    = 100, // 25 samples @250k
    .long_limit     = 100, // FSK
    .reset_limit    = 250, // Maximum gap size before End Of Message [us].

Gives you (add `bitbuffer_print(bitbuffer);`)

    a.cu8: {144} fc b3 2b 32 ab 4b 2d 55 4d 52 aa d4 cc b4 aa ac b5 2c
    b.cu8: {143} fc cc aa ad 2a b4 d4 b3 4a ad 55 54 cc ca cc cd 2d 2c
    c.cu8: {144} fc cd 55 55 4a b4 d4 b2 b2 ad 55 2b 33 4b 55 4d 54 ac
    d.cu8: {144} fc b3 2b 32 ab 4b 2d 55 52 ad 55 2b 33 4b 55 52 d5 2c

in tpms_toyota_callback().
Which looks correct but has no simple preamble. E.g. the buffers start:

    1111110010110011
    1111110011001100
    1111110011001101
    1111110010110011

So maybe pick a preamble of `11111001` (admitting for a missed first bit-period and neatly fitting in a byte): 0xF9.

I.e.

    static const unsigned char preamble_pattern[2] = {0xf9}; // 8 bits
    …
    // Find a preamble with enough bits after it that it could be a complete packet
    while ((bitpos = bitbuffer_search(bitbuffer, 0, bitpos, (const uint8_t *)&preamble_pattern, 8)) + 134 <=
            bitbuffer->bits_per_row[0]) {
        events += tpms_toyota_decode(bitbuffer, 0, bitpos + 8);
        bitpos += 2;
    }

Now you can look at the DMC-decoded data:

    {67} f3 c2 b4 0c 41 3e a0 69 60
    {67} e0 50 b2 e8 40 3f 9f d5 60
[...]

PR #825, opened 2018-10-11
Add PMV-107J TPMS

zuckschwerdt:


zuckschwerdt:
Analyzed and discussed in #821

PR #826, opened 2018-10-11
MQTT: Add Model ID to devices and script to generate proper MQTT topics

vestom:
**Request for Comment**
Currently the recommended way to pass rtl_433 output to MQTT is via a simple pipe to a topic common for all models and sensors (`rtl_433 -F json -U | mosquitto_pub -t home/rtl_433 -l`). 

However that means MQTT subscribers will have to filter state updates based on the JSON content. This is not the [recommended](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices) way to generate MQTT topics, and not supported by e.g. [Home Assistant](https://www.home-assistant.io), which have [rejected](https://github.com/home-assistant/home-assistant/pull/12666) adding such a feature.

Since the Model names in rtl_433 are long and contain spaces and slashes, it is not possible to make a canonical MQTT topic like `home/rtl_433/<model>/<id>`, which would generate a separate topic for each sensor update and make integration with e.g. Home Assistant easy.

I therefore propose adding a `mid` or "Model ID" to each device, consisting of a unique short string (Max ~8 characters). That will make it possible to generate a topic like `home/rtl_433/<mid>/<id>` or example `home/rtl_433/FOWH0530/183` (for the "Fine Offset Electronics, WH0530 Temperature/Rain sensor" with sensor ID 183).

A sample Python script for piping rtl_433 to MQTT topics is included such that the new recommended way to output to MQTT will be `rtl_433 -F json -U | tests/rtl_433_mqtt.py`. Furthermore `mid` is added to a couple of devices. I will add it to the rest (~100), if the proposal is accepted.

### DIFF
--- a/src/devices/calibeur.c
+++ b/src/devices/calibeur.c
@@ -82,6 +82,7 @@ static int calibeur_rf104_callback(bitbuffer_t *bitbuffer) {
 		local_time_str(0, time_str);
 		data = data_make("time",          "",            DATA_STRING, time_str,
 						"model",         "",            DATA_STRING, "Calibeur RF-104",
+						"mid",           "MID",         DATA_STRING, "CALRF104",
 						"id",            "ID",          DATA_INT, ID,
 						"temperature_C", "Temperature", DATA_FORMAT, "%.1f C", DATA_DOUBLE, temperature,
 						"humidity",      "Humidity",    DATA_FORMAT, "%2.0f %%", DATA_DOUBLE, humidity,
--- a/src/devices/chuango.c
+++ b/src/devices/chuango.c
@@ -71,6 +71,7 @@ static int chuango_callback(bitbuffer_t *bitbuffer) {
     data = data_make(
             "time",     "",             DATA_STRING, time_str,
             "model",    "",             DATA_STRING, "Chuango Security Technology",
+            "mid",      "MID",          DATA_STRING, "CHUANGO",
             "id",       "ID",           DATA_INT,    id,
             "cmd",      "CMD",          DATA_STRING, cmd_str,
             "cmd_id",   "CMD_ID",       DATA_INT,    cmd,
--- a/src/devices/danfoss.c
+++ b/src/devices/danfoss.c
@@ -138,6 +138,7 @@ static int danfoss_CFR_callback(bitbuffer_t *bitbuffer) {
 		data = data_make(
 			"time",		"",		DATA_STRING,	time_str,
 			"model",		"",		DATA_STRING,	"Danfoss CFR Thermostat",
+			"mid",		"MID",		DATA_STRING, "DANCFR",
 			"id",		"ID",		DATA_INT,	id,
 			"temperature_C", 	"Temperature",	DATA_FORMAT,	"%.2f C", DATA_DOUBLE, temp_meas,
 			"setpoint_C",	"Setpoint",	DATA_FORMAT,	"%.2f C", DATA_DOUBLE, temp_setp,
--- a/src/devices/fineoffset.c
+++ b/src/devices/fineoffset.c
@@ -41,6 +41,7 @@ static int fineoffset_WH2_callback(bitbuffer_t *bitbuffer) {
     char time_str[LOCAL_TIME_BUFLEN];
 
     char *model;
+    char *mid;
     int type;
     uint8_t id;
     int16_t temp;
@@ -51,22 +52,22 @@ static int fineoffset_WH2_callback(bitbuffer_t *bitbuffer) {
             bb[0][0] == 0xFF) { // WH2
         bitbuffer_extract_bytes(bitbuffer, 0, 8, b, 40);
         model = "Fine Offset Electronics, WH2 Temperature/Humidity sensor";
-
+        mid = "FOWH2";
     } else if (bitbuffer->bits_per_row[0] == 55 &&
             bb[0][0] == 0xFE) { // WH2A
         bitbuffer_extract_bytes(bitbuffer, 0, 7, b, 48);
         model = "Fine Offset WH2A sensor";
-
+        mid = "FOWH2A";
     } else if (bitbuffer->bits_per_row[0] == 47 &&
             bb[0][0] == 0xFE) { // WH5
         bitbuffer_extract_bytes(bitbuffer, 0, 7, b, 40);
         model = "Fine Offset WH5 sensor";
-
+        mid = "FOWH5";
     } else if (bitbuffer->bits_per_row[0] == 49 &&
             bb[0][0] == 0xFF && (bb[0][1]&0x80) == 0x80) { // Telldus
         bitbuffer_extract_bytes(bitbuffer, 0, 9, b, 40);
         model = "Telldus/Proove thermometer";
-
+        mid = "TELLTHER";
     } else
         return 0;
 
@@ -110,6 +111,7 @@ static int fineoffset_WH2_callback(bitbuffer_t *bitbuffer) {
     if (b[3] == 0xFF) {
         data = data_make("time",          "",            DATA_STRING, time_str,
                          "model",         "",            DATA_STRING, model,
+                         "mid",           "MID",         DATA_STRING, mid,
                          "id",            "ID",          DATA_INT, id,
                          "temperature_C", "Temperature", DATA_FORMAT, "%.01f C", DATA_DOUBLE, temperature,
                          "mic",           "Integrity",   DATA_STRING, "CRC",
@@ -120,6 +122,7 @@ static int fineoffset_WH2_callback(bitbuffer_t *bitbuffer) {
     else {
         data = data_make("time",          "",            DATA_STRING, time_str,
                          "model",         "",            DATA_STRING, model,
+                         "mid",           "MID",         DATA_STRING, mid,
                          "id",            "ID",          DATA_INT, id,
                          "temperature_C", "Temperature", DATA_FORMAT, "%.01f C", DATA_DOUBLE, temperature,
                          "humidity",      "Humidity",    DATA_FORMAT, "%u %%", DATA_INT, humidity,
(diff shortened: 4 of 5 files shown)


vestom:
Example configuration for Home Assistant "configuration.yaml":
```
sensor:
  - platform: mqtt
    name: "Garage Temp"
    state_topic: "home/rtl_433/CALRF104/1"
    value_template: "{{ value_json['temperature_C'] }}"
    unit_of_measurement: "°C"
    force_update: true
  - platform: mqtt
    name: "Garage Humi"
    state_topic: "home/rtl_433/CALRF104/1"
    value_template: "{{ value_json['humidity'] }}"
    unit_of_measurement: "%"
    force_update: true
````

zuckschwerdt:
Something like this is very much needed! It could clean up Time-Series DB (or plain RDMS) logging and e.g. CSV as well.

We should perhaps deprecated model and then make it available only with an option switch someday. With plenty of time for people to catch up and switch of course.

Consistent output of "type" (currently only for TPMS) would also be great.

Also "brand" isn't used widely (mainly "OS" and 2 more) it could be dropped, the first few letters in your example would convey the same information.

For MQTT perhaps <mid>/<channel> is even easier to consume, with a fallback to <mid>/<id> if there is no channel (maybe even <type>/<mid>/<channel> if available).

zuckschwerdt:
Perhaps 8 chars is a little short? Is there some prefered length on consumers?
I'm thinking about shortening e.g. Acurite, Akhan, Alecto, Brennenstuhl, Bresser, Calibeur: AC, AK, AL, BS, BR, CA? It'll get messy.
Would `"one word for manufacturer" - "one word for model"` work? I would like to have that convention on model strings already.

Also if you want it really short (but only machine readable) the protocol-id is available already. There was already a proposal to output that with the data.

Just some thoughts.

vestom:
Good questions. I think adding a lot of intelligence to mid should be avoided. The primary purpose is for it to be unique and machine readable. I also considered just doing a hash (e.g. MD5) of the model string in the Python script instead. But it just makes configuration on the receiver side rather user unfriendly, so some human familiarity is nice :-)

Protocol-id could also be fine, as It would make a nice correlation between the command line configuration (`-R <pid>`). But it is unfortunately not unique, as some protocols have multiple sensor models. It has always been an eye-sore, that we have both 'device.name' and 'model', which are often just subtly different. But could we somehow clean that up?

8 chars is not a hard limit, but we should avoid making it a _third_ largish model string. Then we should just shorten up the model strings and make them machine friendly as you propose instead.

PS. IMHO 'brand' is superfluous and 'type' ought to have some associated rules. But that is a whole different can of worms :-) 'channel' could easily be added as an optional extra parameter to the script (e.g. `home/rtl_433/<mid>[/<id>[/channel]]`)

merbanan:
@vestom I will only accept changes if you add a native mqtt client :) 

Anyway could the mid be generated automatically based on the model name and maybe hash output? That way it would always be there and we could generate more then 8 characters. I'm thinking one could take the protocol-id + upper case model name letters + all the numbers and then pad with some hash of the model name. It would not be that pretty but we have the model name for that.

zuckschwerdt:
Yes, I use the channel instead of id, if available, on my setup. Keeps the TSDB logging and Home Assistant happy when I make my yearly rounds changing batteries.

There is already a (mostly) working [MQTT support](https://github.com/merbanan/rtl_433/tree/feat-mqtt). I just have to get around to think up the data to topic mapping…

vestom:
@merbanan Be careful what you wish for :-) At one point I toyed with the idea using the paho libraries. However it would require rtl_433 to become multi-threaded and require yet another boat-load of command line parameters. 

I actually quite like the current "do one thing and do it well" approach of rtl_433 and then use some Python or shell scripting to pipe to MQTT, SQL databases,  or whatever makes you tick :-) 

I think the important part is getting the data format really consistent and well defined. Given the diversity of devices that is not an easy task...

vestom:
@zuckschwerdt Heh. There you go. Well done :-)

vestom:
It was my impression that "id" is the main differentiator between devices of same model. On some devices the id is non-volatile, on others it changes at each power up, and sometimes it is even configurable by a switch. However, I can see some devices have _both_ an id and a channel, so I suppose they might have both an internal ID and a switch? In that situation, it of course makes sense to add the extra 'channel' parameter.

... We really need an interface specification, before adding new data fields... :-/

merbanan:
Some devices have fixed bits, they could be considered as an id. Some devices randomizes some bits during power up when inserting a battery, I consider that as a random id (rid for short). Then we have devices that have the ability to set bits to fixed patterns, I consider those bits as channel bits.

merbanan:
All these can be used to uniquely identify sensors. The random id is less useful but sometimes you take what you got and make the best of it.

zuckschwerdt:
Here is a list of current model strings for reference (might be missing a few odd cases), if we strip e.g. "sensor" postfix or shorten excessive strings a bit it isn't in a too bad shape.

    "Acurite Rain Gauge"
    "Acurite 609TXC Sensor"
    "Acurite Lightning 6045M"
    "Acurite tower sensor"
    "Acurite 5n1 sensor"
    "Acurite 5n1 sensor"
    "Acurite 3n1 sensor"
    "Acurite 986 Sensor"
    "Acurite 606TX Sensor"
    "Akhan 100F14 remote keyless entry"
    "AlectoV1 Wind Sensor"
    "AlectoV1 Rain Sensor"
    "AlectoV1 Temperature Sensor"
    "Ambient Weather F007TH Thermo-Hygrometer"
    "blyss dc5-uk-wh"
    "Brennenstuhl RCS 2044"
    "Bresser 3CH sensor"
    "Calibeur RF-104"
    "Cardin S466"
    "Chuango Security Technology"
    "CurrentCost TX"
    "CurrentCost Counter"
    "Danfoss CFR Thermostat"
    "Dish remote 6.3"
    "DSC Contact"
    "Efergy e2 CT"
    "Efergy Optical"
    "Elro DB286A"
    "emonTx"
    "Esperanza EWS"
    "Fine Offset WH24"
    "Fine Offset Electronics, WH25"
    "Fine Offset Electronics, WH0530 Temperature/Rain sensor"
    "Fine Offset WH1050 weather station"
    "Fine Offset Electronics WH1080/WH3080 Weather Station"
    "Fine Offset Electronics WH1080/WH3080 Weather Station"
    "Fine Offset Electronics WH3080 Weather Station"
    "Ford Car Remote"
    "FT-004-B Temperature Sensor"
    "GE Color Effects Remote"
    "Generic motion sensor"
    "Generic Remote"
    "Generic temperature sensor 1"
    "GT_WT_02 sensor"
    "HIDEKI TS04 sensor"
    "HIDEKI Wind sensor"
    "HIDEKI Rain sensor"
    "Honda Remote"
    "Honeywell Door/Window Sensor"
    "HT680 Remote control"
    "IBIS beacon"
    "inFactory sensor"
    "Inovalley kw9015b"
    "Interlogix"
    "Intertechno"
    "Kedsum Temperature & Humidity Sensor"
    "Kerui PIR Sensor"
    "LaCrosse TX Sensor"
    "LaCrosse TX Sensor"
    "LaCrosse TX141-Bv2 sensor"
    "LaCrosse TX141TH-Bv2 sensor"
    "TX29-IT"
    "TX35DTH-IT"),
    "LaCrosse WS"
    "LaCrosse WS"
    "LaCrosse WS"
    "LaCrosse WS"
    "Wireless M-Bus"
    "ET-732/ET-733"
    "Mebus/433"
    "New Template"
    "KlikAanKlikUit Wireless Switch"
    "Nexa"
    "Nexus Temperature"
    "Nexus Temperature/Humidity"
    "Oil Ultrasonic STANDARD"
    "Oil Watchman"
    "THGR122N"
    "THGR968"
    "WGR968"
    "BHTR968"
    "RGR968"
    "THR228N"
    "THN132N"
    "RTGN129"
    "RTGN318"
    "THN129"
    "BTHGN129"
    "Oregon Scientific UVR128"
    "THGR810"
    "THN802"
    "UV800"
    "PCR800"
    "PCR800a"
    "WGR800"
    "CM160"
    "CM180"
    "CM180"
    "Oregon Scientific SL109H"
    "OSv1 Temperature Sensor"
    "Philips outdoor temperature sensor"
    "Prologue sensor"
    "Proove"
    "Quhwa doorbell"
    "RadioHead ASK"
    "Sensible Living Plant Moisture"
    "RF-tech"
    "Rubicson Temperature Sensor"
    "S3318P Temperature & Humidity Sensor"
    "Schrader"
    "Schrader Electronics EG53MA4"
    "Silvercrest Remote Control"
    "SimpliSafe Sensor"
    "SimpliSafe Keypad"
    "SimpliSafe Keypad"
    "Smoke detector GS 558"
    "Solight TE44"
    "Springfield Temperature & Moisture"
    "Steelmate"
    "TFA pool temperature sensor"
    "TFA-Twin-Plus-30.3049"
     MODEL,
     MODEL,
    "Citroen"
    "Ford"
    "PMV-107J"
    "Renault"
    "Toyota"
    "Vaillant VRT340f Central Heating Thermostat"
    "Vaillant VRT340f Central Heating Thermostat (RF Detection)"
    "Waveman Switch Transmitter"
    "WG-PB12V1"
     MODEL,
    "WT0124 Pool Thermometer"
    "WT450 sensor"
    "X10 Security"

vestom:
That is true! Maybe we should just sanitize the model strings according to your <Manufacturer> <Model> proposal above, remove problematic characters (/, & etc.) and strip redundant words like "sensor", "wireless", "weather station" etc. Then we have a lean model identifier and the device.name string can be the more descriptive one.

Keep It Simple.

I will close this proposal and work at bit on the description in https://github.com/merbanan/rtl_433/pull/827 before we do anything drastic :-)

PR #827, opened 2018-10-11
Proposed Interface Specification for data output

vestom:
Purpose is to remove some ambiguity and achieve some commonality across different sensors to improve machine readability and ease integration with third party software.

This is sure to spark some discussion :-)

### DIFF
--- a/DATA_FORMAT.md
+++ b/DATA_FORMAT.md
@@ -58,16 +58,16 @@ Various data fields, which are common across devices of different types
 Due to the large variance in sensor types this list of common values is non-exhaustive. Additional data value fields should follow the form: `<Type>_<Unit>`, where *Unit* should be in sensors native units insofar possible with no conversion.
 Examples:
 
-* **temperature_C** (double) (Optional)
-  * Temperature from a temperature sensor in degrees Celcius.
+* **temperature_C** (**temperature_f**) (double) (Optional)
+  * Temperature from a temperature sensor in degrees Celcius (Fahrenheit).
   
-* **setpoint_C** (double) (Optional)
-  * Thermal set point of a thermostat device in degrees Celcius.
+* **setpoint_C** (**setpoint_f**) (double) (Optional)
+  * Thermal set point of a thermostat device in degrees Celcius (Fahrenheit).
   
 * **humidity** (double) (Optional)
   * Humidity from a hygrometer sensor in % relative humidity
   
-* **wind_dir_deg** (int) (Optional)
+* **wind_dir_deg** (double) (Optional)
   * Wind direction from wind sensor in compass direction degrees.
   
 * **wind_speed_ms** (double) (Optional)
@@ -76,9 +76,12 @@ Examples:
 * **gust_speed_ms** (double) (Optional)
   * Gust wind speed from wind sensor in m/s. 
 
-* **rainfall_mm** (double) (Optional)
-  * Rainfall from rain sensor in mm since last reset. Reset method is device dependant.
+* **rain_mm** (**rain_in**) (double) (Optional)
+  * Rainfall from rain sensor in mm (inches) since last reset. Reset method is device dependant.
   
-* **pressure_hPa** (double) (Optional)
-  * Air pressure from barometer or Tire Pressure Monitor in hPa (millibar)
+* **rain_rate_mmph** (**rain_rate_inph**) (double) (Optional)
+  * Rainfall rate from rain sensor in mm per hour (inches per hour).
+  
+* **pressure_hPa** (**pressure_psi**) (double) (Optional)
+  * Air pressure from barometer or Tire Pressure Monitor in hPa (psi)
   
--- a/DATA_FORMAT.md
+++ b/DATA_FORMAT.md
@@ -66,10 +66,10 @@ Examples:
 * **wind_dir_deg** (double) (Optional)
   * Wind direction from wind sensor in compass direction degrees.
   
-* **wind_speed_ms** (double) (Optional)
+* **wind_speed_mps** (double) (Optional)
   * Average wind speed from wind sensor in m/s. Averaging time is sensor dependant.
 
-* **gust_speed_ms** (double) (Optional)
+* **gust_speed_mps** (double) (Optional)
   * Gust wind speed from wind sensor in m/s. 
 
 * **rain_mm** (**rain_in**) (double) (Optional)
--- a/DATA_FORMAT.md
+++ b/DATA_FORMAT.md
@@ -44,24 +44,20 @@ particular device model.
 ## Common Device Data
 Various data fields, which are common across devices of different types 
 
-* **battery** (string) (Optional)
-  * Battery status indication
-  * Possible values: "OK", "LOW"
+* **battery** (double) (Optional)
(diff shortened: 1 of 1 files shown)


roger-:
+1!

I suggest having a `device_type` field to identify the type of sensor (e.g. `weather`, `temperature`, `motion`) to make the data more machine friendly. 

Maybe also a `payload` field to include full the undecoded payload data, in the event that someone wants to subsequently decode a field that isn't supported by rtl_433.

vestom:
I do not think it is really feasible to add a `device_type` field. Many of the device are some combination of sensors. E.g thermometer/rainfall, thermometer/hygrometer/barometer, thermometer/anemometer/windspeed. Basically we would have to add a device type for each sensor value returned by the device. And in that case you can just as well infer that a device that returns a `temperature_C` contains a thermometer, `humidity` contains a hygrometer etc.

roger-:
I see your point, but you could classify the sensors you mentioned as type “weather”. 

But what you say is true, so it may not be necessary.

merbanan:
@vestom we could add a bit field with what sensors a device has. Not sure it would be worth it though. It's obvious from the parameters types. Anyway Temperature_F can be present also if the native representation is that.

zuckschwerdt:
I wanted to present TPMS as a good case for `type` but that could easily be a model prefix, e.g. `TPMS-PM107J`, `TPMS-Ford`, … Not sure anymore what good be gained from a type output now.

merbanan:
Maybe we should give up on typing the devices and instead just provide unit measures?

merbanan:
But then again it would help users identify/classify things which is nice and maybe really user friendly.

vestom:
IF we want to add a "type", I think it would be best to add it in a separate field and keep it orthogonal to the "model" field. The primary purpose of the "model" field is to identify/recognize the device, whereas a "type" field would supposedly be some form of "class" or "template" telling something about the data fields you can expect from the sensor and maybe how to interpret them. I have no strong feelings about it, other than I will recommend to define clearly what the different types signify and what you can expect from a given type :-)

Regarding Temperature_F, the big decision is, whether we still want to allow all kinds of units, or we want to standardize to a limited set? If we do not want to standardize, I will change the proposal to reflect the status quo.

However. The main purpose of the proposal is to define some rules for the "model" field, to make this mandatory field a bit more uniform. Hopefully we can agree upon something about this starting point and then later evolve the spec to include possible new changes or additions?

roger-:
I think ultimately it would be nice if the output had enough information to be automatically consumable by things like Home Assistant. 

I’ve noticed that there are certain temperature sensors that aren’t weather sensors (e.g. freezer temperature), so looking at the presence of a temperature field wouldn’t be enough to determine that.  

Perhaps there should be a “subtype” field to put more specific information? So type would be more of a category.

zuckschwerdt:
While some sensors have a clear purpose, the placement and usage is not easily discoverable. (Some of my Temp./Hum. sensors are outdoor, indoor or even in a fridge)

The protocol is not enough to determine the sensor manufacturer or usage also. There are too many clones with sometimes different purpose, sadly.
So "manufacturer" ("brand") will go away where still used, best we can do is guess the kind of transmitter/protocol.

merbanan:
@vestom I'd like it to default to si-units. But decoders should output the native values.

rct:
Regarding converting units - The problem has been rounding errors that creep in when doing multiple conversions.  This is due to pretty printing going on even for what's intended to be machine readable output like JSON.  At some appropriate point, I think we should stop rounding some of these values to 1 or 2 decimal places.

DATA_FORMAT.md
```diff
@@ -0,0 +1,77 @@
...
+    they are prone to excessive false positives.
+  * Possible values:
+    * "CRC" - Cyclic Redundancy Check.
+    * "CHECKSUM" - Accumulated sum of data.
+    * "PARITY" - Parity bit (odd, even, multiple)
+
+## Common Device Data
+Various data fields, which are common across devices of different types 
+
+* **battery** (string) (Optional)
```
rct: This is a case where if we are writing a specification we should attempt to set a standard we want everything to follow.

I propose 
- `battery` (with no units) should be considered deprecated. 
- `battery_low` or `battery_ok` should be added as a boolean (or an int is essentially a boolean).

zuckschwerdt:
As far as I see pretty printing is only for KV and CSV. Specifically JSON disregards DATA_FORMAT.
Reasoning here is KV and CSV are primarily for human consumption, JSON (and other networked ones) are primarily for machines.

I'd say we don't need a default conversion. Users (and packages) can preset options using config files soon.

DATA_FORMAT.md
zuckschwerdt: This is a hot topic. The idea is clear, "low battery" is an "event" thus "1" anything else is default and disregarded, hence "0".
The opposite reasoning is:
- devices that report a voltage or known steps should use `battery_mV`
- other devices should report a level between 0 (empty) and 1 (full)
- if only a flag is available this degrades to quasi-bool 0/1

DATA_FORMAT.md
```diff
@@ -0,0 +1,77 @@
...
+    Depending of device model it may be a non-volatile value programmed into the device, 
+    a volatile value that changes at each power on (or battery change), or a value configurable by
+    user e.g. by switch or jumpers. No assumptions should be made to the id value other than it contains
+    a unique sequence of alphanumeric characters.
+
+* **channel** (string) (Optional)
+  * Secondary device identification. For devices with more than one identification value 
+    (e.g. both an internal value and a switch).
+
+* **mic** (string) (Optional)
```
rct: In terms of a data architecture, MIC as currently defined is redundant data.  

-- It will never change between messages on the same device.  So there is no point in sending and storing it *on every message*.
-  It isn't an attribute of the message., It's an attribute of *the device/the decoder*.

The use case of disabling devices that don't meet the specified level seems like a runtime option for rtl_433. 

Alternatively rtl_433 should output (on start up) a machine readable message about what devices are enabled, and their attributes like MIC.

merbanan: The idea is that the consumer can add more trust to a message if mic is present. I get your point but in a blind scanning scenario when you pickup unknown transmitters it is much more user friendly to directly see that the protocol has an integrity check.

DATA_FORMAT.md
```diff
@@ -0,0 +1,83 @@
...
+For some devices these are the *only* fields contained in the message, as the message itself consitutes an event from this
+particular device model.
+
+* **time** (string) (Required)
+  * Time stamp. String containing date and time of when the message was received. Format is dependent on
+    current locale unless the `-U` command line argument is used.
+
+* **model** (string) (Required)
+  * Device model. Human readable string concisely describing the device by manufacturer name 
+    and manufacturers model designation according to the following syntax: `"<Manufacturer>, <Model>"`.
```
zuckschwerdt: Can you change this to read "<Manufacturer>-<Model>"?

vestom: Regarding units, I suggest changing from using 'p' as division ('per') to underscore '_':
Examples:
m_s (meters per second)
mi_h (miles per hour)
in_h (inches per hour)
The reason is to avoid ambiguity like the common "mph".

zuckschwerdt: How many keys would that concern? At a glance I only see acurite, and hideki, 3 keys total. Worth the change.

PR #828, opened 2018-10-12
Fix fileformat.c

winterrace:
Currently, rtl_433 can not be compiled any more by using MS VisualStudio 2015 and 2017 (tested both).
These 2 changes are required in fileformat.c to make it compile again.

Because:
- strings.h is not part of standard C. For MS compilers, I defined stncasecmp as _strnicmp having the same syntax.
- variable length arrays are not yet supported by MS compilers. I wrote a malloc/free variant which is only used when using a MS compiler.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,118 @@
+# Building rtl_433
+
+## Linux
+
+Compiling rtl_433 requires [rtl-sdr](http://sdr.osmocom.org/trac/wiki/rtl-sdr) to be installed.
+
+Depending on your system, you may also need to install the following libraries.
+
+Debian:
+
+    sudo apt-get install libtool libusb-1.0.0-dev librtlsdr-dev rtl-sdr build-essential autoconf cmake pkg-config
+
+Centos/Fedora/RHEL (for Centos/RHEL with enabled EPEL):
+
+    sudo dnf install libtool libusb-devel rtl-sdr-devel rtl-sdr
+
+### CMake
+
+Installation using cmake:
+
+    cd rtl_433/
+    mkdir build
+    cd build
+    cmake ../
+    make
+    make install
+
+### autoconf
+
+Installation using autoconf:
+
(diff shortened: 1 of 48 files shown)


merbanan:
VLA's are evil, replace it with a fixed large size buffer. Don't riddle the code with ifdefs.

winterrace:
Even better, thank you.

winterrace:
As requested, I removed the ifdefs within the code by fixing the array to 512 bytes (should be enough).

winterrace:
By the way, I suppose this routine will yield undesired behavior when windows style paths are used (e.g. C:\path\filename.am.s16) because of the colon.
Could we use another separator here, e.g. |? Further alternatives (characters which are forbidden in file names) would be *, ? or < / >.
Shall I open a new issue?

zuckschwerdt:
Oh, that colon was a dumb choice then, indeed. I would like to enter it without the need of shell escapes. Any other ideas, esp. not `; ! * ? < > $ |`? Maybe tilde, plus, or hash – though I don't seem to be the best judge here ;)

merbanan:
OCTOTHORPE

merbanan:
but really ~ is fine

winterrace:
Though, in principle, hash or tilde could be part of valid file names, I could also live with one of these.
I don't see any significant use cases that we block this way...

zuckschwerdt:
Just throwing another idea in the ring, double colon `::`…

winterrace:
What about this PR's state? Is the current fix for the "won't compile" problem OK?
Should I change something like reducing the 512 byte buffer a bit further?

And who will address the separator problem?

zuckschwerdt:
I pushed a small fix (no backslash after colon allowed) and will get on this again later.

zuckschwerdt:
Can you test again? Should be properly fixed now. A colon followed by a backslash is now disregarded.

PR #830, opened 2018-10-12
Suppress irritating/false "WARNING: Failed to set frequency correction"

winterrace:
Since the introduction of the new sdr.c this warning shows up on each rtl-sdr initialization because rtlsdr_set_freq_correction is now called with ppm 0 by default, which equals rtl-sdr's default value.

(rtlsdr_set_freq_correction returns -2 if the value is already set - this is not really an error here).


zuckschwerdt:
Perhaps we should skip setting a PPM of 0 altogether? (i.e. abort the set if the default of `0` is given)
I don't know the semantics for rtlsdr here – is that a real error?

merbanan:
No, ppm correction is bogus in this case anyway.

zuckschwerdt:
If no option for PPM correction is given we take the default setting of `0` and apply that. Should we remove that and never set a PPM = 0. Or maybe change the No-setting value to `-1` and ignore that?

merbanan:
never set ppm=0

zuckschwerdt:
@winterrace can you change the patch to exit `sdr_set_freq_correction()` early if the ppm argument is zero?

winterrace:
I just updated it.
However, I placed this check at the caller (program start after command line evaluation).
At least once rtl_433 will have evolved to a more lib-compatible state, it will be useful to correct the PPM value in later calls, so sdr.c should not refuse a zero PPM in general.

src/rtl_433.c
```diff
@@ -1435,7 +1435,7 @@ int main(int argc, char **argv) {
             r = sdr_set_tuner_gain(dev, gain, !quiet_mode);
         }
 
-        r = sdr_set_freq_correction(dev, ppm_error, !quiet_mode);
+        if(ppm_error) r = sdr_set_freq_correction(dev, ppm_error, !quiet_mode);
```
zuckschwerdt: space after keywords `if (` and line break after condition please.

winterrace:
Coding style has been updated :)

zuckschwerdt:
Looks good. Thanks!

PR #831, opened 2018-10-16
Add support for TFA 30.3211.02

ionum:


src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
+#include "rtl_433.h"
+#include "util.h"
+#include "data.h"
+
```
zuckschwerdt: place includes below the header comment

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
...
+/*
+ TFA 30.3211.02 
+
```
zuckschwerdt: add your name, copyright, and gpl notice

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
+#include "rtl_433.h"
+#include "util.h"
+#include "data.h"
+
+/*
+ TFA 30.3211.02 
+ 
+ 1970us pulse with variable gap (third pulse 3920 us)
+
```
zuckschwerdt: we need to investigate if that pulse is preamble/sync or some bitstuffing

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
...
+     C = Channel
+     T = First digit temperatur
+     t = Second digit temperatur
+     u = Third digit temperatur
+     N = Negative temperatur
+     
+
+ */
+
+static int tfa_303211_callback (bitbuffer_t *bitbuffer) {
```
zuckschwerdt: function braces on new line

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
...
+    data_t *data;
+    uint8_t *b      = bitbuffer->bb[0];
+    const uint16_t length = bitbuffer->bits_per_row[0];
+
+    /* length check */
+	if (74 != length) {
```
zuckschwerdt: drop length and put the bits_per_row in the check

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
...
+		if(debug_output) fprintf(stderr,"tfa_303211 wrong size (%i bits)\n",length);
+		return 0;
+	}
+
+    /* shift preamble */
+    int i;
```
zuckschwerdt: better use bitbuffer_extract_bytes

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
...
+    for (i = 0; i < 9; i++) {
+      uint8_t b1 = b[i] << 2;
+      uint8_t b2 = (b[i+1] & 0xC0) >> 6;
+      b[i] = b1 | b2;
+    }
+
+    const float temp        = ((b[2] & 0x0F) * 10) + ((b[3] & 0xF0) >> 4) + ((b[3] & 0x0F) *0.1F);
```
zuckschwerdt: use type const instead of const type

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,101 @@
...
+    const int battery_low   = 0;
+    const int channel       = (b[1] & 0x30) >> 4;
+
+    float tempC = (minus == 1 ? temp * -1 : temp);
+    {
+        char time_str[LOCAL_TIME_BUFLEN];
+        local_time_str(0, time_str);
+
+        data = data_make("time",          "",            DATA_STRING, time_str,
+            "model",         "",            DATA_STRING, "TFA 30.3211.02",
```
zuckschwerdt: place the first args on a new line (nicer alignment), use two indents (8 spaces)

src/devices/tfa_30.3211.c
ionum: done

src/devices/tfa_30.3211.c
ionum: sample g008.... contains long third pulse (above 79 %)

I tested my code from 30% to 98% and -9.7° to 21.7°

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,103 @@
...
+    int const battery_low   = 0;
+    int const channel       = (b[1] & 0x30) >> 4;
+
+    float tempC = (minus == 1 ? temp * -1 : temp);
+    {
+        char time_str[LOCAL_TIME_BUFLEN];
+        local_time_str(0, time_str);
+
+        data = data_make(
+        "time",          "",            DATA_STRING, time_str,
```
zuckschwerdt: sorry, i meant two indents for continuations from the previous line, i.e. 16 spaces here.

src/devices/tfa_30.3211.c
```diff
@@ -0,0 +1,103 @@
...
+    float const temp        = ((b[2] & 0x0F) * 10) + ((b[3] & 0xF0) >> 4) + ((b[3] & 0x0F) *0.1F);
+    int const minus         = (b[1] & 0x08) >> 3;
+    int const hum           = (b[0] & 0x70) >> 4;
+    int const humidity      = (hum < 2 ? hum + 8 : hum) * 10 + (b[0] & 0x0F);
+    int const sensor_id     = ((b[1] & 0x07) << 4) | ((b[2] & 0xF0) >> 4);
+    int const battery_low   = 0;
+    int const channel       = (b[1] & 0x30) >> 4;
+
+    float tempC = (minus == 1 ? temp * -1 : temp);
+    {
```
zuckschwerdt: did some `if ()` got lost here or why the `{ }` block?

ionum: No if lost....

zuckschwerdt: you can remove the block braces `{` and `}` then ;)

zuckschwerdt:
Oh! Apparently ./maintainer_update.py does not sort alphabetically on your platform. I'll need to fix that!

zuckschwerdt:
Can you Rebase this PR onto current master?
(https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa should help)

ionum:
@zuckschwerdt Rebase done...

zuckschwerdt:
Thanks! You need to use `git push -f` to write the changes to GitHub now.

ionum:
git push -f already reports "Everything up-to-date". Github is having problems actually....

zuckschwerdt:
I guess the `>79%` business is to do with the first bit being lost. Your fix `(hum < 2 ? hum + 8 : hum) * 10 + (b[0] & 0x0F)` tries to recover that bit by assuming that Humidity below 20% is actually above 80%.

The problem seems to be that the gap preceeding a pulse encodes the data, which is not easily extracted for the first pulse after the preamble (3900 µs for 0, 5900 µs for 1 there where every other bit is 1900 µs vs 3900 µs).
I'll see if I can fix that.

If you pop [the data from the 5 samples into BitBench](http://triq.net/bitbench/?c=d7a53215a85acdea99&c=d7a53214a85acdebaa&c=d7a53214a85acdebaa&c=d7a53214a85acdebaa&c=819001427e6ffebd68&f=?:b+HUM:7h+?:2b+CH2d+TEMP_BCD:8b8h.4h+|+8h+8h+8h+8h+8h)
You can play around with the decoding and get a nice visualization.

ionum:
You guessed well.

Bitbench is a really nice tool..

zuckschwerdt:
Can you post a number of extracted codes (like the BitBench example), say 10-15 codes? I'd like to see if I can figure out a checksum or CRC.

zuckschwerdt:
Hah! Look at [this](http://triq.net/bitbench?c=57a53215a85acdea99&c=57a53214a85acdebaa&c=819001427e6ffebd68&f=HUM%3A08h%20%3F%3A2b%20CH2d%20TEMP_BCD%3A8b8h.4h%20%7C%20~8h%20~8h%20~8h%20~8h%208h) do you notice something? I'd say that takes care of the first bit issue :)

ionum:
Great!
Did you change flex params (i used "tfa_test:OOK_PPM_RAW:2900:6000:36500")?

I have a ton of samples as grouped binary values (with temp and hum) in csv-format like: 

2018-10-14 19:14:04;01 1101 0010 1001 0001 0001 0010 0010 1000 1010 1101 0110 1110 1110 1101 1101 0111 0001;22.8;52
2018-10-14 19:15:03;00 1101 0010 1001 0001 0001 0010 0010 1000 1010 1101 0110 1110 1110 1101 1101 0111 0001;22.8;52

zuckschwerdt:
I just used the plain codes from rtl_433_tests/tests/tfa_30_3211_02
Can you post 10-20 (unique) of those lines, or mail me the csv?

zuckschwerdt:
Looking at some codes in hex (first 2 bits padded to nibble)
    088900179776ffe865
    188900179776ffe865
we can guess that the initial 2 bits are not part of a check.
Your example CSV is missing the last nibble though ;) Can you generate a new one, preferably in hex (should be 9 bytes / 72 bits or 74 bits as 10 bytes).

ionum:
Got the samples?

zuckschwerdt:
Yes, thanks. But a first stab at the CRC wasn't successful :/ Needs more thought.

zuckschwerdt:
Still working on the checksum. If you look at the BCD digits there is clearly an additive relation to the sum. Not sure exactly how though.

zuckschwerdt:
Nearly got your Checksum. Some kind of add (with carry) and perhaps shift/feedback. This is just one single bit off:
```
        x += (b[3] & 0x5);
        x += (b[3] & 0xf) << 4;
        x += (b[3] & 0x50) >> 4;
        x += (b[3] & 0xf0);

        x += (b[2] & 0x5);
        x += (b[2] & 0xf) << 4;
        x += (b[2] & 0x50) >> 4;
        x += (b[2] & 0xf0);

        x += (b[1] & 0x5);
        x += (b[1] & 0xf) << 4;
        x += (b[1] & 0x50) >> 4;
        x += (b[1] & 0xf0);

        x += (b[0] & 0x5);
        x += (b[0] & 0xf) << 4;
        x += (b[0] & 0x50) >> 4;
        x += (b[0] & 0xf0);
```
`x+chk` (mod 256, i.e. uint8_t) then is 0xff (or 0x0f, add 1 and then we are just a single bit off…)
That's basically adding each nibble to the upper sum and half the nibble bits to the lower sum. Which could make sense if expressed another way that I'm currently not seeing.

merbanan:
Maybe it is a rotation. Look at the BSD checksum. The sensors have very little cpu power available.

zuckschwerdt:
Have a play ;) [tfa-30.3211.02.c](https://github.com/merbanan/rtl_433/files/2567598/tfa-30.3211.02.txt)
The bits seem to stay, I can see no rotation/shift register happening.

zuckschwerdt:
Merged with bit-wise parity checks. PR a change when Checksum or unknown bits are known. Thanks!

zuckschwerdt:
The TFA 30.3211.02 here is a rebranded Ambient Weather TX-8300. See #846.
I unified both PRs and this is now reported as "Ambientweather TX-8300" model.

zuckschwerdt:
Sorry for eventual confusion, I've changed the model string to "AmbientWeather-TX8300" to reflect the new direction for model names.

merbanan:
for (i=0 ; i<4; i++) {
            x += (b[i]&0xF) + ((b[i]&0xF0)>>4);
            y += (b[i]&0x5) + ((b[i]&0x50)>>4);
}
        t = (~(x&0xf))&0xF;
        t2 = (~(y&0xF))&0xF;
        w = t<<4 | t2;

merbanan:
[tfa-working.txt](https://github.com/merbanan/rtl_433/files/3010198/tfa-working.txt)

zuckschwerdt:
btw. this works well for speedy popcount and parity
```
#ifdef __GNUC__
static int popcount(unsigned x)
{
    return __builtin_popcount(x);
}
static int parity(unsigned x)
{
    return __builtin_parity(x);
}
#else
#warning No builtin popcount
static int popcount(unsigned x)
{
    x = x - ((x >> 1) & 0x55555555);
    x = (x & 0x33333333) + ((x >> 2) & 0x33333333);
    return (((x + (x >> 4)) & 0x0F0F0F0F) * 0x01010101) >> 24;
}
static int parity(unsigned x)
{
    x = (x & 0x55555555) + ((x >> 1) & 0x55555555);
    x = (x & 0x33333333) + ((x >> 2) & 0x33333333);
    x = (x & 0x0F0F0F0F) + ((x >> 4) & 0x0F0F0F0F);
    x = (x & 0x00FF00FF) + ((x >> 8) & 0x00FF00FF);
    x = (x & 0x0000FFFF) + ((x >> 16) & 0x0000FFFF);
    return x & 1;
}
#endif
```

merbanan:
Yeah, I know there are faster stuff. I think we need to create some helper tools to aid in analyzing cases like this. And maybe some kind of standardized output. At least output that we can feed directly into reveng and tools alike.

zuckschwerdt:
I made (really hacked) some tools. E.g. a `revdgst` complimentary to `reveng`, also a `bitbrk` to breakout bit changes and relate to checksum changes. Need to find some time to clean that up though ;)

I was thinking of maybe some web UI like BitBench where you drop code rows (in rtl_433 native format of {len}hex) and then fiddle with bits and setting and task reversing tools from there. (I'll just embed them in a webserver -- probably Python to make it hackable and cut down on development time). Definitly on my list (that list isn't ever getting shorter…)

merbanan:
I think a complete tool is almost the size of rtl_433 :) The are so many metrics one could  generate automatically. And it would be nice to have a tool that can find datagrams that differ with just one bit. Or duplicate checksums etc.

zuckschwerdt:
Your checksum code went in with 9ee7579.

merbanan:
Goodie. Anyway this checksum conforms to the low complexity these kind of devices usually implements. The nibblesum only give 4 bits so to get 4 more bits they mask a few bits in the nibble and redo the nibblesum.

zuckschwerdt:
In some sense related to Flechter-8, two nibble wide sums, just the checksum nibbles are not related. I suspect the Jansite TPMS does something like that, the bits seem to map nibble wide into the checksum. But the mapping isn't clear and I can't even see if it's adding with or without carry.

merbanan:
Yeah, I might take a stab at that one also. But I think I used my external expert help card for a while. I think we need to write some checksum reverse engineering documentation. There are quite a few different ones implemented. What I am missing is how you should analyze and think about this.

zuckschwerdt:
I tried to normalize and refactor all but the very exotic ones to [util.c](https://github.com/merbanan/rtl_433/blob/master/src/util.c) / [util.h](https://github.com/merbanan/rtl_433/blob/master/include/util.h) (note that crc8le is a misnomer it should be crc8lsb or crc8r[ef[lect[ed]]] ;) ). Even the simple add, xor and parity checks in decoders should reference there.

PR #832, opened 2018-10-16
Added option to skip the tests to be built.

NicolasJourden:
Hello,

I just propose this option since tests are not required all the time like if you cross-compile it.

The option aims to be able to disable it at will.

cheers,
Nicolas


zuckschwerdt:
Good point, tests need a rework someday.

PR #833, opened 2018-10-17
Replace CRC implementation

emontnemery:
The CRC algorithm is in fact standard CRC-8 with parameters:
width=8  poly=0x31  init=0x6c  refin=false  refout=false  xorout=0x00

Update rubicson.c to reflect this.

This was simple to find by using this excellent tool: http://reveng.sourceforge.net/

### DIFF
--- a/src/devices/rubicson.c
+++ b/src/devices/rubicson.c
@@ -13,27 +13,13 @@
  * unk1 is always 0 probably unused
  * temp is 12 bit signed scaled by 10
  * F is always 0xf
- * crc1 and crc2 forms a 8-bit crc
+ * crc1 and crc2 forms a 8-bit crc, polynomial 0x31, initial value 0x6c, final value 0x0
  *
  * The sensor can be bought at Kjell&Co
  */
 
 
-/* Bitwise implementation of CRC-8 with parameters:
- * width=8  poly=0x31  init=0x6c  refin=false  refout=false  xorout=0x00
- */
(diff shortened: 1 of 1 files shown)


merbanan:
I never got my head around how to do bit wise crc. I assume the patch is correct.

merbanan:
bb[1] is the first repeat of the message, because of demodulator quirks one bit might be dropped in the initial message (bb[0]). That was the reason initially, things might have changed. Dumping all repeats might show this.

Anyway in theory all copies (initial and the repeats) of the message should be the same (signal collisions make that ). We should use the repeats to validate messages better. But if a protocol has a crc we only need to verify that once to be pretty sure that the message is valid.

For protocols without crc we should/could use the repeats to validate the transmission better. We could also repair bit errors with the help of repeats. But no one has derived the classification process for this. The 433MHz band has lots of collisions so another sensor reading will be transmitted later anyway. It is ok to miss a few.

But feel free to work on it if you want.

emontnemery:
@merbanan OK, but then bb[1] should actually be used both for CRC check and data extraction?
For 8-bit CRC there's of course a 1/256 chance of matching CRC when receiving broken data, so it would make sense to check the buffer for 1-2 repetitions as well?

merbanan:
Yes and yes. I think there is a function that can validate several rows.

zuckschwerdt:
It looks like Rubicson and Solight are the same type of transmitter (or even sensor). I need to take a closer look but we should combine the files and unifiy where possible (as you already did with the CRC).

PR #834, opened 2018-10-17
WIP - Add CRC check to solight_te44

emontnemery:
It turns out solight_te44 is using same CRC8 algorithm as rubicson (#833) according to the tests.

However, I would like some guidance on how to refactor solight_te44.c.

### DIFF
--- a/src/devices/solight_te44.c
+++ b/src/devices/solight_te44.c
@@ -38,6 +38,8 @@
 #include "rtl_433.h"
 #include "util.h"
 
+extern int rubicson_crc_check(bitrow_t *bb);
+
 static int solight_te44_callback(bitbuffer_t *bitbuffer) {
 
     data_t *data;
@@ -70,6 +72,8 @@ static int solight_te44_callback(bitbuffer_t *bitbuffer) {
             return 0;
         }
     }
+
+    if (!rubicson_crc_check(bb)) return 0;
 
     local_time_str(0, time_str);
 


emontnemery:
@merbanan this was not quite ready for merge IMHO, i think there are too many checks now.
``` C
    // No need to check for 12 lines anymore?
    if (bitbuffer->num_rows != 12) {
        return 0;
    }

    // Unnecessary to check for 12 equal lines AND CRC check?
    for (int i = 0; i < 12; i++) {
        int bits = i < 11 ? 37 : 36; // last line does not contain single 0 separator

        // all lines should have correct length
        if (bitbuffer->bits_per_row[i] != bits) {
            return 0;
        }

        // all lines should have equal content
        // will work also for the last, shorter line, as the separating bit is always 0 anyway
        if (i > 0 && 0 != memcmp(bb[i], bb[i - 1], 5)) { // 5 bytes to compare 40 bits
            return 0;
        }
    }

    // This will do CRC check on line 1 only
    if (!rubicson_crc_check(bb)) return 0;
```

merbanan:
Yeah, crc can be used instead of all the code.

merbanan:
Oh, you need to add the mic=CRC to the output also.

emontnemery:
> you need to add the mic=CRC to the output also

Sorry, can you explain a bit more?

merbanan:
"mic",           "Integrity",   DATA_STRING, "CRC",

merbanan:
When we have a crc that validate the message, we add that to the output message to signal that the message can be trusted to some level.

PR #840, opened 2018-10-23
Fix possible reading/printing of uninitialized memory

winterrace:
The behavior of rtlsdr_get_device_usb_strings for large i's (i.e.: if the specified device index is larger than the available devices) is to return 0 without touching the buffers for vendor, product and serial. This leads fprintf to interpret uninitialized memory.
This condition can be triggered with "rtl_433 -d 7", for example.

Fixed this way...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,10 @@ build/
 
 *.orig
 *~
+*.tlog
+*.ipdb
+*.iobj
+*.idb
+*.lastbuildstate
+*.db
+*.opendb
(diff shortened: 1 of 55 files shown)


winterrace:
Just to quote the full context:

`
    char vendor[256], product[256], serial[256];
`
...
`
        rtlsdr_get_device_usb_strings(i, vendor, product, serial);
`
`
        if (verbose)
            fprintf(stderr, "trying device  %d:  %s, %s, SN: %s\n",
                    i, vendor, product, serial);
`

Issue #780, opened 2018-07-29
Ambient Weather TX-8365: need help identifying coding


roger-:
Hi,

I'm trying to decode this sensor: https://www.ambientweather.com/amrc8500.html
FCC: https://fccid.io/OQH-000000-13-01

Output is temperature and battery status only, and it supports one of three different 433 MHz channels.

rtl_433 doesn't seem to recognize it, so I captured this recording (with `-a -t` and temperature around 20 C): https://files.fm/u/xef593fx

Output of `-A`:
```
Test mode active. Reading samples from file: g001_433.92M_250k.cu8
Detected OOK package	@ @0.000000s
Analyzing pulses...
Total count:   75,  width: 90339		(361.4 ms)
Pulse width distribution:
 [ 0] count:   74,  width:   492 [490;495]	(1968 us)
 [ 1] count:    1,  width:   979 [979;979]	(3916 us)
Gap width distribution:
 [ 0] count:   39,  width:   483 [482;486]	(1932 us)
 [ 1] count:   35,  width:   972 [969;974]	(3888 us)
Pulse period distribution:
 [ 0] count:   39,  width:   976 [974;980]	(3904 us)
 [ 1] count:   34,  width:  1464 [1462;1467]	(5856 us)
 [ 2] count:    1,  width:  1951 [1951;1951]	(7804 us)
Level estimates [high, low]:  15994,     42
Frequency offsets [F1, F2]:    2499,      0	(+9.5 kHz, +0.0 kHz)
Guessing modulation: Manchester coding
Attempting demodulation... short_limit: 492, long_limit: 0, reset_limit: 975, sync_width: 0
pulse_demod_manchester_zerobit(): Analyzer Device 
bitbuffer:: Number of rows: 1 
[00] {76} 10 00 00 00 00 00 00 00 00 00 
```

Output is OOK, but the data doesn't look correct. I looked at the samples in Audacity and I'm not sure what the coding is. Any ideas?

zuckschwerdt:
Try http://triq.net/iqs to visualize the signal. Could be PPM or (rather unlikely) PCM. I.e.
```
rtl_433 -q -F json -R 0 -X 'tx8365:OOK_PPM_RAW:3000:5000:6000' -r g001_433.92M_250k.cu8

{74}63a45c983c5ba367e00
```
or
```
rtl_433 -q -F json -R 0 -X 'tx8365:OOK_PCM:1968:3888:6000' -r g001_433.92M_250k.cu8

[...]

PR #846, opened 2018-10-31
Add Ambient Weather TX-8300 Thermometer

roger-:
Seems to work!


zuckschwerdt:
Discussion is in #780

src/devices/ambient_weather_tx8300.c
```diff
@@ -0,0 +1,136 @@
+/* Ambient Weather TX-8300 Thermometer
+ * contributed by Roger
+*/
+
+#include <byteswap.h>
```
zuckschwerdt: Put the includes below the protocol info

src/devices/ambient_weather_tx8300.c
```diff
@@ -0,0 +1,136 @@
+/* Ambient Weather TX-8300 Thermometer
```
zuckschwerdt: The device should be named ambientweather_tx8300 -- We'll slowly move to just one word for manufacturer and one for device to simplify parsing in consumers.

src/devices/ambient_weather_tx8300.c
```diff
@@ -0,0 +1,136 @@
...
+    data_t *data; 
+    int count;
+    int crc;
+    
+	if(bitbuffer->num_rows != 1 || bitbuffer->bits_per_row[0] != BITS_PER_MSG) {
+	    if (debug_output > 1) {
+	        fprintf(stderr, "wrong number of bits");
+	     }
+	        
+		return 0; // Unrecognized data
```
zuckschwerdt: Indentation seems off here

src/devices/ambient_weather_tx8300.c
```diff
@@ -0,0 +1,136 @@
...
+Packet format (74 bits):
+   [2 bit preamble] [1 bit start bit] [31 bit payload] [1 bit start bit] [31 bit payload (inverted)] [8 bit CRC]   
+   HH1PPPPP PPPPPPPP PPPPPPPP PPPPPPPP PP1QQQQQ QQQQQQQQ QQQQQQQQ QQQQQQQQ QQCCCCCC CC   
+   
+Preamble format (2 bits):
+    [1 bit (0)] [1 bit rolling count]    
+    0R
+
+Payload format (31 bits):
+   [9 bit unknown (humidity?)] [2 bit channel number] [1 bit negative flag] [7 bit ID]  [12 bit BCD temperature (C, 1 decimal point)]
```
zuckschwerdt: Best to take the payload as 32 bit and mask out the fixed top bit, humidity needs only 7 bits, might be a full byte, but havn't ever seen 9 bits

zuckschwerdt:
Good work. Please change payload to an aligned uint8_t b[] and get all data from that.

roger-:
Thanks. Actually did it that way at first but thought it might be more readable to change it to  use your helper functions (but the endianess issues and extra copied bits defeated the benefits). 

I’ll take a stab at rewriting it though.

src/devices/ambient_weather_tx8300.c
roger-: I’m not sure anymore what that “humidity” field is anymore. Could be a combination of battery status, unused humidity or device code or something.

zuckschwerdt:
also `b[0] = (b[0] & 0x7f) | (b[5] & 0x80);` to fix the first bit

roger-:
PR updated.

roger-:
How does it look now? I’ve been running it for a week and all seems fine.

zuckschwerdt:
Oh, I just merged the TFA 30.3211.02 and it seems this is the same device? I'll check and unify both if it turns out to work.

zuckschwerdt:
The Ambient Weather TX-8300 is rebranded as TFA 30.3211.02. See #831.
I unified both PRs and this is now reported as "Ambientweather TX-8300" model.

roger-:
Thanks!

Can you re-add the rolling count?

zuckschwerdt:
Is it then really useful? What are you using that value for?

merbanan:
Sequence number, just let the consumer care about what to use it for.

zuckschwerdt:
It's just a single bit in the preamble, not secured by the checksum. Looks more like an artifact then proper data -- thus my hesitation ;)

roger-:
I’ve been using it to detect (single) dropped packets, but admittedly it isn’t the best way to do that.

zuckschwerdt:
Perhaps just measure the interval between packets, around one minute or such. Some devices vary the time with the selected channel setting. The different times then would be co-prime to ensure the least amount of overlap.

zuckschwerdt:
Sorry for eventual confusion, I've changed the model string to "AmbientWeather-TX8300" to reflect the new direction for model names.

klohner:
FYI: `fprintf(stderr, "H: %02x, F:%02x\n", b[0], b[1] & 0xc0);` in `ambientweather_tx8300.c`
is throwing `H: 0e, F:80` to stderr before each detection I see.

'''
H: 0e, F:80
{"time" : "2018-12-09 19:04:06", "model" : "AmbientWeather-TX8300", "id" : 34, "channel" : 1, "battery" : 2, "temperature_C" : 1.400, "mic" : "CHECKSUM"}
H: 0e, F:80
{"time" : "2018-12-10 13:11:07", "model" : "AmbientWeather-TX8300", "id" : 34, "channel" : 1, "battery" : 2, "temperature_C" : 2.800, "mic" : "CHECKSUM"}
H: 0e, F:80
{"time" : "2018-12-10 13:43:06", "model" : "AmbientWeather-TX8300", "id" : 34, "channel" : 1, "battery" : 2, "temperature_C" : 2.900, "mic" : "CHECKSUM"}
'''

zuckschwerdt:
Yes, that needs to be behind a device->verbose check. Adding it now ;)

PR #849, opened 2018-11-02
Digitech XC-0324 temperature sensor device handler

Geoff99:
This will almost certainly require some refinement (it's my first attempt at a device handler, and I'm a novice (to say the least) writer of C.  But it seems to work for me :-)

There are really three components to the pull request::

* the actual code to decipher the XC-0324 transmissions (essentially trivial in hindsight, but quite highly commented)
* an adaptation of `bitbuffer_print` that, with some additional trickery, lets me batch process the test data files and generate "debug to csv" output easily, as an aid to my reverse engineering process, and
* the xc0324.README.md file (currently in the src directory).  It started as a short explanation of the device handler aimed at a newbie like myself, and sort of grew out of control into a cookbook (again for a complete newbie) containing everything I wish I'd known at the start about how to set up and develop a new device handler.  Maybe I should just delete it from the pull request - or maybe (suitably edited and corrected) it could be useful for the wiki.

I'd be interested in any and all feedback you have to give about the elements of the pull request.  I fully expect that (if it has any use at all) improvements will have to be made before it is suitable for acceptance into the master, so feel free to criticise / correct or suggest alterations and improvements.

Thanks in advance for your consideration of the pull request, and your patience in looking through it:-) 

PS I think I've managed to rebase my fork to be up to date with merbanan/rtl_433/master, but if not let me know and I'll take care of that.


include/rtl_433.h
```diff
@@ -43,7 +43,7 @@
 
 #define MINIMAL_BUF_LENGTH      512
 #define MAXIMAL_BUF_LENGTH      (256 * 16384)
-#define MAX_PROTOCOLS           110
+#define MAX_PROTOCOLS           111 // ie plus for my prototype xco324 protocol decoder
 #define SIGNAL_GRABBER_BUFFER   (12 * DEFAULT_BUF_LENGTH)
```
zuckschwerdt: don't add a comment to MAX_PROTOCOLS :)

include/rtl_433_devices.h
```diff
@@ -113,7 +113,9 @@
 		DECL(m_bus_mode_f) \
 		DECL(wssensor) \
 		DECL(wt1024) \
-		DECL(tpms_pmv107j)
+		DECL(tpms_pmv107j) \
+		DECL(xc0324)
```
zuckschwerdt: the file/device should probably be named digitech_xc0324. One more info to find it by.

zuckschwerdt:
That's quite an impressive effort! "by a newbie, for a newbie" is exactly the thing that will lower the barrier and help many interested people.
I'll edit the "Howto" and code to give you some suggestions, but that'll take some time ;)

To merge the actual device the code needs to be streamlined to match existing style and without too much debugging or comments, i.e. state only facts and don't explain. (We have to keep 100+ device files in check and need to grasp facts quickly…)

So perhaps this "full package" should be an example how to engineer device code, accompanying a blog post on your Howto.
The code for merge should look like, e.g. [Bresser_3CH](https://github.com/merbanan/rtl_433/blob/master/src/devices/bresser_3ch.c).

Geoff99:
Thanks for responding so promptly! and encouragingly :-)

I'll move the "by newbie, for newbie" md file to a separate branch and submit a separate pull request on it.  It should probably be in the wiki I think, so I'll clone that repository and submit it there, if that's OK.  

That separation should make it easier for me to clean up the actual device handler to be more akin to the nice concise example you pointed me at.  It might take me a few goes to get it exactly right, but I'll do my best to get close so it's "not too hard" for you to nominate what else needs improving.

Once I've got the device handler to an acceptable state, I'll circle back and tidy up anything that's needed in the "by newbie, for newbie" file.  Changing its location will mean a few of the cross references, and some minor parts of the introductory text need a bit of mechanical alteration.  Plus of course any corrections and suggestions you have time to provide.  But I'll wait and see what exactly needs doing in the device handler pull request before doing too much on the .md "documentation".

Hope that's OK - if not just let me know what you'd like changed (I'll try and keep as much of the mechanical effort off you so you can focus on identifying the essential bits that need fixing).

include/rtl_433_devices.h
Geoff99: Done :-)

include/rtl_433.h
Geoff99: Fixed.

Geoff99:
Hi @zuckschwerdt 

I have moved the "by newbie for a newbie" tutorial to a separate pull request #857 , and have parked it for now, until this pull request is completed, at which stage I will revisit it to ensure it remains consistent with the final version of the digitech_xc0324 code.

I have done a (fair) bit of refactoring, to isolate the "debug 2 csv" code as much as possible from the actual device handler code.  In particular, I have moved all the code for the utility "debug to csv" functions into a separate include file, `digitech_xc0324.debug2csv.c`.  They are nothing special, really 

* just an adaptation of `bitbuffer_print` to produce prettily formatted bytes, each byte shown in hex and bit pattern format with a trailing comma (ie csv format), plus 
* wrapper routines for various trace messages, written using `fprintf(stderr, …)` again with trailing commas.  I have given the functions names which reflect their purpose, so that when they are used in the main code, it should be pretty clear what they are doing without needing to refer to the definitions

The main device handler code is in `digitech_xc0324.c`.  It is not as concise as the example you provided, because I have retained calls to the debug2csv code.  But is isn't (I don't think) too verbose now.  I would like to keep the debug2csv code in there, as that is fairly central to the substance of "newbie" tutorial.

Anyway, let me know what you think - and please feel free to suggest any changes or improvements you find necessary to fit with the style and substance of the main rtl_433 code.

Geoff99:
PS A question about `data_make` and `data_free`. I've tried reading the codebase but it is too advanced for my limited knowledge of C.  

Should there be a matching `data_free` after the result of each `data_make` has been sent to `data_acquired_handler`,  to avoid progressive consumption of memory?  

I haven't had a problem yet without a `data_free`, but I've only run the handler on captured test files, and in realtime test mode for 24 hours or so.  

I did experiment with putting a `data-free` into the xc0324 handler code, but obviously did something wrong - it segfaulted on me almost straight away (2nd message in 1st transmission received :-( )

zuckschwerdt:
The `data_free()` is in the [`data_acquired_handler()`](https://github.com/merbanan/rtl_433/blob/master/src/rtl_433.c#L470) -- you relinquish ownership of the memory when you pass it to the handler. Things like that should really be noted the comments.

The device code already looks good. I see why the debugging code is useful, especially the possibility to add labels is neat. But I'd rather see that in an accompanying tool. rtl_433 would pass raw debug data to stdout or network and you then could attach all kinds of analyzers and UIs. We can add that output as simple helper code.

```
if (debug_output > 1)
        bitbuffer_trace(bitbuffer, "Temp was %4.1f", temperature);
// -or-
        buffer_trace(b, "Temp was %4.1f", temperature); // on: uint8_t *b
```

would output the bitbuffer with a comment. It would look like the flex output (`-X`).

You then collect, tag, and process all the bitbuffers (I'd use a python script), update your decoder code, and test your decoder by inputing the codes again (use `-y <code>`).

Geoff99:
Hi @zuckschwerdt 

Thanks for the information about `data_free` and `data_acquired_handler`.  Makes sense to me (I was being protected from myself, even before I realised :-) )

Apologies for the length of this post, but I'm struggling a bit (on several levels) to see where to go next, so I thought I'd let you know what I've tried and seek your advice.  

Generally I prefer code to spreadsheets any day (much more repeatable and decipherable).  However, for the particular interative deciphering approach I used, a spreadsheet was much simpler and quicker than writing python code (all I needed was to stack columns (aka bytes) above each other and sort the lines, then focus on some particular columns (hide columns works well)).  My main objective was to get to a working rtl_433 ditech_xc0324 device handler - the debug2csv approach, and the newbie tutorial just happened along the way.

Anyway :

* I have had a try at writing a `bitbuffer_trace` function.  Lots of fun :-) (sort of).  I learnt a new word (*variadic* arguments), and some new C functionality (`va_list` and friends, and the joys of having 2 versions of a C function, the ellipsis `...`  wrapper version and the `va_list` functional version) - much simpler in python!.  I (eventually) got `bitbuffer_trace` working as a wrapper around `bitbuffer_print`.  But (unless I dig deeper) it doesn't really do what I'm hoping for  - trivial reasons, but

  * it starts newlines for each row of a multi row `bitbuffer`, which spoils my tactics for getting information in a csv file all on a single line, and 
  *it only gives me hex, not hex and bit pattern byte by byte.

* Undaunted, I started to rework my existing `...2csv` code, separating the parts that are specific to the xc0324 device (eg package and message length), from the parts that are really just pretty printing rows and parts of rows from the `bitbuffer`

  * pretty printing lets me print hex and bit patterns side by side - bits are very useful in spotting sync errors in my test files, and also when trying to decipher how a field value has been encoded; while hex is great for seeing the bigger picture, eg which columns are (more or less) constant, which vary a lot
  *  working at the row level and having control over when line endings are sent lets me manage the csv files the way I like.  

* I have also created `_trace` versions of the pretty print routines (I **do** like that as a naming approach - it is exactly what I am doing).  It works, but is getting further and further away from the simple "newbie" approach I began with :-(

* Along the way (while browsing rtl_433 source code for inspiration and education) I discovered how I can tweak the declaration and definition of `in_filename` to make it `extern`.   That means I can acquire it in my `get_csv_label` function instead of using the `alarm` and `fgets`  (I was quite pleased when I got that `fgets` approach working, but I've been worried that by setting my own signal handler I might well be over-riding something crucial deeper in the `rtl_433` code base.

I've sort of reached a dead end though.  Any advice about the simplest way to complete the pull request would be much appreciated.

Thanks!

zuckschwerdt:
Oh, glad you had fun uncovering the depths of C's variadics. But I didn't mean for you to write that code. If we can lay out a flexible and working approach I'll provide the plumbing.
My idea was:
- if you only need a few messages dumped (maybe just one file with `-r`), mostly early in development: sprinkle a few bitbuffer_print (or a new buffer_print for extracted bytes) in temporary.
- if you need a bigger run of messages (perhaps a batch of captures): use the -X flex decoder to get neat rows of data quick.
- if you want to run the code longer but also alert about some edge cases: run in debug (trace) mode and use bitbuffer_trace/buffer_trace to write the lines to the configured output handlers. That way you can log to database or something with a structured format (csv, json, ...) for days or weeks.

If this fits your needs and seems be a good approach to document in your tutorial, then I'll add the helper functions in this way and your new device can go in clean and without "baggage".

Geoff99:
Hi again @zuckschwerdt 

Many thanks indeed for such a quick and encouraging response.  I confess I was losing energy but your suggestions revived my enthusiasm to get a nice clean elegant solution, without too much baggage.

Unfortunately today (and probably over the next couple of days as well :-(  ) I've been a bit short of time to work on this , so this is mainly a "thanks" and a "I'll get back to you properly in a few days" comment.  Plus a tiny bit of a progress report  (mostly for my own benefit - if you are busy - just stop reading here).

I did get chance to experiment with the -X flex decoder.  It's pretty neat but it doesn't quite match the way I'm thinking at the moment.  Could of course be a problem with how **I'm** viewing the challenge.  

It has highlighted to me that the underlying issue bothering me was a (process) design issue - or in your words 

>  laying out a flexible and working approach

Once that's done, plumbing can be attended to.

Anyway, looking at the flex decoder, and thinking about this has :

* reminded me of a few stages in the newbie tutorial that I had glossed over  - won't take much to correct, really just need to explain

  *  how (`rtl_433` will help you) to get "aligned" data, and later on
  *  how to use / what's going on with  `data_make` and `data_acquired_handler` and  `.fields = output_fields`

* given me (at least) 2 ideas about possible working approaches that could be both easy to explain and understand, and also fit within the existing `rtl_433` code base without too much extra effort

    * one of which is "if I really want csv output, why not look at using `data_make` etc plus `-F csv`!"  May not be totally straightforward because the output_fields might vary as I want to add and subtract trace columns, but fun for me to look into anyway.

Geoff99:
Hi @zuckschwerdt 

I think I've made some progress, on several fronts.

#### Thoughts about the broad approach
I like the `-X` option, especially coupled with `-F csv`, but I couldn't
find all the features I wanted to make my "newbie style process" work.

* I am very keen on being able to see bit patterns (logic, I have limited
  "CPU and working memory" in my brain.  When I'm focussing on reverse
  engineering, it's too hard to mentally convert hex to bits - and get
  them right anyway :-))
* I couldn't find out how / whether the flex decoder can automagically
  look after repeat messages within a row (I could only find repeat rows)
  (logic - The device I was working on had within row repeats; comparing
  repeats was useful, especially when looking for the `mic`)
* I favour batch approach over interactive (so bitbench is useful, but
  not as helpful to me as a trace message approach)
* I did like the `name:` feature - **very** useful!
* PS I noticed that the `<modulation>` options are slightly different
  from those used in `r_device`

#### Thoughts about the tutorial
I have clarified in my own mind how the final tutorial might flow:

iteration 0

  * explore `rtl_433`, introduce `-h -R -G -D -DD`

iteration 1

  * discover frequency, modulation type
  * introduce `-X`, `-F csv|json`, package structure - rows and columns

iteration 2

  * version 1 of a callback, introduce `bitbuffer_print`

iteration 3

  * find sync byte|preamble, introduce `bitbuffer_search`

iteration 4

  * start reverse engineering, introduce trace message functionality,
    beginning with `fprint` and then any additional plumbing added to
    `rtl_433`

iteration 5

  * introduce `data_make`, `data_acquired_ handler`, `output_fields`

#### New plumbing

To help me sort my thoughts out, and because learning new stuff is fun,
I have had a go at writing my own "plumbing".  It has 3 main parts:

* a revamped (get_)`bitbuffer_label` function;
* some "pretty_print" routines that (optionally) show bit string as
  well as hex string versions of `bytes`, `buffers`, `bitbuffers`. All
  are (loosely) modelled on the `bitbuffer_print` concept.
* `_trace` extensions of the "pretty_print" routines that support the
  addition of arbitrary trace messages

I'll copy pseudo header functions for these below.

If you think the approach has merit, (and if it's not too presumptuous of
me to ask) I'll bundle them up into a separate (?) pull request for you
to have a closer look at.

Looking forward to hearing back from you :-)

#### Pseudo header functions for plumbing

##### Labels for the bitbuffer traces
```
char bitbuffer_trace_label[LOCAL_TIME_BUFLEN + 48] = {0};

/// Get a label (in_filename or timestamp) for trace messages
char * bitbuffer_label();
```

##### Pretty print routines, with trace versions
```
/// Pretty print a buffer (including trace versions)
/// @param numbits : (max) number of bits printed, bits above numbits are masked
/// @param showbits : if true, print bit strings as well as hex strings
void buffer_pprint(FILE * stream, uint8_t * buffer, 
                   const uint16_t numbits, const bool showbits);
void buffer_pp_trace(FILE * stream, uint8_t * buffer, 
                     const uint16_t numbits, const bool showbits, 
                     char * format, ...);
void vbuffer_pp_trace(FILE * stream, uint8_t * buffer, 
                      const uint16_t numbits, const bool showbits, 
                      char * format, va_list args);
```

```
/// Pretty print part of a bitbuffer row (including trace versions)
/// @param row : the row, part of which will be printed
/// @param bitpos : printing starts here
/// @param numbits : number of bits printed (provided available on row)
/// @param showbits : if true, print bit strings as well as hex strings
void bitbuffer_pprint_partrow(FILE * stream, bitbuffer_t *bits, 
                              const uint16_t row, const uint16_t bitpos, 
                              const uint16_t numbits, const bool showbits);
void bitbuffer_pp_partrow_trace(FILE * stream, bitbuffer_t *bits, 
                                const uint16_t row, const uint16_t bitpos, 
                                const uint16_t numbits, const bool showbits, 
                                char * format, ...);
void vbitbuffer_pp_partrow_trace(FILE * stream, bitbuffer_t *bits, 
                                 const uint16_t row, const uint16_t bitpos, 
                                 const uint16_t numbits, const bool showbits,
                                 char * format, va_list args);
```

```
/// Pretty print all the rows in the bitbuffer (including trace versions).
/// @param showbits : if true, print bit strings as well as hex strings
void bitbuffer_pprint(FILE * stream, bitbuffer_t *bits, const bool showbits);
void bitbuffer_pp_trace(FILE * stream, bitbuffer_t *bits, const bool showbits, 
                 char * format, ...);
void vbitbuffer_pp_trace(FILE * stream, bitbuffer_t *bits, const bool showbits, 
                 char * format, va_list args);
```

##### Underlying machinery
```
/// Format byte as bit string
/// @param numbits : bits above numbits are masked
/// @param result returns bitstring, requires 10 characters (inc the \0)
void byte2bitstr(const uint8_t byte, const uint16_t numbits, char * result);

/// Format byte as hexstr
/// @param numbits : bits above numbits are masked
/// @param result returns hexstring, requires 3 characters (inc the \0)
void byte2hexstr(const uint8_t byte, const uint16_t numbits, char * result);

/// Format byte as string
/// @param numbits : bits above numbits are masked
/// @param showbits : if true, include bit strings as well as hex strings
/// @param result returns hexstring, requires up to 13 characters (inc the \0)
void byte2str(const uint8_t byte, const uint16_t numbits, const bool showbits,
              char * result);
```

PS Just realised I have left `FILE * stream` arguments throughout all the
above.  If required, those could be eliminated, and replaced by
`stderr` hard-coded in the implementations.

PPS As far as I could see how, the implementations of the assorted
functions use each other (so as to eliminate duplicated functionality in the code).

Geoff99:
Hi again @zuckschwerdt 

I realised overnight that some more tidying up on my attempts at writing some plumbing functions is possible (and desirable).  

On reflection 

* the `bitbuffer_.._partrow` group is not really essential, 
* likewise the `..._pretty_print` versions are superfluous (they are really only wrappers around `..._trace` versions with `format` set to `NULL`), 

so with a bit of renaming the new "API" level functions would reduce to `bitbuffer_trace`, `buffer_trace` and `bitbuffer_label` - everything else is really just an implementation detail.  

I'll try that out on my development branch today, then leave this for a while (maybe put a bit more work into polishing the tutorial :-) ).

zuckschwerdt:
Looks like the idea is getting solid. First comments:
One indent is 4 spaces, continued lines always get two indents more than the starting line.
The pointer type `*` should stick to the variable or function name, the qualifier is after the type (i.e. you read the type definition right to left).
`if (in_filename && (strlen(bitbuffer_trace_label) > 0))` should just be `if (in_filename && *bitbuffer_trace_label)`
`strncpy()` doesn't terminate, add e.g. `dst[47] = '\0'`.
The 5 second "Assume it's a new package" generally won't work. That needs more thought.
Talk about spreadsheet not a specific product ;) (and test LibreOffice)
Also set all non-API functions to static.

Geoff99:
Hi @zuckschwerdt 

I think I've found all the formatting changes and fixed them (let me know if not).  And I'm made a mental note to **always** double check the tab and tab to spaces settings in my editor; doubly so when I copy code from websites!).  Also fixed changes to static, mentions of spreadsheets, etc.

I've remove the "Assume it's a new package".  The *price* is the need for an explicit call to (what I now call) `get_trace_label` at the start of my callback handler.  I did investigate (**very** briefly) the possibility of embedding this initialisation upstream, in the code that invokes the callback handler, and very quickly saw that was not a good idea.

Hope I'm getting close now :-)
 

As always, comments and guidance very welcome (I'm learning lots from your help, and the rtl_433 code).

zuckschwerdt:
Getting there! As I said you only need to sketch things and I'll make sure it gets in clean.
Is it acceptable for you if I take over and finalize your ideas in another PR? It will be faster but you won't get the git commit attribution (I'll include a copyright note though).
I have two separate things in mind:

Debugging helpers to be used while hammering out new device decoders:
- `void bitbuffer_debug(bitbuffer_t const *bits)` : always output binary
- `void buffer_print(char const *buf, unsigned len)` : like `bitbuffer_print`
- `void buffer_debug(char const *buf, unsigned len)` : always output binary

And the trace helpers to be used while testing finished code, which you laid out.

Geoff99:
That would be **fabulous** (and no worries whatsoever about git commit attribution – I’d be embarrassed if anyone ever read through the sequence of awful mistakes I made getting this far J).  Please keep trace labels in though, that is central to my approach.

 

FWIW, I have been tinkering (in the last hour or so, mostly for my education) with the concept of a `buffer_make_trace`, which instead of printing the bytes (hex or bit string), puts them into a `data_t` structure and returns them to the reverse engineer, so other fields can be `data_append`ed before invoking a `data_acquired_handler`.  Code is inelegant (still get lost in pointers and de-references) but runs Ok – quite nice for `–F json`, kludgy for `-F csv` because of the need to specify the fieldnames again L

 

BTW Thanks for the tip about  sticking the * to the variable or function name, and reading in from the right.  Makes it **much** easier for me to follow what’s going on.

 

Best

 

Geoff

Geoff99:
PS I have enjoyed trying to implement my ideas.  I really am very new to C, and reading through well written code, trying out small amendments, then discovering what I’ve broken and why has taught me a lot – heaps left to learn though, so still fun ahead for me.

zuckschwerdt:
I had that in mind too, while buffer_print and debug just print, the trace should output to the output handlers via data_t :)

merbanan:
@zuckschwerdt git commit --author="Geaoff@Geoff" doesn't work ?

zuckschwerdt:
@merbanan if using Git instead of the Github UI that would indeed work nicely.

zuckschwerdt:
@Geoff99 the new "tag" option and batch input in #878 will help running new code on a folder of sample files without the need for the `extern in_filename`.
Helper functions for debug/trace output are also on the way.

zuckschwerdt:
There is some preliminary code in https://github.com/zuckschwerdt/rtl_433/commits/feat-dutil -- but that needs to wait on two more patches for ground work. The "well known" keys for trace messages, "msg", "codes" will be register with csv beforehand liek the "tag" key.

Interesting about the new CSV idea, but how would you write a header if you only aquire the field names later and in random order?

Geoff99:
Hi @zuckschwerdt 

Looking forward to seeing those groundwork patches complete :-)

On my new CSV idea, how and when to write the header is the main design/conceptual challenge :-) 

(For me there are a bunch of C implementation challenges as well, probably not so for more experienced folk)

OK, here goes with my current thoughts ….

json output (for example) writes the field label (key / name) at the same time as the field value, so it doesn't have to remember anything.  When the json output handler is sent a `data_t **` message, it just works its way down the list, outputting whatever field names and values it finds.  If, for example, the device callback handler sends two distinct messages (message1 and then later message 2), no problem at all.

But the csv output handler for a particular device writes the header line at "initialisation" time, and then "filters" the information in any `data_t **` message it is sent, outputting only those fields it knows about (remembers) (from the `data_output_csv_t.fields` entry).  

Obtaining and retaining that **"memory" about the fields named in the headers is essential** in the current csv output handlers.  At the moment the knowledge/memory about the csv fields to be output :

* begins as a property of the device (specified at compile time)
* is copied to become a property of the csv output handler for that device (by `data_output_csv_create`), while rtl_433 is processing run time arguments, at which time `data_output_csv_init` is used to write the header line out, and
* is referred to (but not changed) whenever a `data_t **` message is processed.

The alternatives, under a scenario where the csv field names are generated dynamically from the `data_t **` message by (say my prototype `csv_r`) output handler are :

1) *cheat* - produce csv files with no headers (that in effect is what I did in my very first trace message attempts)

2) provide an API level `csv_header(data_t** …)` function which generates the NULL terminated `char **` list of field names and then writes them out as a header; ie let let the device callback message handler code "choose" when it wants / needs to write out a csv header - I'm close to this stage now (just need to tidy my code, and make sure I haven't messed up `calloc` creations and `free`s in my recursive walk of `data_t **` structures to generate the list of (flattened) field names), or

3) reverse the logic and write header lines based on the output handler remembering what fields the previous `data_t **` instance provided, ie

  * omit the list of fields entirely from the device definition (attraction - gives a cleaner separation between (input) devices  and output handlers),

  * initialise the `data_output_csv_t.fields` entry as NULL when the output handler for the device is created while the rtl_433 args are being processed, and then

  * when a `data_t **` message is received, generate the NULL terminated `char **` list of (flattened) field names, and compare it to the  `data_output_csv_t.fields` list.  If they are the same, proceed to output (all) the values from the `data_t **`; if not the same, first output a new header based on the current `data_t **`, and update the output handlers memory in the `data_output_csv_t.fields` entry before outputting the line of values.

I'm pretty doubtful that this degree of surgery / enhancement is worthwhile for the official rtl_433 project (there are far more important developments underway, as I can see from following the repository activity).  For me it's worthwhile and fun. because I'm learning lots about C - if I was just concerned about the end effect, I'd use json and post process into whatever csv format I wanted using a simple python script (something I'm much more familiar with).  

If you are still interested in the idea, let me know, and when (and if) I get my prototype working, and tidy enough to be not too embarrassing, I'll show you a copy - otherwise I'll just continue with it as a fun learning exercise for myself :-).

zuckschwerdt:
I guess there are many users relying on the current working of the CSV output. It isn't tailored to support a special app or purpose but to be as universal as possible. In short: There needs to be a header; One fixed header per output. I don't think it's important when the header is written as long  as it's the first line, whatever makes things easier.
Another idea: write a short program that takes json input and produces custom csv, very much tailored and optimized for one task.
(Or use the syslog input and even work decoupled and over network -- s.a. https://github.com/zuckschwerdt/rtl_433_relay).

zuckschwerdt:
You can preview the new utility functions in #883.
I guess `decoder_output_messagef()`  and `decoder_output_bitrowf()` will be of most use to you.

Geoff99:
Many thanks indeed and a **BIG :-)**  I'll explore the new utility functions, and adopt them into this pull request as soon as I finish a rebase onto the latest master.

The new CSV idea was mainly a learning C exercise for me (apologies for the pun, but I've learnt heaps about memory management). 

Agree **absolutely** about not altering the current CSV API contract.  Even in my playing around, I would have implemented it as a new supplementary "CSV style" output handler.  If you ever wanted to extend the CSV handler by providing an implicit default for CSV output when a device does not have a  `.fields         = output_fields, entry`, it might be possible to generate a default `output-fields` by walking the first `data_t` structure that is acquired - but even that would be a fair bit of effort for dubious gain.  

I'm closing off my  experiments with this for now - learning the new utility functions, and then updating the tutorial accordingly is much more useful.

Once again, thanks for your patience, and encouragement.

Geoff99:
PS **Purely** for info, I got as far as stage 2 in my `csv_r` experiment.

My test `data_t` structure is

```
    data_t *data = data_make(
      "key", "Pretty key", DATA_INT, 42,
      "others", "More data", DATA_DATA, data_make("foo",NULL, DATA_DOUBLE, 42.0, NULL),
      "zoom", NULL, DATA_ARRAY, data_array(2, DATA_STRING, (char*[]){"hello", "World"}),
      "double", "Double", DATA_DOUBLE, 10.0/3,
      NULL);
```

The `csv_fields_make` utility creates an output_fields style `char **` array
```
    char **data_fields =  csv_fields_make(data);
```
which, when suitably formatted using
```
    csv_fields_print(data_fields);
    csv_fields_free(data_fields); // don't forget to free memory when I've finished 
```
prints like this:
```
{
	"key",
	"others_foo",
	"zoom[000]",
	"zoom[001]",
	"double",
	NULL
}
```
And the `csv_headers` utility generates a CSV header line
```
    csv_headers(data);
```
which looks like this:
```
key,others_foo,zoom[000],zoom[001],double
```

Geoff99:
Hi @zuckschwerdt 

After having done my first ever `git fetch` / `merge` from a repository that wasn't called `origin`, I've had a good explore of the new utility functions.  Really neat.

Following below : 

* appreciation of some really neat stuff 👍 
* couple of small typos (I think)
* a small suggestion about function argument order, and
* a wishlist request

#### Really neat stuff

I especially like - well almost everything :-), but just to list things, the : 

* `-K` option, (and the way the tag field is *automatically* prepended to the output field list for the csv outputs :-)
* the ability to process multiple files from the rtl_433 command line
* the layout of the output from the` *_printf` and `*_debugf` functions (and the clever way even the `_printf` functions provide the binary output if there's room);
* the way all the `decoder_output_*` functions 

  * cleverly pack their output into an appropriately big enough array, 
  * **and** then condense it into a simpler string for the csv outputs (compared to my own recent attempts saves an **enormous** amount of effort (and risk of errors) messing around with lots of little bits of heap memory)  
  * **and** then again automatically does the prepend of the `codes` and `msg` field names for the user :-)

I'm almost spoilt for choice about how to do what I want to!

#### Couple of typos

Before I forget, I think I've spotted a couple of minor typos / corrections : 

In *about* line 253 of src/bitbuffer.c 
```
        print_bitrow(bits->bb[row], bits->bits_per_row[row], highest_indent, 0);
```
I think the `0` is meant to be `always_binary`, ie 
```
        print_bitrow(bits->bb[row], bits->bits_per_row[row], highest_indent, always_binary);
```
And in *about* line 217 of src/decoder_util.c 

```
    data = data_make(
            "time", "", DATA_STRING, time_str,
            "msg", "", DATA_STRING, msg,
            "code", "", DATA_STRING, row_code,
            NULL);
```
I think `"code"` is meant to be `"codes"` so it gets recognised by the automatic prepend that occurs, ie
```
    data = data_make(
            "time", "", DATA_STRING, time_str,
            "msg", "", DATA_STRING, msg,
            "codes", "", DATA_STRING, row_code,
            NULL);
```

I have one suggestion, and one item for my wishlist

#### The suggestion
For the `decoder_output_bitbuffer`, `decoder_output_bitbuffer_array`, and `decoder_output_bitrow` the API would be more consistent with the `*f` versions if the `char const *msg` was the last argument rather than the first, eg
```
void decoder_output_bitbuffer(char const *msg, bitbuffer_t const *bitbuffer);
```
would become
```
void decoder_output_bitbuffer(bitbuffer_t const *bitbuffer, char const *msg);
```
and hence be *more like* (and easier to remember, and to swap between)
```
void decoder_output_bitbufferf(bitbuffer_t const *bitbuffer, char const *restrict format, ...);
```

#### The wishlist request

I like the layout of the `bit*_debugf` functions a lot.  

It would be really nice if the `decoder_output_bit*` functions also produced the same layout, with the binary output following the hex version in the `codes` field.  I can see it increases the memory needed, and, probably more importantly, makes the `codes` field  in the csv (or other) output formats much more verbose, but it would really work nicely for newbies like me.


As always, thanks so much for taking the time and effort to help me so much - I'm very grateful.

zuckschwerdt:
Thanks for the thorough check and suggestions! Both fixes went in.
About the API, I thought about that also. I briefly had the idea of

    decoder_output_bitbufferf("format string", bitbuffer, args, ...)

since that looks more like printf -- but it's rather unexpected that bitbuffer is an extra arg dropped in the middle there.

The idea was to have the text first as it's printed first (and looks more like printf), but consistency is really more important.

I'm thinking about long term logging (with files and databases) for the decoder_output_ that's why I try to keep it compact. Perhaps I can offer you a converter script, something like `rtl_433 ... | show_bits.py` -- would that work? (could be fiddly with CSV) Or perhaps show the binary only if verbosity is set to debug?

Community projects can only grow if users invest time to add comments, ideas and stick around to see those thought through and implemented -- so praise goes to you.

sidenote: if you want to rebase and don't care about the history you can `git rebase -i` (interactive) first and *squash* the accumulated commits together. Use `git push -f` after rewriting history.

zuckschwerdt:
One more note: I don't recommend the _array output, and will likely remove "num_rows" and "rows" from the flex decoder, it's a nice idea but seems to complicated to process. Having only "codes" is also more consistent.

And a heads up: the decoder functions will gain a extra first argument shortly, to modularize  the decoders it will be:

    decoder_output_bitbufferf(decoder, bitbuffer, "format string", args, ...)

and

    decoder_output_data(decoder, data)

instead of the current

    data_acquired_handler(data)

Geoff99:
Relieved I haven't over used your helpfulness yet !  I rather suspect this might be the longest pull request conversation, for the smallest actual decoder :-)

I understand about the conciseness / log database objective, it's a good idea.

For me, only getting verbose binary decoder output based on the debug level (eg when `debug_output > 0`) would be perfect - I only really want the binary when I'm in develop / debug mode anyway!  

As an exercise for myself, I've melded the relevant parts from current code to create a prototype verbose  `decoder_output_bitrow` to see if it would work, and how it might look (code below, though there's probably a more elegant way of achieving the result), and it seems to work :-)

Agree re the _array output, I had to experiment to see exactly what it was doing - as csv the result is identical to without the array, and in say json, didn't really seem to add anything I might need.  (didn't look at key / value style output though)

Am interested to see how to use the extra first argument.  There've been several times in my experiments when I've wished I could *access* other upstream parts of the code from my current position. (though possibly that's the opposite of what you have in mind!)

I'll keep an eye open for the next developments on the repositories, and look forward to more fun and learning ahead :-)

PS I'll definitely have to look into git rebase and squash - I tend to do lots of trivial commits (always scared I'll damage something I've got working, so tend to commit each time I make even a tiny bit of progress).


##### Info only, my protoype verbose `decoder_output_bitrow`

```
void decoder_output_bitrow(char const *msg, bitrow_t const bitrow, unsigned bit_len)
{
    data_t *data;
    char *row_code;
    char row_bytes[BITBUF_COLS * 3 + 1];
    char row_bits[BITBUF_COLS * 9 + 1];
    char time_str[LOCAL_TIME_BUFLEN];
    unsigned i;

    local_time_str(0, time_str);

    row_bytes[0] = '\0';
    // print byte-wide
    for (unsigned col = 0; col < (bit_len + 7) / 8; ++col) {
        sprintf(&row_bytes[3 * col], "%02x ", bitrow[col]);
    }
    // remove last nibble if needed
    row_bytes[3 * (bit_len + 3) / 8] = '\0';

/**/
    // Print binary values also? only in debug mode
    row_bits[0] = '\0';
    if (debug_output > 0) {
        uint16_t next_col = 0;
        for (unsigned bit = 0; bit < bit_len; ++bit) {
            if (bitrow[bit / 8] & (0x80 >> (bit % 8))) {
                sprintf(&row_bits[next_col], "1");
            } else {
                sprintf(&row_bits[next_col], "0");
            }
            next_col++;
            if ((bit % 8) == 7) { // Add byte separators
                sprintf(&row_bits[next_col], " ");
                next_col++;
            }
        }
    }

/**/
    // a simpler representation for csv output
    row_code = malloc(8 + BITBUF_COLS * 3 + 4 + BITBUF_COLS * 9 + 1); // "{nnn}..\0"
    sprintf(row_code, "{%d}%s : %s", bit_len, row_bytes, row_bits);

    data = data_make(
            "time", "", DATA_STRING, time_str,
            "msg", "", DATA_STRING, msg,
            "codes", "", DATA_STRING, row_code,
            NULL);
    data_acquired_handler(data);

    free(row_code);
}
```

Geoff99:
Hi @zuckschwerdt 

I've been changing my local version of the XC0324 decoder to use the new utility functions (though I won't push the changes until you add the new first argument mentioned in the heads up earlier).

Using the utilities is **much** cleaner and more straightforward than any of my earlier attempts.  

I can get tantalisingly close to something that very easily supports the newbie approach I have drafted for the wiki, but I'm just not quite there :-( yet, because I can't get the nicely formatted binary outputs (as per `bitrow_debugf`) into the csv files.  That's sort of central to the way I (explain how to) decipher the temperature field, with it's least significant bit first encoding.  I've investigated several different variations using the new utilities, but so far, unsuccessfully.

I'm hoping I can persuade you to add one more `decoder_` style utility , say `decoder_output_verbose_bitrowf` that would include binary as well as hex in the `codes` entry, (or perhaps add a "verbose" flag to the existing `decoder_bitrow…` functions, or even just allow the debug level to switch on a verbose version).

Thanks in advance for your consideration.

zuckschwerdt:
I'm still pondering the idea, where/how to support that so it isn't too verbose and doesn't clutter the output. Do you know the [BitBench](http://triq.net/bitbench) tool I made? Would that fit in your development process? Maybe some easy way to transfer bitbuffer there could help?

Geoff99:
Mmmm, sorry to cause problems and more work.  It's so close now though, that I thought I had to ask.

The core of the development process (which I adapted from some of the references mentioned in the newbie tutorial) is to **batch** process a set of test files which have useful values (in this case a sequence of temperature values which differ by small - ideally one reporting increment - values).  Then put the complete batch into a spreadsheet, sort by the true value observed when the test files were collected, and then look carefully at the sequence of bitbuffer values (or bitrow values once the preamble has been identified) and search for which bytes / part of bytes / bits change in step with the changes in the true observed value.

Unless I misunderstand what bitbench does (quite possible), it doesn't easily fit this "batch process then sort" strategy.  

It's not **too** hard to implement "batch process then sort then look for linked patterns" from scratch in a one-off case (the way I did to begin with) - what's **really** nice about the utility functions you have created is that (**provided** you can get the binary version into the `codes` field) a newbie could follow this pattern with very little effort indeed, purely from within rtl_433, without even needing to play with bash scripts, grep etc.


I do understand the brevity objective (unfortunately not my strength though:-) ), which is why I suggested one additional specialised utility function - it really is the `*bitrow` binary I'm hoping for - the `bitbuffer` is easily (and in fact better) dealt with by an explicit loop over the rows in my planned tutorial.

zuckschwerdt:
This is all good stuff to get a stream lined developer experience, no worries.

I'd say "looking carefully" isn't generally easier with a string of bits in a spreadsheet.

Wouldn't a good process after acquiring some samples be:
- Use the Analyzer to figure out Modulation and Timing.
- Write a flex spec (or ideally copy from the analyzer output)
- Run the flex on a batch of files
- Transfer the codes with comments on the expected values to BitBench
- Fiddle with the BitBench formatter until it works

I usually collect interesting output from a batch run to a text file and annotate the lines (`[ my comment ]` works in BitBench).

All done without coding anything. The thought is to make coding a last and optional step. People are ingenious writing scripts and config files, coding and compiling seem to be a brick wall usually though.

(I'm even thinking about making the BitBench formatter codes compatible with the flex spec.)

Geoff99:
Steps 1-3 are spot on to figure out and test the (de)modulation, and my latest (unpiublished) version of the newbie tutorial advocates exactly that.  

It's at step 4, when I'm trying to decipher the encoding of the individual field values that I approach things a different way.

I can see I'll have to learn more about how Bitbench works. How does your step 4 work (Transfer the codes with comments on the expected values to BitBench)  Can it be automated on a batch of data, or is it manual?

I make **lots** of mistakes, especially when learning stuff, (aka learning C **or** deciphering a new device) so I'm much happier with a batch process that I can rerun from scratch (including compiling code if necessary), so long as it's fairly quick, rather than anything interactive (which I usually manage to make additional mistakes in :-( , which really confuses the learning process).  I make many incremental additions as I go, as I work out a part of the puzzle, and doing the same manual step iteratively doesn't really sit well with that - too tedious and error prone.

PS Maybe I'm not an absolutely typical newbie ? (I have messed about with coding before in Python and  a bit of Ruby, not much C though!!!).  And I do like to figure out how and why things work, as well as arrive at the end result.    You've probably had a wider variety of dealings with different people's approaches to this than I have (my sample size is exactly 1, namely me :-) )

PPS agree string of bits on their own is overwhelming and problematic - hence hex first (to narrow down which bytes change a lot => suggests temperature or check field) and which ones are pretty constant (suggests preamble, sensor id), then (for eg temperature)  look at bits from *suspicious* bytes to try and figure out **how** temperature is encoded into those bytes. Part of my retrospective learning (which I tried to explain in the tutorial) is that with a bit of forethought, you can develop an idea of **what** *patterns* to look for to find a particular type of field, even before you see a single bit of real data.

zuckschwerdt:
Ok, like you wrote earlier you don't want/need to have a whole bitbuffer as bits but just some rows or fields to get a better feel for those while writing the code. I guess a debug output of bits is a good way for that. I'll add it, with the caveat not to use that in working production code ;)

Geoff99:
That would be brilliant, many thanks indeed!!!!!!

zuckschwerdt:
The change is in. You need:
- only `#include "decoder.h"`
- the callback now takes `(r_device *decoder, bitbuffer_t *bitbuffer)`
- `debug_output` is now `decoder->verbose`
- instead of `data_acquired_handler` use `decoder_output_data(decoder, data)` 
- `json_callback` is now properly named `decode_fn`

Adding the debug output function with bits is next ;)

Geoff99:
That's a nice neat parcel of improvements.  The decoupling is good, and the renames make what's happening (even) easier to follow.

I've got a few other things to take care of over the next few days - but I'm going to do a rebase and start making the appropriate alterations to my XC0324 code straight away anyway, because the changes are just too enticing to wait to explore.

Geoff99:
Hi @zuckschwerdt 

This pull request is **almost** ready to go now, I hope :-)  I've amended the decoder to use the new utility functions and it looks quite compact and tidy (to me at least - though they do say beauty is in the eye of the beholder - as always, feel free to suggest changes and improvements).  

It does include a "patch" to `decoder_output_bitrow` to make it produce verbose binary output when running with `decoder->verbose > 0`, mostly so I can test out the outputs, and get ready to amend the "by newbie, for newbie" tutorial to agree with the current decoder.  I'll remove that patch when you have time to produce the official version (I see from activity on the repo that you have a lot of improvements underway at the moment!).

Changing to use the new utilities was quite painless.  What did take me some time was learning about squashing commits, and then actually doing that on my local repo. (I had a bit of a tangled mess of commits, rebases, renames of files, ….).  I did get it to work (as far as I can see) on my local repo - but I've just noticed it hasn't been carried over to my Github fork of rtl_433 :-( .  Back to the drawing board for me, to see what I missed out.

zuckschwerdt:
First (and few) things I see: Raise max protocols to 114 not 115. Use µs values in the r_device, not `*4` sample counts. `reference_values_written` needs to go, strict rule of not keeping state in the decoders. Remove all `time` related output and lines, this is handled by the program now.

zuckschwerdt:
Also note that  short/long _limits to are now properly named _width to better reflect the updated demodulators.
Does [`decoder_output_bitrow_debug()`](https://github.com/merbanan/rtl_433/blob/ae2fb32742a5f5e00879d853e02e8b94017f049e/include/decoder_util.h#L40) fit? Happy to change it to make sure you the most value on debugging there.

Geoff99:
Everything mentioned above fixed (I think)

*  Raise max protocols to 114 not 115. done (my mistake, lost count while resolving merge conflicts in squash) 
* Use µs values in the r_device, not *4 sample counts. done 
* reference_values_written needs to go, strict rule of not keeping state in the decoders. done (I'm guessing this is so threading and such like can occur?) 
* Remove all time related output and lines, this is handled by the program now. done
* short/long _limits to are now properly named _width . changed (but see below for something new I learnt)

I have left the DDD style "reference value" output in there (for now), but since it's main purpose is to provide information for the rtl_433_tests repo and regression tests, it started me thinking about what actually would be the ideal information to save for that purpose (I understand the concept of regression tests; and consider them almost essential; but not the mechanics of setting them up, so at the moment I do a lot of manual reruns and cross checks).  Do you have any recommendations about what should be saved for future regression tests? maybe a json output file would be best?

`decoder_output_bitrow_debug()` and `decoder_output_bitrow_debugf()` do almost what I need for my approach.  Blank spacers after every 8th bit are pretty well essential (just so I can keep track of which bits are which; it also stops the spreadsheet from automatically (and very unhelpfully) converting the field into a (very large) floating point number).  Our earlier discussion about hex versus binary has made me realise that I use them both simultaneously - hex to hone in on **which** bytes vary enough to be (eg) the temperature field, and binary to try and figure out **how** temperature has been encoded.  So what would be **perfect** would be `hh hh hh hh … : bbbbbbbb bbbbbbbb bbbbbbb …`

I also realised (and have fixed) that `OOK_PULSE_PPM_RAW` has gone, and been replaced by `OOK_PULSE_PPM`.  

That (and/or the change to *_width) did lead to an unexpected problem though.  I had set `short_limit` to half way between the  (average) length of short gaps and long gaps.  That used to work fine, but with the latest changes it worked fine for 2 of the 3 collections of test data, but totally failed to trigger at all (well actually was decoded as `00 00 00 00 00 00 00 …`) for the third.  When I tracked that down (took a while because I though it was something else I'd changed) all I had to do was set the `short_width` to the correct value, likewise the `long_width`, and everything worked beautifully again.  Not sure why (more precisely, not sure why I got away with the old values), but anyway the new names and values align much better with what `-a` and `-A` output produces.

PS I'm pleased to see in #889 that there might be a first user for the newbie tutorial.  I'll have to get onto bringing that up to date asap.  As it stands it is really quite convoluted compared to the way the same results could be achieved now, using the flex decoder, and the decoder_utils.

zuckschwerdt:
re PPM_RAW, oops, I' didn't notice you were using that. The _limit business was confusing. I updated and switched all PPM decoders to use proper _widths -- like you discovered the hard way. Sorry!

zuckschwerdt:
On the binary issue, if you also need hex, wouldn't [HEX2DEC and DEC2BIN](https://support.office.com/en-us/article/convert-numbers-to-different-number-systems-880eeb52-6e90-4a9d-9e56-acaba6a27560) be a solution then?

Geoff99:
Mmm, HEX2DEC and so forth could *perhaps* work.  

But it's one more thing that would have to be explained to a newbie about the reverse engineering approach, and more importantly, it's one more thing that would have to be edited into the csv file for each iteration / attempt at deciphering a field.  I'd be embarrassed to confess how many attempts I had before I "spotted" that the temperature was lsb first - and deciphering the checksum took me at least as long.  Suffice to say I spent days peering at screens, and printouts of part of spreadsheets.  In **hindsight**  (now I have a *bit* more experience with C, and with rtl_433, and likewise with reverse engineering simple fields) it **should** have been simple, and that's partly what motivated me to write the "bynewbie4newbie" tutorial.  

What is so nice about the flex decoder **and** the decode-util functions, **and** the -K argument, **and** the tag ing, **and** the "process multiple files" changes you have already made is that together they mean that it is possible to do all the deciphering work just using rtl_433. And most of the tutorial becomes an introduction to the nice things the `rtl_433` codebase provides, plus some "in hindsight" advice about how to decipher a field.

In short, I'm still hoping I can convince you that `hh hh… : bbbbbbbb bbbbbbbb ..` is not too big a change to make to `decoder_output_bitrow_debug` .   I'm sure there are prettier ways, but here's an example I've written that tries to follow the pointer based approach in the current (bits only) version.

```
void decoder_output_bitrow_debug(r_device *decoder, bitrow_t const bitrow, unsigned bit_len, char const *msg)
{
    data_t *data;
    char *row_bits, *p;
    unsigned i;
    unsigned byte_len = (bit_len + 7) / 8;

    p = row_bits = malloc(3*byte_len + 2 + bit_len + byte_len + 1); // "hh .. : bbbbbbbb ..\0"

    // print hex, with byte separators
    for (unsigned i = 0; i < byte_len; ++i) {
        sprintf(p, "%02x ", bitrow[i]);
        p++; p++; p++;
    }
    // Rewind and remove last nibble if necessary
    if (((bit_len + 7) % 8) < 4) {
        p--; p--; *p++ = ' ';
    }
    //Add hex : binary separator
    *p++ = ':'; *p++ = ' ';
    
    // print bit-wide, with byte separators
    for (unsigned i = 0; i < bit_len; ++i) {
        if (bitrow[i / 8] & (0x80 >> (i % 8))) {
            *p++ = '1';
        }
        else {
            *p++ = '0';
        }
        if (!((i + 1) % 8)) *p++ = ' ';

    }
    *p++ = '\0';

    data = data_make(
            "msg", "", DATA_STRING, msg,
            "codes", "", DATA_STRING, row_bits,
            NULL);
    decoder_output_data(decoder, data);

    free(row_bits);
}
```

zuckschwerdt:
@Geoff99 the output functions (`decoder_output_bitbuffer`, `decoder_output_bitbufferf`, `decoder_output_bitrow`, `decoder_output_bitrowf`) can now output bits (nibble grouped) if requested with the `-M  bits` option!
The output is in a separate column/key `bits` along then `codes` column/key.
Thanks for being insistent, I think this will be useful.

Geoff99:
Hi @zuckschwerdt 

The  `-M "bits"`  approach is just **perfect**.  Thanks so much for your patience!

A few non rtl_433 commitments that I've been avoiding have caught up with me :-( , so it will be a day or two before I can meld these changes into the tutorial properly (but I couldn't wait to try the "bits" option out, so I *stole* a few minutes to do a rebase and give them a try :-).  Will make a tiny change to the XC_0324 decoder (so it only emits the first successful message, instead of up to three repeats, unless in debug / tutorial mode) when I get to that as well.

BTW I did get a few minutes to try the IQ spectrogram using Chrome.  It **works** there, (albeit with some grumbles about some aspect or other related to security ???).  **Very nice** !! When I get free of my *distractions* I'll have a better look at what's going on there as well, and provide feedback.

zuckschwerdt:
Going to merge this now. Do we have sample data in rtl_433_tests for this?

zuckschwerdt:
Sorry, yes, we have. [rtl_433_tests/tests/XC-0324](https://github.com/merbanan/rtl_433_tests/tree/master/tests/XC-0324).

PR #858, opened 2018-11-06
support for hideki/cresta temperature sensor

Mindavi:
This patch adds support for the cresta temperature sensor module.
The module talks with the HIDEKI protocol, but doesn't contain the
humidity/wind/rain data (afaict). This means we can easily parse those
using the existing infra. Only a new 'module type' needs to be added.

### DIFF
--- a/src/devices/hideki.c
+++ b/src/devices/hideki.c
@@ -15,7 +15,7 @@
 //    00000000  11111111  22222222  33333333  44444444  55555555  66666666  77777777  88888888 99999999 AAAAAAAA BBBBBBBB CCCCCCCC DDDDDDDD
 //    SYNC+HEAD cha   RC                Nr.?    1° .1°  VB   10°    1° .1°  VB   10°  .1mh 1mh  ?? 10mh   ????    w°  ??    ????     ????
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks. Do we have some samples in rtl_433_tests?

Mindavi:
Didn't have them earlier, but added them just now.

Mindavi:
This also triggers on the SIGMA-TE130NL and mebus_te204nl sensors. Both of those sensors look a lot like the sensor I've got at home, so I think that's correct. This might also mean that the https://github.com/merbanan/rtl_433/blob/master/src/devices/mebus.c should be deleted. It doesn't trigger anymore when I enable it with `rtl_433 -R 13 -r rtl_433_tests/tests/mebus_te204nl/01/gfile001.cu8`, so I'm not sure about the state of that code.

zuckschwerdt:
I don't think we have samples for "Mebus 433" from mebus.c. I guess that was a different device.
No comments, no samples, it's likely to be removed someday.
At least the PPM expected there doesn't match the MC samples in mebus_te204nl/01 and shouldn't ever have.

PR #863, opened 2018-11-14
Refactor AM analyzer

zuckschwerdt:
Refactors the AM analyzer (-a) as well defined module.
No functional changes except the analyzer can now run alongside the decoders.
The analyzer code is mostly unchanged but isolated and properly wrapped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,9 @@ Usage:	= Tuner options =
 	[-x <value>] Override long value in data decoder
 	[-n <value>] Specify number of samples to take (each sample is 2 bytes: 1 each of I & Q)
 	= Analyze/Debug options =
-	[-a] Analyze mode. Print a textual description of the signal. Disables decoding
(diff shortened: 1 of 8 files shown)


merbanan:
ok

merbanan:
Does it allow the signal recorder code (-t) to either filter out the signals that a decoder picks up? Or change the name of a recorded signal when a decoder identifies it?

zuckschwerdt:
Yes `-a` now honors the `-I` setting (`rtl_433 -a -I 1 -t` will only grab unknown signals). Changing the name sounds good, I'll work on that afterwards.

merbanan:
Maybe even place them in specific folders. signals/known | signals/unknown, placing them in the current directory is maybe not the best option.

zuckschwerdt:
After the analyzer is refactored I plan to split the signal grabber off and use the auto-level framing instead of that analyzers framing. Also using proper ring buffers with the input will eliminate the need to continuously copy all the input buffers :)

PR #866, opened 2018-11-16
am_analyze.c: Fix compilation under MS VS 2015/2017

winterrace:
MS environment does not provide <unistd.h>.

I chose an equivalent solution than already present in rtl_433.h


zuckschwerdt:
Thanks!

PR #881, opened 2018-11-22
Fix automake build.

kukabu:
Fixes: 0daff85 ("Add a conf file parser (#790)")
Signed-off-by: Michael Tatarinov <kukabu@gmail.com>


kukabu:
I think it would be great to add Travis CI for checking such trivial bugs.

zuckschwerdt:
Thanks! This should be caught by the ./maintainer_update.py -- but I forgot to run that :/

PR #886, opened 2018-11-26
Decouple decoders

zuckschwerdt:
This decouples decoders from the CLI by replacing global references with a context.
Basically just adds a decoder reference to the device callbacks to reach the verbosity flag and the data_acquired callback.

### DIFF
--- a/include/decoder.h
+++ b/include/decoder.h
@@ -26,7 +26,4 @@
 extern int debug_output;
 extern float sample_file_pos;
 
-/** Pass the data structure to all output handlers. Frees data afterwards. */
-void data_acquired_handler(data_t *data);
-
 #endif /* INCLUDE_DECODER_H_ */
--- a/include/decoder.h
+++ b/include/decoder.h
@@ -23,7 +23,4 @@
 #define FSK_PULSE_PWM                   17 // FSK, Pulse Width Modulation. Short pulses = 1, Long = 0
 #define FSK_PULSE_MANCHESTER_ZEROBIT    18 // FSK, Manchester encoding
 
-extern int debug_output;
-extern float sample_file_pos;
-
 #endif /* INCLUDE_DECODER_H_ */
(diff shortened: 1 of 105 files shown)


zuckschwerdt:
Now also moves the common "time" (time_str) generation to data_acquired_handler to allow better control over meta-data.

merbanan:
I see that you moved out the time parameter generation. But I don't see where the actual time stamp is generated. It should be generated as soon as the signal is acquired before we try to demodulate the signal. That way we get as good as possible timestamp.

And related to that do we output the signal duration in the meta data?

merbanan:
Oh and patch looks good. As always.

zuckschwerdt:
Great ideas! Currently the timestamp is generated after-the-fact (and tweaked for sample pos when reading files). (I moved the scattered `local_time_str(0, time_str)` calls to `data_acquired_handler()` in rtl_433.c.)
There currently is no duration meta data. With the new approach we could easily grab a precission timestamp and the duration and then prepend that in the acquired handler. Other sources of "event" or time meta-data can be placed there too (GPS?)

merbanan:
I had plans to place out receivers in several locations where I live and collect them in a central place it might be a good idea to add a receiver id to the payload. That way you don't need to sort the messages based on the underlying transport parameters but instead just look at the messages.

zuckschwerdt:
Already done :) Use `-K <receiver_id>` (or `-F FILE|PATH` with `-r <file>`).

merbanan:
Look at that. Now I only need to get the MQTT code working.

zuckschwerdt:
Since this change is blocking other work I'll be merging this shortly -- if no objections or other ideas pop up ;)

merbanan:
No objections here.

zuckschwerdt:
The change is in. Decoders now do
- only `#include "decoder.h"`
- the callback now takes `(r_device *decoder, bitbuffer_t *bitbuffer)`
- `debug_output` is now `decoder->verbose`
- instead of `data_acquired_handler` use `decoder_output_data(decoder, data)` 
- `json_callback` is now properly named `decode_fn`

PR #893, opened 2018-12-04
Change raincounter_raw field to rain_inch for acurite

zuckschwerdt:
This removes the derived value "rainfall_accumulation_inch" and "wind_dir".
Both are better handled in a consumers presentation logic/mapping.
Also changes "raincounter_raw" to "rain_inch" with proper unit.

The key "rain_inch" might better be "rainfall_inch" and should be discussed in #827 

This is a breaking change.


merbanan:
Looks good.

PR #894, opened 2018-12-04
Add rtl_tcp input

zuckschwerdt:
Basically working but commands like setting frequency need to be added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Usage:	= General options =
 	[-V] Output the version string and exit
 	[-c <path>] Read config options from a file
 	= Tuner options =
-	[-d <RTL-SDR USB device index> | :<RTL-SDR USB device serial> | <SoapySDR device query>]
+	[-d <RTL-SDR USB device index> | :<RTL-SDR USB device serial> | <SoapySDR device query> | rtl_tcp]
 	[-g <gain>] (default: auto)
 	[-f <frequency>] [-f...] Receive frequency(s) (default: 433920000 Hz)
 	[-H <seconds>] Hop interval for polling of multiple frequencies (default: 600 seconds)
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -155,7 +155,7 @@ static void usage(r_device *devices, unsigned num_devices, int exit_code)
             "\t[-V] Output the version string and exit\n"
             "\t[-c <path>] Read config options from a file\n"
             "\t= Tuner options =\n"
-            "\t[-d <RTL-SDR USB device index> | :<RTL-SDR USB device serial> | <SoapySDR device query>]\n"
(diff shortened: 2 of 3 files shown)


zuckschwerdt:
Commands are implemented and rtl_tcp support looks finished.

merbanan:
Patch looks fine. Merge at will.

ClayF:
Any reason why the host and port and hard coded? I'm trying to connect to a remote rtl_tcp server.

zuckschwerdt:
From `rtl_433 -d`:
```
rtl_433: option requires an argument -- d
	RTL-SDR device driver is available.
	SoapySDR device driver is available.
[-d "" Open default SoapySDR device
[-d driver=rtlsdr Open e.g. specific SoapySDR device
	To set gain for SoapySDR use -g ELEM=val,ELEM=val,... e.g. -g LNA=20,TIA=8,PGA=2 (for LimeSDR).
[-d rtl_tcp[:host[:port]] (default: localhost:1234)
	Specify host/port to connect to with e.g. -d rtl_tcp:127.0.0.1:1234
```
So `-d rtl_tcp:127.0.0.1:1234` should work for your case.

We need a better manual. Where did you expect to see this info, where did you look for this help? We need to put it there. Thanks!

ClayF:
Thanks! I didn't realize -d had its own help page.

```rtl_443 -h``` currently shows the following:

```
[-d <RTL-SDR USB device index> | :<RTL-SDR USB device serial> | <SoapySDR device query> | rtl_tcp]
```

How about changing it to reflect that it accepts a host and port for conformity:
```
[-d <RTL-SDR USB device index> | :<RTL-SDR USB device serial> | <SoapySDR device query> | rtl_tcp[:host[:port]]]
```

**Comment by zuckschwerdt** [comment]
Thanks for the feedback. The note at the end isn't prominent enough then :)
```
		 Use -d, -g, -R, -X, -F, -M, -r, or -w without argument for more help
```
I tried to unclutter the main help page by spliting of help specific to the main options.
Maybe putting `help` as first possible argument on all those is a better suited.
```
…
	[-d help | <RTL-SDR USB device index> | :<RTL-SDR USB device serial> | <SoapySDR device query> | rtl_tcp]
	[-g help | <gain>] (default: auto)
	[-f <frequency>] [-f...] Receive frequency(s) (default: 433920000 Hz)
	[-H <seconds>] Hop interval for polling of multiple frequencies (default: 600 seconds)
	[-p <ppm_error] Correct rtl-sdr tuner frequency offset error (default: 0)
	[-s <sample rate>] Set sample rate (default: 250000 Hz)
	= Demodulator options =
	[-R help | <device>] Enable only the specified device decoding protocol (can be used multiple times)
		 Specify a negative number to disable a device decoding protocol (can be used multiple times)
…
```

ClayF:
That's a much better alternative. Having help as the last argument should be good enough, though.

zuckschwerdt:
Done. The text should give better hints now and the full help on all options is now included in the README also.

PR #896, opened 2018-12-05
Modernize the signal grabber

zuckschwerdt:
Use pulse detect to track and grab frames, instead of the fixed level AM Analyzer.
Deprecate option `-t` and `-I` for new `-S none|all|unknown|known`.

There is a deprecation note when using -t or -I instructing the user to use -S.

The signal grabber can now be run in normal decoding without analyzers.

Instead of `-a -t` users should now be adviced to use `-S all` or `-S unknown`.


merbanan:
Looks good.

PR #899, opened 2018-12-06
prologue: add Alecto checksum check to get rid of collisions

merbanan:
If the Alecto checksum matches we are quite sure this is not a
prologue signal.


merbanan:
This structure is also present with the rubicson protocol and the change is just a split out of the current code to a function so it could be used from another place.

PR #900, opened 2018-12-06
Add generic RC power socket (PT2262/PT2272)

anistor:
This pull request adds support for some generic remote control power sockets based on PT2262/PT2272 IC.

Here's the user manual of the one I tested with: https://www.somogyi.hu/data/product_documents/12121_03_01.pdf

My sample seems to emit on 433.85Mhz rather than on the documented 433.9Mhz frequency.

There's another one from another manufacturer that seems identical based on pictures (but I have not tested it): http://www.avidsen.com/cache/documents/product/fp-packde2prisestelecommandees-101444-avidsen-689.pdf

There are certainly many other clones of this product, so I called this 'generic'.


anistor:
Please note this device is somewhat similar to 'generic remote' defined in https://github.com/merbanan/rtl_433/blob/master/src/devices/generic_remote.c  but it adds proper address/data decoding for this particular application of the 2262 IC and also uses different timing to be able work with this RC power socket.

src/devices/generic_rc_power_socket.c
```diff
@@ -0,0 +1,162 @@
...
+            case 0x03: *p++ = '1'; break;
+
+            default: return 0;            // found an illegal bit combination, give up
+        }
+    }
+    *p = '\0';
+
+    data_t *data = data_make(
+            "model",        "",             DATA_STRING, "Generic Remote Controlled Power Socket (PT2262/PT2272)",
+            "remoteId",     "Remote Id",    DATA_INT,    remoteId,
```
anistor: Should all numbers sent to data_make be declared ```int``` ? ```remoteId```, ```socketId```, ```cmd``` are not int and it seems to work, but not sure...

src/devices/generic_rc_power_socket.c
```diff
@@ -0,0 +1,162 @@
...
+ * Copyright (C) 2018 Adrian Nistor
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ */
+
+#include "decoder.h"
+
+static int generic_rc_power_socket_callback(r_device *decoder, bitbuffer_t *bitbuffer)
```
anistor: This function returns 0 for broken packets. The packets emitted by the remote when multiple buttons are pressed, either both On/Off for same socket or for multiple sockets, are considered invalid because the socket also seems to reject them and does not act on them.

zuckschwerdt:
Your code looks good but simple devices (using x1527 or PT2262) are better implemented as flex decoders (There are some in the `rtl_433.example.conf`).
In your example something like
```
name=PowerSocket,
modulation=OOK_PWM,
short=160,
long=440,
reset=4336,
gap=412,
tolerance=150,
bits=25,
repeats=2,
invert,
get=@4:{4}:cmd:[4:on 1:off]
…
```
You'll likely also want a custom `match=...` to filter your specific device id.

I'm about to rewrite the getter syntax to something like the [BitBench](http://triq.net/bitbench) format synax to make things more obvious.
Supporting the common tristate-bits (`filter=tristate` perhaps) is also on my wish list.

Does that approach work for you, and what improvements to the flex decoder do you perhaps need?

anistor:
It all depends what you need to use rtl_433 for, @zuckschwerdt. The flexible parser is great for experimenting and reverse engineering rf protocols but it is also a bit limited. After you've figured out the protocol it makes sense to move a C-coded parser that is able to aggressively validate the packet and extract fields and output them in a both human readable and machine readable format.

The flexible parser is a bit too flexible for me, as is does not allow me to express the whole list of constraints I want for the packet contents. If the cmd bits match my expectation (```4:on 1:off```) then I get the proper ```on``` or ```off``` value, otherwise I get the actual numeric value of those bits. I'd rather not get any match in the second case.

Also, I'd want to have multiple ```match=``` statements to be able to filter on remote id and also to enforce the last bit of the packet to be 1. That's probably already doable, just never tried it.

There are some other rather complex constraints that cannot be expressed with the flex parser, like there are never any zeroes in the remote id, just 1s an Zs. The only way to do it right now is to match on a pre-generated value, but translating from the remote id to the actual sent bits requires a bit of legwork for the user so the flex parser is a bit uncomfortable to use right now. This could be worked-around with a ```get``` for the remote id field, so the user can easily see the extracted value and then add it to a match statement.

I seem to get multiple matches for a single button press, even if I press it very quickly. This does not happen with the C implementation I wrote because it emits a single data_t for the entire burst, provided that there are at least 2 identical repeats at the beginning, no mater how many identical or non-identical rows are following. This is probably easy to overcome with the flex parser too, I just don't know how to do it.

A side effect of the very simple electrical design of these remotes is that the last 2-3 transmitted rows are usually broken because the command bits have random values when the button is released. Also, if the user presses multiple buttons at once we'll get a packet that is invalid, but the power socket seems to properly ignore any such invalid packets and I'd like to put these validations into code of some sort, be it C or flex syntax.

zuckschwerdt:
Very good points. I try to address that soon in the flex spec. The general thought with the flex decoder is that those x1527/PT2262 packets will have different interpretation for all the remotes out there (buttons, PIR, window sensors, …) and the timings should be user-configurable as those differ among device implementations.
More checks, rules (tristate), and filters are on the wish list!

In summary: we do want your implementation to be easily modifiable by users to build upon instead of having single-use decoders for every one of those slightly different general-purpose signals :)

zuckschwerdt:
I'm going to add a `preset=` option, starting with the very common case of x1527.
It should include most of the checks and coversions you listed.
Then you only need to define timings and what the buttons or events on a specific device map to.

PR #904, opened 2018-12-08
Proposal of an MS-aware version of default config path determination

winterrace:
Hi,
this is a proposal how to try meaningful default paths for the configuration files under Windows.
Currently it looks like this:

```
D:\Coding\rtl_433_win\vs15\x64\Release>>rtl_433.exe
Trying conf file at "rtl_433.conf"...
Trying conf file at "(null)/.rtl_433.conf"...
Trying conf file at "/usr/local/etc/rtl_433.conf"...
Trying conf file at "/etc/rtl_433.conf"...
Registering protocol [1] "Silvercrest Remote Control"
...
```

With the patch it looks like this:

```
D:\Coding\rtl_433_win\vs15\x64\Release>rtl_433.exe
Trying conf file at "D:\Coding\rtl_433_win\vs15\x64\Release\rtl_433.conf"...
Trying conf file at "C:\Users\myusername\AppData\Local\rtl_433\rtl_433.conf"...
Trying conf file at "C:\ProgramData\rtl_433\rtl_433.conf"...
Registering protocol [1] "Silvercrest Remote Control"
...
```

p.s.: in the upper parts it's just some indentation fix (spaces instead of tab)

Please note that I did not test this with the MingW build script. If there will be any problems with my solution, it should work to replace both new "#ifdef _WIN32" with "#ifdef _MSC_VER". This way it should compile with MingW as before (but not profit from the new default paths)


merbanan:
I am not against this but I am against #ifdefs riddled all over the code. There should be a platform file that can contain all this. Maybe a target_linux.c and target_windows.c that is used on the different platforms. And you need to rebase this PR and maybe squash it, currently it does not apply cleanly.

PR #905, opened 2018-12-08
Add precision time and time report options

zuckschwerdt:
Like @merbanan suggested this snapshots the current time when a buffer is received and reports time with exact frame start precision.
Adds meta data options to add microseconds on time output, select sample position or data/time output.
Sample position output now shows the exact frame start instead of buffer start.


zuckschwerdt:
A gettimeofday() implementation for win32 needs to be added.

merbanan:
Looks good. Commit at will.

zuckschwerdt:
Additionally this deprecates the `-U` option in favor of `-M utc` to get all meta data options in one place.

zuckschwerdt:
@winterrace can you check if the gettimeofday implementation for win32 compiles cleanly and works?

winterrace:
Of course. I will do so tomorrow

zuckschwerdt:
@winterrace Thanks! I've just merged this, so you can PR a fix ;) cleanly on top of master.

PR #907, opened 2018-12-10
Fix precision time and time report options

winterrace:
As requested I checked the Win compatibility of the time report PR.

I had to change the includes to fix compilation for Windows.

MS compiler does not seem to recognize 
`#include <sys/time.h>`

but
`#include <time.h>`

works well.

util.h already had time.h included, so I removed the sys/time.h one.
in rtl_433.c I just removed the "sys/" and now its fine again.


zuckschwerdt:
Great! It's mostly working then :) Did the `-M hires` microsecond resolution timestamps work?
I'm afraid we can't drop the [sys/time.h](http://pubs.opengroup.org/onlinepubs/7908799/xsh/systime.h.html) on Linux. It's needed to define the timeval struct and `gettimeofday`. [time.h](http://pubs.opengroup.org/onlinepubs/7908799/xsh/time.h.html) defines the tm struct. It could be that we accidentally pull in sys/time.h somewhere else but time.h usually doesn't do that.
I guess sys/time.h then needs to be _WIN32 ifdef'd.

zuckschwerdt:
Thanks! Just use `-M hires` (e.g. `rtl_433 -M hires`) and all reports should have a dot and six digits after the seconds in the "time" field.

winterrace:
Damn, my good old E4000 rtl-sdr just went def suddenly. I'll try with my 2nd stick tomorrow which lies at my working place.

I just had a closer look at the code. rtl_433 does not call gettimeofday for _WIN32 because you already provided a replacement. And struct timeval is obviously being found already via some indirection from winsock.h which contains it:

```
/*
 * Structure used in select() call, taken from the BSD file sys/time.h.
 */
struct timeval {
        long    tv_sec;         /* seconds */
        long    tv_usec;        /* and microseconds */
};
```

I'll try to suggest an adapted version of this PR soon.

zuckschwerdt:
Sorry, my comments were meant for the non _WIN32 case. If I drop sys/time.h here both the struct and prototype for gettimeofday are missing,

winterrace:
I understood you correctly. I was just stating why the Windows version compiles fine without sys/time.h while it is required under Linux.

I just made a first quick test of your Windows variant of the high precision timing.
Currently it crashes here:

`size_t l = strftime(buf, LOCAL_TIME_BUFLEN, "%Y-%m-%d %H:%M:%S", tm_info);`
because tm_info is a NULL pointer.

The reason is in the line above:
`    tm_info = localtime(&tv->tv_sec); // note: win32 doesn't have localtime_r()`

There seems to be some definition problem that I will check this evening. Struct timeval *tv has 32 bit members (tv_sec here) while localtime expects a 64 bit time_t. So localtime interprets 4 bytes more as intended which yields a NULL pointer as result (--> crash).

A quick fix like this works fine:
```
    time_t time_sec = tv->tv_sec; // this will blow up 32 bit tv_sec to 64 bit time_sec
    tm_info = localtime(&time_sec); // note: win32 doesn't have localtime_r()
```
I will have some closer look at this in the evening and update the PR then.

winterrace:
I close this PR, there will be a reworked and fresh one soon.

PR #910, opened 2018-12-11
Reworked and merged PR: fix hi-res timestamps and default search paths on Win using platform files

winterrace:
Fixed feature:

- hi-res timestamps to not lead to a crash on Windows platform
- hi-res timestamp reading on windows (cast to 64 bit before <<32)

New feature:
- Default search paths regarding rtl_433 config files for Windows platform

Architectural change:
- To reduce #ifdefs: Introduced platform.h and platform_(win|lin).c for functions needing different treatment under Linux and Windows.

Comments:
- The new files (platform*.*) still have to be added to the other build scripts but I'm confident it should compile.
- We can add more stuff to the platform files later to further reduce the #ifdefs (e.g. the watchdog stuff)

Any comments?


zuckschwerdt:
Looks good. I'll think about it a few days. I think using compat headers per topic would be best, i.e. compat_time.h, compat_net.h, compat_path.h. That way we can see which code files reley on features.

winterrace:
OK, that would be fine for me as well.

zuckschwerdt:
About the `struct timeval` can't we use the struct win32 defines? The [notes on that](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/ns-winsock-timeval) (if I found the right docs) seem to fit. You mentioned something about wrong field size?

winterrace:
Maybe my statement was misleading.
The definition from winsock.h is not conflicting in the strict sense. I even presume it is completely compatible.

The problem is that winsock.h is (indirectly) visible from some source files while it is not from others.
If you include platform.h before or after the (indirect) inclusion of winsock.h the compiler will stop with an redefinition error.

Including winsock.h in all source files seemed a bit too much overhead for me (and it also did not work in a quick test).

Then I tried to always place the "platform.h" includes after the last  <...system...> includes (so that winsock.h will always be first) and only define struct timeval in platform.h if it is still unknown. But there's no such thing as "ifdef struct timeval" on precompiler level. Therefore I finally came to the conclusion with the typedef...

zuckschwerdt:
Maybe I'm thinking wrong here, but can't we include winsock.h when sys/time.h is needed and thus have struct timeval? E.g. compat_time.h with
```
// pull in struct timeval
#ifdef _WIN32
#include "winsock.h"
#else
#include "sys/time.h"
#endif
```

winterrace:
That's what I tried. It led to another compilation error. I'll re-check this evening what the exact problem was.

zuckschwerdt:
Chances are *0*. The timestamp is the exact sample count to the start of the buffer. I'd say this is the same demod with different decoders representing the data, a false positive on Prologue here.

gvanem:
I'd say it's the *Oregon* that is a false positive. The temperature here now is more like 3.9 C (not 0.2 C).

merbanan:
So the Oregon has a valid checksum and reports a humidity of 29%. The Prologue reports 102% humidity which sounds strange. Relative humidity is the percentage that the air can hold water in gas form relative to the temperature. 29% sounds much more reasonable in this case.

Anyway unless you own both devices and can verify the values against a real receiver I'd say the signals can't be trusted at all.

gvanem:
They're not mine (some close-by neighbour). 

But would it be possible to create a *False positive* counter based on such close timestamps? And print that in each case (with the KV-output).

merbanan:
When looking through the code it looks like the code needs some rework. It would be nice if you could record one of these dual emit signals. I am quite sure I can make both decoders more robust.

zuckschwerdt:
I plan to rework the OS special case demods to precise timings and sync option. That could perhaps also help.

PR #913, opened 2018-12-13
Plug a mem-leak in sdr_close()

gvanem:
A small fix to free the memory allocated in `rtltcp_open()`, `sdr_open_rtl()` or `sdr_open_soapy()`. 

(There are more leaks. More PRs coming up?).

### DIFF
--- a/src/sdr.c
+++ b/src/sdr.c
@@ -728,6 +728,7 @@ int sdr_close(sdr_dev_t *dev)
 #endif
 
     free(dev->buffer);
+    free(dev);
     return ret;
 }
 


zuckschwerdt:
Important fix, keep 'em coming! Is close() the right place to free dev? I think something called free() would be better. But since the memory comes from open() it's not unexpected I guess ;)

gvanem:
> Important fix, keep 'em coming!  

I really have no simple fix to plug all the mem-leak sources. There are tons of places where leaks happen due to `help_meta()`, `helpdevices()` etc. are calling `exit()`. And since memory is allocated early in `main()`, it would be a mess to plug all those leaks. 

Similar to what I wrote about this issue last year. Pseudo-code:
```c
{
  struct dm_state *demod = NULL;
  int    rc = parse_cmd_line_and_config_file();

  if (rc == 0)
  {
    rc = init_demodulator (in_filename, out_filename, &demod);
    if (rc == 0)
    {
      if (demod->in_file)
         rc = run_infile_loop (demod);
      else if (sync_mode)
         rc = run_live_sync_loop (demod);
      else
         rc = run_live_async_loop (demod);
    }
  }
  main_cleanup (demod); // free 'demod' and all output-handlers
  return rc;
}
```

These `run_X_loop()` functions should IMHO never be allowed to call `exit()`;
`#define exit(n) fprintf(stderr, "FATAL: not allowed.\n")`

zuckschwerdt:
Yes, still on my list. The main program loop is getting smaller and more manageable. Still some refactoring to do though. The hint about no exit, just loop abort is good.
Although I only really fear runtime leaks, not exit leaks. Those are well defined.

gvanem:
Even a `rtl_433 -R?` leaks now. No decoder looping involved.

PR #918, opened 2018-12-15
Always print version, register decoder ranges, quiet default add -v

zuckschwerdt:
Always print the version.
Print registered decoder ranges.
Quiet by default.
Add -v verbosity option, deprecated -q and -D

Output looks like
```
rtl_433 version 18.05-355-g4faf491 branch feat-quiet at 201812151704
Registered 95 out of 119 device decoding protocols [ 1-4 8 11-12 15-21 23 25-26 29-36 38-60 62-64 67-71 73-100 102-103 108-116 ]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,9 @@ Running:
 
 ```
 Usage:	= General options =
-	[-q] Quiet mode, suppress non-data messages
-	[-D] Print debug info on event (repeat for more info)
 	[-V] Output the version string and exit
+	[-v] Increase verbosity (can be used multiple times).
+		 -v : verbose, -vv : verbose decoders, -vvv : debug decoders, -vvvv : trace decoding).
 	[-c <path>] Read config options from a file
 	= Tuner options =
 	[-d <RTL-SDR USB device index> | :<RTL-SDR USB device serial> | <SoapySDR device query> | rtl_tcp]
--- a/rtl_433.example.conf
+++ b/rtl_433.example.conf
@@ -19,13 +19,9 @@
 ## General options
 
 # as command line option:
-#   [-q] Quiet mode, suppress non-data messages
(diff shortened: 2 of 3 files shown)


gdt:
First, replacing the vastly long list of decoders with a single line that has the information is great.  The real problem with the list is that the ratio of useful information to lines printed was too small.

Before your one-line decoder list, I had wanted three modes:
 * quiet (data only, nothing else)
 * normal (useful metadata, meaning version, input source, frequency, samplerate)
 * verbose (add decoder list)

With the compact decoder list, I think there needs to be an option to just output the decoder list with names and if it's default or not, separately from actually running.   This is something people will want occasionally, but not in normal operation.

Then, there's the question of whether quiet vs normal should be the default.  I think that things like program version, input source, frequency and sample rate, the one-line decoder summary are all things that someone would want to have alongside the data, almost always, and they aren't that big.  So they should be on by default, and perhaps prefixed with a # or something so they can be grepped out.

I don't understand why people would want to omit all metadata by default.   Is that about piping to mqtt, or something?   Just "run it and save the output" doesn't fit this, and I think that's what people getting started would do.   Plus, it means this data  is more likely available when they ask for help.   This is why I specifically brought up demoting the decoder list, as having many many lines and not that much information, compared to the other summary metadata.

zuckschwerdt:
The `-R` (no argument or `help`) will output all the decoders :) Lot's of help texts have been extended recently!

Quiet won't ever be real quiet. There is information that is always helpful, version, protocol state, startup. But all that is on stder. One can just `2>/dev/null` to get the data only. Or better even use something like `-F csv:file.csv`.

So my change specifically only quiets the verbose things. Like you mention users will ask for help and the minimum output should still always tell enough :)

PR #920, opened 2018-12-16
Few fixes to re-enable compilation with MS VC after latest updates

winterrace:
Also, some unreferenced variables have been removed...


zuckschwerdt:
Thanks! Please don't change Mongoose, it's a distribution file. A unistd.h should be available on win32? Was there a reasone to remove it from term_ctl.c?

winterrace:
- unistd.h: Unlike the other standard C includes, an "#include <unistd.h>" causes the MS VC compiler to "fatal error C1083: "unistd.h": No such file or directory". It can safely be left out here without any problems, that's why i moved it into the existing #ifndef _WIN32 block.

- mongoose: It also compiles without my fix but it has an ugly taste.  Mongoose include these Windows-overrides...
```
#ifdef _WIN32

int rmdir(const char *dirname) {
  return _rmdir(dirname);
}

unsigned int sleep(unsigned int seconds) {
  Sleep(seconds * 1000);
  return 0;
}

#endif /* _WIN32 */
```
Both are cleanly included with the standard C support since Visual Studio 2015. In the first case, this unnecessary redefinition leads to a warning C4273: "rmdir": inconsistent dll linkage.

In addition, sleep() is not used anywhere in the code so I don't have any idea what's its purpose here.

zuckschwerdt:
Ok, not sure if mingw has rmdir() but let somebody find out then :)

PR #922, opened 2018-12-17
Add Digitech XC0346 support to Fine Offset WH1050

projectgus:
Digitech XC0346 is a rebranded Fine Offset WH1050 on 433.920MHz, except on the unit I have the preamble pulse is 7 bits long not 8 bits.

Possibly the first bit is swallowed by the transmitter module while it is waking up...?

I couldn't find any WH1050 samples in the tests repo to compare, but even rolling back rtl_433 to the initial commit which added WH1050 support didn't work with signals from my Digitech one. This patch shouldn't change the behaviour of "standard" WH1050s.

Tests repo PR with samples: https://github.com/merbanan/rtl_433_tests/pull/249

### DIFF
--- a/src/devices/fineoffset_wh1050.c
+++ b/src/devices/fineoffset_wh1050.c
@@ -35,14 +35,18 @@
 #include "decoder.h"
 
 #define CRC_POLY 0x31
-#define CRC_INIT 0xff
+
+// If you calculate the CRC over all 10 bytes including the preamble
+// byte (always 0xFF), then CRC_INIT is 0xFF. But we compare the preamble
+// byte and then discard it.
+#define CRC_INIT 0x00
 
 static unsigned short get_device_id(const uint8_t* br) {
-    return (br[1] << 4 & 0xf0 ) | (br[2] >> 4);
+    return (br[0] << 4 & 0xf0 ) | (br[1] >> 4);
 }
 
 static char* get_battery(const uint8_t* br) {
-    if (!(br[2] & 0x04)) {
+    if (!(br[1] & 0x04)) {
         return "OK";
     } else {
         return "LOW";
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
If I get this right the preamble is in br[0]. Can you update your patch to lower every br[n] to br[n-1] and remove the preamble when it is detected? I.e. use bitbuffer_extract_bytes(bitbuffer, 0, 8, br, 72) or bitbuffer_extract_bytes(bitbuffer, 0, 7, br, 72) depending.

merbanan:
And it would be nice to actually output the Digitech XC0346 model name in the output.

zuckschwerdt:
@merbanan I don't think that is robust enough (original with a dropped first bit will now also be accepted) and it's only rebranded.

merbanan:
Ok, just leave it in the code comments.

projectgus:
> If I get this right the preamble is in br[0].

You've got it. The preamble in br[0] is either 0xFF or 0xFE (for 8 or 7 1s in a sequence).

(I think the preamble might actually be either 7 or 8 `1` bits followed by `4` 0 bits, as the upper 4 bits of br[1] are currently not used for anything. But I don't know this for sure.)

> Can you update your patch to lower every br[n] to br[n-1] and remove the preamble when it is detected? I.e. use bitbuffer_extract_bytes(bitbuffer, 0, 8, br, 72) or bitbuffer_extract_bytes(bitbuffer, 0, 7, br, 72) depending.

Sure, can do.

>  I don't think that is robust enough (original with a dropped first bit will now also be accepted) and it's only rebranded.

Yes, without more seeing captures from different models it's hard to know if this is a quirk of Digitech vs Fine Offset branding, a quirk of my particular transmitter, a quirk of one particular 433MHz transmitter used in one factory batch, etc.

zuckschwerdt:
Thanks! You can just put more commits on top or squash and force push.

projectgus:
@zuckschwerdt Sorry for the extreme lag in updating. I actually made the changes the day after you asked for them, then thought "I should test this for a bit first", and suddenly here we are...

(The tests have been going pretty well, though. :laughing: )

zuckschwerdt:
Great! Thanks for cleaning up! No worries about lag. And just ping us about progress if we in turn haven't responded in a while ;)

PR #923, opened 2018-12-17
Support for Windows console colours

gvanem:
Call `SetConsoleTextAttribute()` to set console colours mapped from the ANSI sequence numbers in `enum term_color`. Also supports setting a background colour (although this is not used by `src/data.c`yet).

### DIFF
--- a/src/term_ctl.c
+++ b/src/term_ctl.c
@@ -17,64 +17,203 @@
 
 #include "term_ctl.h"
 
+#ifdef _WIN32
+#include <io.h>
+#include <limits.h>
+#include <windows.h>
+
+#ifndef STDOUT_FILENO
+#define STDOUT_FILENO   1
+#endif
+
+#ifndef STDERR_FILENO
+#define STDERR_FILENO   2
+#endif
+
+static CONSOLE_SCREEN_BUFFER_INFO console_info;
+static BOOL                       console_redirected = FALSE;
+static HANDLE                     console_hnd = NULL;
+static FILE                      *console_file = NULL;
+static WORD                       active_fg, active_bg;
+
+static WORD _term_get_win_color(BOOL fore, term_color_t color)
+{
+    switch (color) {
+       case TERM_COLOR_RESET: /* Cannot occur; just to supress a warning */
+            break;
+       case TERM_COLOR_BLACK:
+       case TERM_COLOR_BRIGHT_BLACK:
+            return (0);
+       case TERM_COLOR_BLUE:
+            return (1);
+       case TERM_COLOR_GREEN:
+            return (2);
+       case TERM_COLOR_CYAN:
+            return (3);
+       case TERM_COLOR_RED:
+            return (4);
+       case TERM_COLOR_MAGENTA:
+            return (5);
+       case TERM_COLOR_YELLOW:
+            return (6);
+       case TERM_COLOR_WHITE:
+            return (7);
+       case TERM_COLOR_BRIGHT_BLUE:
+            return (1 + FOREGROUND_INTENSITY);
+       case TERM_COLOR_BRIGHT_GREEN:
+            return (2 + FOREGROUND_INTENSITY);
+       case TERM_COLOR_BRIGHT_CYAN:
+            return (3 + FOREGROUND_INTENSITY);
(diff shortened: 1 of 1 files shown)


src/term_ctl.c
```diff
@@ -17,64 +17,203 @@
...
+       /*
+        * Call this again as the screen dimensions could have changes since
+        * we called '_term_init()'.
+        */
+       CONSOLE_SCREEN_BUFFER_INFO c_info;
+
+       GetConsoleScreenBufferInfo(console_hnd, &c_info);
+       return (c_info.srWindow.Right - c_info.srWindow.Left + 1);
+    }
+    return (INT_MAX);
```
zuckschwerdt: "80" sounds like a better fallback

src/term_ctl.c
```diff
@@ -17,64 +17,203 @@
...
+#endif
+
+#ifndef STDERR_FILENO
+#define STDERR_FILENO   2
+#endif
+
+static CONSOLE_SCREEN_BUFFER_INFO console_info;
+static BOOL                       console_redirected = FALSE;
+static HANDLE                     console_hnd = NULL;
+static FILE                      *console_file = NULL;
```
zuckschwerdt: Can't we pass this as parameter?

src/term_ctl.c
```diff
@@ -17,64 +17,203 @@
...
+static WORD                       active_fg, active_bg;
```
zuckschwerdt: Can you pull this from wAttributes when needed?

src/term_ctl.c
gvanem: My thinking was that on redirection (`rtl_433.exe what-ever > foo`), there is no limitation. But what is the KV-output supposed to look like then?

src/term_ctl.c
```diff
@@ -17,64 +17,203 @@
 
 #include "term_ctl.h"
 
+#ifdef _WIN32
+#include <io.h>
+#include <limits.h>
+#include <windows.h>
+
+#ifndef STDOUT_FILENO
+#define STDOUT_FILENO   1
```
zuckschwerdt: Isn't this identical to "stdout" from stdio.h?

src/term_ctl.c
gvanem: Were? It's just a copy of `fp` from `term_init()` and `term_has_color()`. Probably `stdout`.

src/term_ctl.c
zuckschwerdt: Exactly right on the redirection. But for column wrapping there needs to be a sensible limit. Something between 80 and 160 looks good. I'd say 80 is always a good limit.

src/term_ctl.c
zuckschwerdt: Just pass fp as parameter then and remove console_file here?

src/term_ctl.c
zuckschwerdt: I see. We would need to set fg and bg with a single call then, or use this static.

src/term_ctl.c
gvanem: I don't understand. What about `_term_exit()`? An `atexit()` function is `void`.

gvanem: A global `struct r_cfg cfg` with some added terminal stuff then?

zuckschwerdt: The global r_cfg is nearly gone. It won't be available anymore. But we can pass a void *ctx to every term_ call.

gvanem: I'll patch that later if you just merge this PR. Then I have something to edit. (I'm still a *git n00b* and won't be able to edit this PR now).

zuckschwerdt:
No need for edits. I'll merge and rewrite with a ctx pointer later. Thanks for writing this and explaining it!

zuckschwerdt:
Can you test if https://github.com/zuckschwerdt/rtl_433 works (commit https://github.com/zuckschwerdt/rtl_433/commit/b91c5e58d9e795032b2dc857c58df3614ed4060e)
Mostly written blindly because I don't have the compiler ;)

gvanem:
After your patch, there is in `term_get_columns()`, a:
```c
 if (fd == fileno(console->file)) 
```

This test is unneeded now.

PR #924, opened 2018-12-17
Make an exception for TERM_COLOR_BLACK

gvanem:
Otherwise this would make the foreground colour GREY when OR'ed with FOREGROUND_INTENSITY.
But who has terminal settings where BLACK colours are visible?


zuckschwerdt:
My terminal is slate gray and thus black is visible there :p No ninjas hiding in shadows for me!

PR #927, opened 2018-12-17
Proposal of addressing "TODO: check MAX_DUMP_OUTPUTS"

winterrace:


zuckschwerdt:
I would like to keep the dumper loops as is, those will be changed to a real list_t soon. add_dumper should keep r_cfg as parameter. There are some stray tabs in the patch ;)
Otherwise, yes, needs to be done.

zuckschwerdt:
I was thinking of something like this: https://github.com/zuckschwerdt/rtl_433/commit/a2cd9b83ba5909aa3602148f2dc1b8c39b581c8b

The r_cfg is a catch-all collection of all the former global variables. It needs to be sorted, yes. Good ideas on how to separate cfg/state/demod, thanks.
I'm in the middle of refactoring rtl_433.c into a proper api, so this cleanup has to wait -- but not long.

zuckschwerdt:
Mostly complete refactor to list_t is in #929

zuckschwerdt:
Thanks for pointing to this, a broader enhancement went in with #929

PR #928, opened 2018-12-17
Free output files of data printer modules

winterrace:
The output files should be closed.

This isn't too important for now, since the program will shortly be shutting down anyway.
But it will be important with rtl_433's  progress towards a lib.

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -470,6 +470,9 @@ static void data_output_json_free(data_output_t *output)
     if (!output)
         return;
 
+    if (output->file != stdout)
+        fclose(output->file);
+
     free(output);
 }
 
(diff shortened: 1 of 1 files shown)


winterrace:
By the way, in my working state I splitted the printers to different files:

- data.c (the generic stuff)
- data_printer_csv.c
- data_printer_json.c
- data_printer_kv.c
- data_printer_udp.c

In my eyes it is much more readable, so you could consider the same for the main branch.
I can place another PR if you agree.

(During this modularization I also splitted some shared functions shared between different modules, which might cause problems sooner or later)

merbanan:
Um, can we apply this patch? It looks good to me.

zuckschwerdt:
The data module didn't open the file, it shouldn't close it. I'll merge this (low-prio) for attribution once it's clear how to do it properly.

Issue #377, opened 2016-05-14
What do do with "OOK_PWM_D"? 


khkremer:
I have old code that was able to receive data from an Acurite 00606TX temperature sensor. However, my code uses the OOK_PWM_D demodulator. How can I port this to the current version of rt._433? Is there a comparable demodulator that I can just swap in, or do I have to go back to square one to figure out what the data is like? I used to use this device configuration: 

```
r_device acurite606TX = {
    /* .id             = */ 10,
    /* .name           = */ "Acurite 606TX",
    /* .modulation     = */ OOK_PWM_D,
    /* .short_limit    = */ 875,
    /* .long_limit     = */ 1750,
    /* .reset_limit    = */ 3750,
    /* .json_callback  = */ &acurite_606TX_callback,
    /* .json_callback  = */ // &debug_callback,
    /* .disabled       = */ 0,
};
```

Thanks.

merbanan:
Answer on mailing list.

PR #929, opened 2018-12-18
Add generic list type

zuckschwerdt:
Adds a resizing generic list type and replaces fixed length list

### DIFF
--- a/include/list.h
+++ b/include/list.h
@@ -1,0 +1,28 @@
+/**
+ * Generic list
+ *
+ * Copyright (C) 2018 Christian Zuckschwerdt
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ */
+
+#include <stddef.h>
+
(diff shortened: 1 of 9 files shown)


winterrace:
Looks well, compiles well, runs well :)
My vote is yours :)

klohner:
Under MinGW-w64, getting `Registered zu out of 96 device decoding protocols` in output.  MinGW-w64 doesn't seem to like `%zu` format in `fprintf(stderr, "Registered %zu out of %d device decoding protocols",` by default.

Possible remediation tips here: (https://stackoverflow.com/questions/25670522/c-program-shows-zu-after-conversion-to-windows)

klohner:
As explained at [#377 C library bug: Modifiers %jd %ju %zu not working](https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/), I was able to add `#define __USE_MINGW_ANSI_STDIO 1` before including stdio.h to mitigate this.

klohner:
Perhaps not?  I just made a fresh build in Windows and was getting the `Registered zu out of 96 device decoding protocols`

zuckschwerdt:
Of course you are right :) https://github.com/merbanan/rtl_433/blob/master/src/rtl_433.c#L1543 I'll fix that.

PR #932, opened 2018-12-22
Use local_time_str() for syslog-printer

gvanem:
Instead of having calls to `gmtime_x()` and `strftime()` in several places, 
use the `src/util.c` function `local_time_str()` to obtain the local time string.

I'm not sure how the `data-test` program then should be linked. It would need `src/util.c`.

The syslog message received in my syslog-client now looks like:
```
2018.22.12-16:27:53 127.0.0.1, 1 2018-12-22 16:27:53 Intel-I7 rtl_433 - - - {"time":"2018-12-22 16:27:51.789813", ...
```
I.e. no longer UTC; all times are local. Is this okay? Or maybe use `mg_gmt_time_string()`.

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -53,6 +53,7 @@
 #include "term_ctl.h"
 #include "abuf.h"
 
+#include "util.h"
 #include "data.h"
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Looks good. I won't have time to test this for a few days ;) though.

zuckschwerdt:
Good idea but I'd rather work towards isolated (library like) units. Also `local_time_str()` is for presentation and might have a different format than required for syslog (we just recently changed local_time_str and might do that again).
We should look into this again when the work towards a stable API is done.

PR #935, opened 2018-12-23
Print to terminal with colour-codes inlined (1)

gvanem:
Prototypes for added term_ctrl.c functions.
The diff for term_ctrl.c is in another PR 

(since I fails to figure out here on Github how to add both into the same PR).


zuckschwerdt:
Great! Thanks! I won't get to this for a few days, sorry.

gvanem:
Not sure what you mean, but maybe both ways. 

My idea was that codes like `~1` could be used every-where. Not just in `usage()`.

PR #936, opened 2018-12-23
Print to terminal with colour-codes inlined (2)

gvanem:
The diff for allowing an inline print of text and colours.
My idea was to change the `usage()` screen into something more readable like this:
rtl_433-help-2

Accomplished by:
```c
   void *term = term_init(stderr);
   term_printf(term, "\n  ~1General options~0:\n");
   term_puts(term,
              "    ~2-V~0 Output the version string and exit.\n"
              "    ~2-v~0 Increase verbosity (can be used multiple times).\n"
              "       ~2-v~0: simple verbosity, ~2-vv~0: verbose decoders, ~2-vvv~0: debug decoders, ~2-vvvv~0: trace decoding).\n"
              "    ~2-c <path>~0 Read config options from a file.\n");
....
```
The actual colours for index 2 etc. could come from the *config file*?

### DIFF
--- a/src/term_ctl.c
+++ b/src/term_ctl.c
@@ -10,6 +10,7 @@
  */
 
 #include <stdio.h>
+#include <stdarg.h>
 #ifndef _WIN32
 #include <unistd.h>
 #include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


winterrace:
Agreed, looks better.

However, I'd prefer to be able to (de)activate coloring at least at compile time. I redirect text output to a GUI element (ListView log) which does not support color codes.

I can live with the existing coloring in the key value printed, though. I use my own output printer as interface to the GUI.

zuckschwerdt:
Yes, maybe a setting `color=auto|yes|no` would be good. Currently you'll only get color on a plain unredirected TTY.

zuckschwerdt:
Yes, output to stdout should only come from a configured data printer or a dumper. No random printf *should* exist.

PR #941, opened 2019-01-01
Add RTHN129

0BuRner:
Oregon Scientifc RTHN129 (same as THN129 but with Real-time clock). More info on this gist: https://gist.github.com/0BuRner/7efe536c430a0d49fc84a839e0f85891


0BuRner:
As you can see in the linked gist, I had to mask the first "value" of the sensor_id because its a cycling value that change every request and therefore could not be considered a part of the sensor_id.
Looking at the code, it seems some other models act the same (e.g. ID_RTGN129, ID_RTGN318, ...)

zuckschwerdt:
Could you PR some test samples to rtl_433_tests?

0BuRner:
Do you want some tests about the file I edited or about the data I analyzed in the gist ? Do you have any example of what you're expecting ?

zuckschwerdt:
Something for the CI regression tests. Like this https://github.com/merbanan/rtl_433_tests/tree/master/tests/oregon_scientific/thn129

0BuRner:
I didn't know about this side project containing test data nor the "-r" command line parameter to generate those files. So I haven't recorded cu8 files for any of my use cases mentioned in my gist and I haven't time to create them all again.
But I can add 2 samples with differents temperature.
@zuckschwerdt will you merge this PR to master if I pull request those 2 samples on the other project ?

zuckschwerdt:
Sure, just post a zip with files here. Doesn't matter if they are cu8, cs8, cs16, cf32 (or complex, complex16u, , complex16s) -- just annotate the center frequency used and the sample rate.
We do need sample files for the regressions tests. I also plan to add regression tests based on bitstrings (like in your gist) but that's not ready yet and won't be as robust.

0BuRner:
[RTHN129_433.92M_250k.zip](https://github.com/merbanan/rtl_433/files/3000603/RTHN129_433.92M_250k.zip)

zuckschwerdt:
Works great. Thanks!

PR #942, opened 2019-01-04
Add support for Silverline doorbells

chhenni:


zuckschwerdt:
Sorry, can you please write this as flex decoder? See e.g. https://github.com/merbanan/rtl_433/blob/master/conf/steffen_switch.conf

zuckschwerdt:
Oh, I see, this is a x1527 transmission with tristate bits and you want to map those to simple bits. Support for that is planned for the flex decoder. Just use a comment how this needs to be changed for now.

chhenni:
@zuckschwerdt, Just so i dont make a fool out of me: You want a comment in the implementation here, stating how I think the setup for the flex decoder should be?

zuckschwerdt:
Well, good you ask. No, I think this device is better implemented as flex decoder than C-code. A few lines in a config file are more hackable (timings, id-codes, …) then another fixed doorbell device decoder ;)

chhenni:
Well, I had not seen the flex decoders until you posted the comment. I worked out of this page: github.com/merbanan/rtl_433/wiki/Adding-a-new-remote-device
I do agree on the flex decoders being a cleaner solution. 

But you said: `Just use a comment how this needs to be changed for now.`
Where do you want this? 
Should I rewrite to use a flex decoder, and then add a remark stating how to decode the values properly when support is added. Or do you mean that I should in the current pullrequest write down the setup for when support is added(to remove it at that point)

zuckschwerdt:
The Wiki needs updates badly, sorry.

Yes, just a note for me in the .conf that the `get=` isn't what you want and that I need to add a tristate decoding. But we can leave this PR open for about a week and I'll add the tristate logic.

zuckschwerdt:
The flex spec now supports selecting random bits with a getter mask. E.g.:
If like in this x1527 tristate example every odd bit is `1` and you want to get 6 bits "channel" from the first 12 bits (discarding every odd bit) use:
```
get=channel:@1:555
```
Effectively this is 11 bits (starting with 1 bit offset to discard the first bit) because the mask is always aligned on the first set bit (i.e. 555, aaa, 000555, … are identical).

Imagine the 25 bits of the transmission are:
```
1?1?1?1?1?1?1?1?1?1?1?1?1
```
The mask `555` is
```
0000 0101 0101 0101
```
aligned to the first set bit
```
10101010101
```
offset 1 bit (`@1`) it selects like this:
```
1?1?1?1?1?1?1?1?1?1?1?1?1
 10101010101
->
 ? ? ? ? ? ?
```
resulting in that 6 bit value `??????`.

The same way 3 bits "sound" is `get=sound:@17:15` (where `15` here is hex for the 5-bit pattern `10101` -- or use `a8` if you want to left align the pattern in the byte there).

zuckschwerdt:
Automated test using the rtl_433_tests repo are not supported. Maybe we could tweak that. Not sure how to properly integrate things :/

PR #954, opened 2019-01-11
Add XDG_CONFIG_HOME compatability 

tomswartz07:
## Background
The currently implemented config dir search paths do not follow Linux XDG standards.

## Implementation
This PR enables the program to search in the standard `$(HOME)/.config/` path, according to [XDG Standards](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

This PR simply adds an additional path on top of the currently defined config dirs, so there is little chance for regression if the other paths were used.

------ 

Thanks for a great program! Hope to continue contributing where possible in the future!

### DIFF
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -12,6 +12,7 @@
 #
 # If no -c option is given the first found of this list will be loaded:
 # - ./rtl_433.conf
+# - ~/.config/rtl_433/rtl_433.conf
 # - ~/.rtl_433/rtl_433.conf
 # - /usr/local/etc/rtl_433.conf
 # - /etc/rtl_433.conf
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
Thanks. Looks good. You can remove/replace the `~/.rtl_433/rtl_433.conf` path. Better get it right now. Also expand `XDG_CONFIG_HOME` or even add a search for `XDG_CONFIG_DIRS` if you like.

tomswartz07:
@zuckschwerdt Done and done. Dropped the old `~/.rtl_433...` path and replaced it with a check for the `XDG_CONFIG_HOME` var, failing back to the standard location if it's not set.

Thanks!

tomswartz07:
Ah, that's a much more clean way of approaching it.
I'll implement is as suggested.

PR #956, opened 2019-01-13
fineoffset: removed -400 scale from wh5 fixes #955

### LINKED ISSUE #955: Rosenborg 66796 - invalid temperature
I just bought the Rosenborg 66796 (listed [here](https://www.bilka.dk/elektronik/ure-og-vejrstationer/vejrstationer-og-termometre/rosenborg-traadloes-vejrstation-model-66796/p/100183695) - danish website), and it worked perfectly with this software, EXCEPT that the temperature was measured wrong.

It is detected as a `Fine Offset WH5 sensor`, and from that it gathers that it needs to offset the temperature reading by `temp -= 400`, which produces the invalid temperature.

https://github.com/merbanan/rtl_433/blob/d09827abe4d0065c5bb3a892936c1d315481a5ea/src/devices/fineoffset.c#L93-L96

Commenting out that line, fixes the problem..

I don't know if this is a "one off" issue, but it might be relevant to just remove that code, as it does not provide the original function anymore.

EDIT: I made a PR for this.

eyJhb:
This commit will fix #955

### DIFF
--- a/src/devices/fineoffset.c
+++ b/src/devices/fineoffset.c
@@ -90,9 +90,6 @@ static int fineoffset_WH2_callback(r_device *decoder, bitbuffer_t *bitbuffer) {
             temp &= 0x7FF;	// remove sign bit
             temp = -temp;	// reverse magnitude
         }
-    } else { // WH5
-        // The temperature is unsigned offset by 40 C and scaled by 10
-        temp -= 400;
     }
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
This will break WH5 devices.

eyJhb:
Different issue.

PR #959, opened 2019-01-15
Draft decoder for IKEA Sparsnäs Energy Meter Monitor

slundell:
This is a draft (for review) of my decoder for the IKEA Sparsnäs Energy Meter Monitor. It works great, but there are a few problems with the implementation. 

The PD_MIN_PULSE_SAMPLES needs to be lowered. 5 is the highest value that gets any reading. No readings seems to be missed.

The user needs to set IKEA_SPARSNAS_SENSOR_ID and IKEA_SPARSNAS_PULSES_PER_KWH in the source as per the current implementation. I guess this is not the intended way. Any pointers how to handle this would be greatly appriciated.


src/devices/ikea_sparsnas.c
```diff
@@ -0,0 +1,251 @@
...
+    for (i = 0; i < 8; i++) {
+      if (((crcReg & 0x8000) >> 8) ^ (crcData & 0x80))
+        crcReg = (crcReg << 1) ^ 0x8005;
+      else
+        crcReg = (crcReg << 1);
+      crcData <<= 1;
+    }
+  }
+  return crcReg;
+}
```
merbanan: We have our own crc functions. So eventually this needs to be replaced.

src/devices/ikea_sparsnas.c
```diff
@@ -0,0 +1,251 @@
...
+amazing repository of the IKEA Sparsnäs. Everything is studied with greay
+detail. Many thanks to kodarn!
+
+*/
+
+
+
+#define IKEA_SPARSNAS_SENSOR_ID 617633
+#define IKEA_SPARSNAS_PULSES_PER_KWH 1000
+
```
merbanan: We need to figure out how to pass along decoder parameters. This needs to be in a config file.

src/devices/ikea_sparsnas.c
```diff
@@ -0,0 +1,251 @@
...
+static int ikea_sparsnas_callback(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+
+    if (decoder->verbose > 1)
+        decoder_output_bitbufferf(decoder, bitbuffer, "IKEA Sparsnäs:");
+  
+
+    uint8_t buffer[IKEA_SPARSNAS_MESSAGE_BYTELEN];
+
```
merbanan: Declare the variables first and then add an early termination condition based on the amount of bits and the preamble check. Then add the crc check. If that succeeds continue with decoding.

src/devices/ikea_sparsnas.c
zuckschwerdt: Perhaps we can think up a clean way to unify this, the recent WH5 quirk, and flex specs. Something like parametric decoders.

src/devices/ikea_sparsnas.c
```diff
@@ -0,0 +1,251 @@
...
+    data_t *data;
+    data = data_make(
+        "model",         "Model",               DATA_STRING, "IKEA Sparsnäs Energy Meter Monitor",
+        "id",            "Sensor ID",           DATA_INT, rcv_sensor_id,
+        "sequence",      "Sequence Number",     DATA_INT, sequence_number,
+        "battery",       "Battery",             DATA_FORMAT, "%d%%", DATA_INT, battery,
+        "cumulative_kWh", "Cumulative kWh",     DATA_FORMAT, "%7.3fkWh", DATA_DOUBLE,  cumulative_kWh,
+        "effect",        "Effect",              DATA_FORMAT, "%dW", DATA_INT,  effect,
+        "pulses",        "Pulses",              DATA_INT,  pulses,
+        "mode",          "Mode",                DATA_INT, mode,
```
merbanan: Add mic = CRC here. Look in other decoders that have mic.

merbanan:
All in all it looks real good. Maybe reduce the amount of empty lines also.

src/devices/ikea_sparsnas.c
slundell: OK. Changed to crc16 in utils.c and verified that it works.

src/devices/ikea_sparsnas.c
```diff
@@ -0,0 +1,251 @@
...
+on the front of the meter. This value goes into IKEA_SPARSNAS_PULSES_PER_KWH 
+in this file. The sender also has a unique ID which is used in the encryption
+key, hence it is needed here to decrypt the data. The sender ID is on a sticker
+in the battery compartment. There are three groups of three digits there. The
+last six digits are your sender ID. Eg "400 617 633" gives you the sender id 
+617633. This number goes into IKEA_SPARSNAS_SENSOR_ID in this file.
+
+
+The data is sent using GFSK modulation. It requires PD_MIN_PULSE_SAMPLES in 
+pulse_detect.h to be lowered to 5 to be able to demodulate at 250kS/s. The
```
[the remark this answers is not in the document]
slundell: Changed

src/devices/ikea_sparsnas.c
slundell: I am not entirely sure what you mean here. 

Is it that you want variables declared at the top of the function? 

The other things are done in the order they can be done.
1. Preamble checking and extraction
2. CRC Check (needs extracted packet without preamble)
3. Decryption (No need to decrypt if CRC fails)
4. Additional integrity checks (Needs decrypted packet)
5. Data extraction and interpretation (Needs a decrypted and valid packet)

src/devices/ikea_sparsnas.c
slundell: OK

src/devices/ikea_sparsnas.c
merbanan: The point is that these checks are done for every fsk signal that is detected. So things that are not matching needs to fail fast with easy checks.

src/devices/ikea_sparsnas.c
zuckschwerdt: I already snuck that in when refactoring the decoders ;)
There is a private context (flex currently, but good for all parametric decoders): https://github.com/merbanan/rtl_433/blob/master/include/rtl_433_devices.h#L151
Usage example is https://github.com/merbanan/rtl_433/blob/master/src/devices/flex.c#L104
We now "just" need a way to declare and instantiate parametric decoders in a general and enumerable way.

src/devices/ikea_sparsnas.c
merbanan: The if (decoder->verbose > 1) block should be moved after the bitbuffer size check and preamble search. It looks like a Sparsnäs message has between 180 and 200 bits. That means that we only need to search for the preamble if we get that amount of bits from the demodulator. And the preamble search could be reduced to 0xAA, 0xD2, 0x01 for cases where parts of the preamble is missing.

zuckschwerdt:
@slundell the idea currently is to make this some kind of parametric decoder (generator). It will take parameters and create a decoder instance for each set of parameters. Just like the flex decoder.

merbanan:
@slundell I think we can limit the decoder to one sensor. Otherwise we need to support lists and what not. 

I think brute forcing the keys would be more interesting then lists. With the new code it is easier to see the actual relations from the key to the decoded data. But we still need a way to configure constant each pulse is equal to.

Anyway lets focus on getting the code merged, then we can start tinkering some more.

slundell:
I may have a look at brute forcing. 16M combinations IIRC, but might be possible to reduce some more. One way to circumvent the need for pulse configuration is to add attributes for each of the common configs
eg
```
{
...
  cumulative_kwh_500: 20.0,
  cumulative_kwh_1000: 10.0,
  cumulative_kwh_2000: 5.0 
...
}
```

merbanan:
The pulse configuration should match to the revision of the sensor. Might be possible to infer that from the crypto key.

zuckschwerdt:
@merbanan I think the sensor will be stuck to the S0 output of an existing power meter, and we can't know how that is configured. Though 1000 is common and there won't be that many other values (just 500 and 2000 maybe).

slundell:
This fifth byte is sent as in an init package when the sender is first powered on. It would be best if we did not need to reset the sender as that might affect running displays, or that we might not have physical access to the sender.

zuckschwerdt:
I have a good idea now how to generalize the concept of the flex decoder to support arbitrary parametric decoders. That should cover the case where one or multiple keys are known.

For unknown keys the state and decoding attempts should be done in some script. Always think of rtl_433 as stateless. It could restart any time and will never derive values in time (e.g. most rain counters are absolute and not per unit time -- a downstream client has to calculate that value).

slundell:
I included brute-forcing of the sender ID if it is not known. This takes 2.5ms on a Raspberry Pi 3+ in release build. At least for my key, it is possible that it takes more or less time on another key. It is not guaranteed that the correct key is found. There are usually 1-4 keys that that are able to decrypt the packet. The byte of the key that is not used to encrypt the sensor ID is used to decrypt the battery level. In my case his resulted in a battery level higher than 100% in for all invalid keys. It is of course possible that that is not the case for all senders, and multiple keys might be found. Only the first key found is used for decryption so there might be cases where the packet cannot be decrypted. This could be improved further if needed.  There might also be situations where two different sensors are sending data that will confuse this decoder. I have not been able to test that.

The pulse count setting is fixed at 1000 imp/kWh. It should be trivial to convert in the users application code, if other pulse count settings are used. 

@zuckschwerdt If user settings are used, these can plug in to the variables `ikea_sparsnas_sensor_id `and `ikea_sparsnas_pulses_per_kwh `and the decoder will use them. Then, no key-cracking or user application code changes are needed.

bilbolodz:
It's not connected with rtl_433 but I will ask: where could I buy these meter? I was trying to google it but I've only found manual at IKEA web page. Strange......

zuckschwerdt:
Settings per decoder are on their way. I'll have the patch this weekend.

slundell:
> It's not connected with rtl_433 but I will ask: where could I buy these meter? I was trying to google it but I've only found manual at IKEA web page. Strange......
I don't know really. They seem to be out of stock everywhere. I don't know if this is because of supply issues or that they are being phased out. Some lucky IKEA visitors have been able to pick up a few and are selling them on ebay/blocket.se etc.

molobrakos:
> It's not connected with rtl_433 but I will ask: where could I buy these meter? I was trying to google it but I've only found manual at IKEA web page. Strange......

I believe that IKEA Sparsnäs was a re-brandeded version of a previous white label version of this product: https://lifesmartnordics.com/products/eliq-electricity-meter-sensor If you are lucky they are still compatible, but I don't know that.

zuckschwerdt:
You can test that but nothing needs to be changed right away. I can just add the changes to this decoder after it is merged :)

slundell:
fixed

zuckschwerdt:
It's important to note that this patch currently changes the `PD_MIN_PULSE_SAMPLES` default (analogous to removing a low-pass filter). It's probably better to sample at 1024k rate and not rely on that override.

zuckschwerdt:
Can you grab a sample file to see what goes wrong in the decoding?

slundell:
Added a few at different sample rates to https://github.com/slundell/rtl_433_tests/tree/master/tests/IKEA_Sparsnas/samples/unidentified

None of these are correctly identified. The sample analyser option sees it as one long pulse.
This is with PD_MIN_PULSE_SAMPLES = 10

According to my very limited knowledge, the sample rate should not affect the decoding, as long as it is high enough to convey the message?

zuckschwerdt:
Yes, could you try to lower the gain? The signal has strange clipping.
g002_868m_1024k

merbanan:
@slundell well the thing with the current fsk demodulator is that it generic. You can feed it random signals and it will automatically just work. To make this magic work you need to feed it enough samples. So while in theory the signal message is recoverable with fewer samples per second we some times need to feed it more so we can keep the generic demodulation property.

slundell:
@zuckschwerdt I added samples at gain 21 and different sample rates. 21 is the lowest gain where the signal is received (I changed antenna yestarday). I had a quick glance through sigrok and noticed that sometimes Q is flat.

@merbanan OK, so there is no reason as to why it should work at 250k but not at 1024k?

zuckschwerdt:
If Q seems flat that's likely the scaling. It's sometime auto-set to something large.

zuckschwerdt:
The problem with the higher sample rates is increased accuracy, I guess. The FSK frequencies are estimated closer and don't make the fixed minimum separation. If you change `FSK_DEFAULT_FM_DELTA` to at most `4000` it will work. Most of these values work pretty well but might be just *something that works*. I might need to evaluate performace and really tune the constants. I already planed a test-suite to measure the OOK and FSK performance in precise terms of RSSI, SNR dB and such.

slundell:
I just wanted to let you know that with the recent changes I am now able to receive and demod reliably in 250k. That was previously only possible in 1024k. 

I suspect this has to do with the recent sample_rate - related commits by @zuckschwerdt . 

I still need a PD_MIN_PULSE_SAMPLES of max 5.

zuckschwerdt:
Great to hear! I going to try and improve this further, but it's going to take some time.

slundell:
Updated to current version. What needs to be done to get this merged?

include/pulse_detect.h
```diff
@@ -1,13 +1,3 @@
-/** @file
-    Pulse detection functions.
-
-    Copyright (C) 2015 Tommy Vestermark
-
-    This program is free software; you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation; either version 2 of the License, or
-    (at your option) any later version.
-*/
```
merbanan: This looks like a stray removal. Put it back.

include/pulse_detect.h
```diff
@@ -18,7 +8,7 @@
 
 #define PD_MAX_PULSES 1200      // Maximum number of pulses before forcing End Of Package
 #define PD_MIN_PULSES 16        // Minimum number of pulses before declaring a proper package
-#define PD_MIN_PULSE_SAMPLES 10 // Minimum number of samples in a pulse for proper detection
+#define PD_MIN_PULSE_SAMPLES 5 // Minimum number of samples in a pulse for proper detection
```
merbanan: This change I need a confirm by @zuckschwerdt if it is ok to add. I am sort of for doing that as a stop gap solution for now. If @zuckschwerdt doesn't agree the patch can go in without that change. We then maybe change PD_MIN_PULSE_SAMPLES to an option so that we can get out of the box support of this sensor.

src/devices/ikea_sparsnas.c
```diff
@@ -0,0 +1,305 @@
...
+    uint8_t battery =  decrypted[17];
+    double watt = effect * 24.;
+    uint8_t mode = decrypted[4]^0x0f;
+
+    if(mode == 1){     //Note that mode cycles between 0-3 when you first put in the batteries in
+      watt = ((3600000.0 / ikea_sparsnas_pulses_per_kwh) * 1024.0) / effect;
+    } else if (mode == 0 ) { // special mode for low power usage
+      watt = effect * 0.24 / ikea_sparsnas_pulses_per_kwh;
+    }
+    double cumulative_kWh = ((double)pulses) / ((double)ikea_sparsnas_pulses_per_kwh);
```
merbanan: All doubles should be changed to float.

merbanan:
The current branch has conflicts also. Anyway fix the remaining comments and then I'll merge the code one way or the other (within 5 days). The patch has been lingering for way to long.

include/pulse_detect.h
zuckschwerdt: Yes, that number is a bit fragile. Effectivly a low pass filter for OOK and FSK demod. But it's not normalized to the sample rate. In effect it demands a minimum pulse length of 40µs. This is okay for the fastest timing of 50 µs (mostly TPMS) at 250k. But it's also the reason why IBIS beacon (30 µs) and M-Bus (10 µs, 15 µs…) need at least 1M to work.
I'll run some test to probe the whole range of e.g. 1-15 and see how fragile it really is. Ideally we should just use (and tune) the existing IIR low pass.

zuckschwerdt: Already at `9` there are two regressions (fineoffset_wh25/01/gfile005.cu8, honda_remote/01/hondacar-unlock.cu8) and at `5` a mix of 24 tests fail (alecto, current cost, oil gauge, tpms).
We really need to tune the low pass filter or something else before we can change this.

Doesn't this decoder work with higher sample rates and no change to this constant?

slundell:
OK. I'll fix these comments. I'm sorry I messed up the merge.

src/devices/ikea_sparsnas.c
slundell: Why so? I still need a double in the end to pass it in DATA_DOUBLE?

merbanan: The point is to use the type with enough precision for the job. float has 4 bytes of storage and double usually uses 8 so by using float we save a few bytes and on performance limited platforms (if there are any left) the double emulation is assumed to be more costly compared to the float version. DATA_DOUBLE should be changed to float.

merbanan:
Ok, I merged the decoder but it is disabled by default. It needs the new fsk demod (which still is vaporware). Maybe we can add an option to set the PD_MIN_PULSE_SAMPLES from the command line, but I will focus my energy towards a better fsk demod.

merbanan:
I'll close this PR now. Changes have to be added in new tickets.

merbanan:
@slundell can you try and see if this is decoded by default now? -R 130 should be needed.

PR #975, opened 2019-02-04
Add Alecto WS-1200 v1.0/v2.0 decoders

zuckschwerdt:
WIP, piggy-backed on Fine Offset WH0530. Needs DCF decoding and collision removal (Fine Offset WH3080).


merbanan:
Looks ok to merge.

PR #980, opened 2019-02-07
Fix humidity value bellow 16 on Nexus sensor

pm-cz:
The code was not detecting Auriol 2013/Nexus sensor humidity if it was 0xf (15) or lower due to only checking upper nibble.

Example data for the scenario:
> [00] { 0}                : 
> [01] {36} ac 10 2e f0 d0 : 10101100 00010000 00101110 11110000 1101
> [02] {36} ac 10 2e f0 d0 : 10101100 00010000 00101110 11110000 1101
> [03] {36} ac 10 2e f0 d0 : 10101100 00010000 00101110 11110000 1101
> ...

For comparison, Nexus(-like) sensor in neighbourhood without humidity sensor has last nibble set to 0:
> [00] {36} 91 8f e3 f0 00 : 10010001 10001111 11100011 11110000 0000
> [01] {36} 91 8f e3 f0 00 : 10010001 10001111 11100011 11110000 0000
> ...

### DIFF
--- a/src/devices/nexus.c
+++ b/src/devices/nexus.c
@@ -74,7 +74,7 @@ static int nexus_callback(r_device *decoder, bitbuffer_t *bitbuffer) {
         humidity = (uint8_t)(((bb[r][3]&0x0F)<<4)|(bb[r][4]>>4));
 
         // Thermo
-        if (bb[r][3] == 0xF0 && (bb[r][4] >> 4) == 0x0) {
+        if (humidity == 0x00) {
         data = data_make(
                 "model",         "",            DATA_STRING, "Nexus Temperature",
                 "id",            "House Code",  DATA_INT, id,
--- a/src/devices/nexus.c
+++ b/src/devices/nexus.c
@@ -74,7 +74,7 @@ static int nexus_callback(r_device *decoder, bitbuffer_t *bitbuffer) {
         humidity = (uint8_t)(((bb[r][3]&0x0F)<<4)|(bb[r][4]>>4));
 
         // Thermo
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Isn't that code actually checking for the value labeled "const"? Does "const" related to the fact if it is a T/H or T only transmission?

pm-cz:
Actually, both are valid questions. I have tried to modify the code the least possible way, but nibble 6 is always 0xf (that's the check on line 56). Nibbles 7 and 8 (not 6 and 7 as written wrongly in the comment) contain humidity value (the formula is OK). But on some sensors instead of humidity, there is 0x00 (which actually maps to 20% in receiver).

But the code only check the upper 4 bits being 0, so if humidity is 0-15, it does not take it into account.

Another possibility would be to just re-write the line to:
`if (humidity == 0x0) {`

instead of my suggested modification. But the result should be the same.

zuckschwerdt:
I meant that `bb[r][3] == 0xF0` should always be == 0xF0 as that is the "const" nibble. Your `&& (bb[r][4] >> 4` checks the lower portion of the humidity. So 16%, 32%, ... would fail here, right?
I guess `(humidity == 0)` shows the intent best.

pm-cz:
Yes, it is "const" if no humidity is set. No, it will not fail. The data is always 0xF H L. The original check just checked   0xF H nibbles, but if H==0 and L != 0, it still went to the "no humidity" section.

zuckschwerdt:
Thanks for figuring this out!

PR #985, opened 2019-02-16
Enhance Kedsum, S3318, Esperanza with MIC

zuckschwerdt:
Breaking change: models changed to "Kedsum-TH", "Conrad-S3318P", "Esperanza-EWS".
Also the Esperanza had some bits wrong and showed a slightly wrong temperature.

### DIFF
--- a/src/devices/kedsum.c
+++ b/src/devices/kedsum.c
@@ -1,31 +1,42 @@
-/* Kedsum temperature and humidity sensor (http://amzn.to/25IXeng)
-   My models transmit at a bit lower freq. Around ~433.71 Mhz
+/** @file
+    Kedsum temperature and humidity sensor (http://amzn.to/25IXeng).
+    My models transmit at a bit lower freq. of around 433.71 Mhz.
+    Also NC-7415 from Pearl.
 
-   Copyright (C) 2016 John Lifsey
+    Copyright (C) 2016 John Lifsey
+    Enhanced (C) 2019 Christian W. Zuckschwerdt <zany@triq.net>
 
-   This program is free software; you can redistribute it and/or modify
-   it under the terms of the GNU General Public License version 3 as
-   published by the Free Software Foundation.
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 2 of the License, or
+    (at your option) any later version.
+*/
+/**
+Largely the same as esperanza_ews, s3318p.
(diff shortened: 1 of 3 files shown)


src/devices/kedsum.c
```diff
@@ -41,44 +51,33 @@ static int kedsum_callback(r_device *decoder, bitbuffer_t *bitbuffer) {
...
-            "channel",       "Channel",     DATA_INT, channel,
-            "temperature_F", "Temperature", DATA_FORMAT, "%.02f F", DATA_DOUBLE, temperature_f,
-            "humidity",      "Humidity",    DATA_FORMAT, "%u %%", DATA_INT, humidity,
+            "model",            "",             DATA_STRING, "Kedsum-TH",
+            "id",               "ID",           DATA_INT, id,
+            "channel",          "Channel",      DATA_INT, channel,
+            "battery",          "Battery",      DATA_STRING, battery_str,
+            "temperature_F",    "Temperature",  DATA_FORMAT, "%.02f F", DATA_DOUBLE, temp_f,
+            "humidity",         "Humidity",     DATA_FORMAT, "%u %%", DATA_INT, humidity,
+            "mic",              "Integrity",    DATA_STRING, "CRC",
```
merbanan: s/CRC/LFSR/ ? It is technically not a crc per my understanding.

merbanan:
All good.

src/devices/kedsum.c
zuckschwerdt: Not technically a proper "CRC". But it's of comparable strength. I'd say the difference is that a CRC is strictly linear an a general LFSR is not. This one in particular looks linear though :) It seems a "degenerate" (fixed key) form of the common digest type -- or maybe I'm just not seeing how this is perhaps a broken CRC-4. I definitly going to investigate the algorithm further.
Currently the MIC has values of `PARITY < CHECKSUM < CRC` relating to strength. Perhaps we could suffix a width, e.g. `PARITY-1, PARITY-2, SUM-8 < CRC-4, CRC-8, CRC-16` types should cover it.

merbanan: I leave it up to you to decide. It is also fine as it is. The current model is sort of broken as some CRC's are 4 bits and others are 16 bits. The difference is significant with regards to that.

zuckschwerdt: Ah well, turns out it really is a CRC-4. But the last 4 bits are only folded in, not processed. I can imagine the faulty CRC lagging 4 bits in the processing. Could also be by design but that would be a poor choice.

PR #993, opened 2019-02-26
Include <stdlib.h> ahead of "util.h"

gvanem:
Since <stdlib.h> currently is NOT included in "util.h", this triggers this warning using MSVC:
  f:\ProgramFiler-x86\WindowsKits\Include\10.0.18309.0\ucrt\stdlib.h(1289): warning C4005: 
  'max': macro redefinition (compiling source file src/util.c)
(same for the 'min()' macro).

Fix by including <stdlib,h> in "util.h" AND including "util.h" last.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -9,10 +9,9 @@
     (at your option) any later version.
 */
 
-#include "util.h"
-#include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include "util.h"
 
 uint8_t reverse8(uint8_t x)
 {


zuckschwerdt:
Ah, that bug again. We actually don't want the min/max macros. I thought `-DNOMINMAX` in CMakeLists.txt got rid of those. Can you investigate that?
Reason to include the own header first is to check that the header is self-sufficient.

gvanem:
AFAICS, a `-DNOMINMAX` is important only when a `#include <windows.h>` is used. 
`src/util.c` doesn't include `windows.h`.

My simple change to swap some `.h`-files should fix the above warning. 
Don't you agree that compiler/system headers should IMHO *always* be included before RTL_433's `include/*.h` files?

zuckschwerdt:
Public-to-private is just one way to sort header includes. If you include private headers first you get a free check if they are complete (e.g. not missing something like stdint.h).

More interestingly: where are those min/max defines coming from? Does stdlib.h define them? Is there a way to stop that?

gvanem:
> Does stdlib.h define them? 

For MSVC, yes. It does that unconditionally for C:
```c
// Non-ANSI Names for Compatibility
//
//-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
#if _CRT_INTERNAL_NONSTDC_NAMES
    #ifndef __cplusplus
        #define max(a,b) (((a) > (b)) ? (a) : (b))
        #define min(a,b) (((a) < (b)) ? (a) : (b))
    #endif
```
> Is there a way to stop that?

Not that I know of. Unless `__STDC__` is *not* defined.

But with MinGW, it's easier ways to stop it. But haven't tried that.

zuckschwerdt:
Perhaps best if we rename our own min/max definitions and forget all about some header somewhere defining those.
Does https://github.com/merbanan/rtl_433/tree/fix-minmax fix the issue? It at least makes it more visible that there is a macro …which might evaluate the arguments twice.

gvanem:
How do I get a .diff for this?

zuckschwerdt:
The commit is 77568f2, the diff is at https://github.com/merbanan/rtl_433/commit/77568f2.diff

gvanem:
That .diff work just fine. Feel free to close my PR.

zuckschwerdt:
Thanks for debugging this!

Issue #943, opened 2019-01-05
Support for (generic) FS20 sensors


MarcusWernicke:
I would like to request support for FS20 sensors (motion sensors, door contacts, smoke detectors).

(The ELV device supported (for weather stations) references fs20.c, so it's probably based on a generic decoder. )

Some samples recorded via -s unknown from a smoke detector and a motion detector:
https://drive.google.com/file/d/16xgnUE9xuzWl8uKP5LWzCYcRsYqeR-Ty/view?usp=sharing

Project implementing an FS20 decoder in Python:
https://github.com/eT0M/rtl_sdr_FS20_decoder

zuckschwerdt:
Do you know how the ELV EM-1000 relates to the FS20 family? Is there anyone around who has the ELV EM-1000?

MarcusWernicke:
> Do you know how the ELV EM-1000 relates to the FS20 family?

I don't think it uses the FS20 protocol (or at least its documentation doesn't mention that). It just seems to be a product made by the same German manufacturer.

PR #999, opened 2019-02-28
Add simple fs20 remote decoder

pusch01:
No one fixed eT0M/rtl_sdr_FS20_decoder to work with Gnuradio 3.7 for years. :(
So I added a simple fs20 decoder to rtl_433.


merbanan:
Hi, we need some samples signals and the indentation is of in a few places. I think some build system changes are also needed to hook up the decoder.

zuckschwerdt:
Whatever uniquely identifies a device should be the key "id", directly following "model".
If you are unsure about the files to modify for the build system try ./maintainer-update.py
Also run `make test` (after a cmake build) for some basic checks.

zuckschwerdt:
Also FS20 has been requested in #943. @MarcusWernicke might be able to help testing.

pusch01:
sorry, I don't use git command line tools. I just downloaded and messed with rtl_433-master.zip from here. Of course, the device must be integrated into build system. I did it with ugly, local changes. I used silvercrest.c as template, because it was the smallest device-code. I simply added it to all build files. This effects remote device id, that you can choose with rtl_433 -R <device>. I hope someone of the main developers can trim the build-system.
Maybe I can generate some fs20 cu8-files at weekend. I don't want to make my own housecode public. 
There is no id key, that identifies a fs20 remote anyway. 
BTW: Now I tested some more original remotes and the reset_limit have to be decreased.

zuckschwerdt:
I've just added a clang-format config to master. That should help to follow code style. You should however wrap the `data_make` call in `/* clang-format off */` `/* clang-format on */` and prettify that yourself. Look e.g. at https://github.com/merbanan/rtl_433/blob/master/src/devices/ambientweather_wh31e.c#L87

Just reading your comment. That's ok then, I'll merge and rewrite your commit in a day or so.

zuckschwerdt:
Please test. Needs `-R 122` to select the protocol.

PR #1009, opened 2019-03-13
File input fixes

wentasah:
Few fixes that allow me to process cf32 files.


zuckschwerdt:
Sorry, I somehow missed your PR and incidentally fixed the CF32 issue already. Thanks for fixing the scale issue.

kpreid:
Could we have a tagged release including the cf32 fix sometime soon? (I'd like to be able to point to a compatible version number and expect it to show up in package managers, for kpreid/shinysdr#128.)

merbanan:
@kpreid 201904 release good enough? I'd like a release before we rework the fsk decode procedure.

kpreid:
That sounds fine.

wentasah:
Out of curiosity, how do you plan to rework FSK decoding? I'm also playing with it to decode signal from my device.

PR #1015, opened 2019-03-20
Add TS-FT002 decoder

zuckschwerdt:
LSB/MSB and nibble swapping are confusing. Needs careful testing from device owner.


merbanan:
I we have signal samples I'd say merge the decoder and wait for bug reports.

zuckschwerdt:
Ok, I'll rebase and merge tomorrow.

PR #1018, opened 2019-03-23
kerui: break out additional fields from "state"

roger-:
This breaks out the following fields from "state": 

*    "motion",
*    "closed",
*    "tamper",
*    "battery_low" / "battery_ok"

Hopefully this isn't too verbose, but it is more consistent with other modules like `dsc.c`.

FYI the following are redundant and could be removed: 

* "state"
* "cmd"


zuckschwerdt:
Thanks. No need to copy the _X-macro voodoo. That's only for compatibility with old expectations on keys. `battery_ok: 1` is a good choice.
I'll have to look through other security devices and see if this is a good general scheme. Perhaps `closed` should be `opened` as that is the alarm state, same as `motion` and `tamper`?

roger-:
Sure, will update the PR.

`opened` does seem like a better choice; I was just trying to be consistent with dsc.c

roger-:
Bump.

merbanan:
Any update on this?

zuckschwerdt:
The patch is ok. The question is: can we agree on a policy wether and where we want

- "cmd": "motion" / "cmd": "open" / "cmd": "close" / "cmd": "tamper" / "cmd": "battery" **OR**
- "motion": 1 / "opened": 1 / "opened": 0 / "tamper": 1 / "battery_ok": 0

That's "event" logging versus "state" logging. Slight problem in the later however, "motion", "tamper", and "battery_ok" will never be reset and that's no good for time series logging.

kpreid:
My 2 cents:

I have an app which displays data from rtl_433 generically in a GUI. It basically assumes what you called the state logging model: it shows the last-received value of every field not coded as a special case.

Hence, my suggestion would be to do what you called "state" logging, but also perhaps have a catalog of field metadata (that can optionally be output on startup), one part of which would communicate "this field implicitly returns to zero/unset/absent", possibly with an expected time-scale. Then tools that want to log events can just note the event, and tools that want a time series can use the suggested time interval, or their own if they have a natural granularity, for synthesizing a return to zero.

Another advantage of this mechanism would be that in a case like "battery_ok", you can use the metadata to handle _both_ sensors that report a boolean status within packets that contain other data (use the latest value) and ones that send them as separate events (use the event-to-time-series mechanism, but the same field name).

zuckschwerdt:
Thanks, I'm also in favor of "state" logging and would attribute a "higher" (thus easier interface) abstraction ("raw" events are translated and grouped to trackable states). E.g. the `battery_ok` case shows this -- we already do this in other sensors.

We should be clear however that most states here only go one way (`battery_ok: 0`) and that might break expectations. We really like to keep decoders free of stateful logic and things like timeouts and defaults would need an adapter or bridge script.

And that's my concern here, the "state"-like output here is really just "events" and that might cause grief with users implicitly expecting more. On the other hand, uniform keys with other sensors is always a goal.

I guess in practice nobody should expect a "motion:0" or "tamper:0" form a regular PIR and "battery_ok" goes along with that, maybe.

kpreid:
In case clarification is needed: I'm proposing that rtl_433 and its decoders would continue to be free of statefulness, but there would be metadata — just static compiled-in information about what the decoders are producing — that external postprocessors can use to know when things are event-like.

Introducing this metadata concept could also provide other useful information, like for example the set of all possible fields for each decoder, or how frequent transmissions are expected to be (timeout for assuming a transmitter has gone offline).

zuckschwerdt:
I see. Yes, we do have the `output_fields` in every decoder now. A way to query those would be nice.

roger-:
Any updates on where we stand with this?

merbanan:
@zuckschwerdt this looks ok to me, can we merge this?

zuckschwerdt:
PR #1384 does something similar, at some point we should normalize this "flag output" style.

PR #1019, opened 2019-03-25
Normalize weather (rain/wind) keys

zuckschwerdt:
This normalizes weather (rain/wind) related keys. The change is selected only with the `newmodel` option.
The new keys are:
- `rain_mm`
- `rain_in`
- `rain_rate_mm_h`
- `rain_rate_in_h`
- `wind_avg_km_h`
- `wind_avg_mi_h`
- `wind_avg_m_s`
- `wind_max_km_h`
- `wind_max_mi_h`
- `wind_max_m_s`
- `wind_dir_deg`

Should "gust_speed_" perhaps be named "wind_gust_" to relate it closer to "wind_speed_"? Not sure, please comment.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -120,7 +120,7 @@ static int acurite_rain_gauge_callback(r_device *decoder, bitbuffer_t *bitbuffer
         data = data_make(
             "model",    "",        DATA_STRING,    _X("Acurite-Rain","Acurite Rain Gauge"),
             "id",        "",        DATA_INT,    id,
-            "rain",     "Total Rain",    DATA_FORMAT,    "%.1f mm", DATA_DOUBLE, total_rain,
+            _X("rain_mm","rain"),     "Total Rain",    DATA_FORMAT,    "%.1f mm", DATA_DOUBLE, total_rain,
             NULL);
 
         decoder_output_data(decoder, data);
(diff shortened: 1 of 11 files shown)


steepleian:
The use of wind together with gust (rather than gust on its own) seems to
be the more prevalent expression in the weather community.

zuckschwerdt:
Thanks! Also nicely groups all wind related keys with a `wind_` prefix. Would it be canonical to use `wind_gust_` with `wind_speed_` or rather `wind_avg_` then?

steepleian:
Not sure. Maybe there are some other weather enthusiasts who can contribute to the discussion.

zuckschwerdt:
Browsing some meteorology forums I get the impression that the measurements would best described as `avg` and `peak` (or `max`). The term "gust" seems to be some specific fast wind event, not the highest value over some period.
Could really use more insight here.

steepleian:
Both the weather data logger (WeeWX) and website template (weather34) I use
have max for wind speed gust as default.

zuckschwerdt:
Okay, `wind_avg_` and `wind_max_` it is!

steepleian:
OK

dgrostoto:
I am currently making domoticz newmodel compliant, during the analysis,
I have found :
       4 "direction_deg" 
      1 "wind_direction"
can you check if it is correct ?

zuckschwerdt:
Thanks! Yes those should be changed. I'll do this right away.

Issue #460, opened 2016-12-18
Incorrect pressure Oregon Scientific BTHGN129 (ID: 5d35)


ROV28:
Hi,

I'm trying to add the Oregon Scientific BTHGN129 (ID: 5d35) sensor to RTL_433, but I.m not getting the correct value for the pressure, if I use the same method as for the BHTR968 [ ((msg[7] & 0x0f) | (msg[8] & 0xf0))+856 ], I get a pressure of 1071mbar, which is impossible:
a) According to the manual the range is 700-1050mbar. Where does die +856 come from? Adding 856 makes it impossible to get a value lower than 856. 
b) 1050mbar is the all-time high in this country :-), at the moment is should be around 1034mbar.

Ok, adjusting the offset could be and quick 'fix' but then the next question arises, a range of 700-1050mbar with a 1mbar resolution is not possible with 8 bit?!

About the protocol I found:

ID Code(s)                           | Message Length (nibbles)
_______________________________________________________
5D35                                    | 21
_______________________________________________________
Nibbles | Contents               | Temp/RH plus Barometer
_______________________________________________________
10..8     | Temperature         | LSD is 0.1 degC
11         | Temperature Sign | Non-zero for negative values
13..12   | Relative Humidity  | Percent
16..14   | Pressure               | Binary (not BCD) in units of 0.01 inHg
17..18   | Unknown
_______________________________________________________

But I assume this table is incorrect: IMO nibble[14] is the 'Comfort' nibble and nibble[18] is the 'Forecast' nibble. And 'inHg', does not match with the 1mbar resolution?!

This is how far I got, a logging:
Nibble[0]: 5 
Nibble[1]: 13 =>d
Nibble[2]: 5 
Nibble[3]: 3   ==> SensorID: 5d35
Nibble[4]: 2   ==> Channel: 2
Nibble[5]: 4 
Nibble[6]: 11   ==> Rolling code: 114
Nibble[7]: 0   ==> Flags 1: 0 
Nibble[8]: 5 
Nibble[9]: 0 
Nibble[10]: 2 
Nibble[11]: 0   ==> Temperature: +20.5 
Nibble[12]: 3 
Nibble[13]: 6   ==> Humidity: 63% 
Nibble[14]: 4   ==> Comfort: 4 
Nibble[15]: 7 
Nibble[16]: 13   ==> Pressure: 215
Nibble[17]: 0 
Nibble[18]: 2   ==> Forecast: 2 
Nibble[19]: 5 
[...]

PR #1023, opened 2019-03-26
Fix BTHGN129 pressure reading

jamaron:


zuckschwerdt:
Are you sure? Can you check the discussion in #460?

jamaron:
This is the first part of the BTHNG129 fixes, i.e. fixing the pressure reading that was the original problem with BTHNG129 driver, as you can see at the beginning of the discussion you referenced. I will then make another pull request for adding the checksum checking for BTHNG129, as discussed at the end of that same discussion.

PR #1028, opened 2019-04-01
Fix return values for nexa and proove decoder

Mindavi:
This issue is related to #1028, but does not actually solve the issue there. It does solve an issue found upon researching that issue however.

Fix return values for the nexa and proove decoder, so they'll signal data is found. Took the opportunity to introduce the values given in https://github.com/merbanan/rtl_433/blob/42560751c80147f62fff306338645ddea0cde9e9/include/r_device.h#L23-L33

Reordered the fields so they're equal in the proove and nexa sensor, giving somewhat nicer output.
proove-next
Before the fields were ordered differently, giving a different ordering in the output.

Should an issue be opened to do all of the sensors with these return values or does something like that exist already? Is a good issue for the tag `first-issue` or `up-for-grabs` or `beginner` or something. I might also do it if it's something that is wanted.

### DIFF
--- a/src/devices/nexa.c
+++ b/src/devices/nexa.c
@@ -22,7 +22,7 @@ static int nexa_callback(r_device *decoder, bitbuffer_t *bitbuffer) {
 
     /* Reject codes of wrong length */
     if (bitbuffer->bits_per_row[1] != 64 && bitbuffer->bits_per_row[1] != 72)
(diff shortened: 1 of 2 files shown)


merbanan:
Looks good. But make it 2 commits next time, 1 functional and 1 with cosmetics.

PR #1031, opened 2019-04-02
Minor fixes to SI conversion and new model reporting

helgew:
This PR addresses the fact that the SI conversion for the Acurite 5n1 sensor was not working for the accumulated rain gauge readings. For historical reasons (?), that was reported as "rain_inch" instead of "rain_in". Two other minor changes convert battery status reports from "Ok" to "OK".


helgew:
It's been a while! Great work on this project, everyone! Thanks for keeping this so well supported.

zuckschwerdt:
Thanks, looks like the new model changes rattled some loose bits ;)

Issue #1032, opened 2019-04-02
Streamline sensor id keys


helgew:
Related to https://github.com/merbanan/rtl_433/pull/986, I noticed that many sensors report IDs in some form or the other. However, the keys for those IDs are a mix of "sensor_id", "id", "device", etc.

I am happy to tackle this in my ample spare time, but wanted to first elicit some feedback on whether a) it should be done, and b) what the best approach should be?

For b) `_X("sensor_id", "id")` seems to be the obvious answer.

zuckschwerdt:
It should be `_X("id","device")` so users can switch that mit newmodel. I think I already looked for those id-like keys. Maybe I missed some?

PR #1034, opened 2019-04-03
conversion of some device IDs and other fields to the new model

helgew:
fixes https://github.com/merbanan/rtl_433/issues/1032

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,8 @@ build*/
 _*/
 
 rtl_433_tests
+
+.idea
+cmake-build-debug/
+
+.DS_Store
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -555,7 +555,7 @@ static int acurite_txr_callback(r_device *decoder, bitbuffer_t *bitbuf)
 
                 data = data_make(
                     "model",        "",   DATA_STRING,    _X("Acurite-5n1","Acurite 5n1 sensor"),
-                    "sensor_id",    NULL, DATA_INT,       sensor_id, // @todo normaiize to "id" at 1.0 release.
+                    _X("id", "sensor_id"),    NULL, DATA_INT,       sensor_id,
                     "channel",      NULL,   DATA_STRING,    &channel_str,
                     "sequence_num",  NULL,   DATA_INT,      sequence_num,
                     "battery",      NULL,   DATA_STRING,    battery_low ? "OK" : "LOW",
@@ -578,7 +578,7 @@ static int acurite_txr_callback(r_device *decoder, bitbuffer_t *bitbuf)
 
                 data = data_make(
                     "model",        "",   DATA_STRING,    _X("Acurite-5n1","Acurite 5n1 sensor"),
-                    "sensor_id",    NULL, DATA_INT,  sensor_id, // @todo normalize to "id" at 1.0 release.
+                    _X("id", "sensor_id"),    NULL, DATA_INT,  sensor_id,
                     "channel",      NULL,   DATA_STRING,    &channel_str,
                     "sequence_num",  NULL,   DATA_INT,      sequence_num,
                     "battery",      NULL,   DATA_STRING,    battery_low ? "OK" : "LOW",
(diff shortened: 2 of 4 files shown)


src/devices/x10_rf.c
```diff
@@ -69,8 +69,8 @@ static int x10_rf_callback(r_device *decoder, bitbuffer_t *bitbuffer) {
 
     data = data_make(
             "model",    "", DATA_STRING, "X10-RF",
-            "houseid",  "", DATA_STRING, housecode,
-            "deviceid", "", DATA_INT, bDeviceCode + 1,
+            _X("channel", "houseid"),  "", DATA_STRING, housecode,
+            _X("id", "deviceid"), "", DATA_INT, bDeviceCode + 1,
```
zuckschwerdt: Can you swap both lines? model, id, channel is the common format.

src/devices/acurite.c
```diff
@@ -930,7 +930,7 @@ static int acurite_00275rm_callback(r_device *decoder, bitbuffer_t *bitbuf)
             else if (probe == 2) {
                 ptempc = (((signal[0][7] & 0x0f) << 8) | signal[0][8]) * 0.1 - 100;
                 data = data_append(data,
-                        "ptemperature_C",  "Celsius",      DATA_FORMAT,    "%.1f C",  DATA_DOUBLE, ptempc,
+                        _X("temperature_C", "ptemperature_C"),  "Celsius",      DATA_FORMAT,    "%.1f C",  DATA_DOUBLE, ptempc,
```
zuckschwerdt: This (and below) is actually a distinct value from the probe, not the base temp/hum value. It's a collision with the base value.

src/devices/x10_rf.c
helgew: Will do... I wish every change request could be done in three key strokes!

src/devices/acurite.c
helgew: Got it. Would `temperature_1_C` and `temperature_2_C` be more appropriate?

zuckschwerdt: not sure, so far I only normalized the keys that sorely stood out.

helgew: I had a look at all of them and there were only a couple others that may benefit from clarification. I think the `temperature_<n>_C` approach is the way to go for this one since it seems to be a device that can report multiple temperatures from different probes.

zuckschwerdt:
Looks good, thanks. Lets wait a few days if someone wants to get in on the `ptemperature_C`, `phumidity` issue.

.gitignore
```diff
@@ -56,3 +56,8 @@ build*/
 _*/
 
 rtl_433_tests
+
+.idea
```
Mindavi: Maybe you should add this to your own global gitignore instead of adding it in here. Also the .DS_STORE can be in there. In the original gitignore file the build*/ folder is already ignored, so you could do `mkdir build && cd build && cmake .. && make` instead of making a cmake-build-debug folder. Just some minor comments, but they avoid cluttering this file for every project you want to contribute to.

helgew: > Maybe you should add this to your own global gitignore instead of adding it in here.

That is a good idea although I thought that others might benefit from the addition of the default ignores for a fairly popular IDE. Are you familiar with IdeaLabs' CLion?

> Also the .DS_STORE can be in there. 

Great... but who died and made you queen?

> In the original gitignore file the build*/ folder is already ignored, so you could do `mkdir build && cd build && cmake .. && make` instead of making a cmake-build-debug folder.

Thanks. I am quite familiar with the build structure of this project. I fixed it back in 50cfd95da14b37678326fdf17fe2ec18213bfb3c. That folder, as mentioned above, is a default folder created by the IDE.

> Just some minor comments, but they avoid cluttering this file for every project you want to contribute to.

Are you really worried about the addition of 4 lines to a file that no one really reads or edits except for when they want to add lines to it they think might benefit others?

zuckschwerdt: Please try a friendly and open discussion. Handling your IDE excludes in a global setting does indeed seem apropriate -- with the benefit to get it for free in every repo you use. (btw. that IDE isn't free (FOSS), so it's likely not on most peoples shortlist.)

helgew: Where was I unfriendly? I replied with humor to what seemed a rather condescending comment. Maybe the humor got lost...

As for the proposed changes, I don't really care one way or the other. I was just trying to help out.

[CLion actually is free (as are all other JetBrains tools) for FOSS projects like rtl_433](https://www.jetbrains.com/store/terms/comparison.html).

Issue #1014, opened 2019-03-20
Nexa/Proove ID is inverted?


lasseo:
I am a first time user of RTL-433. I wanted to sniff existing communication between Nexa equipments to be able to add this to Hass.io. However, unsure if I interpret data wrong or if "RTL-433" is giving me wrong info.

I have a Nexa dimmer unit, and Nexa wallbuttons to controll the dimmer (self-learning)
Sniffing traffic when turning light off with the Nexa buttons gave with RTL-433 gave the following output.

time      : 2019-03-19 22:19:16
model     : Proove       House Code: 67084961
Channel   : 3            State     : OFF           Unit      : 2             Group     : 1
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2019-03-19 22:19:16
model     : Nexa         House Code: 67084961
Group     : 1            Channel   : 3             State     : OFF           Unit      : 2
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

However, this does not match with information received with TelldusCenter ( sniffing )
Remote Code / House code: 23902  Unit code: 2

Even TelldusCenter sniffing got some random values first, but then it stabilised on correct and working values.  

I feel there is something wrong in RTL-433 parsing, but it could equally be me not understanding the application good enough. Please reject issue if incorrect. I am happy to provide more "data" if needed.

zuckschwerdt:
`67084961` is `23902` inverted ("bit flipped"). Can't say which one is "right". Can you reference the code to something "official"?

lasseo:
What do you mean with official?

- Products are the following:
[...]

Issue #1035, opened 2019-04-03
prologue sensor, unexpected output


lasseo:
This was seen when sniffing:

time      : 2019-04-04 00:17:41
model     : Prologue sensor                        id        : 9
rid       : 84           Channel   : 1             Battery   : LOW           Button    : 0             Temperature: 5.40 C       Humidity  : 204 %
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

- Looks wrong to have "Button"
- Humidity should be shown as 20.4%

I do not own this sensor so I can not confirm what device it is.

merbanan:
The prologue protocol is very simple and used by alot of devices. It also has no checksum so it is hard to verify the data except repeats.

Anyway this sensor reports 204 for humidity. That is CC in hex. Devices without humidity sensors has CC in the field where the humidity data should be. This is a bug that should be fixed as the check is easy to perform.

With regards to the button, I have a device which has a button on the back to manually send a transmission. When I do that the button bit is set. So this parameter is correct for my device.

So keep this issue open until I fixed the humidity==0xCC issue.

zuckschwerdt:
Can you confirm that the 5.4 C temperature reading is sensible for this device? Was this just a one off, or a series of similar readings?

lasseo:
Thanks for the information.
I am kind of new to this area, so did not know about the button option. With that explanation I guess it make sense to display it.

zuckschwerdt: I was getting this reading several times, during the hour I was playing with RTL-433. As I wrote in the issue report, I do not own the sensor ( must be one of my neighbours ). However my own temperature sensor displayed 5.7 C so for me the 5.4 C value should be considered correct.

Thanks for your support, and sorry if some of my reported issues are already known issues. I am trying to search in open and closed issues before posting.

zuckschwerdt:
Humidity 0xCC is fixed (and general clean up) with 579213f.

PR #1036, opened 2019-04-04
Change Nexa/Proove ID, group cmd

zuckschwerdt:
This inverts the Nexa/Proove ID and group cmd flag.
S.a. #1014


Mindavi:
Related to #1014 and not #1035?

zuckschwerdt:
Thanks!

merbanan:
Looks ok.

PR #1037, opened 2019-04-05
Remove the dependency on the order of command line arguments

helgew:
Currently, device specific configuration parameters (e.g. verbosity or `-M newmodel`) have to be listed on the command line before the devices are listed. This commit makes the order of command line arguments irrelevant.

The issue can be easily reproduced:
`rtl_433 -M newmodel -vv -R 20` -> verbosity and new model parameters are correctly reflected in the output
`rtl_433 -R 20 -M newmodel -vv` -> output is non-verbose and old model parameters are used


zuckschwerdt:
Good idea!

helgew:
I was scratching my head for the longest time before I realized why my new model keys were not showing up :D

PR #1052, opened 2019-04-29
added example for mqtt format option...

krikk:
this format option was not clear for me after first reading...


zuckschwerdt:
We probably need a wiki page to explain all the intricacies :)

PR #1054, opened 2019-05-03
CMake: Switch back to _GNU_SOURCE

neheb:
This fixes compilation with uClibc. Otherwise strdup is undefined.

A few solutions here. strdup can be replaced with strndup. _BSD_SOURCE can also be used. But replacing POSIX_C_SOURCE causes other functions to be undefined. _GNU_SOURCE is easiest.

edit: https://downloads.openwrt.org/snapshots/faillogs/arc_arc700/packages/rtl_433/compile.txt


zuckschwerdt:
Thanks! Can you keep the posix line, commented out and annotated, e.g.

    # for strdup, setenv, use either
    #ADD_DEFINITIONS(-D_POSIX_C_SOURCE=200809) # does not work with uClibc
    ADD_DEFINITIONS(-D_GNU_SOURCE)

neheb:
Done

PR #1061, opened 2019-05-17
Thermopro TP12 also supports TP20

mikeage:


mikeage:
I don't know if you want PRs for things like this; I assume there are also other Thermopro models that work, but just in case, here it is. I was pleasantly surprised when this just worked out of the box!

zuckschwerdt:
Thanks. More documentation which models, brands, resellers use a protocol is always appreciated.

PR #1063, opened 2019-05-22
Add DirecTV RC66RX Remote Control

klohner:
This adds support for the DirecTV RC66RX Remote Control device.   Seems to test well with my remote directly as well as samples submitted to [rtl_433_tests](https://github.com/merbanan/rtl_433_tests/pull/280).  It's a rather unusual decode, though.  Happy to discuss.


src/devices/directv.c
```diff
@@ -0,0 +1,532 @@
...
+    return label;
+}
+
+
+static inline int bit(const uint8_t *bytes, unsigned bit)
+{
+    return bytes[bit >> 3] >> (7 - (bit & 7)) & 1;
+}
+
+// similar to bitbuffer_search in bitbuffer.h, but allows us to search a custom bitrow_t
```
merbanan: Maybe move to bitbiffer.h?

src/devices/directv.c
```diff
@@ -0,0 +1,532 @@
...
+SYNC + DATA + EOR (repeated at least once, but could continue indefinitely as
+long as button is held down)
+
+SYNC is either 000111111111100 (first row in signal) or 0001111100
+(subsequent rows in transmission)
+
+DATA (20 data packets) and EOR (1 data packet) is converted to bits or EOR
+marker as such:
+
+| Raw Data Packet Bits | Decoded Logical Data Bits |
```
merbanan: I think the proper name is symbol. You transmit symbols that translate to bits or special exceptions.

zuckschwerdt:
I didn't look too closely but the symbols look like DMC, right? short transition -> `0`, long transition -> `1`.

src/devices/directv.c
```diff
@@ -0,0 +1,532 @@
...
+  
+    // Signal is reset by rtl_433 before recognizing rows, so in practice, there's only one row.
+    // It would be useful to catch rows in signal so that there'd only be one decoded row per
+    // signal and we could count repeats to report the length of the signal, but that's not
+    // supported yet.  It seems the gap between rows exceeds the ook_hysteresis threshold in
+    // pulse_detect.c:  int16_t const ook_hysteresis = ook_threshold / 8; // ±12%
+    // and changing this value isn't the right direction, nor does this even work for this signal.
+    // Grouping rows in a signal like this will need to be supported in some other way, and this 
+    // support is not yet available in rtl_433.
+    // For now, we'll decode the signal in the bitbuffer assuming it is only one row.
```
merbanan: Proper grouping and segmentation of rows are an issue in other signals as well. They break the heuristics that is used. And when we go for FSK signals they are a magnitude shorter then the OOK ones. All in all the grouping properties doesn't apply. It's hard to decode all kinds of random signals.

src/devices/directv.c
```diff
@@ -0,0 +1,532 @@
...
+            }
+            return 0;
+        }
+        // will be either 10 or 11 at the end of bitrow here.  Not good.
+        if (decoder->verbose > 1) {
+            fprintf(stderr, "directv: bitbuffer contains unexpected %s bits at end of bitrow.\n", dtv_twobits==0?"10":"11");
+        }
+        return 0;
+    }
+    // We've received an acceptable decoded dtv value here.
```
merbanan: Oh, this was kinda messy. Lets hope the DMC decoder can be used to simplify the logic.

zuckschwerdt: I didn't check properly, kind-of-DMC rather as the bits are of variable length presumably. Something like "DPWM" -- if that makes sense. A new demod  derived from DMC/PWM should do.

src/devices/directv.c
```diff
@@ -0,0 +1,532 @@
...
+};
+
+r_device directv = {
+        .name        = "DirecTV RC66RX Remote Control",
+        .modulation  = FSK_PULSE_PCM,
+        .short_width = 600,  // 150 samples @250k
+        .long_width  = 600,  // 150 samples @250k
+        .gap_limit   = 30000, // gap is typically around 27,600μs, so long that rtl_433 resets 
+                              // signal decoder before recognizing row repeats in signal
+        .reset_limit = 50000, // maximum gap size before End Of Row [μs]
```
merbanan: The maximum signal length heuristics messes up the row splitting here. One could think of using some kind of buffering here to get the actual rows. But that would mess with the realtime aspects of the remote. The only good idea I have here is to perform a duplicate check and suppress the duplicate events. That way we get a behavior that is usable and reliable for a consumer.

merbanan:
Impressive work with regards to the decoding step.

src/devices/directv.c
klohner: Right.  I've updated the description to use more consistent and appropriate terms.  Does it make more sense now?

src/devices/directv.c
klohner: Good point.  After thinking about it this way (as some kind of differential PWM), I realized I was going about decoding all wrong.  I've completely rewritten the decoding by only looking at symbol transitions, making it so much more robust (works for symbol inversions) and generalized.

src/devices/directv.c
klohner: My thought on how to implement this:

- Provide a function that will allow for generation of a hash value given any arbitrary text.
- Decoders can use this to create a hash value of whatever fields in their decode would be considered part of the same decoding if it were seen in the subsequent decoding.
- Allow decoders to put this hash value into their output data structure
- Provide this hash value somewhere accessible in the device or bitbuffer for the decoder to access.

In implementation, I'm thinking something like:

```
static int template_callback(r_device *decoder, bitbuffer_t *bitbuffer)
{

    [...]

    // Each decoder, if it wants, can add something like these two lines:
    sprintf(decode_identifying_data, "[New Template|%d|%d]", sensor_id, value);
    decode_hash = make_hash_value(decode_identifying_data);
    
    // Decoder has an option, if desired, to act on the repetition itself
    prev_decode_hash = bitbuffer->prev_decode_hash;
    if (decode_hash == prev_decode_hash) {
        do something
    }

    // But, it should just add the new hash here:
    data = data_make(
            "model", "", DATA_STRING, "New Template",
            "id",    "", DATA_INT,    sensor_id,
            "data",  "", DATA_INT,    value,
            "mic",   "", DATA_STRING, "CHECKSUM", // CRC, CHECKSUM, or PARITY
            "decode_hash", "", DATA_HASH, decode_hash,   // provide HASH here
            NULL);

    [...]
```

The main program can just choose to show all decodings as normal, but maybe 
there's a new flag to rtl_433 that will have it suppress all repetitions (just see
if decode hash is same as previous hash), or maybe count all the the repetitions
it sees and output them all as one second event with a field to denote this:

```
time      : 2019-05-23 21:56:26
model     : DirecTV-RC66RX                         device_id : 123456
button_id : 0x25         button_name: [SELECT]     event     : INITIAL
mic       : CHECKSUM
time      : 2019-05-23 21:56:28
model     : DirecTV-RC66RX                         device_id : 123456
button_id : 0x25         button_name: [SELECT]     event     : REPEAT
mic       : CHECKSUM     repetition_count: 15
```

zuckschwerdt:
The DPWM seems to work well? It would be great if we can fit it in [pulse_demod.c](https://github.com/merbanan/rtl_433/blob/master/src/pulse_demod.c), (the last addition was PIWM with a similar goal -- use pulse and gap lengths). Since we do not need clock recovery we could also switch to the `short`, `long`, and `sync` widths with `tolerance`.

src/devices/directv.c
```diff
@@ -0,0 +1,411 @@
...
+ */
+
+/**
+DirecTV RC66RX Remote Control decoder.
+
+The device uses a FSK to transmit a PCM signal TRANSMISSION.  Its FSK signal
+seems to be centered around 433.92 MHz with its MARK and SPACE frequencies
+each +/- 50 kHz from that center point.
+
+A full signal TRANMISSION consists of ROWS, which are collections of SYMBOLS.
```
merbanan: TRANSMISSION

merbanan:
We need dupcheck support. Not needed for merge though,

#908

src/devices/directv.c
klohner: Thanks.  Fixed that.

klohner:
The DPWM function works very well for what I needed.  From my testing, it even parses just as well if fed either just the OOK mark pulses or just the space pulses.  My only other concern/thought for making it more general purpose would be to have it return at the first non-empty message in the row and also return a position pointer to the beginning of the next sync in the row.  That way it could be efficiently wrapped in a bitbuffer function that would turn a single row of data into multiple rows.

merbanan:
Code merged.

PR #1066, opened 2019-05-26
Add support for Ecowitt temperature sensor

mhansen:


mhansen:
Tests, images, manual here: https://github.com/merbanan/rtl_433_tests/pull/284

src/devices/ecowitt.c
```diff
@@ -0,0 +1,115 @@
+/*
```
merbanan: Add proper copyright header.

src/devices/ecowitt.c
```diff
@@ -0,0 +1,115 @@
...
+ * 55-bit one-row data packet format (inclusive ranges):
+ * |  0-14 | Always 111111101010011
+ * | 15-22 | Random(?) Sensor ID, reinitialized on boot
+ * | 23-24 | Always 00
+ * | 25-26 | 2-bit sensor channel, selectable on back of sensor {00=1, 01=2, 10=3}
+ * | 27-28 | Always 00
+ * | 29-38 | 10-bit temperature in tenths of degrees C, starting from -40C. e.g. 0=-40C
+ * | 39-46 | Always 1111 1111
+ * | 47-54 | Probably a checksum. Haven't decoded this yet. Not a simple adder, probably a CRC.
```
merbanan: Can you try with reveng or post a set of sample payloads?

merbanan: Or try with CRC-8 with poly 0x31 init 0x00
   bitbuffer_extract_bytes(bitbuffer, 0, 7, b, sizeof(b) * 8); // Skip first 7 bits

    // Verify checksum
    int crc = crc8(b, 7, 0x31, 0);

Ecowitt is some kind of brand for Fine Offset. They are probably very similar. Maybe you can merge this decoder to one of the fine offset ones.

merbanan: fineoffset_WH2_callback looks very very similar to this.

zuckschwerdt:
The buffer looks mostly aligned. Can you copy the style of other decoders and swap `bitbuffer_extract_bytes` for bitmasks on the buffer (see e.g. [nexa](https://github.com/merbanan/rtl_433/blob/master/src/devices/nexa.c))?

src/devices/ecowitt.c
```diff
@@ -0,0 +1,115 @@
+/*
+ * 55-bit one-row data packet format (inclusive ranges):
+ * |  0-14 | Always 111111101010011
```
merbanan: This matches this code:

    } else if (bitbuffer->bits_per_row[0] == 55 &&
            bb[0][0] == 0xFE) { // WH2A
        bitbuffer_extract_bytes(bitbuffer, 0, 7, b, 48);
        model = _X("Fineoffset-WH2A","Fine Offset WH2A sensor");

Maybe the crc check rejects the data.

merbanan:
The Fineoffset-WH2A code but with a the WH5 bias offset seems to match this sensor.

mhansen:
Thanks for the great comments folks. Heads up, I'll be away for a bit, but I'll come back to this and address your comments.

src/devices/ecowitt.c
mhansen: Wow, that suggested checksum worked a treat. Nice!

src/devices/ecowitt.c
mhansen: Done

mhansen:
> The buffer looks mostly aligned. Can you copy the style of other decoders and swap `bitbuffer_extract_bytes` for bitmasks on the buffer (see e.g. [nexa](https://github.com/merbanan/rtl_433/blob/master/src/devices/nexa.c))?
> 
> Although I do like this more declarative idea, we should perhaps look into that some day.

Done: I put one bitbuffer_extract_bytes at the top (to skip 7 bits, mostly-aligning everything), then the rest is done with bitmask operations.

There's probably some even more declarative way of doing this than how I was doing it - might be worth a try some day.

src/devices/ecowitt.c
mhansen: Yeah, looks very similar. I wonder if we could merge these, but my guess is perhaps not: They're slightly different: the WH2A has a 4-byte checksum, but this has a 5-byte checksum.

mhansen:
PTAL. Thanks very much for your help figuring out the checksum.

PR #1081, opened 2019-06-18
Add support for HCS200 KeeLoq encoder

667bdrm:
Added Microchip HCS200 based remotes decoder

Sample data: https://github.com/merbanan/rtl_433_tests/pull/287

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,8 @@ Supported device protocols:
     [124]  LaCrosse/ELV/Conrad WS7000/WS2500 weather sensors
     [125]  TS-FT002 Wireless Ultrasonic Tank Liquid Level Meter With Temperature Sensor
     [126]  Companion WTR001 Temperature Sensor
-    [127]  Microchip HCS200 KeeLoq Hopping Encoder based remotes
+    [127]  Ecowitt Wireless Outdoor Thermometer WH53/WH0280/WH0281A
+    [128]  Microchip HCS200 KeeLoq Hopping Encoder based remotes
 
 * Disabled by default, use -R n or -G
 
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ Supported device protocols:
     [124]  LaCrosse/ELV/Conrad WS7000/WS2500 weather sensors
     [125]  TS-FT002 Wireless Ultrasonic Tank Liquid Level Meter With Temperature Sensor
     [126]  Companion WTR001 Temperature Sensor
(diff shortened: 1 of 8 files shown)


zuckschwerdt:
Looks good, thanks.
- Is `serial` unique per device, then it should be `id`. Or is it some kind of message "counter"?
- The `battery` key should be `battery_ok` with an `int`value.
- `(b[1] << 24)` should be `((unsigned)b[1] << 24)` (others are fine as-is)
- For `serial` reduce every shift by 4 there and don't shift and mask "at the end"
- Both `%04X` overflow and should be `%08X`
- Please use 4 spaces as indent
- Continued lines (after data_make, output_fields, r_device) should have two indents
- Add a trailing comma after that last field in output_fields and r_device

667bdrm:
> Is serial unique per device, then it should be id. Or is it some kind of message "counter"?

Yes, it should be unique id

> For serial reduce every shift by 4 there and don't shift and mask "at the end"

If I do so there will be data from the battery status byte. Per datasheet page 8 there are 28 bit serial field and there are button status data next ` >> 4`at the end is used to cut off this, `& 0x0fffffff` is used to remove 0xf from the beginning.

zuckschwerdt:
The `& 0x0F` should probably be `& 0xF0`, because the former with `>>4` is effectivly `& 0x00`
I.e. you actually want `(b[5] << 20) | (b[6] << 12) | (b[7] << 4) | (b[8] >> 4)`, right?

667bdrm:
Yes, it is. The serial (id) decoding now the same for the test data.

zuckschwerdt:
Can you add the GPL notice like in other decoders, if that license is okay with you?

667bdrm:
Yes, sure. GPL is ok.

merbanan:
@zuckschwerdt can we merge this? Code looks clean.

zuckschwerdt:
Needs a rebase but otherwise good to merge!

nunojpg:
@667bdrm why are you looking for buttons at b[8] == 2,4,6,9? My Remote (Gibidi) uses 1,2,3,4. Maybe this is a artifact of some changes done in the value content? With current master I definitely only get button 1 and button 2 to show as ON, button 3 and 4 alias to button 2, and I checked athat b[8] becomes 1,2,3,4.

zuckschwerdt:
See #1512 for a discussion on relabeling the buttons to fit the protocol description.

PR #1087, opened 2019-06-27
Adding two EV1527 based sample configurations

curlyel:
This PR wants to add two configuration examples to adopt EV1527 based devices.


zuckschwerdt:
Thanks!

PR #1090, opened 2019-06-27
Add support for Eurochron weather station sensor

olanwe:
Hi everyone, I wrote a device driver for my Eurochron weather station:
 
eurochron

It is running for two weeks now (24/7) without any problems or major dropouts, so I thought it might be useful for others as well.

BTW, these are my very first steps with Git, so I hope I did not mess anything up yet... :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ Supported device protocols:
     [126]  Companion WTR001 Temperature Sensor
     [127]  Ecowitt Wireless Outdoor Thermometer WH53/WH0280/WH0281A
     [128]  DirecTV RC66RX Remote Control
+    [129]* Eurochron (temperature and humidity sensor)
 
 * Disabled by default, use -R n or -G
 
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ Supported device protocols:
     [126]  Companion WTR001 Temperature Sensor
     [127]  Ecowitt Wireless Outdoor Thermometer WH53/WH0280/WH0281A
     [128]  DirecTV RC66RX Remote Control
-    [129]* Eurochron (temperature and humidity sensor)
+    [129]* Eurochron temperature and humidity sensor
 
 * Disabled by default, use -R n or -G
 
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -314,7 +314,7 @@ stop_after_successful_events false
   protocol 126 # Companion WTR001 Temperature Sensor
   protocol 127 # Ecowitt Wireless Outdoor Thermometer WH53/WH0280/WH0281A
   protocol 128 # DirecTV RC66RX Remote Control
-# protocol 129 # Eurochron (temperature and humidity sensor)
+# protocol 129 # Eurochron temperature and humidity sensor
 
 ## Flex devices (command line option "-X")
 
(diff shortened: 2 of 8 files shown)


zuckschwerdt:
Thanks, some notes:

- The new file needs to be added to the VC build scripts (run `./maintainer_update.py` to see what needs changing)
- Use a space after `if`, put the `{` inline with the `if`.
- Temp is 12 bit 2's complement right? Use sign extend, i.e. `(int16_t)((b[3] << 8) | (b[4] & 0xf0)) >> 4`
- Scale by `* 0.1` instead of `/ 10` preferably
- Simplify to `battery_low = b[1] >> 7`
- Similar for TX-Key
- Don't use `_X` macro -- that's only for legacy code
- Use `"battery_ok", "Battery", DATA_STRING, !battery_low,`
- Use `"button"` instead of `"key"`
- Use just `output_fields` instead of `eurochron_csv_output_fields`
- Add a trailing comma to the last field of `output_fields` and `r_device`
- Add one indent more to `output_fields` and `r_device` content
- If possible and there is some model marking on the sensor then qualify "Eurochron" with something, e.g. Eurochron-TH42

olanwe:
Thanks for your code review and notes! I have applied the changes now. Please let me know in case I missed something.

> The new file needs to be added to the VC build scripts (run ./maintainer_update.py to see what needs changing)

There are no changes on the VC scripts, because I have already done those changes manually before (I compared the last device pull request to see what files have been changed there). 
However, README.md would be changed at multiple positions as well, but not because of my device addition.
For example, the last device description changes to:
```
-    [127]  Ecowitt Wireless Outdoor Thermometer WH53/WH0280/WH0281A
+    [127]  Ecowitt Temperature Sensor
```
as well as changed options and the additional notice for "newmodel".
I have **not** added these changes yet, because they were not part of my code change. However, I can add them to this pull request (or open a new one afterwards) if you like.

> Use "battery_ok", "Battery", DATA_STRING, !battery_low,

I have changed this to DATA_INT (together with the "button" line), because DATA_STRING will not work here. I hope this is OK.

> If possible and there is some model marking on the sensor then qualify "Eurochron" with something, e.g. Eurochron-TH42

There is no model information on the sensor, not even in the battery compartment. There is only the manufacturers address and an article number for re-ordering on the back of the device.

zuckschwerdt:
You are right, I mixed up `DATA_STRING`/`DATA_INT`. No need to change the Readme in any other place. Well implemented, looks good.
We could reasonanly use a model name of `Eurochron-TH` -- we used the `-TH` suffix for some other devices with temperature/humidity capability to have the common `Yyy-Zzz` model name scheme. And perhaps drop the parens on `.name`.
Did you PR some sample files to [rtl_433_tests](https://github.com/merbanan/rtl_433_tests)?

olanwe:
OK, I have changed the model name.
> Did you PR some sample files to [rtl_433_tests](https://github.com/merbanan/rtl_433_tests)?

Not yet, but I have already prepared some sample files. Do you have any directory name preferences (as it looks a little bit mixed up in there)? Is "eurochron-th/01" OK?

zuckschwerdt:
Sure, that looks good. I guess someday we'll decide on a prefered scheme (maybe in a effort to generate some meta-data and compile the readme's) but for now "whatever works" ;)

PR #1094, opened 2019-07-02
DirecTV - use standard "id" field name

klohner:
I think I should be using the standard "id" field name in this device driver to better align with output from other device drivers.


zuckschwerdt:
That one must have slipped through, good catch. Perhaps best to use the `_X` macro to migrate this with `newmodel`.

zuckschwerdt:
I meant `_X("id", "device_id")`, but I missed that this is a very recent addition, so probably no harm to just switch the key without the macro. Sorry for the confusion.

klohner:
Yeah, this one is so new, I'm probably the only one who would notice the change.

PR #1097, opened 2019-07-05
add channel output to lacrosse_TX141TH_Bv2

rsnel:
The channel is encoded in the lower two bits
of the flags nibble. We extract it and add
it to the output. Channels known to me:

- flags = xx01 (channel 1)
- flags = xx10 (channel 2)


zuckschwerdt:
Great! Thanks!

PR #1102, opened 2019-07-10
Fix a crash in hcs200, due to parameter type mix-up in a call to data_make

zmeiresearch:


zuckschwerdt:
Oops. Need to check if regression tests are misssing ;)

PR #1120, opened 2019-08-11
cmake: use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

nunojpg:
CMAKE_SOURCE_DIR refers to the top level CMakeLists.txt, which is not the case
when this CMake project is included in another CMake project

For this case CMAKE_CURRENT_SOURCE_DIR or PROJECT_SOURCE_DIR can be used.

Signed-off-by: Nuno Goncalves <nunojpg@gmail.com>


zuckschwerdt:
Recommended best practice. Thanks! Do you use rtl_433 as a dependency in your project?

nunojpg:
I just compiled to run on board the PlutoSDR and I had a master project including rtl_433, SoapySDR and SoapySDRPluto. Thanks for merging.

PR #1121, opened 2019-08-14
Create rtl_433.1 manual page

alexmyczko:
this is called a man page, putting it into `/usr/share/man/man1/`, `man rtl_433` will show it nicely. you might want to move it into a sub folder, and improve the contents with the output of the `help2man --no-discard-stderr rtl_433` command.

### DIFF
--- a/rtl_433.1
+++ b/rtl_433.1
@@ -1,0 +1,29 @@
+.TH rtl_433 1 "July 25 2018"
+.SH NAME
+rtl_433 \- decode information from a wireless sensor using librtlsdr
+.SH SYNOPSIS
+.B rtl_433
+.RI [ options ]
+.SH DESCRIPTION
+This manual page documents briefly the
+.B rtl_433
+command.
+.PP
+\fBrtl-433\fP is a generic data receiver, mainly for the 433.92 MHz, 868 MHz (SRD),
+315 MHz, and 915 MHz ISM bands. It works with RTL-SDR and/or SoapySDR. Activly tested
+and supported are Realtek RTL2832 based DVB dongles (using RTL-SDR) and LimeSDR
+(LimeSDR USB and LimeSDR mini engineering samples kindly provided by MyriadRf),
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Do you also want hook this up to the Makefile and CMake to get installed? Otherwise I'll take a look at that later.

alexmyczko:
Probably that makes sense but only for Linux. You need to check where macOS and FreeBSD want the file. Windows, no idea... as you wish, it only makes sense to keep it, if someone fills up details about the parameters and config file, otherwise not so much (so if nobody wants to edit it, just drop/remove the PR :)

zuckschwerdt:
I guess we wouldn't want to edit roff text by hand. Maybe Asciidoc(tor) or Pandoc could be used (as a maintainer tool, not a dependency). Input would be mostly markdown then.

zuckschwerdt:
I just had a go with asciidoctor. Good results with a simple syntax. I'll merge this, change to asciidoc and hook up in the install scripts in a few days.

alexmyczko:
And tag it a new release?

zuckschwerdt:
That's overdue, yes.

alexmyczko:
19.08 or 19.09?

merbanan:
19.08 should be ok. I have nothing new to add. And among the PR's there is nothing that seems really complete for merging that we could wait for. @zuckschwerdt do you have anything to add before a release?

zuckschwerdt:
I don't see anything big waiting. I'll add a man page (asciidoc format works well) and then we'll slap the release tag on it next week.

petterreinholdtsen:
For the record, FreeBSD keep manual pages in /usr/share/man/man1/ too.  I believe MacOSX do too, but could not verify it right now.

klohner:
For Windows documentation, maybe create [CHM Files From AsciiDoc](http://germaniumhq.com/2018/09/29/2018-09-29-Creating-Chm-Files-From-AsciiDoc/)?

PR #1130, opened 2019-08-19
Remove all (4) unreferenced variables

winterrace:
see above, minor stuff...


merbanan:
You need to rebase/rework this. Changes looks ok. But 75 commits are abit to much.

winterrace:
OK, I have to check how to do this...

winterrace:
Hm, it was the same with my last pull requests. But each of them ended up in a single commit to rtl_433. So the long commit history of my source brances has no negative effect to the rtl_433 repo. However, I try to get rid of this...

zuckschwerdt:
Yes, we can squash merge this, but the PR looks really messy. Perhaps you can reset your master branch, pull our upstream master branch and then in the future branch your changes off that.

PR #1138, opened 2019-08-29
Added 'unique' to the flex decoders

curlyel:
Using 'unique' in the flex decoder produces a cleaner out e.g. for direct MQTT publish.

Thank's Christian for making 'unique' available to it!


zuckschwerdt:
Glad to hear, thanks.

PR #1151, opened 2019-09-15
Added support for QUIGG GT-TMBBQ-05

glageo:


zuckschwerdt:
Thanks. Looks good. I'm going to rewrite this a little for code style on merge.

zuckschwerdt:
Can you add a header comment to the source with your copyright and licence? See e.g. https://github.com/merbanan/rtl_433/blob/master/src/devices/s3318p.c

glageo:
Done.

zuckschwerdt:
@glageo To figure out if there is an ID, can you reset the battery (disconnect for a few seconds) and see if the suspected id field changes?

glageo:
Already tried. Does not change. I assume it's either hard wired - although I don't think this is reasonable in this price range - or the device is just single channel. 
I also noticed that my receiver really has a bad reception. It hardly catches up signals through a window. The DVB-T stick with it's bad antenna catches the signals without any problems. Maybe this is on purpose so you don't catch the signals from your neighbours?  If I manage to get another device I can tell..

PR #1159, opened 2019-10-02
Add documentation for influxdb sink

azrdev:
both variants, MQTT consumer plugin, and rtl433json_to_influx.py

### DIFF
--- a/docs/INTEGRATION.md
+++ b/docs/INTEGRATION.md
@@ -118,7 +118,8 @@ See [rtl_433_graphite_relay.py](https://github.com/merbanan/rtl_433/tree/master/
 
 ### InfluxDB
 
-TBD.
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks. Take a look at our [examples](https://github.com/merbanan/rtl_433/tree/master/examples). You really shouldn't use a log file to communicate. UDP with the syslog protocol will tolerate faults better. Also please try `-M newmodel`, the old keys are a mess -- you probably noticed ;)

azrdev:
@zuckschwerdt look at the time stamps, I built this script quite a while ago, still works though.

and it's not using a file but a named pipe :)

zuckschwerdt:
Thanks, I'll update the docs soon. We also now have a very minimal [example script](https://github.com/merbanan/rtl_433/blob/master/examples/rtl_433_influxdb_relay.py).

zuckschwerdt:
We now have native InfluxDB support with #1192. Documentation will soon be updated to show all possibilities to interface.

zuckschwerdt:
Documentation on some options for InfluxDB went in with dac4e63.

PR #1160, opened 2019-10-04
Convert read OOK pulse_data to current sample rate

halfbakery:
The saved OOK dumps weren't been decodable due to the sampling rate mismatch. Upon dump it is converted to usec based values, so we should convert back during load.

### DIFF
--- a/include/pulse_detect.h
+++ b/include/pulse_detect.h
@@ -68,7 +68,7 @@ void pulse_data_print_vcd_header(FILE *file, uint32_t sample_rate);
 void pulse_data_print_vcd(FILE *file, pulse_data_t const *data, int ch_id);
 
 /// Read the next pulse_data_t structure from OOK text.
-void pulse_data_load(FILE *file, pulse_data_t *data);
+void pulse_data_load(FILE *file, pulse_data_t *data, uint32_t sample_rate);
 
 /// Print a header for the OOK text format.
 void pulse_data_print_pulse_header(FILE *file);
--- a/src/pulse_detect.c
+++ b/src/pulse_detect.c
@@ -98,14 +98,15 @@ void pulse_data_print_vcd(FILE *file, pulse_data_t const *data, int ch_id)
         fprintf(file, "#%.f 0/\n", pos * scale);
 }
 
-void pulse_data_load(FILE *file, pulse_data_t *data)
+void pulse_data_load(FILE *file, pulse_data_t *data, uint32_t sample_rate)
 {
     char s[256];
     int i    = 0;
     int size = sizeof(data->pulse) / sizeof(int);
 
     pulse_data_clear(data);
-    data->sample_rate = 1000000; // assumes 1us timescale
+    data->sample_rate = sample_rate;
+    double to_sample = sample_rate / 1e6;
     // read line-by-line
     while (i < size && fgets(s, sizeof(s), file)) {
         // TODO: we should parse sample rate and timescale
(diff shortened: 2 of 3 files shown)


zuckschwerdt:
Great catch. Thanks! You must been one of the first few people actually using the OOK feature. Is it working well otherwise? Do you have a good use-case as example for others maybe?

halfbakery:
Most of the ESP8266/ESP32 based projects lack of a good RF433 support decoding complex protocols used by thermometers, wheather stations etc. Your project has a fantastic support of them, but for simple OOK based protocols the whole software radio stack is quite heavyweight.

My idea is to build a system where an Arduino/ESP8266 or ESP32 with a simple 433 OOK decoder does the message separation, and transmits the result via network in OOK format to rtl_433 for further processing. The ESP side could be quite simple and maintainance-free, and the rtl_433 could be used anywhere on the network, even in the cloud.

```
$ socat OPEN:/dev/ttyACM0,b115200,echo=0,echoe=0,echok=0,echoctl=0,echoke=0,onlcr=0,raw - | ./src/rtl_433 -r ook:-
rtl_433 version unknown inputs file rtl_tcp RTL-SDR
Use -h for usage help and see https://triq.org/ for documentation.
Trying conf file at "rtl_433.conf"...
Trying conf file at "/home/halfbakery/.config/rtl_433/rtl_433.conf"...
Trying conf file at "/usr/local/etc/rtl_433/rtl_433.conf"...
Trying conf file at "/etc/rtl_433/rtl_433.conf"...

        Consider using "-M newmodel" to transition to new model keys. This will become the default someday.
        A table of changes and discussion is at https://github.com/merbanan/rtl_433/pull/986.

Registered 108 out of 138 device decoding protocols [ 1-4 8 11-12 15-17 19-21 23 25-26 29-36 38-60 63 67-71 73-100 102-103 108-116 119 121 124-128 131-138 ]
Test mode active. Reading samples from file: <stdin>
time      : @0.000000s
model     : WT0124 Pool Thermometer                Random ID : 74
Channel   : 1            Temperature: 22.2 C       Integrity : CHECKSUM
Data      : 60
```

merbanan:
Ideally add an input protocol for this.

zuckschwerdt:
@merbanan we have that :) It's the `rtl_433 -r ook:-` example posted.

@halfbakery our OOK format and support is new. If you encounter something odd just request that we change things ;)

One thing to note: I did a test with one of those cheap 1-bit modules (basically a little more than an open wire on a digital pin). The missing framing is rather bad, rtl_433 isn't made to scan a continuous bit stream for patterns. You would need to detect a preamble and only then request decoding from rtl_433.
We might also, some day, offer the decoder library as a package for inclusion in other projects. You would need some scaffolding for the data output though.

zuckschwerdt:
@merbanan or likely you are already one step further. Sure we could add something like the rtl_tcp input for OOK data.

merbanan:
@zuckschwerdt yeah, that was what I thought of. Maybe even something with framing capability, perhaps listening on a multicast address. Then the devices can just broadcast whenever data is received and rtl_433 can handle it as a regular wireless signal.

halfbakery:
@zuckschwerdt My current arduino sketch is doing exactly this. Searching for frame candidates between two gaps with some heuristics. My idea is exactly the same as yours: adding something like rtl_tcp or implement something over MQTT if my proof of concept would work.

halfbakery:
Proof of concept sketch is available at https://github.com/halfbakery/Arduino-RF433-OOK-test

Issue #1161, opened 2019-10-07
Add support for Lidl Auriol hg02832


OverSave:
Hello, Please add the device Lidl Auriol hg02832.

`rtl_433 -r g001_433.93M_250k.cu8 -a -A
rtl_433 version 19.08-18-g8eecdbb branch master at 201909241811 inputs file rtl_tcp RTL-SDR
Use -h for usage help and see https://triq.org/ for documentation.
Trying conf file at "rtl_433.conf"...
Trying conf file at "/home/pi/.config/rtl_433/rtl_433.conf"...
Trying conf file at "/usr/local/etc/rtl_433/rtl_433.conf"...
Trying conf file at "/etc/rtl_433/rtl_433.conf"...
Consider using "-M newmodel" to transition to new model keys. This will become the default someday.
A table of changes and discussion is at https://github.com/merbanan/rtl_433/pull/986.
Registered 108 out of 138 device decoding protocols [ 1-4 8 11-12 15-17 19-21 23 25-26 29-36 38-60 63 67-71 73-100 102-103 108-116 119 121 124-128 131-138 ]
Test mode active. Reading samples from file: g001_433.93M_250k.cu8
Detected OOK package    @0.211772s
Analyzing pulses...
Total count:   44,  width: 40.22 ms             (10056 S)
Pulse width distribution:
 [ 0] count:   21,  width:  252 us [244;308]    (  63 S)
 [ 1] count:    3,  width:  860 us [852;864]    ( 215 S)
 [ 2] count:   20,  width:  616 us [608;620]    ( 154 S)
Gap width distribution:
 [ 0] count:    4,  width:  848 us [840;856]    ( 212 S)
 [ 1] count:   20,  width:  600 us [600;608]    ( 150 S)
 [ 2] count:   19,  width:  236 us [236;248]    (  59 S)
Pulse period distribution:
 [ 0] count:    1,  width: 1148 us [1148;1148]  ( 287 S)
 [ 1] count:    3,  width: 1708 us [1700;1720]  ( 427 S)
 [ 2] count:   39,  width:  852 us [848;860]    ( 213 S)
Level estimates [high, low]:   1676,     17
RSSI: -9.9 dB SNR: 19.7 dB Noise: -29.6 dB
Frequency offsets [F1, F2]:     566,      0     (+2.2 kHz, +0.0 kHz)
Guessing modulation: Pulse Width Modulation with sync/delimiter
[...]

PR #1166, opened 2019-10-14
Add Auriol-HG02832 support

zuckschwerdt:
S.a. #1161

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ Read the Test Data section at the bottom.
     [138]  Globaltronics GT-WT-03 Sensor
     [139]  Norgo NGE101
     [140]  Elantra2012 TPMS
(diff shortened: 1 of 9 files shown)


zuckschwerdt:
@OverSave please test if this works.

merbanan:
No need to hold this up. Commit at will.

OverSave:
looks like if everything works, thanks good and fast work

OverSave:
have now found a small bug, sometimes the temperature 00001 is displayed too much. Example 16.700001

zuckschwerdt:
That's just how a float works. But KV and CSV should only output a single decimal, JSON three decimals. Where do you see 6 decimals?

zuckschwerdt:
The one place I didn't check. Good catch, fixed with c124af4 now.

OverSave:
alright thanks

PR #1167, opened 2019-10-14
Update brew packages in macOS

septs:
Command Line Tools for Xcode in macOS Catalina, Not include `pkg-config`.


zuckschwerdt:
Thanks!

PR #1170, opened 2019-10-17
Add preamble check for microchip hcs200 to reduce false positives

Mindavi:
The microchip decoder gives a reasonable amount of false positives on messages that don't have the preamble mentioned in the source file. I've added a check that checks that preamble. This seems to work fine with the provided files in the rtl_433_tests project, while reducing false positives.

Something that would trigger the decoder would be

`rtl_433 -y {78}f0f82145d6685ec6505b767c178e75caeb65ec9a3b23d7580f54711647c0e67b019c97b82c64e407330870b524397e7aadc355f5cd4af93aa3a0282dc7d1ea193ef30c18de7fa28fbafd0bc2e514 -v`

With this patch it does not trigger anymore, it will now trigger _only_ if the first 12 bits are ones.

### DIFF
--- a/src/devices/hcs200.c
+++ b/src/devices/hcs200.c
@@ -26,6 +26,10 @@ rtl_433 -R 0 -X 'n=name,m=OOK_PWM,s=370,l=772,r=14000,g=4000,t=152,y=0,preamble=
 
 #include "decoder.h"
 
+// the preamble mentioned above, 0xfff (12 bits)
+static const uint8_t preamble_bits = 12;
+static const uint8_t preamble[2] = {0xff, 0xf0};
+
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Wouldn't it be easier to just check the first two bytes with a simple mask instead of the full bitbuffer_search if we demand a match at zero anyway? If not can you move the preamble definition inside the decoder function (like [honeywell](https://github.com/merbanan/rtl_433/blob/master/src/devices/honeywell.c#L43)), otherwise the doc-comment will attach to the preamble not the function itself.

Mindavi:
Yeah that's ok too, I don't think it matters much which of those are chosen. I just thought it would be cleaner to use the search. I'll update it to just check the first bytes with a mask.

Issue #1174, opened 2019-10-23
Bug in elantra tpms decoder


Mindavi:
Reproduced by: `rtl_433 -y '{480} 9c 55 5c 55 5c 55 5c 55 5c 55 55 55 55 55 55 55 55 55 55 55 55 55 55 55 55 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 00 00 00 00 00 00 00 00 00 00 00 00 00 00'`.

The problem is that the decoder counts all the events together and returns that as 1 value. This shouldn't be a problem if 0 is nothing found and 1 is something found, but the new return values count up as a big negative number. This means that the elantra tpms is doing it wrong, but the events accounting is also not defensive enough.

zuckschwerdt:
I also want to simplify that two stage pattern (pattern find and then calling the real decoder).

zuckschwerdt:
`exit(1)` would also be fine, it's a condition that must not happen.

PR #1176, opened 2019-10-23
Fix elantra tmps return value with incorrect decodes

Mindavi:
Test case: `rtl_433 -y '{480} 9c 55 5c 55 5c 55 5c 55 5c 55 55 55 55 55 55 55 55 55 55 55 55 55 55 55 55 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 c5 55 00 00 00 00 00 00 00 00 00 00 00 00 00 00'`

I'm not sure if this is ideal or if another way might be used to bubble up the accounting events. At least it doesn't misbehave any more.

See  #1174


zuckschwerdt:
Better to fix it now, and make it pretty later ;)

PR #1181, opened 2019-10-25
Fix bug where the radiohead decoder tried to read at position -1 of an array

Mindavi:
reproduction test set:
[radiohead_ask_crashes.zip](https://github.com/merbanan/rtl_433/files/3773132/radiohead_ask_crashes.zip)

or one of these bitbuffers:
`rtl_433 -y '{62} aa aa aa aa a3 9b 63 80'`
`rtl_433 -y '{190} aa aa aa aa a3 9b 63 96 59 65 d9 65 96 2d 8a d3 2c aa ca ac 5a 5d 85 80'`
`rtl_433 -y '{64} aa aa aa aa 39 b6 39 60'`
... (and more, but this should suffice)


zuckschwerdt:
Great stuff, as always!

PR #1185, opened 2019-10-31
Fix issue where bt_rain decoder uses -1 as index

Mindavi:
reproduce by doing rtl_433 -G -y {0}0.
ASAN will warn about this.
```
rtl-sdr/433mhz/rtl_433/src/devices/bt_rain.c:39:9: runtime error: index -1 out of bounds for type 'uint16_t [25]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior rtl-sdr/433mhz/rtl_433/src/devices/bt_rain.c:39:9 in 
rtl-sdr/433mhz/rtl_433/src/devices/bt_rain.c:39:53: runtime error: index -1 out of bounds for type 'uint16_t [25]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior rtl-sdr/433mhz/rtl_433/src/devices/bt_rain.c:39:53 in 
```

I've checked other decoders for this, but this is the only offender.
Also updated return codes while I was at it.

### DIFF
--- a/src/devices/bt_rain.c
+++ b/src/devices/bt_rain.c
@@ -35,14 +35,16 @@ static int bt_rain_callback(r_device *decoder, bitbuffer_t *bitbuffer) {
     float temp_c, rainrate;
 
     row = bitbuffer_find_repeated_row(bitbuffer, 4, NUM_BITS);
+    if (row < 0)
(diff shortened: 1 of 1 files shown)


merbanan:
Thanks.

PR #1186, opened 2019-11-01
fix bug where no closing brace would cause a crash

Mindavi:
This happens because the length check is only done after checking if there's a closing brace. Moved the length check so it's always normalized.

test case `rtl_433 -G -y {970000`


zuckschwerdt:
Thanks.

Issue #1114, opened 2019-07-20
WS4945 showing "DSC row 0 invalid bit count 11"


garyvdm:
Thank you for the awesome work you have done on this project

I'm trying to pickup a bunch of DCS WS4945 transmitters that were installed at a house I bought.

![IMG_20190720_154723](https://user-images.githubusercontent.com/352175/61579507-393d8f80-ab06-11e9-9ca1-f567d593625d.jpg)

Here is a recording from `rtl_433 -S all`. 
[g001_433.92M_250k.zip](https://github.com/merbanan/rtl_433/files/3413680/g001_433.92M_250k.zip)

When I run `rtl_433 -vvv -R 23 g001_433.92M_250k.cu8`, I see the following line a few times in the output: 

    DSC row 0 invalid bit count 11

I do not get any decoded signals.

garyvdm:
From some info from #645 , I was able to get it to working with this patch:

```
diff --git a/src/devices/dsc.c b/src/devices/dsc.c
index 05bb383..8a08b00 100644
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -220,9 +220,9 @@ static char *output_fields[] = {
 r_device DSC = {
     .name          = "DSC Security Contact",
     .modulation    = OOK_PULSE_PCM_RZ,
-    .short_width   = 250,    // Pulse length, 250 µs
-    .long_width    = 500,    // Bit period, 500 µs
-    .reset_limit   = 5000, // Max gap,
+    .short_width   = 536,    // Pulse length, 536 µs
+    .long_width    = 1072,    // Bit period, 1072 µs
+    .reset_limit   = 6000, // Max gap,
     .decode_fn     = &dsc_callback,
     .disabled      = 0,
     .fields        = output_fields,
```

merbanan:
Send a PR for one more r_device with the correct timings. Wrap the callback so that different model names can be used. This device can use "DCS WS4945" as model names.

garyvdm:
Ok. I'll give it a try.

dxdc:
@garyvdm @merbanan I submitted a PR for this. I'm having some other issues with 2 DSC contacts (others are working fine) so need to do some more troubleshooting. May be unrelated.

merbanan:
Fix committed. Regarding the 2 contacts the signal quality is lacking. Maybe try offsetting the frequency a little bit. It was pretty close to dc.

PR #1188, opened 2019-11-01
Fix for #1114

dxdc:
WS4945 showing "DSC row 0 invalid bit count 11"

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -31,6 +31,7 @@
     DECL(calibeur_RF104) \
     DECL(X10_RF) \
     DECL(dsc_security) \
+    DECL(dsc_security_ws4945) \
     DECL(brennenstuhl_rcs_2044) \
     DECL(gt_wt_02) \
     DECL(danfoss_CFR) \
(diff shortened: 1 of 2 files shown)


merbanan:
The patch is incomplete, the decoder registration is missing and I'd rather use a different callback so we can differentiate between the models. Can you grab some signals also so we can add those to the test repo?

dxdc:
@merbanan thanks, just diving into the world of 433 and really love this tool.

> decoder registration is missing 
Fixed

> I'd rather use a different callback so we can differentiate between the models
I'm not sure this is needed? I think the decoding is the same, just a different pulse width/configuration. It seems like a waste to duplicate the code; although, perhaps there is a way to reference a model# as part of the callback. If you're able to offer some guidance on the best way to do that I'm happy to make those edits. I didn't see other device set up like that, however.

> Can you grab some signals also so we can add those to the test repo?
I'm attaching 3 files:
1. 'g001' from @garyvdm, which is what this patch was originally designed for. It doesn't seem to help my situation though.
2. 'g066' and 'g068' from me.

I'm having a particular problem with g066/g068 in that the `closed` state is not reporting reliably. Only 1 out of 10+ transmissions seems to be picked up by rtl_433. This is just an issue with 2 out of about 16+ sensors. 

Based on the details in #1114, I'm wondering if (in general?) the pulse timings need to be better tuned? I'd be grateful if you're able to share why the g066 and g068 signals are so unreliable for me.

[cu8_files.zip](https://github.com/merbanan/rtl_433/files/3800915/cu8_files.zip)

PR #1191, opened 2019-11-01
fsk: add Y option for selecting different fsk low pass coefficients

merbanan:
Prepatory work for new fsk pulse detector.

FSK demod low pass filter selection.

### DIFF
--- a/include/baseband.h
+++ b/include/baseband.h
@@ -63,10 +63,10 @@ void baseband_low_pass_filter(uint16_t const *x_buf, int16_t *y_buf, uint32_t le
     @param len: number of samples to process
     @param[in,out] state: State to store between chunk processing
 */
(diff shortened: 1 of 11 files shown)


merbanan:
@zuckschwerdt this is as clean as I could make it. The -r sample file option will not work though. I think the file name parser needs to set frequency and sample rate cfg values to get it to work the same.

merbanan:
Anyway the code works, there are some tiny changes before I can merge. I just want overall input of the approach.

merbanan:
With this patch several decoders should be able to activate by default.

zuckschwerdt:
Overall looks tidy enough. The `-r` option should be fixable later.

merbanan:
Ok, the is now fairly useful. I'll merge it now and improve it later.

PR #1192, opened 2019-11-02
Add InfluxDB output

palstek:
This commit adds direct support for InfluxDB line protocol (https://v2.docs.influxdata.com/v2.0/reference/syntax/line-protocol/). I tested it with InfluxDB 2.0 alpha 17-19. But it should work with InfluxDB 1.x too.

Support for TLS encryption is currently missing. How complicated/easy is it to enable this support in Mongoose?

Any comments are welcome.

Best regards,
  Daniel

### DIFF
--- a/include/output_influx.h
+++ b/include/output_influx.h
@@ -1,0 +1,21 @@
+/** @file
+    InfluxDB output for rtl_433 events
+
+    Copyright (C) 2019 Daniel Krueger
+    based on output_mqtt.c
+    Copyright (C) 2019 Christian Zuckschwerdt
+
+    This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


zuckschwerdt:
Looks very interesting and well written, thanks!
I'll review and test, but no obvious things stand out at first glance ;)

merbanan:
Looks good to me.

zuckschwerdt:
Should other fixed meta-data also be tags? E.g. `mic`(Integrity) and `mod`(Modulation) are not really "values". What do you think @palstek?

palstek:
I think that tags are values which identify different sensors. Fields are values which are measured in some way. `mic` and `mod` are measurement values. Ok, they won't change. However, in theory they might.

PR #1201, opened 2019-11-15
InfluxDB output module was missing from the traditional compilation

halfbakery:
The new InfluxDB output module was only added to cmake, but missing from the traditional make toolchain.


zuckschwerdt:
Oops. Thanks for the heads-up. I forgot to run `maintainer_update.py`. I'll update all files right away.

PR #1203, opened 2019-11-19
Fix Acurite 899 rain_mm conversion value

ke6zfi:
The 899 documentation rounded the conversion value. The 899 appears to measure in 0.01" per count. Using 0.254 now matches the display even after a large number of count. The print output was wrong with the %d.


zuckschwerdt:
Great! Thanks! Can you also update the comment a few lines above that to read:
` // one tip is 0.01 inch, i.e. 0.254mm`

ke6zfi:
@zuckschwerdt - Comment corrected.

Issue #1206, opened 2019-11-25
ISO-8601 shows no time zone


LiberationFrequency:
Is it possible to implement the time zone for -M time:iso . This way you cannot see at what time the dataset was created. You may want to display the local time, but still use the data across different time zones. The best format would be time+/-hh:mm and timeZ for UTC. If this is not possible due to portability, then it would be nice if at least the Z is written behind for UTC.
Thanks in advance.

zuckschwerdt:
Checkout and discuss in #1207

PR #1207, opened 2019-11-25
Add timezone offset print option

zuckschwerdt:
Picks up the idea from #1206 to add the timezone offset to ISO-8601 outputs.
The offset will always be `+1234` (never `Z`).

Currently optional with `tz` flag, perhaps `iso` should always output TZ?

### DIFF
--- a/include/r_util.h
+++ b/include/r_util.h
@@ -24,7 +24,7 @@
     #endif
 #endif
 
-// buffer to hold localized timestamp "YYYY-MM-DD HH:MM:SS.000000"
+// buffer to hold localized timestamp "YYYY-MM-DD HH:MM:SS.000000+0000"
 #define LOCAL_TIME_BUFLEN 32
 
 /** Get current time with usec precision.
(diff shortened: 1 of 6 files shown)


LiberationFrequency:
It's okay, but the Z is also defined in the standard. In my humble opinion it would be better to take a colon ±00:00 as separator between hours and minutes just for the sake of beauty. ±0000 is also allowed, but it sounds a little bit military.

zuckschwerdt:
This output is straight from `strftime`'s `%z`. We could transform the output, or use a value from `tzset`. But I'd rather not put too much effort into that unless it doesn't conform.

Generally I'd recommend UTC if you need to consider different zones.

LiberationFrequency:
Like I said, ±0000 is perfectly all right. Complies with the standard. Don't bother! Thanks.

zuckschwerdt:
Are you primarily interested in the current timezone offset or in tagging UTC ("Z" or "+0000")?

LiberationFrequency:
For the sake of completeness, it would be ideal if the offset to UTC were always displayed. This is the advantage of ISO-8601 that it is an international date format. But I'd be happy if  only UTC was tagged. Whether with "Z" or "+0000" does not matter.

PR #1216, opened 2019-11-30
Add pulse analyzer support for read OOK data

halfbakery:
The pulse analyzer was not available on OOK data read from a file. This patch adds support for it.


merbanan:
Looks good to me.

PR #1229, opened 2019-12-08
Wrong measures with Efergy Optical

cescp:
Efergy Optical transmitter can operate in three modes:
- red led: information is sent every 30s
- orange led: information is sent every 60s
- green led: information is sent every 90s

To get the mode: short-push the physical button on transmitter.
To set the mode: long-push the physical button on transmitter.

The mode information is present in byte[3]:
- red led (every 30s):    bytes[3]=64 (0100 0000)
- orange led (every 60s): bytes[3]=80 (0101 0000)
- green led (every 90s):  bytes[3]=96 (0110 0000)

This information was not taken into account. This pull request solves this issue.

Also added: entry for 4000 impulses/kwh.

### DIFF
--- a/src/devices/efergy_optical.c
+++ b/src/devices/efergy_optical.c
@@ -13,16 +13,23 @@ Efergy IR is a devices that periodically reports current energy consumption
 on frequency ~433.55 MHz. The data that is transmitted consists of 8
 bytes:
 
-- Byte 0-3: Start bits (0000), then static data (probably device id)
-- Byte 4-6: all zeros
-- Byte 7: Pulse Count
-- Byte 8: sample frequency (15 seconds)
-- Byte 9: seconds
+- Byte 0-2: Start bits (0000), then static data (probably device id)
+- Byte 3: seconds (64: 30s - red led; 80: 60s - orange led; 96: 90s - green led)
+- Byte 4-7: all zeros
+- Byte 8: Pulse Count
+- Byte 9: sample frequency (15 seconds)
 - Byte 10-11: bytes 0-9 crc16 xmodem XOR with FF
 
 if pulse count <3 then energy =(( pulsecount/impulse-perkwh) * (3600/seconds))
 else  energy= ((pulsecount/n_imp) * (3600/seconds))
 
+Transmitter can operate in 3 modes (signaled in bytes[3]):
+- red led: information is sent every 30s
+- orange led: information is sent every 60s
+- green led: information is sent every 90s
+
+To get the mode: short-push the physical button on transmitter.
+To set the mode: long-push the physical button on transmitter.
 */
 
 #include "decoder.h"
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Very nice, thanks!
- `pulsecount` should be an `int`
- `"pulses", "Pulse-rate",     DATA_FORMAT, "%i", DATA_INT, imp_kwh[i],` -> `"pulses", "Pulse-rate",     DATA_INT, imp_kwh[i],`
- `"pulsecount",  "Pulse-count",   DATA_FORMAT, "%.01f", DATA_DOUBLE, pulsecount,` -> `"pulsecount",  "Pulse-count",   DATA_INT, pulsecount,`
- add `pulsecount` to `output_fields`

cescp:
I think that given values should be Power (kW) instead of Energy (kWh), as they are a measure for instant consumption, not cumulated energy.
As seen in https://efergy.com/e2-classic/ the mode "Energy now" (the one captured here) gives kW.
See also: https://en.wikipedia.org/wiki/Kilowatt_hour#Confusion_of_kilowatt_hours_(energy)_and_kilowatts_(power)

zuckschwerdt:
Looks good for merge. One minor thing: please use 4 spaces for indent, not tabs ;)

zuckschwerdt:
Let's get this merged then open another PR to change energy to power. Note that it's a breaking change in the output, thus needs to sit open for a while for other people to notice and maybe get a word in ;)

merbanan:
@cescp Power needs 2 readings and the time duration between them. The over the air data is pulses, and pulses are kWh*factor. How do you get Power from that without a state in the decoder?

cescp:
Values read from transmitter are number of pulses (energy) during a certain amount of time: energy/time = power (kW·h/h = kW)
For example, if transmitted values are 50 pulses (pulsecount=50) during 90 seconds (seconds=90), the calculation is:
(50 pulses / 90 s) * (3600 s / 1 h) * (1 kW·h / 4000 pulses) = 0.5 kW

zuckschwerdt:
First of: it is a confusing topic, so no blame and no guarantees.

It's called an energy meter, not a power meter. Could be a marketing term but I'd say it's the literal function description. Measuring energy means integrating power over time -- which is the main task of meters.

Power is only a useful quantity if paired with a duration. Instant power would only help to gauge your cables and fuses. So we really do want an energy (work) reading if possible.

And I guess we are getting that, the report is a number of impulses at a set factor of impulses per kW/h.

What I don't grasp is why we scale that with `3600/seconds`. That would in fact reduce the reading to average power over an arbitrary time. So, yes, the current calculation smells like power, it should be however.

(My energy meter "ticks" away at 1W/h used -- I'm getting those real-time, but could sum those up over some reporting period without changing the units.)

PR #1230, opened 2019-12-08
Add Auriol AFW2A1 support

LiberationFrequency:
This adds support for the Lidl Auriol AFW 2 A1 sensor. (IAN 311588)
Test files were requested to be merged. (rtl_433_tests/pull/319)


src/devices/auriol_afw2a1.c
```diff
@@ -0,0 +1,178 @@
...
+
+    Copyright (C) 2019 LiberationFrequency
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+/*
```
zuckschwerdt: should be `/**` so it works as doc-tag

src/devices/auriol_afw2a1.c
```diff
@@ -0,0 +1,178 @@
...
+    int channel;
+    int battery_low = 1;
+    int tx_button = 1;
+    int16_t temp_raw;
+    float temp_c;
+    int humidity;
+
+    row = bitbuffer_find_repeated_row(bitbuffer, 12, 36);
+    if (bitbuffer->num_rows != 12)
+        return DECODE_ABORT_EARLY;
```
zuckschwerdt: Do you really need this? It would easily break if the reception isn't prefect.

src/devices/auriol_afw2a1.c
```diff
@@ -0,0 +1,178 @@
...
+        battery_low = 0;
+        tx_button = 1;
+    }
+
+    /* clang-format off */
+    data = data_make(
+            "model",            "",                  DATA_STRING, "Auriol-AFW2A1",
+            "id",               "",                  DATA_INT,    id,
+            "channel",          "Channel",           DATA_INT,    channel,
+            "battery",          "Battery",           DATA_STRING, battery_low ? "LOW" : "OK",
```
zuckschwerdt: make this `"battery_ok" ... !battery_low,`

src/devices/auriol_afw2a1.c
```diff
@@ -0,0 +1,178 @@
...
+            "button",           "Button",            DATA_STRING, tx_button ? "true" : "false",
```
zuckschwerdt: make this `DATA_INT, tx_button,`

zuckschwerdt:
Overall looks good. I like that you took time to get our coding style right :) (and presumably ran maintainer_update.py)
Can you split the flags this way, saves a lot of code:
```
     battery_ok = (b[1] & 0x80) >> 7; 
     tx_button  = (b[1] & 0x40) >> 6; 
     channel    = (b[1] & 0x30) >> 4; 
```

src/devices/auriol_afw2a1.c
LiberationFrequency: No, we can leave that out, if you say so. It should prevent that unnecessarily wrong sensors are read out.

zuckschwerdt: Usually if you expect to see e.g. 12 rows we search for 7 identical ones (more than half).
But since there is no checksum you can be a little more restrictive (exactly 12 x 36 seems to fragile though)

LiberationFrequency:
Okay, it's done. Please take a look into the battery_low issue. It works, but I wrote it vice versa, because I run into trouble.

zuckschwerdt:
Is `battery_low == battery_ok` really intended? I.e. is the bit set on low condition (battery_low flag), or is it 0 on low condition (battery_ok flag).

Sorry about confusion with the check. I meant to say: remove the `bitbuffer->bits_per_row[3] != 36` stuff, keep a `row = bitbuffer_find_repeated_row(bitbuffer, 7, 36); `.

LiberationFrequency:
No, it's not supposed to be that way. But I don't know how to write it when I split the flags. I think 0 should stand for low, and 1 for ok. 
That's the first time I've ever written anything in C. 
I'll change that with the bitbuffer tomorrow.

zuckschwerdt:
Your first-time C looks very clean. Just use `battery_ok` as variable name instead of `battery_low` to make it clear. Looks fine otherwise.

LiberationFrequency:
Thanks, that makes me proud. It was my first sensor and my first C. I've learned a lot in the process.
I changed that, but I'm pretty sure we should do some validation checks to early abort to rule wrong sensors out. This produces a lot of "Not an Auriol-AFW2A1 device" errors.

andrerodr:
I have compiled LiberationFrequency's code and I wasn't able to reproduce "Not an Auriol-AFW2A1 device" error a single time, it just seems to work perfectly.

zuckschwerdt:
@andrerodr Thanks for the testing!

PR #1242, opened 2019-12-28
Added support for FineOffset/ECOWITT WH51

0x4d0x44:


merbanan:
mic needs to be added.

zuckschwerdt:
Note:
- mic needs to be added to `output_fields_WH51` too.
- key order should be "model", "id", "battery", "moisture", "boost", "ad", "mic" (i.e. battery up, boost down).
- battery should be battery_mV.
- `ad` is rather short, there might be a better key name.
- `sprintf( id, "%2.2x%2.2x%2.2x", ` should be  `sprintf(id, "%2x%2x%2x", `

merbanan:
Committed a fix.

zuckschwerdt:
We don't have samples for this?

PR #1246, opened 2019-12-30
Update m_bus.c with device type strings from OMS

antonmeyer:
Update m_bus.c with device type strings from OMS


merbanan:
Looks good, thnx.

Issue #909, opened 2018-12-11
Add option for GPS coordinates in messages


merbanan:
There should be an option for static or dynamic GPS coordinates to be added to the messages. If you have a fixed sdr receiver then adding the coordinates statically to a config would be fine. If you are moving reading values from a NMEA compatible GPS over a serial device should be supported. And we need a webservices method for android to be able to update the position on that platform as reading the coordinates from users space seems to be hard.

zuckschwerdt:
I have some early gpsd-based code here https://github.com/zuckschwerdt/rtl_433/tree/feat-gpsd
I just need to verify and clean that. Also the caveat here is that building with gps support needs C11.

merbanan:
Code looks good. I do think that we should add the gps parameters statically and copy the values if libgps support is added. That way we can update the values via a webservice as the android location api not only derive the location based on gps only. And static configuration would be possible.

merbanan:
It would be nice with a read position from gps command line option and output the text needed for static configuration. That way the setup of a receiver could be very straight forward.

gdt:
I second the use of gpsd rather than reinventing it.

A complexity is that it would be nice to get gps timestamps before and after and interpolate, assuming the system clock is synchronized.  This could delay messages up to a second or so, but get better locations.

zuckschwerdt:
We have some experimental code for gpsd. I guess the question was more towards a push than a pull data flow for gps. I.e. We could have some gps provider (gpsd, api, webservice) push the info to use, rather than the output pulling it from exactly gpsd. I'll rework the code in that direction.

zuckschwerdt:
The idea was not to tie the gps output to one source (gpsd) but leave that open for alternative ways to acquire gps data. (push/pull meaning the internal call chain, not an external interface.)

I don't have gps hardware so I'm speculating, but I guess the libgps/gpsd timestamps related to the instant they were read from the hardware not the time we read and output them?

Any help here to get things right and improve accuracy is welcome!

gdt:
[...]

PR #1250, opened 2020-01-03
GPSd support

mckaymatthew:
This pull request partially address #909. I understand the desire for a more flexible approach for alternative data sources as discussed in the issue, but this covers many use cases and is fairly reasonable. It is also fairly loosely integrated and doesn't lock future work. 

Please take a look and provide feedback. 

When compiled, and when enabled, messages received have the GPS Status, Fix Mode, and LLA appended to it. 

A couple of notes I have:
- Optionally compiled support. Takes the "AUTO" approach SoapySDR support uses. 
- Command line argument, -u requires a "hostname:port" argument. 
- Failure cases considered:
  - Startup behavior is to exit if gpsd connection is not successful. 
  - Runtime behavior is to auto-reconnect if gpsd connection was lost. This does print to stderr at runtime.
- Builds on all my platforms, all tests pass
  - Centos 7. x86_64. 1663 records tested, 0 have failed
  - Ubuntu 18.04.3 LTS. x86_64. 1663 records tested, 0 have failed
  - Raspbian GNU/Linux 9. armv7l. 1663 records tested, 0 have failed


zuckschwerdt:
Thanks. It'll take some time for me to review, lots of backlog.
One first note: if not for legacy reasons CMake switches should not be a magic AUTO. It is reasonable to require people to set a YES and then break if the lib isn't found.

src/rtl_433.c
```diff
@@ -1054,6 +1063,35 @@ static void parse_conf_option(r_cfg_t *cfg, int opt, char *arg)
...
+        if (arg && strlen(arg) != 1) {
+            //arg = arg + 1;
+            int i, count;
+            for (i = 0, count=0; arg[i]; i++) {
+              count += (arg[i] == ':');
+            }
+            if(count != 1) {
+                (void)fprintf(stderr,"GPS requires a host and port.\n");
+                usage(1);
+            }
```
merbanan: How about to default to localhost and port 2947 if no arguments are given?

merbanan:
One more comment is the use of ifdefs. I really dislike them in common code but I'm not sure it is worth the complexity to get rid of them. Anyway the implementation looks clean.

src/rtl_433.c
gdt: I agree that defaulting to localhost:2947 is good.   My guess at the normal use case is a laptop that is moving, so having gpsd on the same laptop is the obvious approach.

mckaymatthew:
Thanks for the quick feedback. 
- I made it a "NO" option by default.
- I added a small section to BUILDING.md mentioning the option.
- I agree ifdefs are undesirable. The other option I considered was runtime dynamic linking but this has its own problems. Sounds like ifdef is fine for now.
- Made -u have an optional argument. I've never used getopt before... this proved to be confusing to implement. 
- Fixed stylecheck error.

src/rtl_433.c
alphafox02: @merbanan it doesn’t seem like this was ever added to main? Was there not enough interest or need?

alphafox02: Actually I didn’t read closely enough 
https://github.com/merbanan/rtl_433/pull/1636

PR #1259, opened 2020-01-10
Update BUILDING.md

jdwhite:
CentOS/Fedora/RHEL require the libusbx and libusbx-devel packages to build. The libusb-devel package lacks the function definition for libusb_error_name() in libusb.h required to build rtl_433.


zuckschwerdt:
Thanks. The analysis is interesting. Seems we only recently introduced that dependency with 14c5b9a. Maybe we need to find a different way there?
Can you list the versions of libusb and libusbx used, is libusbx perhaps just newer?

Issue #1062, opened 2019-05-21
Add fuzz tests


mhansen:
I think this project could get a lot of use out of fuzz tests - tests that send random input to the decoders and check they don't crash or segfault while running under ASAN.

Given that rtl_433:
- accepts untrusted input (over the radio)
- passes the input to over 100 decoders
- each decoder is written in C

I think fuzz tests might find some issues pretty quickly?

I imagine this would be a bit of work to set up, but wanted to file an issue to track any ideas. The first place I'd try: AFL Fuzz, which is pretty neat - AFL instruments the binary to discover new inputs that reach new internal binary states http://lcamtuf.coredump.cx/afl/

zuckschwerdt:
Sounds interesting. There is already the `-y` option to feed arbitrary bitbuffers to the decoders. That mechanism could easily be used programmatically. I already planned to use that for quick(er) regression tests on each build -- the full regression tests with rtl_433_tests are to heavy and slow for that.

mhansen:
That's pretty cool, I didn't see the -y button. Sounds like a good addition to the regression tests for sure.

Mindavi:
I created a small script that uses the -y option to mash random data into rtl_433. This helped me find some decoders that were greedy (matched on bogus data) and fix them. It's not a smart script but might help with finding bugs related to decoders that match on too many inputs.

Of course, some decoders don't have any checksums so will happily accept any data at all.

https://gist.github.com/Mindavi/3b972526384dac1248e4aded49e8aff0

zuckschwerdt:
Great to see this finally done! Some devices will only trigger with N identical bitbuffers, some only with a minimum number of rows, some need a first row with sync (something like a single bit). Maybe some ideas to explore?

mhansen:
This is really cool, Rick! I'm impressed that you found some greedy
decoders so quickly.

Mindavi:
I'm otherwise not very familiar with instrumented fuzzing, but if I find the time I'd really like to learn how to do it and implemented it for this application. I'm sure it will be helpful in finding some interesting edge cases.

Mindavi:
Well, according to the interwebz afl should run a lot faster, but it does find stuff anyway. These are my first results after just a couple of minutes:

```
rick@rick-GL552VW:~/hobby/rtl-sdr/433mhz/afl-fuzz$ afl-whatsup output/
status check tool for afl-fuzz by <lcamtuf@google.com>

Individual fuzzers
==================

[...]

PR #1263, opened 2020-01-12
Add test pattern and demo mode

zuckschwerdt:
This adds test pattern to each decoders r_device and a demo mode option (-D) to run all test pattern.

### DIFF
--- a/src/rtl_433.c
+++ b/src/rtl_433.c
@@ -924,7 +924,7 @@ static void parse_conf_option(r_cfg_t *cfg, int opt, char *arg)
             cfg->report_meta = atobv(arg, 1);
         break;
     case 'D':
-        fprintf(stderr, "debug option (-D) is deprecated. See -v to increase verbosity\n");
+        cfg->demo_mode = atobv(arg, 1);
         break;
     case 'z':
         if (!arg)
@@ -1331,6 +1331,22 @@ int main(int argc, char **argv) {
         r_free_cfg(cfg);
         exit(!r);
     }
+
+    // Special case for test pattern demo mode
+    if (cfg->demo_mode) {
+        r = 0;
+        for (void **iter = demod->r_devs.elems; iter && *iter; ++iter) {
+            r_device *r_dev = *iter;
+            if (cfg->verbosity)
+                fprintf(stderr, "Showing demo data for device %s.\n", r_dev->name);
+            for (char **code = r_dev->demo_pattern; code && *code; ++code) {
+                r += pulse_demod_string(*code, r_dev);
+            }
+        }
+        r_free_cfg(cfg);
+        exit(!r);
+    }
+
     // Special case for string test data
     if (cfg->test_data) {
         r = 0;
--- a/include/r_device.h
+++ b/include/r_device.h
@@ -54,6 +54,7 @@ typedef struct r_device {
     struct r_device *(*create_fn)(char *args);
     unsigned disabled;
     char **fields; ///< List of fields this decoder produces; required for CSV output. NULL-terminated.
+    char **demo_pattern; ///< List of demo pattern. NULL-terminated.
 
     /* public for each decoder */
     int new_model_keys; ///< TODO: temporary allow to change to new style model keys
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -54,6 +54,7 @@ typedef struct r_cfg {
     int ppm_error;
     uint32_t out_block_size;
     char const *test_data;
+    int demo_mode;
     list_t in_files;
     char const *in_filename;
     int do_exit;
--- a/src/devices/bresser_3ch.c
+++ b/src/devices/bresser_3ch.c
@@ -94,6 +94,11 @@ static char *output_fields[] = {
     NULL
 };
 
+static char *demo_pattern[] = {
+        "{41}e8d9f3bf7680{41}e8d9f3bf7680{41}e8d9f3bf7680",
+        NULL,
+};
+
 r_device bresser_3ch = {
     .name           = "Bresser Thermo-/Hygro-Sensor 3CH",
     .modulation     = OOK_PULSE_PWM,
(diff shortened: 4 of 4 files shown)


zuckschwerdt:
An idea to have fixed test pattern for each possible decoder output and a demo mode to run all these.
I.e. each decoder should have one known good pattern per `data_make` branch (say wind, rain, ... messages) to test and show-case all possible outputs.

Not really intended for built-in regression tests (would need many more patterns and edge-cases with an expected output) but rather as quick demo to users what output to expect from decoders and as a way to generate a show case of all our decoders.

Opinions, ideas, discussion?

mhaas:
Sounds like a good idea to me, especially as the data model is not really normalized across all decoders.

On the other hand, why not use the example data from rtl_433_tests? There is example data in JSON format already.

zuckschwerdt:
Yes, it would be a good idea to pull the demo data from the test repo. Just once with some python hack.
The idea with the demo, opposed to the test repo, is to allow instant demo output in every format the user selects. E.g. the user wants to see what the output for a custom MQTT setup is for a given device.

Something I also want to later do is to output the help text from the source comments. It could be compiled in with some transformation step. Then `-R 52 -D` could tell you exactly what this device is, how it works, ... and some example outputs.

Mindavi:
This would also help when trying to fuzz rtl_433 (see #1062 ), as the demo inputs can be used as a seed to run the decoders.

zuckschwerdt:
Playing devil's advocate here: we could also collect the test/demo pattern in an external file, transform that into a header to include and use it as a library (e.g. `global_pattern_lib[52]`).
Not sure what fits best.

merbanan:
Merge at will.

src/rtl_433.c
```diff
@@ -1331,6 +1331,22 @@ int main(int argc, char **argv) {
...
         exit(!r);
     }
+
+    // Special case for test pattern demo mode
+    if (cfg->demo_mode) {
+        r = 0;
+        for (void **iter = demod->r_devs.elems; iter && *iter; ++iter) {
+            r_device *r_dev = *iter;
+            if (cfg->verbosity)
+                fprintf(stderr, "Showing demo data for device %s.\n", r_dev->name);
```
Mindavi: If all devices are enabled, does this show 'Showing demo data' for each device? This doesn't seem right to me, as most devices (currently) wouldn't have any test patterns.

zuckschwerdt: Yes, that should be considered. Anyway, I'll let this PR sit until I have a script to import patterns for most of the devices from the test repo. Won't be too much an issue then.

Mindavi:
I've gathered some data to seed my fuzzer. I will attach them here.

These are gathered by using the rtl_433_tests library and grepping the output/manually looking at output.

There are 2 zips, one has manually selected samples(correct), the other was created by doing some automated work(scraped).

[correct.zip](https://github.com/merbanan/rtl_433/files/4156106/correct.zip)
[scraped.zip](https://github.com/merbanan/rtl_433/files/4156107/scraped.zip)

**The scripts below are quickly hacked together, I've just listed them to give insight in my methods. Would be good to formalize this method to generate bitbuffers, maybe by building such functionality into rtl_433 or creating a cleaner python script.**

Script used for scraping usable bitbuffers:

```
#!/usr/bin/env bash
# script that runs rtl_433 over all files (run in rtl_433_tests) and outputs result as json

echo running rtl_433
rtl_bin="/path/to/rtl_433"
find . -name "*cu8" -exec $rtl_bin -r {} -G 4 -F json -vv \;
```

pipe the output of that script to a file, split it into multiple files by using `split -l input_file`.

```
#!/usr/bin/env python3
"""
script to 

import os
import subprocess
import shutil

dir = 'split/'
out_dir = 'split-filtered/'

def main():
    for f in os.listdir(dir):
        filepath = dir + f
        result = subprocess.run(['/path/to/rtl_433', '-G', '4', '-y', '@'+filepath, '-F', 'json' ], capture_output=True)
        if result.returncode == 0:
            with open(filepath) as file:
                line = file.readline()
            print(line, end='')
            print(result.stdout.decode('utf8'), end='')
            shutil.copy(filepath, out_dir + f)

if __name__ == "__main__":
    main()
```

This should give a directory with bitbuffers that generate useful output.

The last step is to collect these bitbuffers 

```
#!/usr/bin/env python3

import json
import os
import subprocess
import collections

input_dir = '/home/rick/hobby/rtl/tests/split2-filtered/'
rtl_binary = '/home/rick/hobby/rtl/asan/build/src/rtl_433'

def main():
    results = collections.defaultdict(list)
    counter = 0
    for filename in os.listdir(input_dir):
        path = input_dir + filename
        #print(filename)
        result = subprocess.run([rtl_binary, '-G', '4', '-y', '@' + path, '-F', 'json'], capture_output=True)
        if result.returncode == 0:
            with open(path) as file:
                line = file.readline().rstrip()
            output = result.stdout.decode('utf8').rstrip()
            outputs = output.splitlines()
            for output_line in outputs:
                data = json.loads(output_line)
                model = data['model']
                results[model].append(line)
#                print(model, len(results[model]))
        counter += 1
        if counter % 100 == 0:
            print()
            total = 0
            for model, data in results.items():
                total += len(data)
                print(model, len(data))
            print('total:', total)
    
    result_string = json.dumps(results)
    with open('split2-data-points.json', 'w') as wr_file:
        wr_file.write(result_string)

if __name__ == "__main__":
    main()
```

This leaves you with a json file that has as key the names of the decoders, as value a list of bitbuffers that trigger the decoder.

Mindavi:
[rfraw-samples.zip](https://github.com/merbanan/rtl_433/files/5507219/rfraw-samples.zip)

This is an archive with rfraw samples for all the decoders with a known signal.

When I find the time I'll look into making a sort of testing/demo mode into the r_devices. That demo / testing mode should also be usable for testing, so should return a 1 when one of the decoders doesn't work as it should (anymore).

In that way the samples are right in the decoder and someone making a new decoder can directly add some testing to it to prevent future regressions.

PR #1274, opened 2020-01-27
Fix type of battery_mv in Fineoffset-WH51

hackery:
https://github.com/merbanan/rtl_433/commit/b96d3463a8f4dfe0a62fe962a84365edf3dfe1d0 added `battery_mv` but passed the wrong type into data_make


zuckschwerdt:
Not the first time I botched that :/ Thanks for the fix!

PR #1285, opened 2020-02-01
Insteon

evilpete:
Here a decoder for  insteon's home automation  wireless protocol.


evilpete:
What did I miss?

merbanan:
I retract my comment.

zuckschwerdt:
Great work! Some first notes:
- Can you run ./maintainer_update.py to update build files (e.g. Makefile is sorted alphabetically)?
- `static uint8_t preamble` should be in the decode function as `uint8_t const preamble`
- `verbose` and the env setup should be `decoder->verbose` (-vv option)
- `gen_ext_crc` isn't a crc, you should use `add_bytes()` util, e.g. `uint8_t chk = ~add_bytes(&b[7], 15) + 1`
- can you descibe what `gen_crc` does? Is it an LFSR?
- remove `byte_to_binary`, use `bitrow_print...` if you really need debugging output
- function's `{` should be on the next line, generally try clang-format if possible
- no need to `bzero` results, right?
- all the `static`s look off, is that all state data?

evilpete:
Will do, I'll make the requested updates in a day or so (I cranked this out
this out quickly lifting stuff from older project)

The "crc" code is from some stuff I did a few years ago when I reverse
engineered the protocol (I but a like to the POC code in the readme)

The bzero is out of habit :-)

The gen_crc is a type of LFSR (I remember it being pain in ass to reverse)
I have a descriptions in slides from a old talk.

Insteon uses three different validation checks, one for regular 13byte
wireless communications , a different one for power line communications,
and a third for 'extended' 32byte messages.

The over use of statics is for any data passed out of the local call
environment (I have not read the supporting infrastructure code and wanted
to play it safe)

zuckschwerdt:
I see, no need for static then, everything given to `data_make()` is dup'ed.

The `gen_crc` is really interesting, the way a nibble of the running chk is folded "up" could also be some variant of Adler/Fletcher style arithmetic.
We currently have two unknown checksums that kind of look like this. @merbanan

zuckschwerdt:
It really does look like some unknown checksums we have. All lower nibbles are plain parity, and the upper nibbles are …complicated.

evilpete:
I'll send you a write up in the morning

evilpete:
https://github.com/evilpete/insteonrf/blob/master/Doc/crc.txt

merbanan:
We need to list which protocols that has missing checksum documentation so that people could look into them. 

The algorithm in crc.txt is within the complexity offered by these sensors. We should add it to revdgst with arbitrary shifting and rotation.

merbanan:
I added the xor_shift function to revsum and it worked fine on this data but not on the Jansite TPMS data. I remember there was another sensor with only 4 bits of xorsum that.

zuckschwerdt:
I remember one or two decoders with that kind of problem (only the lower nibble is clearly parity), but I can seem to locate those. Perhaps I'll review each decoder and tag the current state of MIC.

evilpete:
I added to the crc.txt file to hopefully make it clearer for you

https://github.com/evilpete/insteonrf/blob/master/Doc/crc.txt

evilpete:
ok, i've updated most of the code as suggested.

evilpete:
FYI : Once this is merged,  I plan  to replace the (admittedly) primitive FSK demodulation utility in my [Insteon RF](https://github.com/evilpete/insteonrf) toolkit with a reference to [rlt_433](https://github.com/merbanan/rtl_433).

merbanan:
@evilpete Regarding crc.txt, getting to the map is easy to understand. But how did you go from the map to the algorithm. Do you remember the logic behind that?

I think we currently can reverse linear codes pretty good, others have taken lots of resources. I'd like to document some strategies on how to figure them out.

merbanan:
Code looks good to me.

evilpete:
@merbanan  I sorta walk though it in the file  [crc.txt](https://github.com/evilpete/insteonrf/blob/master/Doc/crc.txt)
I was able to generate packets where only the last byte changed ( in this case turning on a light and increasing the brightness by one each packet)

Starting with one packet 
(in this case `23 80 25 13 11 78 2B 11 04 : 42` )

I XORed it with another packet, since the only change were bits in the last byte,  I was able to observe that change that occurred in a single iteration  

(note the 2nd to the last byte differs and the last byte is the CRC )
Eg:

`23 80 25 13 11 78 2B 11 04 : 42`Orig pkt:  
`23 80 25 13 11 78 2B 11 14 : 52 `  XOR
`00 00 00 00 00 00 00 00 10 : 10` . Result

showing  0x10 is passed without bit shifting ( just XOr)

the table at [line 53 in crc.txt](https://github.com/evilpete/insteonrf/blob/master/Doc/crc.txt#L55) shows this for all values in the upper nibble (included below)

I do this for 10 -> F0  in the upper nibble my XOR'ing first line with the 16 following lines and showing the delta 


```
= 23 80 25 13 11 78 2B 11 04 : 42       This gets XORed with each following lines
=
    23 80 25 13 11 78 2B 11 14 : 52    XOR with 23 80 25 13 11 78 2B 11 04 : 42 
    00 00 00 00 00 00 00 00 10 : 10  = Result
--
    23 80 25 13 11 78 2B 11 24 : 62    XOR with 23 80 25 13 11 78 2B 11 04 : 42 
    00 00 00 00 00 00 00 00 20 : 20  = Result
--
    23 80 25 13 11 78 2B 11 34 : 72    XOR with 23 80 25 13 11 78 2B 11 04 : 42 
    00 00 00 00 00 00 00 00 30 : 30  = Result
--
    23 80 25 13 11 78 2B 11 44 : 02    XOR with 23 80 25 13 11 78 2B 11 04 : 42 
    00 00 00 00 00 00 00 00 40 : 40  = Result
--
    23 80 25 13 11 78 2B 11 54 : 12     ... and so on....
    00 00 00 00 00 00 00 00 50 : 50
--
    23 80 25 13 11 78 2B 11 64 : 22
    00 00 00 00 00 00 00 00 60 : 60
--
    23 80 25 13 11 78 2B 11 74 : 32
    00 00 00 00 00 00 00 00 70 : 70
--
    23 80 25 13 11 78 2B 11 84 : C2
    00 00 00 00 00 00 00 00 80 : 80
--
    23 80 25 13 11 78 2B 11 94 : D2
    00 00 00 00 00 00 00 00 90 : 90
--
    23 80 25 13 11 78 2B 11 A4 : E2
    00 00 00 00 00 00 00 00 A0 : A0
--
    23 80 25 13 11 78 2B 11 B4 : F2
    00 00 00 00 00 00 00 00 B0 : B0
--
    23 80 25 13 11 78 2B 11 C4 : 82
    00 00 00 00 00 00 00 00 C0 : C0
--
    23 80 25 13 11 78 2B 11 D4 : 92
    00 00 00 00 00 00 00 00 D0 : D0
--
    23 80 25 13 11 78 2B 11 E4 : A2
    00 00 00 00 00 00 00 00 E0 : E0
--
    23 80 25 13 11 78 2B 11 F4 : B2
    00 00 00 00 00 00 00 00 F0 : F0
```

Thus can see that 0x10 -> 0x0F is passed without bit shifting ( just XOr)

the table at [line 109](https://github.com/evilpete/insteonrf/blob/master/Doc/crc.txt#L109)  shown below, I do the same but only change the lower nibble and observe the change in the CRC

(in this table I only show the Xor'ed results, with zero's replaced with dots in improve readability )

```
.. .. .. .. .. .. .. 01 .. 31
.. .. .. .. .. .. .. 02 .. 62
.. .. .. .. .. .. .. 03 .. 53
.. .. .. .. .. .. .. 04 .. C4
.. .. .. .. .. .. .. 06 .. A6
.. .. .. .. .. .. .. 07 .. 97
.. .. .. .. .. .. .. 08 .. 88
.. .. .. .. .. .. .. 09 .. B9
.. .. .. .. .. .. .. 0A .. EA
.. .. .. .. .. .. .. 0B .. DB
.. .. .. .. .. .. .. 0C .. 4C
.. .. .. .. .. .. .. 0D .. 7D
.. .. .. .. .. .. .. 0F .. 1F
```

this table show that the lower nibbles are \ passed without bit shifting 

**BUT** it also shows the effect the lower nibble has on the upper nibble

from this I was able to construct a lookup table to recreate the algorithm see [line 154 of crc.txt](https://github.com/evilpete/insteonrf/blob/master/Doc/crc.txt#L154)
then eventually reverse engineer the  


Does that help / make any sense  ?

merbanan:
Ok, well looking at
01 0000 0001 -> 30 0011 0000
shows a 4 position shifting operator and some operation that turn one 1 bit to two.

The real magic here is translating the table to:
            r ^= dat[i] ;
            r ^= (( r ^ ( r << 1 )) & 0x0F) << 4 ;

In this case a 16 byte table isn't really an issue but with 8 real bits the mapping table is 256 elements long.

evilpete:
I agree a table is not a good solution particularly for more then a handful of bits,  for me it was just a step in the process since it gave me a functional model

merbanan:
It is the process I'm after. Generating the table can be done automatically if you have a large enough sample set. We have [revdgst](https://github.com/triq-org/revdgst) with the ambition to add all integrity check algorithms currently used in rtl_433. There aren't that many ways to generate integrity checks that are easy to compute on low complexity devices.

zuckschwerdt:
Since it's add without carry we can follow the bits all the way through the calculation easily. It's likely that the algorithm above is what's implemented. Here is another way to see where the bits actually go:
Add all odd rows (indexes 0, 2, ...) to `result0`, add all even rows (indexes 1, 3, ...) to `result1`.
Add `result0`, `result1`, `result0 << 4`, `result0 << 5`. I.e. the shifting doesn't need to be in the loop. It effectivly just folds half the (final) lower bits up.

merbanan:
@zuckschwerdt ok that makes sense, I found another metric/rule that can be generated automatically. In the case of one bit differences you can look in the output how many bits are set. It can be observed that in 3 cases there is 1->2 bit relation, this indicates an additive step. In the 4th case it is 1->1, imo this 2,2,2,1 relation indicates a shift to the left.

Just as a theoretical example if it had been a rotation operator the bit relation had been 2,2,2,2. (And no hint on what way to rotate can be observed from that).

Anyway I think we can conclude this discussion and focus more on the issue. @zuckschwerdt can we merge the code?

zuckschwerdt:
Sure, I'll merge this and put some style updates on top right after that.

zuckschwerdt:
@evilpete can you explain about `inverted`? You invert the buffer at the start if an odd number of previous buffers had a succsessful search for a pattern, but then for that buffer undo the invert?
Is that really meant to be stateful across different messages? Are the buffers alternating inverted or did you just mean to try two different match pattern, plain and inverted?

evilpete:
The inverted testing is legacy from the standalone decoder I originally, since. it took input from different sources/demodulators I always to a inverted test.

In the submitted code in this push request  I retest it incase the first "hit" was false.

I'll update the code and remove the 'inverted' test.  tonight or tomorrow and let you know

zuckschwerdt:
Can you squash and rebase to master? (`git rebase -i HEAD~6` to interactively squash the last 6 commits). Otherwise I can also resolve the conflicts with the GitHub UI.

evilpete:
done

evilpete:
@zuckschwerdt :  how does this look,  I'm done with this for a while unless there are any issues

zuckschwerdt:
Good. I'll merge in the next days and put some final touches of our code style on it.

merbanan:
@zuckschwerdt ok to merge?

zuckschwerdt:
Guess so :) I wanted to prepare a few style changes but this isn't our usually decoder anyway. I'll merge later today and runc some final checks.

zuckschwerdt:
@evilpete
- stdio.h will already be there and with not allocs we don't need stdlib.h or did I miss something? 
- Can you remove the `(void)`s?
- is INSTEON_PREAMBLE_i needed anymore?
- is `#define INSTEON_PREAMBLE insteon_preamble` needed?
- `insteon_preamble` needs to be static so we don't export it
- to attach the doc-comment it needs to start with `/** @fn int insteon_callback(r_device *decoder, bitbuffer_t *bitbuffer)`
- the first doc-comment sentence should be `     Template decoder for Insteon.` (same as the file comment which documents the license)

evilpete:
done

the (void)s are from dealing with static analyzers that bitch if a return value is silently ignored

zuckschwerdt:
For future commits and merges please make sure the commit message is sensible, e.g. "Add Insteon support".

evilpete:
thank for merging,  I plan to update my [insteonRF](https://github.com/evilpete/insteonrf) tools to use rtl_433 instead of my admittedly shitty FSK demodulator. (it was based on using atan tracking direction of rotation to detect I lead or follow Q)

PR #1286, opened 2020-02-02
Add error codes and some maitainance

merbanan:


src/devices/gt_wt_02.c
```diff
@@ -70,7 +70,7 @@ static int gt_wt_02_process_row(r_device *decoder, bitbuffer_t *bitbuffer, int r
 
     // accept only correct checksums, (sum of nibbles modulo 64)
     if ((sum_nibbles & 0x3F) != checksum)
-        return 0; // DECODE_FAIL_MIC
+        return DECODE_FAIL_MIC;
```
zuckschwerdt: This will mess up `counter`. Should have been commented better.

src/devices/rubicson_48659.c
```diff
@@ -161,10 +162,18 @@ static int rubicson_48659_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
     if (row < 0)
-        return 0;
+        return DECODE_ABORT_EARLY;
 
     if ((bitbuffer->bits_per_row[row] > 33) || (bitbuffer->bits_per_row[row] < 10))
-        return 0;
+        return DECODE_ABORT_LENGTH;
+
+    // Verify checksum
+    uint8_t sum = (add_bytes(bb[row], 3) & 0xff) - bb[row][0];
```
zuckschwerdt: The comment talks about XOR, should that be corrected?

merbanan: Well I wrote the code and I don't remember anything of it. Then this weekend I ran revdgst and this line popped out. So I don't really know, I will test with the live sensor an adapt the comment accordingly.

src/devices/gt_wt_02.c
merbanan: I will add a > 0 check in the return function.

zuckschwerdt: The comment there says "return on first successful", and that's what really happens too.
Guess the logic needs some reworking, saving the last fail code and return on counter > 0 or with the fail code in counter after the loop.

PR #1291, opened 2020-02-04
Feat error codes2

merbanan:
More error codes for the decoders.


zuckschwerdt:
Great! Notes:
- The float casts won't apply, as there are double promotions in those lines
- disabling fordremote needs an update to example conf and readme (can be done later)
- for gt_tmbbq05 I'd argue that a 4-bit check is more of CHECKSUM strength than PARITY (which I think of as 1 or 2 bit strength).
- For each changed/added mic string we likely need to fix the json in tests at the same time

merbanan:
Did all but fixing the json.

PR #1296, opened 2020-02-05
get sample rate from filename

evilpete:
added shell operators to get sample rate from filename

### DIFF
--- a/examples/sigrok-conv.sh
+++ b/examples/sigrok-conv.sh
@@ -25,7 +25,13 @@ fi
 out=$2
 
 if [ -z "$3" ] ; then
-  rate=250
+  # strip the "k.cu8"
+  # ${fil:0:-5}  also works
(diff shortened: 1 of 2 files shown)


Mindavi:
Does this work for filenames not containing sample rate as well? I might imagine people changing filenames before uploading etc., so it should handle that case.

merbanan:
Well as long as you still can override the sample rate it doesn't matter IMO. It is a tool more targeted for developers.

zuckschwerdt:
I have some experimental code to create the Pulseview format directly in rtl_433. We'd have more control over all the options (sample rate, frequency, …) then.
The interesting bit is to call the zip util, I'd rather not include a lib for that.
I see if I can finish that code soon.

zuckschwerdt:
I think enhancing the script will make it to fragile in the end. We should move away from a shell script, there is some prototype code in #1297

Issue #1295, opened 2020-02-05
Add Pulseview demod tracing tool


evilpete:
If possible, include output of  both FSK demodulators

zuckschwerdt:
I like the idea. I already have some kind of enhanced "debugging" version, will adapt that.

merbanan:
@evilpete of course output from both bit slicers are the obvious solution.

zuckschwerdt:
I will add more options for the SR ouput later, please test the prototype code in #1297

PR #1297, opened 2020-02-06
Add native Sigrok writer

zuckschwerdt:
S.a. #1296 #1295


merbanan:
Do you really need to zip the data? We're not really saving any of the files for later. Anyway LGTM.

zuckschwerdt:
Current Pulseview needs all the files in a zip, yes. They plan to drop that for the next version maybe.

Issue #1231, opened 2019-12-10
Adding Rubicson 48686(sensor) Meat Thermometer


Net-time:
First of all this first gives a false positive as a Thermopro TP11 Thermometer (nonsense temperature)
Then it falls in line as Rubicson-48659.
The pulse lengths are a bit shorter though. so perhaps the tolerance is a bit wide for 48659.
rtl_433 -X "n=48686,m=OOK_PPM,s=600,l=1600,r=5000,g=3400,repeats>=3"

"-R 133"  does not work with Domoticz  , I need to use "-R -84 -R 133"
off topic but probably because of the "false positive" (works with rtl_433)

Due to it's function it's not suitable for fast changing measurments as it will send almost continuously.
Which will shorten battery life and block other sensors. (Not really a good sensor but it was on sale :0 )

zuckschwerdt:
So the Rubicson-48686 works with the Rubicson-48659 decoder and shows correct values?
Can you provide a captured sample (maybe also one that false-positives as TP11).
Use `-S all`, check the sample with `-r` then upload here as zip. Thanks!

Net-time:
As I said it "falls in line" so I can only get false positive and the correct one as one file.
As I've had time to sleep on it, I reckoned that the false positive was because of a truncated message at start, as I have on one of my OPUS sensors, but was wrong there too. It is because of a preamble with 3 short and one gap.
[48686.cu8.zip](https://github.com/merbanan/rtl_433/files/3951544/48686.cu8.zip)
![48686](https://user-images.githubusercontent.com/47906827/70642088-f019e380-1c3e-11ea-9517-8d7a62b82768.JPG)

merbanan:
Ok, added integrity checks for Rubicson and Thermopro sensors. The duplicates are gone. Thanks for the bug report.

PR #1303, opened 2020-02-08
Add integrity checks for Thermopro TP11 and TP12

merbanan:
This also fixes https://github.com/merbanan/rtl_433/issues/1231

### DIFF
--- a/conf/car_fob.conf
+++ b/conf/car_fob.conf
@@ -1,0 +1,18 @@
+# Unknown carfob with rolling code
+# Copyright (C) 2020 Benjamin Larsson
+#
+
+
+decoder {
+    name=Car fob,
+    modulation=OOK_PWM,
+    short=428,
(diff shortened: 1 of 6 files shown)


zuckschwerdt:
The existing LFSR utils are Galois already. You can either reverse the bytes and feed them to
the plain `lfsr_digest8` or you can mix both that and `lfsr_digest8_reflect`(which is byte and bit reflected) to create a byte reflected only variant.

merbanan:
I used the sum not the xor. Revdgst did report that both xor xor and xor sub worked. 

Done with g 51 k 04 final XOR 00 using xor xor (100 %)
Done with g 51 k 04 final XOR 00 using xor sub (100 %)

Anyway it turned into something better.

merbanan:
Fixed and committed. Much smaller change.

PR #1305, opened 2020-02-08
radiohead-ask: Fix issue where calloc with negative number leads to segfault

Mindavi:
I caught this by removing all CRC/checksum/parity checking from code.
Even before starting fuzzing one of my corpus inputs triggered this already.


zuckschwerdt:
Great catch. Thanks!

Issue #1271, opened 2020-01-26
Abarth 124 TPMS


TTigges:
### Introduction:

We are receiving signals from the TPMS sensors of our (Fiat) Abarth 124 Spider TPMS sensors. It’s a bit of a unique case because the 124 Spider are actually built by Mazda on the base of the MX-5. However, only the Abarth models seem to have active TPMS sensors. Unfortunately, the car does not display the sensor infos. I am testing with a set of tires sitting inside (controled environment, temperature and pressure are known) as well as a set of tires mounted on the car with copied sensors. I also got codes from friends, also recorded with rtl_433.

### Auto suggested profile/protocol:

With _-G_, rtl_433 suggests the sensors are Jansite sensors but the values are off. So from this, I assume the signals are similar/having a similar preamble? Or how is the fitting device protocol selected? _(More below at **First try at a device protocol** and **What I don't understand.**)_

Where Jansite’s code is
II II II IS PP TT CC

my code seems to be
?? ?? II II SS PP TT

### Received codes and first notes:

Right now, we have checked 3 cars with 4 sets of tires total = 16 sensors.

**Set A 1**
- 0f54771bb79045
- 0f5476e8b78f45
- 0f5476eab79140
- 0f547711b79043

**Set A 2**
- 0f54771bb49835
- 0f5476e8b6ac35
- 0f5476eab5ab35
- 0f547711b48e35

Notice the same IDs (byte 3 + 4). Byte 5 (expected status) behaves a bit weird, though.

**Set B**
- 0f365c01078d40
- 0f365beb079141
- 0f38cb2f078e42
- 0f365c1a079043

Here, the 5th byte is uniform (07) again. I was surprised that the 2nd byte is different for one of the sensors.

**Set C**
- 0c15590367903b
- 0c2990b667913d
- 0cbf53cb679e3c
- 0f492f8367923c

The 5th byte is uniform here, too. But the first 2 bytes are a mess here. I’m surprised by the „0c“.

TTigges:
[...]

PR #1311, opened 2020-02-09
Add Abarth 124 TPMS Sensors

TTigges:
Process to the device protocol: https://github.com/merbanan/rtl_433/issues/1271#
Recorded sensor messages: [Abarth_124_433_recorded-signals.zip](https://github.com/merbanan/rtl_433/files/4177854/Abarth_124_433_recorded-signals.zip)

It is suggested that VDO is the OEM supplier and that the sensors are VDO TG1C.
The Fiat/Abarth 124 is produced by Mazda on the MX-5 ND production line but only few MX-5 ND seem to have TPMS sensors.


merbanan:
Merged manually

PR #1325, opened 2020-02-23
inFactory improvements

hagen-git:
- add CRC computation
- add battery low bit
- add references to other brands
- tighter timings
- debug: add raw format (e.g. for JSON file)


src/devices/infactory.c
```diff
@@ -1,84 +1,127 @@
...
+            crc |= 1; 
+        }
+        if (crc & 0x10) {
+            crc ^= 0x13; // Koopmann 0x9, CCITT-4; FP-4; ITU-T G.704
+        }
+    }
+
+    return (crc == msg_crc);
+}
+
```
merbanan: https://github.com/merbanan/rtl_433/blob/master/src/util.c#L64 can this be used?

src/devices/infactory.c
```diff
@@ -1,84 +1,127 @@
...
-    humidity = (b[3] & 0x0F) * 10 + (b[4] >> 4); // BCD
+    battery  = (b[1] >> 2) & 1;
     temp_raw = (b[2] << 4) | (b[3] >> 4);
-    temp_f   = (float)temp_raw * 0.1 - 90;
+    humidity = (b[3] & 0x0F) * 10 + (b[4] >> 4); // BCD, 'A0'=100%rH
     channel  = b[4] & 0x03;
 
+    temp_f   = (float)temp_raw * 0.1 - 90;
+
+    mic = infactory_crc_check(b) ? "CRC" : "CRC_ERROR"; // we could also return DECODE_FAIL_MIC;
```
merbanan: Correct we only return payloads with crc ok. And we should check the crc before we decode the parameters. Fail as quick as possible.

src/devices/infactory.c
```diff
@@ -1,84 +1,127 @@
...
     /* clang-format off */
     data = data_make(
             "model",            "",             DATA_STRING, _X("inFactory-TH","inFactory sensor"),
             "id",               "ID",           DATA_INT, id,
             "channel",          "Channel",      DATA_INT, channel,
+            "battery",          "Battery",      DATA_INT, battery,
             "temperature_F",    "Temperature",  DATA_FORMAT, "%.02f °F", DATA_DOUBLE, temp_f,
-            "humidity",         "Humidity",     DATA_FORMAT, "%u %%", DATA_INT, humidity,
+            "humidity",         "Humidity",     DATA_FORMAT, "%u %%rH", DATA_INT, humidity,
+            "mic",              "Integrity",    DATA_STRING, mic,
```
merbanan: mic is always CRC when it is present.

src/devices/infactory.c
```diff
@@ -1,84 +1,127 @@
...
+    int i;
+    char c;
+    char *p = msg;
+    for (i = 0; i < 10; i++) {
+        c = ((b[i >> 1]) >> (4 * (1 - (i & 1)))) & 0x0F;
+        c = (c > 9) ? c + 55 : c + 48;
+        *p++ = c;
+    }
+    *p = 0;
+}
```
merbanan: Yeah, it will take me 30 minutes to understand this. Anyway I guess it forms the payload in hex. We have bitbuffer_print for that. Add it under a verbose/debug check if you need it.

src/devices/infactory.c
```diff
@@ -92,17 +135,43 @@ static char *output_fields[] = {
...
+        "mic",
+        "battery",
+#ifdef DEBUG        
+        "rawmsg",
+#endif // DEBUG        
         NULL,
 };
 
+/*
+Analysis using Genuino (see http://gitlab.com/hp-uno, e.g. uno_log_433):
```
merbanan: Impressive analysis. We have a tool revdgst that can take apart 8-bit checksums. I don't think it can handle 4-bit CRCs yet though.

src/devices/infactory.c
```diff
@@ -92,17 +135,43 @@ static char *output_fields[] = {
...
+
+- four preamble pairs '1'/'0' each with a length of ca. 1000uS
+- syncPre, syncPost, data0, data1 have a '1' start pulse of ca. 500uS
+- syncPre pulse before dataPtr has a '0' pulse length of ca. 8000uS
+- data0 (0-bits) have then a '0' pulse length of ca. 2000uS
+- data1 (1-bits) have then a '0' pulse length of ca. 4000uS
+- syncPost after dataPtr has a '0' pulse length of ca. 16000uS (we declare it received at 12ms)
+
+This analysis is the reason for the new r_device definitions below.
+NB: pulse_demod_ppm does not use .gap_limit if .tolerance is set.
```
merbanan: If the old samples decode then I think it is ok.

src/devices/infactory.c
```diff
@@ -1,84 +1,127 @@
...
+
+    mic = infactory_crc_check(b) ? "CRC" : "CRC_ERROR"; // we could also return DECODE_FAIL_MIC;
+    infactory_raw_msg(raw_msg, b);                      // DEBUG: print raw message bits
+
     /* clang-format off */
     data = data_make(
             "model",            "",             DATA_STRING, _X("inFactory-TH","inFactory sensor"),
             "id",               "ID",           DATA_INT, id,
             "channel",          "Channel",      DATA_INT, channel,
+            "battery",          "Battery",      DATA_INT, battery,
```
zuckschwerdt: should be "battery_ok"

src/devices/infactory.c
```diff
@@ -1,84 +1,127 @@
...
             "temperature_F",    "Temperature",  DATA_FORMAT, "%.02f °F", DATA_DOUBLE, temp_f,
```
zuckschwerdt: There should be no degree symbol (it's not ASCII), can you remove it while you are at it.

src/devices/infactory.c
```diff
@@ -1,84 +1,127 @@
 /** @file
     inFactory outdoor temperature and humidity sensor.
+    Other brand names: nor-tec, GreenBlue, DAY
```
zuckschwerdt: Please add the info in the decoder comment below, not in the copyright.

src/devices/infactory.c
zuckschwerdt: yes, please remove this, it's easier to maintain 100+ decoders if they are short and in roughly the same pattern.

src/devices/infactory.c
hagen-git: Unfortunately not - the CRC is computed over 36 message bits, but crc4() only computes over full bytes.

src/devices/infactory.c
hagen-git: OK. Will change.

src/devices/infactory.c
hagen-git: Now I understand why it was defined as a string constant in other places.

src/devices/infactory.c
hagen-git: OK, will remove. I used it to compare rtl_433 output with my own logs. :-)

src/devices/infactory.c
hagen-git: I checked it with my own samples. Will look for the original ones (they are probably still on-site).

src/devices/infactory.c
hagen-git: My first name was weakbattery. Then I copied the name from another device (ambient_weather.c). The bit is the 'battery low' indicator.
In fineoffset.c I see a string 'LOW' or ' OK' - can I use the same pattern?

src/devices/infactory.c
hagen-git: Yes, will do.

src/devices/infactory.c
hagen-git: OK.

src/devices/infactory.c
hagen-git: I needed the difference message analysis, because for CRC computation two bits are at a different position than in the final output. At least I have learned a lot about CRCs in the process. Now I even know how to compute it out-of-order (if needed). :-)

src/devices/infactory.c
zuckschwerdt: battery:OK/LOW is the old convention. The new one is battery_ok:0/1

src/devices/infactory.c
zuckschwerdt: CRC is invariant if you prepend zero bits. I'll have a closer look an give an example tomorrow

hagen-git:
TODO:
- check function is OK with original samples
- simplify CRC check function

src/devices/infactory.c
hagen-git: I found a way to compute the CRC using crc4(). :-)

src/devices/infactory.c
hagen-git: In hideki.c I can see that battery_ok == 0  means 'OK' there, this feels wrong to me.

hagen-git: holman_ws5029.c looks better. I like the negation of battery_low and use the same pattern.

hagen-git:
Found the original samples - they decode, but instead of 6, the first file decodes only 5 datagrams.

Want to compare with my own sensor samples, looks like the tolerance/median settings need a tiny bit of tweaking.

hagen-git:
OK. Looks like a tolerance setting of 750 us should be wide enough. Even with a weak battery, the times do not change much.

src/devices/infactory.c
hagen-git: Checked OK. :-)

hagen-git:
Looks like there are already two infactory directories in rtl_433_test.
Would you like a third one? ;-)

zuckschwerdt:
Ah, sorry, looks good. I'll merge shortly. (I wanted to investigate why the last data bits are xor'ed and not crc'ed, but that seems just a quirk in the protocol)

hagen-git:
> Ah, sorry, looks good. I'll merge shortly. (I wanted to investigate why the last data bits are xor'ed and not crc'ed, but that seems just a quirk in the protocol)

No wonder, I was also puzzled - most algorithms I have seen 'push out' the last data bits. But from the descriptions I have seen in Wikipedia and elsewhere, as the CRC is the remainder of a polynomial division, it is OK to 'just' XOR the last data bits (e.g. the last nybble for a 4-bit CRC) with the CRC computed so far.

Looking at the error returns, I noticed that sometimes I had brackets around the return, sometimes not, so I tried to sync this with my latest push. Hope it is OK this way around.

zuckschwerdt:
With "push out" you mean "pad out"? E.g. turn a nibble `0xN` into `0xN0`? That changes the CRC, only prepending zeros is invariant. But sure, there could be implied zero padding at the end.
Regarding CRC it holds that: `if crc(b[0..n-1]) == b[n] then crc(b[0..n]) == 0`, i.e. feeding the remainder to the crc leaves an output of `0`, but that's not what's used here. The last data bits are xor'ed to the remainder.

hagen-git:
No, what I meant is this: the CRC can be implemented as a shift register with XOR gates that implement the polynome. In a lot of implementations, at the end of a message, something is appended at the end of the message - usually 0 - so that all message bits will actually trigger the polynome XOR part. From the description on Wikipedia (CRC = remainder of polynome division) and the example listed there, you can see that _not_ padding the message with zero at the end is also correct. (I only really understood the mechanics when I followed the article of Gregory Ewing [https://www.cosc.canterbury.ac.nz/greg.ewing/essays/CRC-Reverse-Engineering.html].)

zuckschwerdt:
Ok, putting more bits than input into a shift register to ensure every actual data bit is shifted out I understand. But that's specific to implementation and not a part of the algorithm.
The inital value and the the final xor are used to prevent (false) all-0 (or all-1) messages from succeeding.
CRC doesn't care about byte bounds, you can feed the last nibble no problem (not in our implementation though). But that's not done here, which is strange. Using an xor on something other than the remainder is not CRC. I.e. to add the nibble "by hand" you would need up to  4 xor's (one for each set bit) with the generator.

merbanan:
@zuckschwerdt can we merge this? Looks fine to me.

PR #1336, opened 2020-03-15
Add Honeywell CM921/BDR91/Evohome decoder

wintersteiger:
This adds a decoder for Honeywell CM921 + BDR91 messages. I didn't expect this, but it essentially uses a (small) subset of the Evohome protocol, so I suspect this works for other Evohome setups too. With a little bit more tinkering, this can replace the HGI80 gateway box, as the messages from the HGI80 seem to be basically an ASCII conversion of the binary Evohome messages. I don't have the hardware to test it further and to decode more message types though.

It's fairly reliable for me; I use an RTL-SDR with `rtl_433 -f 868M -g 50 -l 8000 -R 151`.


src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+    data_t *data = data_make(
+            "model",        "", DATA_STRING, "Honeywell CM921",
+            "Packet",       "", DATA_STRING, packet_hex,
+            NULL);
+    /* clang-format on */
+
+    message_t message;
+
+    if (parse_msg(&packet, row, &message))
+        data = interpret_message(&message, data);
```
merbanan: It the checksum fails you should return that.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+#ifdef _DEBUG
+    data = add_hex_string(data, "Header", &message.header, 1);
+    uint8_t cmd[2] = { message.command >> 8, message.command & 0x00FF};
+    data = add_hex_string(data, "Command", cmd, 2);
+    data = add_hex_string(data, "Payload", message.payload, message.payload_length);
+    data = add_hex_string(data, "Unparsed", message.unparsed, message.unparsed_length);
+    data = add_hex_string(data, "CRC", &message.crc, 1);
+    data = data_append(data, "# man errors", "", DATA_INT, man_errors, NULL);
+#endif
+
```
merbanan: Here somewhere you should add that the message has a MIC of type CHECKSUM. Look in other decoders for reference.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+    (at your option) any later version.
+*/
+/**
+Honeywell CM921 Thermostat.
+
+868Mhz FSK, PCM, Start/Stop bits, reversed, Manchester.
+*/
+
+#include <stdlib.h>
+#include <mongoose.h>
```
merbanan: Both these should not be needed,

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
+/** @file
+    Honeywell CM921 Thermostat
+
+    Copyright (C) 2020
```
merbanan: There should be some author to the code.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+                const uint8_t *p = &msg->payload[1 + 3*i];
+                // *(p+1) == 0x01 always?
+                int value = *(p+2);
+                switch (*p) {
+                    case 0xC8: data = data_append(data, "max_flow_temp", "", DATA_INT, value, NULL); break;
+                    case 0xC9: data = data_append(data, "pump_run_time", "", DATA_INT, value, NULL); break;
+                    case 0xCA: data = data_append(data, "actuator_run_time", "", DATA_INT, value, NULL); break;
+                    case 0xCB: data = data_append(data, "min_flow_temp", "", DATA_INT, value, NULL); break;
+                    case 0xCC: /* Unknown, always 0x01? */ break;
+                    default: printf("Unknown parameter to 0x1030: %x02d=%04d\n", *p, value);
```
merbanan: No printfs without verbose guard.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+            }
+            break;
+        }
+        case 0x313F: {
+            UNKNOWN_IF(msg->payload_length != 1 && msg->payload_length != 9);
+            switch (msg->payload_length) {
+                case 1:
+                    data = data_append(data, "time_request", "", DATA_INT, msg->payload[0], NULL); break;
+                case 9: {
+                    X313F* pl = (X313F*)msg->payload;
```
merbanan: Not fond of this construct.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+
+    for (unsigned i = 0; i < msg->num_device_ids; i++)
+        for (unsigned j = 0; j < 3; j++)
+            msg->device_id[i][j] = next(bb, &ipos, num_bytes);
+
+    msg->command = (next(bb, &ipos, num_bytes) << 8) | next(bb, &ipos, num_bytes);
+    msg->payload_length = next(bb, &ipos, num_bytes);
+
+    if (msg->payload_length > 0) {
+        msg->payload = (uint8_t*)malloc(sizeof(uint8_t) * msg->payload_length);
```
merbanan: Can we get rid of all allocations?

The payload is a sequence of (8-bit) bytes represented in hexadecimal (i.e. 00 to FF) and with a maximum length of 48 (minimum length 1).

Here it should be possible to replace it with a 48 bytes long array.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+    { .t = 12, .s = "THm" }, /* Thermostat with setpoint schedule control (DTS92E, CME921) */
+    { .t = 13, .s = "BDR" }, /* Wireless relay box (BDR91) (HC60NG too?) */
+    { .t = 17, .s = " 17" }, /* Dunno - Outside weather sensor? */
+    { .t = 18, .s = "HGI" }, /* Honeywell Gateway Interface (HGI80, HGS80) */
+    { .t = 22, .s = "THM" }, /* Thermostat with setpoint schedule control (DTS92E) */
+    { .t = 30, .s = "GWY" }, /* Gateway (e.g. RFG100?) */
+    { .t = 32, .s = "VNT" }, /* (HCE80) Ventilation (Nuaire VMS-23HB33, VMN-23LMH23) */
+    { .t = 34, .s = "STA" }, /* Thermostat (T87RF) */
+    { .t = 63, .s = "NUL" }, /* No device */
+};
```
merbanan: json is quite talkative. All strings in the comments should be enabled in the json output.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+            for (unsigned j = 0; j < 3; j++)
+                sprintf(did + 2*j, "%02x", msg->device_id[i][j]);
+        }
+        strcat(ds, did);
+        free(did);
+    }
+
+    return data_append(data, "Device IDs", "", DATA_STRING, ds, NULL);
+}
+
```
merbanan: All allocations here should be possible with static buffers.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+#include "decoder.h"
+
+#define _DEBUG
+
+static char *hex(bitbuffer_t *b, int row) {
+    size_t lenB = b->bits_per_row[row]/8;
+    if (b->bits_per_row[row] % 8 != 0) lenB++;
+    char *r = malloc(lenB * 2 + 1);
+    cs_to_hex(r, b->bb[row], lenB);
```
merbanan: Use something like this instead:

```
    char    str_buf[1024];

    // Get time now

    // Make data string
    str_buf[0] = 0;
    for (unsigned n=0; n<out->length; n++) { sprintf(str_buf+n*2, "%02x", out->data[n]); }

```

We might create a helper for it one day.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+    data = add_hex_string(data, "Unparsed", message.unparsed, message.unparsed_length);
+    data = add_hex_string(data, "CRC", &message.crc, 1);
+    data = data_append(data, "# man errors", "", DATA_INT, man_errors, NULL);
+#endif
+
+    decoder_output_data(decoder, data);
+
+    free(packet_hex);
+    free(message.payload);
+    free(message.unparsed);
```
merbanan: I really don't like allocations :). Anyway they should be fairly easy to get rid of.

src/devices/honeywell_cm921.c
```diff
@@ -0,0 +1,449 @@
...
+{
+    // Sources of inspiration:
+    // https://www.domoticaforum.eu/viewtopic.php?f=7&t=5806&start=240
+
+    // preamble=0x55 0xFF 0x00
+    // preamble with start/stop bits=0101010101 0111111111 0000000001
+    //                              =0101 0101 0101 1111 1111 0000 0000 01
+    //                            =0x   5    5    5    F    F    0    0 4
+    // post=10101100
+    // each byte surrounded buy start/stop bits (0byte1)
```
merbanan: buy -> by

merbanan:
The visual studio makefile and README.md changes are missing. We need signals samples also.

Anyway one of the more complex protocols present.

You should not need the -l parameter for FSK signals. Try tune around to see if it can work fine without -l.

src/devices/honeywell_cm921.c
zuckschwerdt: Mongoose is used for `cs_to_hex()`, though actually `hex()` shouldn't be there, it's used for `DATA_STRING, packet_hex,` but please either debug print buffers or output decoded fields.
On that, can you document the packet field structure in a doc comment to `honeywell_cm921_decode()`? I guess there are different packets looking like the struct's?

src/devices/honeywell_cm921.c
zuckschwerdt: Point in case: it's leaking memory.

src/devices/honeywell_cm921.c
zuckschwerdt: It's unused code, `decode_device_id()` isn't called as `style` is fixed to `1`.
Is it four different IDs per packet or is it a four-part ID? Wouldn't the type from this table be really useful?

src/devices/honeywell_cm921.c
wintersteiger: Done!

src/devices/honeywell_cm921.c
wintersteiger: Done!

src/devices/honeywell_cm921.c
wintersteiger: Yeah, I got rid of these, the various hex conversion were redundant anyway. 

The packet field structure is the Evohome structure, there are a couple links in the code that point to webpages that describe some of it. The full specification is not known.

src/devices/honeywell_cm921.c
wintersteiger: Done!

src/devices/honeywell_cm921.c
wintersteiger: Done!

src/devices/honeywell_cm921.c
wintersteiger: Neither am I, it doesn't work in this context. I got rid of it.

src/devices/honeywell_cm921.c
wintersteiger: Not sure 48 is the right number, but I did get rid of the allocations.

src/devices/honeywell_cm921.c
wintersteiger: I've never used the json output, but I can take a look.

It's between 1 and 3 device IDs per packet (don't ask me why the did that...) and the first 6 bit of each ID indicate the device type. I don't like the "type:serial" format (`style==0`) , but some people seem to prefer it. Probably because it's the default output format of the Honeywell HGI80 gateway, which seems to have been popular at some time.

src/devices/honeywell_cm921.c
wintersteiger: That's been fixed too.

src/devices/honeywell_cm921.c
wintersteiger: Yeah, helper would be nice, I think I had ended up with three different ways to convert to hex :) (Fixed now).

src/devices/honeywell_cm921.c
wintersteiger: Done.

src/devices/honeywell_cm921.c
wintersteiger: Done.

wintersteiger:
Thanks for the thorough review! It does seem to work just fine without the `-l`. 

I have plenty of samples, I'll submit those too. There are some very rare cases in which rtl_433 and PulseView seem to disagree on single bits, resulting in rtl_433 not decoding them correctly, but they are display in PulseView correctly, which can be a bit confusing. I'm not sure I'll be able to get one of those quickly, but I suppose there will always be some minor discrepancies between those two.

merbanan:
-Y minmax needs to be added to get the same output from Pulseview.

jfberry:
@wintersteiger First, fantastic work.

I believe that the protocol is called Ramses II this may be a better name than CM921 which is just one of the products using this protocol.

I have a large evohome installation and am using a HGI80 to monitor it with domoticz. I ran up this pull request out of interest, and it is logging things from my installation.

`_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2020-06-26 18:17:35
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 107bc7 064e7d unknown   : 12b0
time      : 2020-06-26 18:17:37
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 107bc7 064e7d unknown   : 12b0
time      : 2020-06-26 18:17:54
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 064e7d 064e7d domain_id : 250          demand    : 0.000
time      : 2020-06-26 18:18:02
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 5b9705 5b9705 domain_id : 0            demand    : 0.000
time      : 2020-06-26 18:18:03
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 5b9705 5b9705 domain_id : 0            demand    : 0.000
time      : 2020-06-26 18:18:18
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 107575 107575 unknown   : 30c9
time      : 2020-06-26 18:18:19
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 5b96e5 5b96e5 domain_id : 0            demand    : 0.000
time      : 2020-06-26 18:18:21
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 5b96e5 5b96e5 domain_id : 0            demand    : 0.000
time      : 2020-06-26 18:18:32
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 107579 107579 unknown   : 30c9
time      : 2020-06-26 18:18:51
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 1075db 064e7d unknown   : 12b0
time      : 2020-06-26 18:18:51
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 1075db 064e7d unknown   : 12b0
time      : 2020-06-26 18:18:52
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 107575 064e7d unknown   : 3150
time      : 2020-06-26 18:18:53
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 107579 064e7d unknown   : 3150
time      : 2020-06-26 18:19:07
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 10764d 064e7d unknown   : 3150
time      : 2020-06-26 18:19:09
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 107649 064e7d unknown   : 3150
time      : 2020-06-26 18:19:56
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 1075a9 1075a9 unknown   : 30c9
time      : 2020-06-26 18:20:10
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 1075a7 1075a7 unknown   : 30c9`

If you would like some assistance improving coverage let me know (or I may cross reference against the domoticz interpreter).

zuckschwerdt:
@jfberry thanks, good pointer! There is good info on the Ramses II protocol and Residential Network Protocol (RNP). We should generalize the decoder with that info.

jfberry:
This Wiki has done a decent job documenting the protocol. I am an rtl_433 beginner (only installed it today), but it would be nice if the HGI80 serial format could be used as an option, then this would be able to act as an alternative input into an ecosystem which already exists.

https://github.com/Evsdd/The-Evohome-Protocol/wiki

merbanan:
@jfberry either create a output plugin or a wrapper.

zuckschwerdt:
Easiest would be to modify https://github.com/merbanan/rtl_433/blob/master/examples/rtl_433_custom.py#L38

wintersteiger:
@jfberry that's great, thanks for the logged messages! My Evohome devices are limited to CM921+BDR91, so I could only test a very small subset of the protocol, but decided to put it up here anyways in case someone wants to extend it. I think it's feasible to extend it to where it can be a replacement for the HGI80, as long as we have devices to test it (especially for the rarer message types of which we might not get many logs).

I'm happy to help adding missing features, I just can't properly test them. If you compile rtl_433 in debug mode, it will log the raw data (Packet, Header, etc, these: https://github.com/wintersteiger/rtl_433/blob/f38bb00f6192e335ee024d7cfc21d5133c278cc5/src/devices/honeywell_cm921.c#L383). From those logs, I think I would be able to add a significant amount of  code without having to test it with actual devices immediately.

As the others suggested, the translation to the HGI80's format is relatively easy. It seems to be just a translation of the raw payload data into ASCII, making it easier to process over serial lines. The fields of the messages seem to be the same though, and their order too.

src/devices/honeywell_cm921.c
wintersteiger: (I'll mark the non-controversial conversations resolved.)

jfberry:
> I'm happy to help adding missing features, I just can't properly test them. If you compile rtl_433 in debug mode, it will log the raw data (Packet, Header, etc, these: https://github.com/wintersteiger/rtl_433/blob/f38bb00f6192e335ee024d7cfc21d5133c278cc5/src/devices/honeywell_cm921.c#L383). From those logs, I think I would be able to add a significant amount of code without having to test it with actual devices immediately.

I will build the debug version now and execute tomorrow. I'll try and recreate a good cross section of messages by changing some set points both from HR92s and the controller, restarting domoticz (which sends messages to the controller to query it etc).  Is there any particular command line you would prefer I use to capture the best information?

jfberry:
@wintersteiger (see previous response) and try as I might I can't work out how to get the _DEBUG flag set through cmake/make - any clues?

jfberry:
@wintersteiger just doing some testing - modified the cm921.c file to define _DEBUG. Got a segmentation fault after a while, so ran with -vv - not sure if this helps or not

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2020-06-28 17:45:40
model     : Honeywell CM921                        MIC       : CHECKSUM
Device IDs: 5b96e5 5b96e5 domain_id : 0            demand    : 0.000         Packet    : 185b96e55b96e5000802000032              Header    : 18
Command   : 0008         Payload   : 0000          CRC       : 32            # man errors: 0
pulse_demod_pcm(): Honeywell CM921 Wireless Programmable Room Thermostat
bitbuffer:: Number of rows: 1
[00] {375} 55 55 55 55 5f f0 05 99 55 59 54 ac ad 4c d4 d4 b4 d3 35 53 34 cd 4d 4b 4d 33 55 33 2a ca b2 ac ad 2a cc b2 ac ab 2a ca b5 2c cb 56 55 55 54
honeywell_cm921_decode: {420} 24 92 92 49 24 84 92 49 24 92 49 24 92 49 92 49 24 93 90 f8 41 f2 1f 08 10 80 40 9c 00 00 00 10 00 10 00 00 00 00 00 00 00 00 01 00 00 00 00 00 00 00 00 00 00
preamble_start=420 start=450 len=-30
honeywell_cm921_decode: {277} 90 20 10 07 f9 09 78 04 f9 c3 87 08 3e 70 3f 92 10 ff fc 38 00 e7 c0 e4 e4 09 3f c0 21 38 49 f0 03 e0 18
preamble_start=277 start=307 len=-30
honeywell_cm921_decode: {347} 24 92 49 24 92 49 24 92 49 24 92 49 24 92 49 24 92 49 c8 7c 20 7c 87 c2 04 09 f0 27 00 02 00 00 00 0e 00 00 00 00 00 00 00 00 00 00
preamble_start=347 start=377 len=-30
honeywell_cm921_decode: {143} 90 20 03 80 01 c0 00 00 00 0f 84 3e 72 49 00 38 7f f0
preamble_start=143 start=173 len=-30
honeywell_cm921_decode: {479} 13 26 66 64 cc cc cc cc cc cc cc 8c cc cc cc cc cc cc cc cc cc cc cc cf 30 fc 20 7e 61 f8 60 61 80 60 ce 00 00 00 e2 00 0c 00 00 00 00 00 00 00 00 00 60 00 00 00 00 00 00 00 00 00
preamble_start=479 start=509 len=-30
honeywell_cm921_decode: {305} e6 18 06 00 33 00 3f 03 cc ce 7f 00 9e 06 1e 70 c0 c3 c0 0c c0 30 c3 f8 03 fe 00 66 7f 01 e4 3c cf ff f8 1e 60 66 00
preamble_start=305 start=335 len=-30
honeywell_cm921_decode: {341} 24 90 92 49 24 92 49 24 92 49 24 92 49 24 92 49 24 93 81 f0 83 e4 3e 10 20 4f 02 70 07 20 00 00 03 80 00 00 00 00 00 00 00 00 00
preamble_start=341 start=371 len=-30
honeywell_cm921_decode: {142} c4 20 03 80 03 80 00 00 00 e4 f8 72 73 98 24 3e 07 f8
preamble_start=142 start=172 len=-30
honeywell_cm921_decode: {317} 55 7f c0 16 65 55 65 4a b2 d4 b3 2b 54 b2 cd 33 2d 4a b3 4c b3 35 55 34 d5 4b 2a cb 54 b4 ab 4a ca b2 ad 33 4a d5 95 50
preamble_start=317 start=347 len=-30
honeywell_cm921_decode: {448} aa aa aa aa aa aa aa aa aa aa aa ab fe 00 b3 2a ab 2a a5 96 a5 59 a6 59 9a aa 9a 6a 59 95 aa 59 66 99 96 aa 59 56 5a a5 a5 5a 96 55 95 65 59 5a a6 a6 a9 99 9a ac aa aa
preamble_start=83 start=113 len=335
Segmentation fault
```

wintersteiger:
The easiest way to enable debug mode is to add it to the CMake command line, e.g. `cmake -DCMAKE_BUILD_TYPE=Debug ..`, which should enable `_DEBUG` everywhere. Cleaning (erasing) the build directory may help so it doesn't mix release and debug object files.

I'll have to look at the segfault - I think I should be able to trigger by manually feeding it that package data.

I think, to trigger a variety of different messages, it would be a good idea to go through all the settings in your controller(s) and changing each of the settings, then change it back to the old value again. That can help detect encoding issues (numbers and IDs etc) and it enables features that you normally don't use. If you have a complex controller though, there may be quite a large number of settings, so this could take some time :-)

wintersteiger:
Oh no, you were right, in this particular project I added `_DEBUG` as a local macro, so you do have to add that manually. There was indeed a problem with the decoder that likely caused the segfault you saw. It's not currently known how many device IDs are sent with each message, but we do know that information is encoded in the header and I had never seen a `3C` header before. I fixed that now and I added the 30C9 message type, so I can tell you that it had 20.56C in your zone 0 at the time. 

I also rebased this branch onto the latest master, which means the protocol now has ID 155 instead of 151. 

@merbanan @zuckschwerdt: is the only reason this hasn't been merged into master the missing json support? (I can add that if necessary, just a bit short on time.)

jfberry:
@wintersteiger A good selection of Evohome Ramses II protocol capture here -- https://pastebin.com/2sKiuTsw
I changed set points from controllers, let heating turn on (capture didn't get until heating turned all the way off again), had a shower so hot water temperature and demand, pressed ID buttons, did some testing, on top of which domoticz restarted which interrogates the controller for zone names and fault logs (it also interrogates temperatures every 5 minutes so there this will be a bit noisier than a regular capture). This sent various demand requests to the boiler relays.
It's certainly enough to get a pretty grounded set of coverage I would think.
https://pastebin.com/2sKiuTsw

wintersteiger:
Thanks for the logs @jfberry that's a much better variety of messages than the ones I get! 

I ran into some trouble with the message header decoding; I believe they may not contain the information I was looking for (e.g. number of device IDs). I factored the decoder out into it's own project ([here](https://github.com/wintersteiger/evodec)) to make it easier to test and tweak, and I added an HGI80-like output format, but I don't expect that domoticz would be able to do anything useful with it yet (but I didn't try it). As input, it takes messages of the form that rtl_433 labels "Packet", e.g. `143255c1000802000694`.

Since you have an HGI80, would it be possible for you to record a log with time-stamped messages of both systems, so that we can correlate them? I think that would make decoding the headers and device IDs much easier. 

Another reason for factoring this out into it's own project is that I started experimenting with 868MHz-specific transceiver modules (e.g. based on TI CC1101) and they do a relatively good job at receiving the raw packets in hardware for very little money. With those, I think it's feasible to build an HGI80 replacement based on a Raspberry PI Zero or something similarly low-end.

jfberry:
> Thanks for the logs @jfberry that's a much better variety of messages than the ones I get!

No problem, happy to help with some correlated logs. I may be able to turn the domoticz debug logs on again (I moved away from the debug build now it is all working) - but these have exactly what you would want and would mean that I don't need to do anything different.

Incidentally there has been quite a bit of work on an open source replacement for the HGI80 - using an RFBee and a UartSbee, including some arduino code to do the decoding and formatting. I think it can do the transmit side of the HGI80 as well, but perhaps not reliably

https://www.automatedhome.co.uk/vbulletin/showthread.php?5085-My-HGI80-equivalent-Domoticz-setup-without-HGI80/page2

vsessink:
Hi, only just entering the discussion. @wintersteiger great work. A short remark: the way the number of headers is calculated, makes things go weird sometimes: msg 0c1d001353284922f1010be1 is interpreted as follows: dev_id 1d0013 532849 22f101, command 0bfc (note the fc being a buffer overflow) and a payload that is even more buffer overflow (fcfcfcfc etc). Wouldn't it be better to do something like: 1. if we happen to know the exact # of dev_id, use that; if we don't: try to figure out if a certain number of dev_id will make the message "complete"? For example: msg above can only be interpreted if 01 is the payload length; all other bytes are too large to be the length (because they are larger than the message itself).  Which, in turn, means "command" is 22f1 and 532849 could be ID2 or timestamp. Should I try code something?

wintersteiger:
Sure, that sounds like a good idea @vsessink. I'm not sure whether rtl_433 would want that in the main branch, but it would definitely help with debugging and testing!

vsessink:
Well all right. After some testing with your test messages , some more test messages from my own setup, the only values I can find are 0x4, 0x8 0xc combined with 0x10, 0x20 and 0x30. Nevertheless, adding
```
                          msg->header == 0x0c ? 2 :
                          (bitrow_get_byte(bmsg->bb[row], 6*8) == (num_bytes - 6 - 2)) ? 1 :
                          (bitrow_get_byte(bmsg->bb[row], 9*8) == (num_bytes - 7 - 2)) ? 1 :
                          (bitrow_get_byte(bmsg->bb[row], 7*8) == (num_bytes - 9 - 2)) ? 2 :
			  (msg->header >> 2) & 0x03; // total speculation.
```
... helps, because the "total speculation" part is less likely to trigger. The above would be enough to implement the possibilities as described in EVOHOME_WirelessProtocol.pdf. Please note that location 7 means that there isn't actually a second ID, there is a timestamp and a command - but I can't check as I haven't seen any packets with 0x16 for header byte. The full evo protocol, if the document called evotouch_protocol.pdf is right, consists of three optional ID's, two optional parameter bytes, then command and length, a total of 12 possible length locations (anywhere between position 1 and position 12). Testing for ```(bitrow_get_byte(bmsg->bb[row], location*8) == (num_bytes - location - 2))``` would still reveal possible locations for the length field. Please feel free to implement the above in either honeywell_cm921 or in your evodec code.

PR #1341, opened 2020-03-25
Added support for sigrok convert on windows

DerekGn:
Change:
Modified write_sigrok.c to create sigrok file, depends on 7zip being installed and on the path.
Modified rtl_433.c to close and flush the dumper file handles due to a file write behavior difference's between windows and linux. The files contain 0 bytes when the sigrok file is being created as the dumper files have not been closed at that point.

### DIFF
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -85,16 +85,12 @@ sample_rate   250k
 # see "decoder" section below.
 
 # as command line option:
-#   [-l <level>] Change detection level used to determine pulses [0-16384] (0 = auto) (default: 0)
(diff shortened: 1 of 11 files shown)


zuckschwerdt:
Great, thanks! Can you copy the dumper fclose loop from r_free_cfg to close_dumpers (and add dumper->file=NULL) instead of changing call order?

DerekGn:
Yes, no problem, I will do this shortly.

DerekGn:
I have updated the code to close the dumpers prior to the sigrok file generation. I have reverted the ordering change in rtl_433.c

PR #1344, opened 2020-04-09
Update template guideline for verbosity

### LINKED ISSUE #1343: r_dev->verbose does not match cfg->verbosity
- According to the following, verbosity specified at the command line is not the same as the one available to the decoder 
https://github.com/merbanan/rtl_433/blob/0e355dc8973c1bb0b9b4dff68a737ef5945741bd/src/r_api.c#L167-L170
- This is inconsistent with the man page and src/devices/new_template.c
- I prefer that verbosity is consistent at both location. I could send a PR to fix this but which one should be the expected behavior?

x0urc3:
Update verbosity guideline for new device templates 

Fixes #1343

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ stamp-h1
 libtool
 Doxyfile
 .dirstamp
+tags
 
 .tarball-version
 .version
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
Thanks.

PR #1345, opened 2020-04-09
Remove list of supported device protocols from man page

x0urc3:
The list of supported device protocols is looong. This PR removes the list of supported device protocols from *nix man pages.  User can use `rtl_433 -R` to get the list of supported devices


merbanan:
@zuckschwerdt I think we want that list don't we?

zuckschwerdt:
Can't say. I never use the man page, and it really is quite long. The request is reasonable, I guess people would just want to look up CLI parameters.

merbanan:
Maybe move it to the end?

PR #1349, opened 2020-04-13
Add support for Visonic Powercode devices

maxlock:
Adds support for Visonic Powercode devices. Still requires packet checksum to be checked, but my C isn't that good yet :) Full details on the protocol included here and in an accompanying pull request in https://github.com/merbanan/rtl_433_tests/pull/342

### DIFF
--- a/src/devices/visonic_powercode.c
+++ b/src/devices/visonic_powercode.c
@@ -1,0 +1,125 @@
+/** @file
+    Decoder for Visonic Powercode devices. Tested with an MCT-302.
+
+    Copyright (C) 2020 Maxwell Lock
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+ */
+
+/**
+The device uses OOK PWM encoding, short pulse 400us long pulse 800us, and repeats 6 times
+You can use a flex decoder -X 'n=visonic_powercode,m=OOK_PWM,s=400,l=800,r=5000,g=900,t=160,y=0'
+
+Powercode packet structure is 37 bits. 4 examples follow
+
+              s addr                       data     cksm
+              1 01101111 01000111 01110000 10001100 1001 - magnet near, case open
+              1 01101111 01000111 01110000 11001100 1101 - magnet away, case open
+              1 01101111 01000111 01110000 00001100 0001 - magnet near, case closed
+              1 01101111 01000111 01110000 01001100 0101 - magnet away, case closed
+              | |                        | |||||||| |  |
+     StartBit_/ /                        / |||||||| \__\_checksum, XOR of preceding nibbles
+     DeviceID__/________________________/  ||||||||
+                                           ||||||||
+                                    Tamper_/||||||\_Repeater
+                                      Alarm_/||||\_Spidernet
+                                     Battery_/||\_Supervise
+                                         Else_/\_Restore
+
+1 bit start bit
+3 byte(24 bit) device ID
+1 byte data
+1 nibble (4 bit) checksum
+
+Checksum is a londitudinal redundancy check of the 4 bytes containing the device ID and data.
+Bytes are split into nibbles. 1st bit of each nibble is XORed and result is 1st bit of checksum,
+then the same for the 2nd, 3rd and 4th bits. To be implemented.
+
+Protocol cribbed from:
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Looks good.
- `brand` should be removed
- `model` should be "Visonic-Powercode"
- order should be model, id, ...
- the bits should be DATA_INT of 0/1
- indent in continued lines (data_make, output_fields, r_device) should be 8 spaces
- last `NULL` in output_fields should have a comma trailing
- data_make should be surrounded by `/* clang-format off */`, `/* clang-format on */`
- the ascii art in the `/**` decode comment needs 4 leading spaces, the list needs dashes

maxlock:
OK, I think that covers the changes.

src/devices/visonic_powercode.c
```diff
@@ -0,0 +1,124 @@
...
+                                      Alarm_/||||\_Spidernet
+                                     Battery_/||\_Supervise
+                                         Else_/\_Restore
+
+1 bit start bit
+3 byte(24 bit) device ID
+1 byte data
+1 nibble (4 bit) checksum
+
+Checksum is a londitudinal redundancy check of the 4 bytes containing the device ID and data. Bytes are split into nibbles. 1st bit of each nibble is XORed and result is 1st bit of checksum, then the same for the 2nd, 3rd and 4th bits. To be implemented.
```
Mindavi, line 37: longitudinal

also, consider splitting into multiple lines for readability (e.g. github mobile doesnt line wrap automatically)

src/devices/visonic_powercode.c
```diff
@@ -0,0 +1,124 @@
...
+
+static int visonic_powercode_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    uint8_t *b; // bits of a row
+    uint8_t msg[32];
+    uint8_t sum[4];
+    data_t *data;
+    char id[7];
+
+    // 37 bits expected, 6 minimum packet repetitions
```
Mindavi, line 55: if one packet is dropped, does that make the whole transmission invalid? according to documentation above the total amount of transmissions is 6. Maybe accept a packet loss of 1 or 2 packets? Might help for people with noisy environments.

src/devices/visonic_powercode.c
```diff
@@ -0,0 +1,124 @@
...
+
+r_device visonic_powercode = {
+        .name        = "Visonic powercode",
+        .modulation  = OOK_PULSE_PWM,
+        .short_width = 400,  // short gap is 132 us
+        .long_width  = 800,  // long gap is 224 us
+        .gap_limit   = 900,  // some distance above long
+        .reset_limit = 5000, // a bit longer than packet gap
+        .decode_fn   = &visonic_powercode_decode,
+        .disabled    = 0, // disabled and hidden, use 0 if there is a MIC, 1 otherwise
```
Mindavi, line 122: disable by default until MIC is implemented

src/devices/visonic_powercode.c
```diff
@@ -0,0 +1,124 @@
...
+        "supervised",
+        "spidernet",
+        "repeater",
+        NULL,
+};
+
+r_device visonic_powercode = {
+        .name        = "Visonic powercode",
+        .modulation  = OOK_PULSE_PWM,
+        .short_width = 400,  // short gap is 132 us
```
Mindavi, line 117: are the comments valid here? otherwise remove or update

src/devices/visonic_powercode.c
Mindavi: if (row < 0), no messages found (iirc). You can return then. Otherwise you might index with -1. Please check if return code of bitbuffer_find_repeated_row can be negative.

maxlock:
@Mindavi I've removed the errant comments, and set to disabled. I've also allowed it to accept fewer packets.  What is the purpose of mic?

Mindavi:
MIC stands for Message Integrity Check. It means that a checksum or parity is checked. Because checksums are typically different between devices, they're a good way to filter out false positives. That's why devices without it are disabled by default.

src/devices/visonic_powercode.c
maxlock: fixed

src/devices/visonic_powercode.c
maxlock: fixed

Mindavi:
Thanks for the contribution!

zuckschwerdt:
@Mindavi also run ./maintainer_update.py to refresh the docs, add to the rtl_433.example.conf, and maybe add the backslash on the last entry in include/rtl_433_devices.h -- not sure when that got lost.

PR #1353, opened 2020-04-15
Bresser Weather Station 5in1 Wind Fixes #1315 & Battery Low Fixes #1323

### LINKED ISSUE #1315: Bresser 5-in-1 wind issue (not displayed correctly)
I have been running the Bresser 5-in-1 for over a year

I have noticed that the wind speed displayed on the console is always higher than what is recorded in rtl_433

For example, the wind speed on console would be 7.2km/h but my rtl_433 stream would show 1.6ms (5.76km/h)

Any ideas where to even start?

### LINKED ISSUE #1323: please review battery_low flag for bresser_5in1
i have a temperature sensor very similar to bresser_5in1 (it has lower features, will add info later) and found that there is a low-battery bit at bit7 of byte 25.

someone with bresser5in1 device may check if this also applies to that device, i guess it's the same there and has been overlooked (as it must be tested specifically

also see
https://groups.google.com/forum/#!topic/rtl_433/G0dou0HomGI

peteakalad:
Bresser 5 in 1 Wind Related Fixes;
- Wind Gust Fix merbanan/rtl_433#1315 Change to use normal binary conversion for gust (for some reason inconsistent with average wind speed).
- Fix for both Gust and Average wind speeds for large values.  Wrong MSB was taken for both.  MSB taken from bits after LSB like other metrics - confirmed working.
- Update initial comment to correct understanding of bits.

Tested from 0-35m/s for both wind gust and wind average speed.

### DIFF
--- a/src/devices/bresser_5in1.c
+++ b/src/devices/bresser_5in1.c
@@ -28,7 +28,10 @@
  * f1 fd 7f ff af ff ef bd fd b7 c9 ff ff 0e 02 80 00 50 00 10 42 02 48 36 00 00 00 00 (from https://github.com/merbanan/rtl_433/issues/719#issuecomment-388896758)
  * ee b7 7f ff 1f ff ef cb fe 7b d7 fc ff 11 48 80 00 e0 00 10 34 01 84 28 03 00 (from https://github.com/andreafabrizi/BresserWeatherCenter)
  * e3 fd 7f 89 7e 8a ed 68 fe af 9b fd ff 1c 02 80 76 81 75 12 97 01 50 64 02 00 00 00 (Large Wind Values, Gust=37.4m/s Avg=27.5m/s from https://github.com/merbanan/rtl_433/issues/1315)
- * CC CC CC CC CC CC CC CC CC CC CC CC CC uu II    GG DG WW  W TT  T HH RR  R  t
+ * ef a1 ff ff 1f ff ef dc ff de df ff 7f 10 5e 00 00 e0 00 10 23 00 21 20 00 80 00 00 (low batt +ve temp)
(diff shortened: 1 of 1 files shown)


peteakalad:
Added update commit for bresser 5in1 battery ok / low

vekexasia:
Tested this branch and seems to be working. Assuming that Battery: 1 means "ok"
image

Thanks @peteakalad

peteakalad:
Thanks for confirming. 

Yes the label ought to be “Battery ok” to be less ambiguous but that’s non-device code setting that.  I made the battery flag changes consistent with the other devices.

Battery value meaning:
1 = Battery Ok
0 = Battery Low

zuckschwerdt:
This still needs regression checking and analysis. I'm not sure if both values were that wrong for the original author. Wind direction should have been obviously wrong? But the MSB on the other values was perhaps never tested?

peteakalad:
Understand. 

Was no bugs with wind direction this was not changed in this PR. 

Issues fixed were for large wind gust and actual wind speed where the right MSB were not being included in the calculation.  This would only be noticed for high wind speeds which I was able to test with a leaf blower.  I suspect this just hadn’t been noticed as most had only seen/tested lower speed winds.  

The battery ok was new addition after working out the bit which carried this with old batteries.

peteakalad:
See #1315

zuckschwerdt:
Ah, right. I just saw `GGGD` changing to `GGDG` and thought the `D` moved, but there are gaps :)
@mercurystorm have you testes this PR? It looks good to merge, @peteakalad did solid detective work here!

mercurystorm:
I have been running it since @peteakalad pushed it and its been running perfectly! Its nice to have the gusts reporting correctly and the same as the Bresser display unit

zuckschwerdt:
Thanks for the great work!

PR #1355, opened 2020-04-16
visonic_powercode: add lrc check and enable by default

Mindavi:
@maxlock could you check if this works for your device? It seems to work for the 2 samples in the rtl_433_tests repository.

### DIFF
--- a/src/devices/visonic_powercode.c
+++ b/src/devices/visonic_powercode.c
@@ -52,8 +52,7 @@ static int visonic_powercode_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     uint8_t msg[32];
     data_t *data;
     char id[7];
-    uint8_t lrc_calculated = 0x00;
-    uint8_t lrc            = 0xff;
+    uint8_t lrc;
 
     // 37 bits expected, 6 packet repetitions, accept 4
     int row = bitbuffer_find_repeated_row(bitbuffer, 4, 37);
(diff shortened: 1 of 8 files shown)


src/devices/visonic_powercode.c
```diff
@@ -52,30 +52,50 @@ static int visonic_powercode_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+    // Take high nibble of lrc
+    lrc            = (lrc & 0xf0) >> 4;
+    lrc_calculated = 0;
+    for (int i = 0; i < 4; i++) {
+        // low nibble
+        lrc_calculated ^= (msg[i] & 0xf);
+        // high nibble
+        lrc_calculated ^= ((msg[i] & 0xf0) >> 4);
+    }
+    lrc_calculated &= 0x0f;
```
zuckschwerdt, line 83: It's possibly easier to just xor the four bytes, then fold the nibbles once at the end. Also you can just add the chk too (expand the extract to 36) E.g. `b = xor_bytes(msg, 5); (b >> 4) ^ (b & 0xf) == 0`

src/devices/visonic_powercode.c
```diff
@@ -120,6 +140,6 @@ r_device visonic_powercode = {
         .gap_limit   = 900,
         .reset_limit = 5000,
         .decode_fn   = &visonic_powercode_decode,
-        .disabled    = 1,
+        .disabled    = 0,
```
zuckschwerdt, line 143: needs the "mic" field added ;)

maxlock:
> @maxlock could you check if this works for your device? It seems to work for the 2 samples in the rtl_433_tests repository.

@Mindavi yes, that's working fine here. Thanks for adding! :)

Mindavi:
@zuckschwerdt I've run the maintainer_update.py script (and updated it to have python3 support, python2 has been deprecated for a while now), but it inserts some weird characters. I've tried running it with python2, but that gave the same results. Is it ok?

zuckschwerdt:
Huh, strange. Those extra characters don't belong (and are harmful in the man page). I'll try to find where they come form.

zuckschwerdt:
I guess it could be the [get_help_text](https://github.com/merbanan/rtl_433/blob/master/maintainer_update.py#L55) regex line, but it doesn't occur in my setup.

Mindavi:
Hey @zuckschwerdt, I fixed the maintainer_update script for my setup. Could you please double-check if it's working at your end as well? Otherwise I think this should be good to go.

zuckschwerdt:
Oh, so it's differences in the getopt implementation. Mine doesn't quote the option character. Perhaps `\'?.\'?`is needed then?

zuckschwerdt:
Can you also handle #1358 -- this PR needs a rebase after that.

zuckschwerdt:
@Mindavi can you also handle #1362 -- looks like a collision ;)

Mindavi:
Will handle both of those (probably) tomorrow and fix the script with the updated regex. Your suggestion sounds like a good one, will try it and and use it if it works :).

PR #1357, opened 2020-04-20
Extend & fix MQTT Home Assistant example

maxlock:
Current implementation sends discovery messages like this:

`homeassistant/binary_sensor/Visonic-Powercode-6f4770-battery/config  :  {"payload_on": "1", "state_topic": "rtl_433/kvm/events", "payload_off": "0", "name": "Battery", "device_class": "battery"}`

The problem is that the sensor names and descriptions that appear in Home Assistant aren't user friendly, nor are they always unique as homeassistant doesn't pick up the device id. This change updates the above message to look like this

`homeassistant/binary_sensor/Visonic-Powercode-6f4770/Visonic-Powercode-6f4770-battery/config  :  {"name": "Visonic-Powercode-6f4770-battery", "payload_on": "1", "device": {"model": "Visonic-Powercode", "identifiers": "6f4770", "name": "Visonic-Powercode-6f4770"}, "device_class": "battery", "state_topic": "rtl_433/kvm/devices/Visonic-Powercode/6f4770/battery", "payload_off": "0"}`

referencing the [homeassistant discovery documentation](https://www.home-assistant.io/docs/mqtt/discovery/)

Firstly the topic is extended to populate the nodeid. Next the name is updated to contain the device type and id making it unique. A device construct is added for the homeassistant device registry containing a unique identifier and name. Lastly the state topic is added as homeassistant no longer supports implied state topics and to take data from the devices topics rather than the event topic.

Additionally I've added some device mappings for rtl_433 generated wireless related data generated by the -M levels option, and additional battery mapping for binary_state devices and some security mappings for devices I'm currently setting up.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -33,7 +33,7 @@ mappings = {
             "device_class": "temperature",
             "name": "Temperature",
             "unit_of_measurement": "°C",
-            "value_template": "{{ value }}"
+            "value_template": "{{ value|float }}"
         }
     },
     "temperature_1_C": {
@@ -43,7 +43,7 @@ mappings = {
             "device_class": "temperature",
             "name": "Temperature 1",
             "unit_of_measurement": "°C",
-            "value_template": "{{ value }}"
+            "value_template": "{{ value|float }}"
         }
     },
     "temperature_2_C": {
@@ -53,7 +53,7 @@ mappings = {
             "device_class": "temperature",
             "name": "Temperature 2",
             "unit_of_measurement": "°C",
-            "value_template": "{{ value }}"
+            "value_template": "{{ value|float }}"
         }
     },
     "temperature_F": {
@@ -63,7 +63,7 @@ mappings = {
             "device_class": "temperature",
             "name": "Temperature",
             "unit_of_measurement": "°F",
-            "value_template": "{{ value }}"
+            "value_template": "{{ value|float }}"
         }
     },
 
@@ -74,7 +74,7 @@ mappings = {
             "device_class": "battery",
             "name": "Battery",
             "unit_of_measurement": "%",
-            "value_template": "{{ float(value) * 99 + 1 }}"
+            "value_template": "{{ float(value|int) * 99 + 1 }}"
         }
     },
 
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Thanks! Looks like good ideas. The canonical battery flag name is `battery_ok`. Note that "localhost" as opposed to "127.0.0.1" might bind on IPv6 which then might not work.

maxlock:
I fixed the host entry. The battery flag is an interesting one. There are examples of both in the rtl_433 devices code. Home assistant has classes for both [regular](https://www.home-assistant.io/integrations/sensor/#device-class) and [binary](https://www.home-assistant.io/integrations/binary_sensor/#device-class) sensors. regular sensors contains a percentage, and binary are essentially booleans.

zuckschwerdt:
`battery` is the old key and converted to `battery_ok` (https://github.com/merbanan/rtl_433/blob/master/src/r_api.c#L502).
See https://triq.org/rtl_433/DATA_FORMAT.html#common-device-data for more info on common keys ;)
With recent rtl_433 you should only see `battery_ok` with in the output.

maxlock:
ok, that's interesting. it would appear that the mqtt code isn't doing that translation, as it still outputs battery.

zuckschwerdt:
I see. `Visonic-Powercode` indeed has `battery` as part of a big set of flags. I already suggested to change that ;)

maxlock:
:) yeah I can change that, but there's others too. So is data_acquired_handler in r_api.c supposed to translate battery to battery_ok? if so it's not doing it for the mqtt output.

zuckschwerdt:
There really shouldn't be, we aim for a common handling of those basic keys. Only "battery" with a string value (OK=1, otherwise=0) will be converted though.
PR welcome, if you spot some offending keys in decoders. I just tested MQTT (`rtl_433 -F mqtt://127.0.0.1:1883`) and I see `rtl_433/host/devices/Bresser-3CH/1/32/battery_ok`.

maxlock:
ok makes sense, I've removed the battery mapping.

maxlock:
just pushed a final update to fix the value templates. they are no longer json.

roger-:
BTW why not reformat `"model": "Visonic-Powercode" ` to `"model": "Visonic Powercode" `?

PR #1358, opened 2020-04-21
Fix visonic device battery reporting

maxlock:
@zuckschwerdt how does this look?


zuckschwerdt:
Thanks! "battery" "LOW" "OK" is the old scheme which we are phasing out. "battery_ok", 0/1 would be better ;)

maxlock:
OK, fixed. :)

Mindavi:
The rebasing was a little uncharted territory for me, but it's merged now with c47bef0815e5240253236c023f21deeff5979058 :)

PR #1364, opened 2020-04-24
build: fix build for FreeBSD

micro3x:


zuckschwerdt:
Thanks! Is that all that's needed? What inputs work, do you use librtlsdr, or SoapySDR? MAybe you also want to add notes to BUILDING.md?

zuckschwerdt:
I guess we should really use
```
    #include <sys/types.h>
    #include <sys/socket.h>
    #include <netdb.h>
    #include <netinet/in.h>
```
unconditionally (except Win32).

Is settting `__BSD_VISIBLE` the proper way? Shouldn't we select the right `_SOURCE` level?
Only setting `_POSIX_C_SOURCE`, `_ANSI_SOURCE` or `_C99_SOURCE` will hide `__BSD_VISIBLE` and we set none of those, right?

zuckschwerdt:
f26830d inserts the missing include. Can you test again and rebase this change if needed?

micro3x:
"Fix socket portability" worked perfectly.
Closing this PR since now it's obsolete.

zuckschwerdt:
Thanks for testing and the feedback!

PR #1371, opened 2020-05-02
Prometheus/OpenMetrics relay

Wilm0r:
Wrote and debugged this over the last week. Should be pretty solid at least for (OS) weather sensors, allowed me to integrate weather info in existing Grafana dashboards and now I'm finally getting alerts when batteries run empty.

### DIFF
--- a/examples/rtl_433_prometheus_relay.py
+++ b/examples/rtl_433_prometheus_relay.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+
+import collections
+import json
+import os
+import sys
+import threading
+import time
+
+import dateutil.parser
+import http.server
+
+
+class rtl_433(object):
+	# These fields are used to separate and tag unique sensors, instead of being exported
+	# as readings/values.
+	_ID_FIELDS = [
+		("brand", str),
+		("model", str),
+		("channel", int),
+		("id", int),
+	]
+	_MIN_REPEAT_SECS = 3600  # See a device at least twice in this timespan before sharing it.
+	_EXPORT_TIMESTAMPS = True
+	
+	_MAX_AGE_SECS = 300  # Used for gc, and max age to show *anything* for given id
+	_BACKLOG_SECS = 60   # If multiple samples within this timestamp, share them all
+
+	_LOG_CLEAN_INTERVAL = 5000  # Number of iterations
+
+	log = []  # [(timestamp, id_fields, variable name, value), ...]
+
+	def __init__(self, stream):
+		self.stream = stream
+
+	def run(self):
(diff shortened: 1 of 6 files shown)


examples/rtl_433_prometheus_relay.py
```diff
@@ -0,0 +1,146 @@
...
+import time
+
+import dateutil.parser
+import http.server
+
+
+class rtl_433(object):
+	# These fields are used to separate and tag unique sensors, instead of being exported
+	# as readings/values.
+	_ID_FIELDS = [
```
zuckschwerdt, line 17: probably should include `type` and `subtype`

zuckschwerdt:
Thanks! Have you considered using a network input (syslog) to decouple this and not rely on pipes? That way you can run each server (rtl_433, relay) on it's own and possibly under a supervisor.

examples/rtl_433_prometheus_relay.py
Wilm0r: Ah! Those were missing on my OS sensors, didn't know. Added.

Wilm0r:
> Thanks! Have you considered using a network input (syslog) to decouple this and not rely on pipes? That way you can run each server (rtl_433, relay) on it's own and possibly under a supervisor.

I did see that this morning when looking at existing exporters, missed it so far. Looks like the inside of it is still just json so that should be an easy fix. Will update the PR soon.

Wilm0r:
> > Thanks! Have you considered using a network input (syslog) to decouple this and not rely on pipes? That way you can run each server (rtl_433, relay) on it's own and possibly under a supervisor.
> 
> I did see that this morning when looking at existing exporters, missed it so far. Looks like the inside of it is still just json so that should be an easy fix. Will update the PR soon.

I've finished this change but will let it run for a day or so here first.

One thing by the way, I saw in at least one of the exporters:

    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)

Was this added under the assumption that automatic multiplexing will happen? Because what actually will happen in all SO_REUSEPORT implementations that I know of, is the packets randomly (or deterministically through port numbers perhaps?) will go to one of the listener sockets.

zuckschwerdt:
Yes, "Syslog" is just the JSON you'd get on stdout with a fixed header (just to adopt an existing protocol).

I don't remember how that SO_REUSEPORT got there, likely left over from some early debugging. It's indeed a cargo-cult way to do things that can't work. Feel free to remove all instances with your PR.

Wilm0r:
Pushed including `SO_REUSEPORT` cleanup.

PR #1379, opened 2020-05-14
fix ERT Endpoint Type extraction

phrrngtn:
Putative fix for ERT Endpoint Type (149) decoding as per discussion on https://groups.google.com/forum/#!topic/rtl_433/loZFvPhCxOE

Output is now consistent wrt rtlamr.

### DIFF
--- a/src/devices/ert.c
+++ b/src/devices/ert.c
@@ -65,7 +65,9 @@ static int ert_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
     /* Extract parameters */
     physical_tamper = (b[3]&0xC0) >> 6;
-    ert_type = (b[3]&0x60) >> 2;
+    /* endpoint type is 4 bits and starts at position 26.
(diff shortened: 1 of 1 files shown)


src/devices/ert.c
```diff
@@ -65,7 +65,9 @@ static int ert_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
     /* Extract parameters */
     physical_tamper = (b[3]&0xC0) >> 6;
-    ert_type = (b[3]&0x60) >> 2;
+    /* endpoint type is 4 bits and starts at position 26.
+     * xref https://github.com/bemasher/rtlamr/wiki/Protocol */
+    ert_type = (b[3]&0x36) >> 2;
```
zuckschwerdt, line 70: if you want a mask of `0011 1100` that's 0x3c?

phrrngtn: You are correct. Sorry about that! I mis-shifted the lower two bits.

PR #1382, opened 2020-05-15
Adding new device specification for Cotech 36-7959

burken-:
I apologise in advance if the code is not top notch... my experience with C is very limitied to say the least. That said... this class decodes and maps the data transmitted by the Cotech 36-7959 weather station.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -160,6 +160,7 @@
     DECL(klimalogg) \
     DECL(visonic_powercode) \
     DECL(eurochron_efth800) \
+    DECL(cotech_36_7959) \
 
     /* Add new decoders here. */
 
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Needs a bit of polish, yes. But the idea comes across. The use of `extract_number()` is smart, we might need to add that as basic tool.
For the preamble it's better to just use:
```
    uint8_t preamble_bits[] = { 0x01, 0x40 }; // i.e. {12}014
    unsigned preamble_len = 12;
```
Can you add the usual (`/**`) header with copyright and documentation, like found in most other files?

burken-:
Thanks for the feedback! Added copyright and documentation and made some changes to the code regarding the preamble bits.

zuckschwerdt:
The `_loop` code effectivly adds a 9th bit, right? Shouldn't the addition then be `256` instead of `255`? Can you test if that results in plausible values?

burken-:
You're of course right

zuckschwerdt:
Do you want to refine the code yourself? Then I'll add detailed notes. Otherwise I can merge as-is and just bring this to our usual style.

burken-:
I fear there will be many rounds back and forth if I do it. If you don't mind you can merge as-is

Issue #1383, opened 2020-05-16
Honeywell-Security devices ignore reed switch.


thetic:
At least some models (eg Honeywell 5816, 2GIG DW10) support both wired contact switches and magnetic reed switches. On cursory inspection, reed switch state is encoded in bit 5 of the event field, but is ignored in decoding. State is [currently assigned](https://github.com/merbanan/rtl_433/blob/master/src/devices/honeywell.c#L98) as `(event & 0x80) >> 7`, but I think it should be `(event & 0x80) && (event & 0x20)`. I'm not sure whether this would be considered a breaking change or if there is a means to support both sensors discretely.

gdt:
Looks like this was fixed with #1384.  Please comment if that's wrong.

PR #1384, opened 2020-05-16
Improve Honeywell sensor support

thetic:
At least some models (eg Honeywell 5816, 2GIG DW10) support both wired
contact switches and magnetic reed switches to detect door/window state.
Contact switch state is encoded in bit 7 of the event field, and reed
switch state is encoded in bit 5.  Previously, only contact switch state
was reflected in the reported state.  This change determines the
reported state from both sensors; if either bit is zero, the state is
reported as "closed."

As reported in #1383.

EDIT: I split the states of the two sensors into discrete fields. This is a breaking change, but I think it is a better approach.

### DIFF
--- a/src/devices/honeywell.c
+++ b/src/devices/honeywell.c
@@ -56,7 +56,7 @@ static int honeywell_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int reed;
     int contact;
     int heartbeat;
-    int alarm;
+    int motion;
     int tamper;
     int battery_low;
 
@@ -94,13 +94,13 @@ static int honeywell_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         return DECODE_FAIL_MIC; // Not a valid packet
 
     event = b[3];
-    // decoded event bits: CATRBHUU
+    // decoded event bits: CTRMBHUU
     // NOTE: not sure if these apply to all device types
     // NOTE: contact switch state is in bit 7 and reed switch state is in bit 5
     contact     = (event & 0x80) >> 7;
     tamper      = (event & 0x40) >> 6;
     reed        = (event & 0x20) >> 5;
-    alarm       = (event & 0x10) >> 4;
+    motion      = (event & 0x10) >> 4;
     battery_low = (event & 0x08) >> 3;
     heartbeat   = (event & 0x04) >> 2;
 
@@ -112,7 +112,7 @@ static int honeywell_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             "event",        "", DATA_FORMAT, "%02x", DATA_INT, event,
             "contact_open", "", DATA_INT,    contact,
             "reed_open",    "", DATA_INT,    reed,
-            "alarm",        "", DATA_INT,    alarm,
+            "motion",       "", DATA_INT,    motion,
             "tamper",       "", DATA_INT,    tamper,
             "battery_ok",   "", DATA_INT,    !battery_low,
             "heartbeat",    "", DATA_INT,    heartbeat,
--- a/src/devices/honeywell.c
+++ b/src/devices/honeywell.c
@@ -56,7 +56,7 @@ static int honeywell_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int reed;
     int contact;
     int heartbeat;
-    int motion;
+    int alarm;
     int tamper;
     int battery_low;
 
@@ -94,13 +94,13 @@ static int honeywell_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         return DECODE_FAIL_MIC; // Not a valid packet
 
     event = b[3];
-    // decoded event bits: CTRMBHUU
+    // decoded event bits: CATRBHUU
     // NOTE: not sure if these apply to all device types
     // NOTE: contact switch state is in bit 7 and reed switch state is in bit 5
     contact     = (event & 0x80) >> 7;
     tamper      = (event & 0x40) >> 6;
     reed        = (event & 0x20) >> 5;
-    motion      = (event & 0x10) >> 4;
+    alarm       = (event & 0x10) >> 4;
     battery_low = (event & 0x08) >> 3;
     heartbeat   = (event & 0x04) >> 2;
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Maybe keep "state" as-is for a transitional period. Would "switch_open" =5|7 help? I guess integrators would know which option they used, so no?

We need a more unified approach for these bit-state outputs. Something that uniformly works for more and different devices. Some (unit) suffix that signifies a "alarm / trigger" and outputs a `1` if present. `_open` is fitting here but perhaps it could be more general?

This has popped up before, I'd like to solve and document it. But it needs perhaps a brain-storm of ideas, please.

thetic:
5|7 is likely to remain true forever unless *both* the contact switch and the reed switch are physically installed.

As far as supporting devices more broadly, my sensors include 2gig DW10 Thin Door/Window, Honeywell 5816 Door/Window, and 2gig PIR1 Motion Detector. No bits go high unless there is something "important": open contact switch, open reed switch, low battery, motion alarm, tampering, heartbeat. These flags seem universal. The motion sensor never reports an open switch and the door sensors never report an alarm. If we could get confirmation from other devices, I think this is a robust approach. I haven't tried interpreting he values of the alarm bits. I can collect that data if you'd like.

thetic:
I support adding the "state" field as previously implemented to maintain existing behavior. Maybe with a deprecation warning.

src/devices/honeywell.c
```diff
@@ -93,11 +94,13 @@ static int honeywell_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         return DECODE_FAIL_MIC; // Not a valid packet
 
     event = b[3];
-    // decoded event bits: AATABHUU
+    // decoded event bits: CATRBHUU
```
zuckschwerdt, line 97: "CTRMBHUU", right?

thetic: Good call.

zuckschwerdt:
What do you think about a "unit"-suffix that can be used elsewhere, something like
- `"contact_open_evt"` or `"contact_evt"`
- `"alarm_evt"`
- `"reed_open_evt"` or `"reed_evt"`
- `"motion_evt"`
- `"tamper_evt"`

Or maybe `"reed_E"` (`E` for Event). `_event` would be possible too, but I prefer a short key.

Processors can then use the common scheme of `"<name> _E"` to detect and handle all event-like keys.

thetic:
I don't know that "event" is the right term here. I think maybe "flag" might be more appropriate. "Event" implies that something has just occurred, but a sensor could send messages with a given bit set repeatedly.

zuckschwerdt:
Agreed. I intended to signify the importance of a `1` (something happened) over the `0` that could be a read as n/a or disregarded.

thetic:
I know it would be a big lift, but have you considered adding boolean types? C99 and JSON both support them.

zuckschwerdt:
Booleans have been discussed and found not to be useful enough. It's not clear that all consumers can deal with anything but numbers and maybe strings well, e.g. can all used databases like TSDBs aggregate them? We currently have a single place to look for type information: the suffix. Agreed, more structure could be nice but we aim more for a good-enough common mimimum ;)

zuckschwerdt:
Thanks for the change. To get this in without much disturbance I'd request another change though: keep all existing fields as-is (esp. `state`) and just add the new ones with `_open`. We'll discuss a common name scheme some other day then.

zuckschwerdt:
Looks good, thanks. "CTRMBHUU" is again a typo ;) Which leads me to a last question (promise!): you changed "alarm" from C+R+M to just M? Is M always motion? Or does it depend on the sensor? I'm thinking we could keep "alarm" like "state" as deprecated and make M "motion"? ("motion_open"?) Does that make sense? I really don't want to upset existing users without warning.

thetic:
I'm not sure whether it's a general alarm flag or it's for motion only. There are glass break sensors that use the same protocol that might use that bit. I don't have one to test with. The term "motion_open" doesn't make any sense.

zuckschwerdt:
Ok. Since this is breaking we'll leave it open for a week if someone wants to test and comment.
Feel free to remind me if I don't merge this within 10 days.

thetic:
reminder to merge

rhinomods:
I am not seeing any of my 5816s at all. I am using  merbanan:master and tried thetic:master to see if it was a git Issue when I have been playing. Sorry I have just restarted playing with rtl_433 today. (this might not even be the correct spot for me to comment)
I do see my5800pir fine and a neighbors sensor every once.

zuckschwerdt:
Try to grab samples from the sensors with `-S all` then check the samples with https://triq.org/iqs/ try to read them back with `-A`, convert to .ook with `-w FILE.ook` study in https://triq.org/pdv/ and if all else fails upload some confirmed sample as zip here for others to help out.

rhinomods:
Thank you zuckshwerdt! 
I appreciate the pointers. I am not going to pretend to under stand exactly what is going on with all the data HOWEVER you helped me notice that my bench testing of the devices was way too close to the antenna. 
Again thank you for taking the time and having patience with a beginner!

zuckschwerdt:
Anytime. Yes, contrary to first instinct: a stronger signal is only good up to a certain point, and also: hitting the signal dead center (on the set `-f` frequency) is not great.

rct:
@rhinomods - did you ever make any progress figuring out what is different about the 5816's signal?   I picked up a 5800mini, a 5800PIR and a 5816. Both the 5800mini and 5800PIR work well with rtl_433 regardless of how close they are to the RTL-SDR.  I thought there was something wrong with the 5816 until I tried moving it 10+ ft away.  Using stats, it looks like the 70 decoder is getting called but may be aborting due to length. 

```
{"time" : "2021-08-15 11:47:47.419692", "enabled" : 1, "since" : "2021-08-15T11:39:59", "frames" : {"count" : 252, "fsk" : 21, "events" : 61}, "stats" : [{"device" : 70, "name" : "Honeywell Door/Window Sensor, 2Gig DW10/DW11, RE208 repeater", "events" : 328, "ok" : 62, "messages" : 62, "abort_length" : 265, "fail_mic" : 1}]}
```

The signals I've captured so far don't seem to be useful, I'm getting errors about overruning the buffer. the `-A` analyzer doesn't seem to pick out any signal properties reliably.  Maybe the signal is just too strong and overloading the receiver.

```
Signal bigger than buffer, signal = 4980736 > buffer 3145728 !!
*** Saving signal to file g001_345M_250k.cu8 (2436206 samples, 3145728 bytes)
*** Saving signal to file g002_345M_250k.cu8 (311306 samples, 655360 bytes)
*** Saving signal to file g003_345M_250k.cu8 (1304176 samples, 2621440 bytes)
*** Saving signal to file g004_345M_250k.cu8 (403089 samples, 917504 bytes)
Signal bigger than buffer, signal = 6422528 > buffer 3145728 !!
*** Saving signal to file g005_345M_250k.cu8 (3186799 samples, 3145728 bytes)
*** Saving signal to file g006_345M_250k.cu8 (422564 samples, 917504 bytes)
*** Saving signal to file g007_345M_250k.cu8 (1479701 samples, 3014656 bytes)
*** Saving signal to file g008_345M_250k.cu8 (474540 samples, 1048576 bytes)
```

Issue #1386, opened 2020-05-18
Problems decoding ERT SCM messages (protocol 149)


phrrngtn:
The problem was discussed on a thread in the google group: https://groups.google.com/forum/#!msg/rtl_433/loZFvPhCxOE/BmvuBqDLAQAJ

There are two parts to this issue. One is adding support for other ERT protocols such as SCM+, IDM. This was requested in  #1378. The protocols are described in the rtlamr wiki https://github.com/bemasher/rtlamr/wiki/Protocol

The other part is that it seems that rtl_433 is missing a bunch of messages that are successfully decoded by rtlamr. See below for details of the attempts to reproduce.  The most promising looking diagnostics came from rtlamr when run with -samplefile. It looks like the program is saving the raw samples to a file but also indicating the sample offset, number of samples and payload of the decode. Given the fixed sample rate (2359296), the baud rate of the signal (32768), the fixed length of the packet 96 bits, it should be possible to find the samples rtlamr is framing.

```
rtl_tcp -a 192.168.68.126

# start the multiplexer that dups traffic on port 1234 to 7678
rtlmux -h 192.168.68.126 -p 1234

# Run the go-based s/w that successfully decodes a lot of ERT traffic
rtlamr -server=127.0.0.1:7878   -msgtype=scm,scm+,idm

# Run rtl_433 as directed but using sampling rate and frequency copied verbatim from rtlamr diagnostics
rtl_433 -d rtl_tcp://192.168.68.126:7878  -s 2359296  -f 912600155 -R 149 -S all

# 
rtl_433   -s 2359296  -r g004_912.6M_2359.3k.cu8 -W test.sr
```

I only saw one pulse there.

I ran rtlamr with -samplefile which saves the samples with length and offset information. I assume that it is saving the data in the same format it is getting it from rtl_tcp but I have not checked the source to verify that. 
`phrrngtn@sdrpi:~ $ /home/pi/go/bin/rtlamr -samplefile=foo.cu8 -server=127.0.0.1:1234   -msgtype=scm,scm+,idm`

16:54:11.635072 decode.go:45: CenterFreq: 912600155
16:54:11.635723 decode.go:46: SampleRate: 2359296
16:54:11.636156 decode.go:47: DataRate: 32768
16:54:11.636211 decode.go:48: ChipLength: 72
16:54:11.636256 decode.go:49: PreambleSymbols: 32
16:54:11.636302 decode.go:50: PreambleLength: 4608
[...]

PR #1387, opened 2020-05-19
Add attenuation histogram output

zuckschwerdt:
Adds optional verbose output to the pulse detector. Should help with e.g. #1386.
The _to_att can be made O(log N) instead of O(N) by nesting the thresholds tree-like. tbd.


zuckschwerdt:
Anybody feel free to improve this, it's just a rough idea to get more debug output. Merging as it won't impact normal output.

Issue #1318, opened 2020-02-15
Classify DSC Security Devices?


jcwren:
I have 6 DSC WS4939 key fobs and 2 DSC DW4917 door/window sensors. I notice that the ESNs for the  WS4939 start with `0x68` and the DW4917's start with `0x21`. Is there enough information out there where the decoder could classify the device type based on the ESN? 

Doing so would allow the 4 buttons on the key fobs to be identified as `button_1`, `button_2`, `button_3` and `button_4` instead of `tamper`, `xtamper1`, `xtamper2`, and `exception`. For the DW4917, which only has the single internal reed switch, it could be identified as `contact` or some-such.

If there's concern about breakage, maybe just the device name could be added to the JSON, or the new names added in addition to the existing ones.

zuckschwerdt:
Yes, Type and ESN are together displayed as `id`. The fist nibble (character) is the type, the remaining five the ESN.
Could you start to build a table what the types are? E.g.

Type | Device | kind
--- | --- | ---
6 |  WS4939 | key fob
2 | DW4917 | door/window sensor

Can other DSC Security owners perhaps join in and list type and device info?

zuckschwerdt:
I've added some preliminary type output. Please add more types and other findings here.

rct:
I see you haven't changed ESN which is good.  

I just wanted to be on record saying, please don't change the definition of the ESN and/or ID. The DSC system uses a 6 digit(*) ESN as the device ID. That's what's printed on the devices, that's how they get enrolled in the system. etc. So the type is an inherent part of the ESN/ID.

Here's what I have for types.  The type is more of a class, then a map to a specific device:

| Type | Description |
| ---- | ------------ |
| 0      | unknown |
| 1    | unknown |
| 2    | Door and window contacts, also Flood sensor, and possibly others |
| 3    | Motion sensor, Glass break detector, and possibly other burglary sensors |
| 4   | Smoke detector, and possibly other fire sensors |
| 5   | unknown |
| 6   | keyfob, possibly only 1-way keyfobs |
| 7   | unknown |
| 8   | Carbon Dioxide detector |
| 9   | unknown |

A few things to note:

[...]

PR #1389, opened 2020-05-21
Added support for WS4904P Pet immune PIR (0x03 subtype) ...

mostorer:
Hey all. I have just discovered this usefull tool to catch my DSC alarm panel (Alexor). As I have also PIR sensors, I would like to add them as supported devices. It's just a subtype addition (0x03). At the same time, I have found that my WS4945 door sensors have the same subtype of already recognized DW4917, so updated the display string.
As I have some more time, I would like also to add support for remote keypad, but it will be not an easy work for me, as I have seen that it is not recognized using DSC security protocol (23) as it is configured.

### DIFF
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -174,7 +174,8 @@ static int dsc_callback(r_device *decoder, bitbuffer_t *bitbuffer)
 
         switch (subtype) {
         case 0x6: subtype_str = "WS4939 key fob"; break;
-        case 0x2: subtype_str = "DW4917 door/window sensor"; break;
+        case 0x3: subtype_str = "WS4904P Pet immune pir detector"; break;
+        case 0x2: subtype_str = "DW4917/WS4945 door/window sensor"; break;
         default: subtype_str = "unknown"; break;
         }
 


merbanan:
Can you change the commit message?

anmd -> and

Looks good otherwise.

rct:
First, adding a description to the subtype to the list doesn't a sensor "supported"  4904P has worked fine since the DSC code was added.  It outputs an open event when it senses motion.

I haven't been paying attention in a while.  I guess the subtypes stuff got merged.  I thought we came to the conclusion in #1318 that it was basically a dead end?   #1318 is still open.

There are multiple devices that emit the same subtype, so encoding specific sensor model numbers is misleading at best. For example the WL912L glass break is also a subtype 3.  There are other motion sensor models that I haven't seen but probably are also subtype 3.  I've got 4 different sensors that are all subtype 2 - door/window contacts.

The one place that needs the subtype support is the keyfob since decoding it with the status bits that all the other contacts use isn't helpful.

rct:
@mostorer - If you are referring to the WT5500 "two-way" wireless keypads they use a very different encoding. I'm pretty sure a 1mhz sample rate is needed to receive it.  They are are also encrypted.  

The two-way keyfobs - WS4989 look like they use a pretty similar protocol.

mostorer:
> Can you change the commit message?
> 
> anmd -> and
> 
> Looks good otherwise.

Done! sorry for that.

mostorer:
Sorry but I'm new also to github. Hope generating not too much problems.

mostorer:
> > First, adding a description to the subtype to the list doesn't make a sensor "supported". 4904P has worked fine since the DSC code was added. It outputs an open event when it senses motion.

Yes, you are right, but as the transmitted info have already been decoded, why do not display a reference name intead of "unknown" ? IMHO it is more clear.

mostorer:
> @mostorer - If you are referring to the WT5500 "two-way" wireless keypads they use a very different encoding. I'm pretty sure a 1mhz sample rate is needed to receive it. They are are also encrypted.
> 
> The two-way keyfobs - WS4989 look like they use a pretty similar protocol.

@rct - Yes, I'm referring to WT5500 and WTK5504. I have both and I'm quite sure they use the same protocol. Could also be interesting to catch messages sent by the panel to keypads and to WT4989 to communicate its status. 
As I will find some spare time I will try. 

Is there some useful tutorial or article explaining how to work on decoding communications ?

zuckschwerdt:
@rct you are right the subtype is a dead end here. It's useful only for the keyfob. The ouput was added in 70f523e to experiment only. The issue then got stalled somehow. We need a clean up and the keyfob handling.
@mostorer see #1318 for a detailed table why the subtype has too little information / too much collision to work.

We can leave the subtype as numeric value, for easier processing than in then ESN. But interpreting the value can't be done by rtl_433 in a general and meaningful way.

zuckschwerdt:
> Is there some useful tutorial or article explaining how to work on decoding communications ?

1. Get a good grip on the coding (e.g. is it really PWM and no information in the gaps? this already came up with some protocols).
2. Grab as many codes as you can and table them together with an event.
3. Squint your eyes hard at the codes, maybe use [BitBench](https://triq.net/bitbench).
(3a. try to find a checksum early on, helps with byte alignment)

There are also some articles around, like https://github.com/merbanan/rtl_433/wiki/digitech_xc0324.README -- perhaps you could make a list while you explore this, would be really nice to have.

mostorer:
> @mostorer see #1318 for a detailed table why the subtype has too little information / too much collision to work.
> 
> We can leave the subtype as numeric value, for easier processing than in then ESN. But interpreting the value can't be done by rtl_433 in a general and meaningful way.

@zuckschwerdt - Had a look at the #1318 and fully agree with you ... too much devices within subtype to list all of them. In any case could be useful to have, in addition to the subtype numeric value by its own, also a general meaningful description as for the table at #1318

mostorer:
> > Is there some useful tutorial or article explaining how to work on decoding communications ?
> 
> 1. Get a good grip on the coding (e.g. is it really PWM and no information in the gaps? this already came up with some protocols).
> 2. Grab as many codes as you can and table them together with an event.
> 3. Squint your eyes hard at the codes, maybe use [BitBench](https://triq.net/bitbench).
>    (3a. try to find a checksum early on, helps with byte alignment)
> 
> There are also some articles around, like https://github.com/merbanan/rtl_433/wiki/digitech_xc0324.README -- perhaps you could make a list while you explore this, would be really nice to have.

I will try asap. Will also put cu8 recording in https://github.com/merbanan/rtl_433_tests if they could be useful.
Thank you so much.

rct:
@mostorer - Regarding the WT5500/WT5504, etc.   I did the DSC security contact decoding/implementation and a couple of other Acurite and LaCrosse devices. 

Even with that background, I wasn't able to make much progress decoding the WT5500.  Admittedly I didn't spend a ton of time on it and it was a number of years ago.  The more general Manchester demod stuff wasn't in rtl_433 yet.  All of the awesome analysis tools that @zuckschwerdt has added weren't available yet. 

Keep in mind that DSC's two-way devices like the WT5500 are encrypted.  So even when it is successfully demodulated into a bit stream, it won't be useful for reverse engineering until it can be decrypted.

A security researcher, Logan Lamb, mentioned he was able to break the encryption pretty easily. However he didn't publish details beyond the paper. See "home insecurity".  Note: the security companies tried to gag him. So even if the encryption is trivially broken, it probably wouldn't be good to make it part of rtl_433 for a couple of reasons.

Also maybe somewhat paranoid, but I haven't been willing to post a recording of the enrolled security keypad that contains the device's ESN on a public repository.  I would advise against you doing so. 

Yes, the chances of someone actually being able to exploit it against you specifically are likely very small, but doesn't seem like a good idea.

I do have recordings from an un-enrolled WS4989 that I'd be willing to post.  It is likely unencrypted because a key exchange hasn't happened yet.  It is probably useful for the modulation, but the message details will likely change once the device is enrolled.

merbanan:
@rct should this PR be closed?

PR #1390, opened 2020-05-22
Fix Home Assistant MQTT auto discovery

DagBertelsen:
* Added support for setting MQTT username and password
* Added config for Oregon wind and rain sensor.
  - Note, the device_class 'weather' is not supported in
    Home Assistant see: https://www.home-assistant.io/integrations/sensor/
* Oregon sensors use both channel and id so the instance
  variable will now support either one of them or both.
* Home Assistant will only allow edit of discovered
  sensors in GUI if they have a unique_id set.
* Identifiers in the 'device' seems to be use to group
  sensors in one group in home assistant
* Manufacturer in the 'device' is required to create a
  group of sensors in Home Assistant.

Showing sensors added to Home assistant
image

Sensor where battery, temp and hum sensor is added to one group
image

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -20,6 +20,8 @@ import paho.mqtt.client as mqtt
 MQTT_HOST = "127.0.0.1"
 MQTT_PORT = 1883
 MQTT_TOPIC = "rtl_433/+/events"
+MQTT_USERNAME = "username"
+MQTT_PASSWORD = "password"
 DISCOVERY_PREFIX = "homeassistant"
 DISCOVERY_INTERVAL = 600  # Seconds before refreshing the discovery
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Wonderful, thanks! @maxlock can you give this a quick test run?

maxlock:
I can spin the code up later this week, just had a quick look and think you need to allow for no username or password being set.

zuckschwerdt:
I am uncomfortable with the unconditional username/password too. At least Mosquitto has no problem to allow arbitrary u/p when no auth is required. Wouldn't bet on that though.
@DagBertelsen maybe having username or password of `None` could skip `username_pw_set`?

DagBertelsen:
Good point. I updated the commit

merbanan:
@zuckschwerdt can this be merged?

zuckschwerdt:
I was hoping for a review/test from @maxlock as he is an active user of this.

maxlock:
Sorry for the huge delay. I've just tested this pull, and it seems to be working OK for me.

PR #1397, opened 2020-05-27
ADD: Missing delay and low battery codes for DS10A door sensor.

azuca:
Tested with several DS10A on min, max and low battery

### DIFF
--- a/src/devices/x10_sec.c
+++ b/src/devices/x10_sec.c
@@ -13,6 +13,12 @@
  * on limited observations, doesn't take into account bits that might
  * be set to indicate something like a low battery condition.
  *
+ * DS10A door/window sensor bitmask : CUUUUDUB
+ *      C = Door/window closed flag.
+ *      U = Unknown. Cleared in all samples.
+ *      D = Delay setting. Min=1. Max=0.
+ *      B = Battery low flag.
+ *
  * Copyright (C) 2018 Anthony Kava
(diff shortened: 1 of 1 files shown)


merbanan:
Hi, the LSB sound like a battery indicator. Mask it out and add it as a separate parameter.

azuca:
Hi, The LSB is a low battery indicator for the DS10A devices but I don't own the other X10 devices handled in this module so I can't verify their use of this bit. Enhancing the switch statement for the DS10A cases that I am able to specifically test seemed the lowest risk approach, especially when dealing with security devices. Open to suggestions/recommendations.

zuckschwerdt:
There is no existing use of the LSB so this should be safe:
`battery_low = b[2] & 1;` and
`switch (b[2] & 0xfe) {`
Handle the battery_low like here: https://github.com/merbanan/rtl_433/blob/master/src/devices/eurochron.c#L75

azuca:
Looks good. Thanks for the pointer to the eurochron code.

azuca:
Some X10 security devices, like MS10A, do not support a battery indicator so I can't output a "Battery OK" message when the LSB is cleared because the battery state is obviously unknown. I could only include the battery indicator in the output data when the LSB is set. This would also maintain backwards compatibility in the output format for DS10A devices. Alternatively, I could explicitly tag events as originating from an X10 security device that I know for sure supports the battery indicator and only output the battery indicator for those events. Any preferences?

merbanan:
Looks good to me

PR #1404, opened 2020-06-08
WH51 minimum packet length

brianmichalk1:
Removed the minimum packet length for WH51 validation.  
Extra symbols existing at the end of the packet due to artifacts, would cause DECODE_ABORT_EARLY.  Removed the packet length statement, and packets are now decoded correctly when extra symbols are appended to the end of the packet.

I have tested this for about two hours, and will continue to do so.

### DIFF
--- a/src/devices/fineoffset.c
+++ b/src/devices/fineoffset.c
@@ -553,11 +553,6 @@ static int fineoffset_WH51_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     uint8_t b[14];
     unsigned bit_offset;
 
-    // Validate package
-    if (bitbuffer->bits_per_row[0] < 136) {  // Minimum length 14 bytes data + 3 bytes preamble
-        return DECODE_ABORT_LENGTH;
-    }
-
     // Find a data package and extract data payload
     bit_offset = bitbuffer_search(bitbuffer, 0, 0, preamble, sizeof(preamble) * 8) + sizeof(preamble) * 8;
     if (bit_offset + sizeof(b) * 8 > bitbuffer->bits_per_row[0]) {  // Did not find a big enough package


merbanan:
Well how about you increase the minimum length a bit?

brianmichalk1:
I'm not sure what's reasonable.  An extra four characters?

merbanan:
@brianmichalk1 each char is 4 bits so that makes 16 which sounds ok.

fgonza2:
Hi, i am trying to make the WH-51 work but can't read anything, can you guys provide input on how to make it work? thanks

merbanan:
Should be fixed with aa7436424394bc5eb0a5b85f827ac4e9d26ad69e.

fgonza2:
thanks, i tested itm this is what i get:

time      : 2020-07-03 12:35:46
model     : soil_sensor  count     : 1             num_rows  : 1             rows      : 
len       : 116          data      : 5100c9f3527f00f815ffffffd6be0
codes     : {116}5100c9f3527f00f815ffffffd6be0

fgonza2:
great!! i can see the MM value as 0x64 (100%) and 0 when not in water.    Do i need to process all this data outside RTL433 or is there a flag for decoding this within RTL433?

merbanan:
@fgonza2 should be fixed in 6bd4b146889a84f0746655a4281a92eb339e5a9f

fgonza2:
`time      : 2020-07-03 13:57:46
model     : Fineoffset-WH51                        ID        : 00c9f3
Battery level: 1.222     Battery   : 1800 mV       Moisture  : 23 %          Transmission boost: 0     AD raw    : 160           Integrity : CRC

`

Awesome, thanks a lot for your library and for your hard work on all this.  It is an excellent project

fgonza2:
hey, i am using now both 433Mhz and 915Mhz versions of this device.   The 915Mhz version doesnt work just by barely chaning the frequency to 915Mhz is there anything else that needs to be done? is there a reason this decoder wont work in the 915Mhz version?

merbanan:
Attach a signal recording and we can take a look at it. -S unknown should get a signal.

fgonza2:
nevermind i found the problem, looks like i need to keep the sample rate @ 250k vs 1M default at that frequency and now works.  Can you explain the dependency of sampling frequency vs. center frequency? thanks

zuckschwerdt:
Sample rate is samples taken per second (and the unit for events per time is Hz).
Since the samples are complex I/Q pairs the sample rate is also the effective frequency window seen (not Nyquist's f/2).
That window is centered around the center frequency given. So at the default sample rate of 250k and center frequency of 433.92M the window seen is 433.795M to 434.045M.

fgonza2:
got it so it is the bandwidth around the center frequency.    So why when you get a wider window (1M) you dont get anything with the same center frequency?

merbanan:
You include some kind of "noise". A stationary signal will confuse the demodulation code.

Issue #1378, opened 2020-05-14
Add support for SCM+


stevenjev:
I notice that support for SCM was added #977 . However, I was also wondering if support for SCM+ could be added. Based on information from the [rtlamr](https://github.com/bemasher/rtlamr) project, I believe that decoding SCM+ should be fairly similar to SCM as SCM+ is "Similar to SCM, allows greater precision and longer meter ID's." 

The rtlamr project's scm+ decoding is available here: https://github.com/bemasher/rtlamr/tree/master/scmplus

I apologize, as this entire field is very new to me. I am trying to provide outputs of the signals captured, but while I am able to see the scm+ signals in rtlamr, I am having difficulty at the moment capturing the scm+ signals in rtl_433.

merbanan:
We need samples, motivation and time to be able to add support for any type of protocol. Try playing around with the gain and frequency when you record the signals.

stevenjev:
When I run rtlamr, default flags, decoding for scm and scm+, multiple signals are found/decoded quite frequently: https://pastebin.com/0eKt19g9

If I run: `rtl_433 -f 912600155 -s 2359296 -S unknown` (so what I believe are the same frequency and sample rate), I get these sample files: [samples.zip](https://github.com/merbanan/rtl_433/files/4630880/samples.zip)

phrrngtn:
Saved sample file of scmplus messages. It looks like rtlamr is just saving the blocks that it has detected messages in. I had previously thought it was saving all the samples but on looking at the file sizes and offsets a bit more closely, it seems that this is not the case.

https://www.dropbox.com/s/36fqt7s8sh1jzqf/scmplus.cu8?dl=0


```
phrrngtn@sdrpi:~ $ /home/pi/go/bin/rtlamr -samplefile=scmplus.cu8 -server=127.0.0.1:1234   -msgtype=scm+
17:46:15.919703 decode.go:45: CenterFreq: 912600155
17:46:15.920352 decode.go:46: SampleRate: 2359296
17:46:15.920400 decode.go:47: DataRate: 32768
17:46:15.920435 decode.go:48: ChipLength: 72
17:46:15.920515 decode.go:49: PreambleSymbols: 16
17:46:15.920553 decode.go:50: PreambleLength: 2304
17:46:15.920586 decode.go:51: PacketSymbols: 128
17:46:15.920619 decode.go:52: PacketLength: 18432
17:46:15.920658 decode.go:59: Protocols: scm+
[...]

PR #1410, opened 2020-06-20
Scmplus

evilpete:
SCM+ decoder

written in response to #1378

tested only on samples as I do not have a local source to test with (my meter uses something different )


evilpete:
@merbanan   :  I named this  scmplus, but it occurred to me that the (regular) scm decoder you already have is named [ert](/merbanan/rtl_433/blob/master/src/devices/ert.c),  is there preferred name for SCM+ for continuity ?

evilpete:
Note: 
Field key names and json format set to mostly match [rtlamr](https://github.com/bemasher/rtlamr) field names.

A [pull request](/merbanan/rtl_433_tests/pull/347) has been made to rtl_433_tests with samples

src/devices/scmplus.c
```diff
@@ -0,0 +1,198 @@
...
+
+    Copyright (C) 2020 Peter Shipley <peter.shipley@gmail.com>
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include <arpa/inet.h>
```
merbanan, line 12: Can we get rid of this ?

src/devices/scmplus.c
```diff
@@ -0,0 +1,198 @@
...
+
+    /* clang-format off */
+    data = data_make(
+            "model",           "",                 DATA_STRING, "SCM+",
+            "ProtocolID",     "Protocol_ID",       DATA_STRING, protocol_id_str,
+            "EndpointType",   "Endpoint_Type",     DATA_STRING, endpoint_type_str,
+            "EndpointID",     "Endpoint_ID",       DATA_INT,    endpoint_id,
+            "Consumption",     "",                 DATA_INT,    consumption_data,
+            "Tamper",          "",                 DATA_STRING, physical_tamper_str,
+            "PacketCRC",       "crc",              DATA_STRING, crc_str,
```
merbanan, line 161: This value is not really needed but can be kept for rtlamr compatibility reasons.

src/devices/scmplus.c
```diff
@@ -0,0 +1,198 @@
...
+    /* 
+        Field key names and format set to  match rtlamr field names 
+
+        {Time:2020-06-20T09:58:19.074 Offset:49152 Length:49152
+        SCM+:{ProtocolID:0x1E EndpointType:0xAB EndpointID:  68211547 Consumption:  6883 Tamper:0x4900 PacketCRC:0x39BE}}
+    */
+
+    /* clang-format off */
+    data = data_make(
+            "model",           "",                 DATA_STRING, "SCM+",
```
merbanan, line 155: The + character might not be such a good mqtt topic. I'm not sure but if so the a change from + to plus would be good.

src/devices/scmplus.c
```diff
@@ -0,0 +1,198 @@
...
+/**
+
+Freq 912600155
+
+Random information:
+
+https://github.com/bemasher/rtlamr/wiki/Protocol
+http://www.gridinsight.com/community/documentation/itron-ert-technology/
+
+ Units: Some meter types transmit consumption in 1 kWh units, while others use more granular 10 Wh units
```
merbanan, line 24: Is there no way to know this ?

src/devices/scmplus.c
```diff
@@ -0,0 +1,198 @@
...
+    if (sync_index >= bitbuffer->bits_per_row[0]) {
+        return DECODE_ABORT_EARLY;
+    }
+
+    if ( (bitbuffer->bits_per_row[0] - sync_index) < 128) {
+        return DECODE_ABORT_LENGTH;
+    }
+
+    if (decoder->verbose) {
+        (void)fprintf(stderr, "%s: row len=%hu\n", __func__, bitbuffer->bits_per_row[0]);
```
merbanan, line 50: void should not be needed

src/devices/scmplus.c
evilpete: not that i can find.  rtlamr has the same comment

src/devices/scmplus.c
stevenjev: Testing this pull request, when using [/model], I received "SCM_" not "SCM+"

merbanan: SCMplus is prefered then.

src/devices/scmplus.c
evilpete: your right I should have used `htobe16` from  `endian.h `?   
although I find `ntohs()` to be more portable.

I would rather not automatically make assumptions on local byte order with  hard a coded
`((t_16 & 0xFF00) >> 8 | (t_16 & 0x00FF) << 8)`

with intermixed with :
`#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__`

src/devices/scmplus.c
```diff
@@ -0,0 +1,198 @@
...
+        (void)fprintf(stderr, "%s: sync_index=%d\n", __func__, sync_index);
+    }
+
+    // bitbuffer_debug(bitbuffer);
+    bitbuffer_extract_bytes(bitbuffer, 0, sync_index, b, 16 * 8);
+
+    if (decoder->verbose) { // print bytes with aligned offset
```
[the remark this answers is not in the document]
evilpete: done

src/devices/scmplus.c
evilpete: yes it was included for theoretical  compatibility.

compatibility is also the reason I convert the fields to hex strings manually as opposed to using DATA_FORMAT

src/devices/scmplus.c
evilpete: done

src/devices/scmplus.c
evilpete: Sorry, old-school habit

src/devices/scmplus.c
zuckschwerdt: I'd say we are used to pick bit and bytes from random places in the decoders, for just two instances of short with not real performance requirement just pick t_16 apart and just bitshift-or.

evilpete: ok...
bitshift-or  manually 
Done

PR #1413, opened 2020-06-21
x10sec: Added more sensors, Tamper Alarm, CRC check and tuned parameters

FredrikFornstad:
Added support for more devices such as KR18 (key-fob), DS18 (Dual Door/Window magnetic & wired sensor), MS18 (Motion Sensor).

Added Tamper Alarm and CRC check and added some verbose logging.

Adjusted timing of short and long bit to what I believe is according to the protocol (The timing of the devices are not very accurate and depending on device and temperature the short bit vary from 520 to 580 from what I have seen).

Refactored the code to make execution faster: aborts earlier and only look at row 1 for the message (I have not seen it come in anywhere else during fairly extensive testing)

I am new to rtl_433, so please check my code in case I have misunderstood something.

I have however noted that the PPM decoding is not very good in noisy environments, but this PR does not affect that. I will do some more testing on that and maybe create a separate issue with some cu8-files attached to it.

### DIFF
--- a/src/devices/x10_sec.c
+++ b/src/devices/x10_sec.c
@@ -2,22 +2,43 @@
  * X10 Security sensor decoder.
  *
  * Each packet starts with a sync pulse of 9000 us and 4500 us gap.
- * The message is OOK PPM encoded with 567 us pulse and long gap (0 bit)
- * of 1680 us or short gap (1 bit) of 590 us. There are 41 bits, the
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Great stuff, looks really good! Minor nitpicking: Can you
- put "mic" in a final `data_append` so it's last.
- replace `µs` with `us` -- ugly but ASCII ;)

FredrikFornstad:
Fixed so that mic is last and replaced text to 'us'

PR #1428, opened 2020-07-11
Add support for Missil ML0757

bilde2910:
This device is a weather station that records wind speed, rainfall and temperature and transmits it back to a receiver device that display this data, along with the receiver's temperature and humidity. This is my first time delving into the `rtl_433` project.

## Row format

<id: 8 bits> <flags: 8 bits> <field 1: 12 bits> <field 2: 8 bits> <tail: 4 bits>

## Flags

1. Battery low bit
2. Unknown
3. Unknown
4. Unknown
5. Unknown
6. Packet type
7. Unknown
8. Unknown

## Data

When bit 6 is **set**, field 1 is accumulated rainfall in steps of 0.45 mm, and field 2 is wind speed as (field value + 2) * 1.4 km/h with special cases for 0-3 km/h.

When bit 6 is **not set**, field 1 is temperature (signed) in 1/10ths °C, and field 2 is **unknown** at this time.

## Temperature packet field 2

I don't know what the 8-bit field in the temperature packet is.

Here are some samples:

```
0110 0111 0000 0000 0000 1110 0000 0011 0110 1111
0110 0111 0000 0000 0000 1110 0011 0011 0110 1111
0010 1100 0000 0000 0001 0000 0011 0011 0101 1111
0110 0111 0000 0000 0000 1110 0100 0011 0110 1111
0110 0111 0000 0000 0000 1110 0100 0011 0110 1111
0110 0111 0000 0000 0000 1110 0101 0011 0110 1111
0110 0111 0000 0000 0000 1110 0101 0011 0110 1111
0110 0111 0000 0000 0000 1110 0101 0011 0110 1111
1000 1011 1000 0000 0000 1111 0011 0011 0110 1111
0110 0111 0000 0000 0000 1111 0000 0011 0110 1111
0110 0111 0000 0000 0001 0010 0110 0011 0101 1111
0110 0111 0000 0000 0001 0010 1100 0011 0101 1111
0110 0111 0000 0000 0001 0010 1111 0011 0101 1111
0110 0111 0000 0000 0000 0011 0100 0011 1000 1111
0110 0111 0000 0000 1111 1100 1100 0011 1000 1111 
0010 1100 0000 0000 0001 0000 0011 0011 0101 1111
0010 1100 0000 0000 0001 0000 1100 0011 0101 1111
0010 1100 0000 0000 0001 0000 1111 0011 0101 1111
```

In all of these, the field is either 0x35, 0x36 or 0x38. What I have ruled out:

- It is **not** humidity. I put it in the bathroom while taking a long shower, with all seals and protective covers removed, and there was no difference.
- It is **not** battery voltage. The unit takes 2xAA batteries and thus receives 3.0V from the batteries.
- It is **not** battery percentage. I put in bad batteries in the unit, which didn't change the field, but raised flag bit 1 instead.
- It does not appear to be a checksum. It remains the same even when the input data varies wildly, and is only picked from a select few possible values. Also, it is only sent with the temperature packet, and not the rain packet, as wind speed uses that field in the rain packet.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -106,6 +106,7 @@ rtl_433_SOURCES      = abuf.c \
                        devices/maverick_et73.c \
                        devices/maverick_et73x.c \
                        devices/mebus.c \
+					   devices/missil_ml0757.c \
                        devices/new_template.c \
                        devices/newkaku.c \
                        devices/nexa.c \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -106,7 +106,7 @@ rtl_433_SOURCES      = abuf.c \
                        devices/maverick_et73.c \
                        devices/maverick_et73x.c \
                        devices/mebus.c \
-					   devices/missil_ml0757.c \
+                       devices/missil_ml0757.c \
                        devices/new_template.c \
                        devices/newkaku.c \
                        devices/nexa.c \
--- a/src/devices/missil_ml0757.c
+++ b/src/devices/missil_ml0757.c
@@ -1,0 +1,151 @@
+/** @file
+    Missil ML0757 weather station with temperature, wind and rain sensor.
+
+    Copyright (C) 2020 Marius Lindvall
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+/**
+The unit sends two different alternating packets, one for temperature and one
+for rainfall and wind. All packets are 40 bits and are transferred 9 times.
+Packet structure appears to be as follows:
+
+                         BIT
+          0   1   2   3   4   5   6   7   8
+      0x0 +---+---+---+---+---+---+---+---+
+          | Device ID                     |
+      0x1 +---+---+---+---+---+---+---+---+
+    B     |BAT| ? | ? | ? | ? |RWP| ? | ? | <-- FLAGS BYTE
+    Y 0x2 +---+---+---+---+---+---+---+---+
+    T     | Data field 1                 >|
+    E 0x3 +---+---+---+---+---+---+---+---+
+          |<Data field 1  | Data field 2 >|
+      0x4 +---+---+---+---+---+---+---+---+
+          |<Data field 2  | 1 | 1 | 1 | 1 |
+      0x5 +---+---+---+---+---+---+---+---+
+
+When flag bit RWP is not set, data field 1 is (temp in °C * 10) as a signed
+12-bit integer, and data field 2 (8 bits) is unknown.
+
+Where when bit RWP is set, data field 1 is accumulated rainfall in number of
+steps as a signed 12-bit integer, where each step is 0.45 mm of rain, but when
+the sign bit flips, the counter appears to reset to 0. Data field 2 is wind
+speed as an 8 bit integer where 0x00 = 0 km/h, 0x80 = 1.4 km/h, 0xC0 = 2.8 km/h,
+and any other value is ((value + 2) * 1.4) km/h. I know this doesn't intuitively
+make sense, but it's what my testing has come up with and it matches what is
+shown on the display.
+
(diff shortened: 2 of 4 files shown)


src/devices/missil_ml0757.c
```diff
@@ -105,7 +105,8 @@ static int missil_ml0757_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     if (flag_rwp) { // Rainwall and wind
         /* clang-format off */
         data = data_make(
-                "model",            "",             DATA_STRING, _X("ML0757-RainWind","Missil 0757 Rain/Wind"),
+                "model",            "",             DATA_STRING, "Missil-ML0757",
+                "subtype",          "Sensor(s)",    DATA_STRING, "RainfallWind",
```
zuckschwerdt, line 109: `subtype` is for e.g. security devices like PIR vs door contact where a single unit is of a specific function. It's ok to not have a subtype here, some messages will have the rain value, some the temp, but overall you get both. In other words subtype should be static for each device.

src/Makefile.am
```diff
@@ -106,6 +106,7 @@ rtl_433_SOURCES      = abuf.c \
                        devices/maverick_et73.c \
                        devices/maverick_et73x.c \
                        devices/mebus.c \
+					   devices/missil_ml0757.c \
```
zuckschwerdt, line 109: Please use 4 spaces instead of tabs.

src/devices/missil_ml0757.c
```diff
@@ -0,0 +1,154 @@
...
+
+    Copyright (C) 2020 Marius Lindvall
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+/**
+The unit sends two different alternating packets, one for temperature and one
```
zuckschwerdt, line 12: The device doc should start with a short sentence. Just copy line 2.

zuckschwerdt:
Thanks, looks great.

bilde2910:
Thanks for the review. I've fixed the indentation issue, copied the short sentence to the doc and removed the subtype in the latest commit. :-)

merbanan:
Visual studio build file changes are missing.

bilde2910:
> Visual studio build file changes are missing.

Which files are those? I added those mentioned in the wiki, but I'm on Linux and don't use Visual Studio, so I'm not sure what files VS adds/modifies. I'd be happy to add them.

zuckschwerdt:
`./maintainer_update.py` should take care of all build files. But you need a clean working directory (`git stash` if needed) and please carefully review the suggested changes and pick only what applies to your PR.

bilde2910:
The script does not appear to work for me. It prints `src_files`, `device_files`, `include_files`, and `r_devices`, then:

- From a fully clean slate (`git reset --hard HEAD`) it fails with `FileNotFoundError: [Errno 2] No such file or directory: './build/src/rtl_433'`.
- After compiling `rtl_433` so the above file exists, it results in a different error:

```
  File "/usr/lib/python3.8/sre_parse.py", line 1039, in parse_template
    this = chr(ESCAPES[this][1])
KeyError: '\\x'

During handling of the above exception, another exception occurred:

  File "./maintainer_update.py", line 116, in <module>
    replace_block(r'```',
  File "./maintainer_update.py", line 40, in replace_block
    filedata = regex.sub(repl, filedata)
  File "/usr/lib/python3.8/re.py", line 327, in _subx
    template = _compile_repl(template, pattern)
  File "/usr/lib/python3.8/re.py", line 318, in _compile_repl
    return sre_parse.parse_template(repl, pattern)
  File "/usr/lib/python3.8/sre_parse.py", line 1042, in parse_template
    raise s.error('bad escape %s' % this, len(this))
re.error: bad escape \x at position 8878
```

Using Python 3.8.3

merbanan:
Code merged in b5672f3717b14daee24aa0e6aabfdddb92221a47.

PR #1433, opened 2020-07-14
Add Sharp SPC775 support

ddrown:


zuckschwerdt:
Thanks, looks good. Can you
- Adapt a header like https://github.com/merbanan/rtl_433/blob/master/src/devices/kerui.c
- The crc is actually a LFSR, the factors are the keys (you can see the shift left in the key stream)
- perhaps our `lfsr_digest8()` works here? You can check for LFSR params with https://github.com/triq-org/revdgst/ or post 10-20 codes here and I'll take a peek.
- the MIC here is "CRC"

ddrown:
how do these changes look?

zuckschwerdt:
Oh, so `lfsr_digest8_reflect()` worked right away? Usually the key stream does not reset after 8 bits/keys, but we have other sensors that do this "truncated LFSR". Your solution to fold all bytes prior to LFSR checking is great!
I still need to look at the row-find-loop to see if we can simplify that or make a utility function for it.

ddrown:
Yup, lfsr_digest8_reflect matched the crc sequence I saw.  This device only uses the first 8 bits from the LFSR.

For the row-find code, I used something similar to the flex.c device.

Issue #1430, opened 2020-07-13
Add support for La Crosse TX141-B


bsharper:
I have a similar transmitter to what pictured in https://github.com/merbanan/rtl_433/issues/1268, but without humidity. The model number is TX141-B. 

I ran the command that @tvass ran in issue #1268 and just changed the n= value. I have no clue what these values mean, but I did get somewhat similar results.

```
rtl_433 -X 'n=lacrosse-tx141-b,m=OOK_PWM,s=256,l=500,r=1888,g=0,t=0,y=748'
[...]
 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2020-07-12 20:44:28
model     : lacrosse-tx141-b                       count     : 15            num_rows  : 15            rows      :
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 32           data      : 2a6d2a6b,
len       : 34           data      : 2a6d2a6bc
codes     : {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {32}2a6d2a6b, {34}2a6d2a6bc

[...]

time      : 2020-07-12 20:44:59
model     : lacrosse-tx141-b                       count     : 16            num_rows  : 16            rows      :
[...]

PR #1434, opened 2020-07-16
Changes to allow LaCrosse TX141B to work properly

bsharper:
This remote temperature sensor is similar to other TX141 sensors, however it uses a 32 bit message length which is shorter than the others. 

The main differences with this sensor are:

1.  No humidity reading.
2. No channel select.

I've verified battery flag and test flag are working.

Pictures and additional information can be found under https://github.com/merbanan/rtl_433/issues/1430

### DIFF
--- a/src/devices/lacrosse_tx141x.c
+++ b/src/devices/lacrosse_tx141x.c
@@ -75,7 +75,7 @@ A count enables us to determine the quality of radio transmission.
 The TX141-BV2 is the temperature only version of the TX141TH-BV2 sensor.
 
 Changes:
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks, looks good.
- you don't need the `_X` macro with new code, just "LaCrosse-TX141B" is fine
- wouldn't min length 32 be better to match the TX141B (32: TX141B, 33-36: TX141BV3. 37+: TX141), currently 31 bits would go to TX141BV3 as well)

bsharper:
@zuckschwerdt Thanks for the feedback, I've tested and committed the changes to the branch in the PR.

Issue #1422, opened 2020-07-02
Support for Auriol AFT 77 A1 temperature sensor


aguedob:
Hi all,

I've just bought this outdoor temp sensor and after taking several samples I managed to identify the temp using rtf_433:

The second and the third byte (f4 0**1 34** 02 27 c0 in this sample) determines the current temperature (0x134 -> 308 -> 30.8 degrees)

p_limit: 143
bitbuffer:: Number of rows: 12 
[00] {42} f4 0**1 34** 02 27 c0 : 11110100 0000000**1 00110100** 00000010 00100111 11
[01] { 1} 00                : 0
[02] {42} f4 01 34 02 27 c0 : 11110100 00000001 00110100 00000010 00100111 11
[03] { 1} 00                : 0
[04] {42} f4 01 34 02 27 c0 : 11110100 00000001 00110100 00000010 00100111 11
[05] { 1} 00                : 0
[06] {42} f4 01 34 02 27 c0 : 11110100 00000001 00110100 00000010 00100111 11
[07] { 1} 00                : 0
[08] {42} f4 01 34 02 27 c0 : 11110100 00000001 00110100 00000010 00100111 11
[09] { 1} 00                : 0
[10] {42} f4 01 34 02 27 c0 : 11110100 00000001 00110100 00000010 00100111 11
[11] { 0}                   : 

Would be great if someone can help supporting it.

Thanks!

zuckschwerdt:
Can you grab some codes with slightly varying tempertures so we can see if there is a checksum in the last bytes? (just the `{42} f4 01 34 02 27 c0` lines will do. Also try to grab a good sample with `-S all` (read it back in with `-r g00...` to check) and attach here as zip.

merbanan:
Hi follow this instruction and we can add support:

https://github.com/merbanan/rtl_433/blob/master/docs/CONTRIBUTING.md#supporting-additional-devices-and-test-data

aguedob:
I've attached a zip file with some samples and a txt file with the matching temps in C degrees. 

[AFT77A1.zip](https://github.com/merbanan/rtl_433/files/4872288/AFT77A1.zip)

Let me know if you need more samples. Thanks!

evilpete:
@merbanan.  
I was taking a look at this and noted there is  reoccurring  demodulation error in all the samples 

Running the command : 
```
rtl_433 \
    -f 433.92M -s 250k \
    -R 0 -a 4 \
    -W output.sr \
[...]

PR #1438, opened 2020-07-20
Auriol AFT 77 A1 temperature sensor

evilpete:
Protocol support for  Auriol AFT 77 A1 temperature sensor

written in reference to #1422

should work between 0°C -> 51.3°C 

 checksum algorithm for temperature below 0°C may fail, will update pull request as info becomes available


zuckschwerdt:
Thanks, this is great to see your findings in practice. Minor nitpicking:
- both comments should start `/**`, the first `/** @file`, the both first lines should be the same and end with a dot.
- the opening brace `{`  for functions needs to be on the next line
- `output_fields` should end with `NULL,` (note the comma), indent (also r_device) should be 8.
(perhaps try clang-format, we mostly follow those suggestions)

evilpete:
Done..

evilpete:
nitpicking edit Done...

zuckschwerdt:
- file doc-comment should have four spaces indent but no star
- first doc senstence needs a full stop
- first sentence should also be in the decoder doc comment
- enumeration in doc comment (`I: Device ID`...) needs a dash in each line
- for scaling `* 0.1f` is prefered to `/ 10.0` (implied double and div is more expensive than mul)
- `"mic",` in data_make has one leading space too much ;)

Otherwise ready for merge!

zuckschwerdt:
Oh, one last thing: the `_X` macro is only for legancy model names. The name here should be just "Auriol-AFT77A1". Or if the "A1" part is the board revision (like in similar devices) then "Auriol-AFT77" (best to match the filename too).

evilpete:
Note that is a work in progress and should not be merged ( yet )

merbanan:
@evilpete what is the status of this? I think it looks good enough to merge.

evilpete:
I does not work, more data samples need, a  preferably a full range Not being power cycled.

I cracked the crc but only for the current ID token which changes every time it is power cycled

merbanan:
And the files actually decode with the Vauno EN8822C decoder since 2 years...

Issue #1424, opened 2020-07-06
Support for Ambient Weather TX-3107 Pool Temperature Sensor


snicker:
I'm a bit inexperienced when it comes to reverse engineering RF, so I apologize if this request is malformed.

I recently acquired this:
https://www.ambientweather.com/amtx3107.html

I know it transmits at 915mhz, but I have no idea where to begin with decoding the signals. I have saved several cu8 files, but using the `-A` flag to replay them attempts to guess the modulation with mixed results (no clue, PCM). I've put the device right next to my RTLSDR radio antenna and filtered signals using the `-Y level=-1.0` option, and the pulses come at regular intervals, so I'm fairly certain that everything is coming from the TX-3107.

However, I don't know what to do next to attempt to decode the information into a useful set of data. Is there anything more I can do to help other than just uploading the cu8 files?

zuckschwerdt:
You can drop the cu8 files on https://triq.org/iqs/ to verify a good signal.
Then demodulate the I/Q data to OOK: `rtl_433 -w FILENAME.ook FILENAME.cu8` and drop onto https://triq.org/pdv/ to inspect the pulses.

snicker:
Alright, I'm quite out of my element here. I was able to verify consistent looking signal but I have no idea how to use the I/Q data once converted to OOK. Is there any kind of ... how to decode spectrograms for dummies resource anywhere?

snicker:
![image](https://user-images.githubusercontent.com/221540/87884888-3b128b00-c9c6-11ea-89c0-462d59a14526.png)
I've managed to get this far in terms of making this easy to look at, but I don't know what to do from here to determine exact pulse length, or what that even means (I'm assuming this converts to binary and then somehow I'd need to turn that into data)

snicker:
found this resource too https://www.osengr.org/WxShield/Downloads/Weather-Sensor-RF-Protocols.pdf

merbanan:
Can you share the ook file?

snicker:
[amtx3107_cu8+ook.zip](https://github.com/merbanan/rtl_433/files/4944683/amtx3107_cu8%2Book.zip)
Here are 3 samples, cu8 and ook

merbanan:
Well pulse coded modulation with a length of 480us looks suitable.

merbanan:
[...]

PR #1439, opened 2020-07-20
Update all header doc comments

evilpete:


zuckschwerdt:
Great start :) Files should look like e.g. https://github.com/merbanan/rtl_433/blob/master/src/devices/kerui.c
- `@file` doc-comment with just one line description (short) and author, copyright. Indented, no leading asteriks.
- doc-comment with one line description (can be longer), no indent, valid markdown.

With the reason that we need copyright information per file and decoder docs per decoder.
Extracting the decoder markdown doc with doxygen isn't great, thus we limit ourselves to a simple text with no indent :)

evilpete:
ok I'll see what I can do about more "cleanup"

As you may have guessed I've been bored,  I'll keep throwing pull requests at you till I get distracted by something else thats new and shiny

merbanan:
@evilpete care to look at https://github.com/merbanan/rtl_433/issues/1424. It is very shiny.

evilpete:
@zuckschwerdt is this better?

zuckschwerdt:
This is really getting us somewhere with the machine-readable doc comments, thanks!

evilpete:
Meh.. No problem, it was less then a hour of editing...

any other low hanging requests??

merbanan:
DOUBLE -> FLOAT wrt datatype name

PR #1440, opened 2020-07-23
Acurite Atlas decoding minor cleanup

rct:
* factored atlas decoding out to separate function for readability
* New model key for `message_type` vs `subtype`
* New model key for `id` vs `sensor_id`
* Changed raw message dump to stderr from an unprotected print to a data field

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -412,21 +412,9 @@ static int acurite_6045_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsign
     return 1;
 }
 
+
+
 /**
-This callback handles several Acurite devices that use a very
-similar RF encoding and data format:
-
-- 592TXR temperature and humidity sensor
-- 5-n-1 weather station
-- 6045M Lightning Detector with Temperature and Humidity
-- Atlas
-
-    CC RR IIII | IIII IIII | pBMMMMMM | pxxWWWWW | pWWWTTTT | pTTTTTTT | pSSSSSSS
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Can you indent the data_* rows by only 8 spaces, as we usually have?
Also you can possibly further simplify the output by using the new `DATA_COND` as shown in e.g. a9b8a46
I.e. pull the `if`s up and e.g. add a `int has_uv`, then use a single data_make with
```
...
        "uv",               NULL,           DATA_COND, has_uv,    DATA_INT, uv,
        "lux",              NULL,           DATA_COND, has_uv,    DATA_INT, lux * 10, 
...
```

rct:
@zuckschwerdt - I fixed the indentation of the data_make and data_appends.

I don't see a good opportunity to use the new `DATA_COND` feature.  I thought the authors who worked on this first (you, @jrhelbert, ...) did a good job of DRYing things up given that there are 6 different message types--3 different sets of data depending on type and the versions with/without the lightning detector installed.) 

It looks like it is a trade off between DRY vs more easily readable data_make with more copy & paste code to maintain.

zuckschwerdt:
I just wanted to point out that there is now an option to simplify conditional data_make rows. It's not required :)

rct:
@zuckschwerdt - Understood. I think `DATA_COND` is a good idea. Looking at it now, I think I should use make the `exception` field use `DATA_COND` so it only becomes part of the message if non-zero.

Issue #1441, opened 2020-07-25
False Positives with Oregon Scientific SL109H


evilpete:
I've noticed I was receiving  a lot of of "hits" for the Oregon Scientific SL109H device across multiple frequencies, after taking a closer look it seems  that any packet with  a length of  38 bits and comprised of all zeros is reported as a valid packet
( the "checksum" is calculated by [simple addition of 5 nibbles  ](https://github.com/merbanan/rtl_433/blob/50c774d1423e6eba833acbc66ab2fc2ab145ad2a/src/devices/oregon_scientific_sl109h.c#L48)).

Suggested fix: 
add a simple sanity check to  reject packet  with something like the following:

```
IF  Id == temp_raw == humidity == channel == 0
   Reject_packet() 
```
(What are the odds of having Id humidity and channel be all be zero?)
If this is an acceptable work around I'll generate a pull request...

This can be easily reproduced with :

`rtl_433  -R 54 -y "{38}0000000000"` or `rtl_433  -R 54 -y "{38}ff000000000"`. (not all zeros)

```
time      : 2020-07-24 23:52:37
Model     : Oregon-SL109H Id        : 0
Channel   : 3            Celsius   : 0.00 C        Humidity  : 0 %           Status    : 0             Integrity : CHECKSUM
```

Actual Received Packet:
```
time      : 2020-07-24 23:38:18
Model     : Oregon-SL109H Id        : 0
Channel   : 3            Celsius   : 0.00 C        Humidity  : 0 %           Status    : 0             Integrity : CHECKSUM
Modulation: ASK          Freq      : 910.0 MHz
RSSI      : -12.1 dB     SNR       : 7.5 dB        Noise     : -19.7 dB
pulse_demod_ppm(): Oregon Scientific SL109H Remote Thermal Hygro Sensor
bitbuffer:: Number of rows: 2
[00] {38} 00 00 00 00 00             : 00000000 00000000 00000000 00000000 000000
[01] {68} 00 00 00 00 00 00 00 00 00
```

evilpete:
Here is a list of decoders found,  I'll generate a pull request for solve these:

Protocol  | Name
--------- | -------
21 |	Calibeur-RF104
47 |	Conrad-S3318P
49 |	Quhwa-Doorbell
54 |	Oregon-SL109H
68 |	Kerui-Security
84 |	Thermopro-TP11
108 |	Hyundai-WS
115 |	Honeywell-ActivLink
[...]

PR #1443, opened 2020-07-25
Fix Globaltronics QUIGG GT-TMBBQ-05 false positives

evilpete:
Addresses same issue described in #1441 regarding false positives with sized all zero packets but in driver [gt_tmbbq05.c](rtl_433/blob/master/src/devices/gt_tmbbq05.c). 

can be reproduced with :

`rtl_433 -R 137 -y '0000000000000 {33}0000000000000 {33}0000000000000 {33}0000000000000 {33}0000000000000 {33}0000000000000 {33}0000000000000 {33}0000000000000' -vvv`


evilpete:
There are probably more,  if i find them. I'll try to submit any other in a single pull req in the future

Issue #1394, opened 2020-05-25
MQTT Birth and Last Will and Testament (LWT) messages


kpine:
Add support for sending Birth and LWT messages. This would make it trivial to know when a rtl_433 is running or has gone offline.

zuckschwerdt:
Good idea. Suggestions on the topic and content?

peteakalad:
For reference tasmota publishes these messages on a reboot;

tele/tasmota_D851A7/LWT Offline
tele/tasmota_D851A7/LWT Online

kpine:
I'm not sure if it's better to have a single topic with different payloads, or separate topics.

But, I would probably prefer the single topic with different payloads, something like `<topic prefix>/status` with payloads `offline` and `online`. That's an approach I've seen elsewhere [[1](https://esphome.io/components/mqtt.html#last-will-and-birth-messages)] [[2](https://www.home-assistant.io/docs/mqtt/birth_will/)]. You could include other information besides the process status in a JSON payload, if desired.

I'm not using any sort of hostname or instance identifier in my topics, since I only have a single service running (maybe I should be). My MQTT config is:

```
output          mqtt://hostname:1883,retain=0,events=rtl_433/events,states=rtl_433/states,devices=rtl_433/[protocol][/channel:0][/id:0]
```
So I'd want to see the status at `rtl_433/status` or so. Not sure if that's possible to derive automatically, or if it would need to be a separate config option (e.g. `,status=rtl_433/status`). Even better, but a different subject, would be able to allow a custom topic prefix.

Also, speaking of Tasmota, I came upon a Github [issue](https://github.com/arendst/Tasmota/issues/7189) related to this. Probably would want offline status being reported even in the case of a DISCONNECT.

zuckschwerdt:
We could introduce `prefix=` to clean things up. Not sure if many people use all of events, states and devices. The events and devices carry the same information in different formats.

Perhaps `prefix/LWT` is better than `prefix/status` as there is already `prefix/states` and that is confusing.

We have three different status messages available:
- connect (birth) message: "Online"
- disconnect (exit) message: "Offline"
- connection lost (LWT) message: "Offline"

We would aim to always disconnect cleanly, thus the abnormal connection lost case could get another, more serious message.

What would parse the message and what are the expectations there?

kpine:
> The events and devices carry the same information in different formats.

Thanks for the reminder, I've removed it, it was a holdover from some previous testing. 

[...]

PR #1451, opened 2020-07-28
Improve: program exit with a special exit code in case of error

psa-jforestier:
Based on discussion here : https://github.com/merbanan/rtl_433/issues/1394#issuecomment-660213824
I added an exit code to indicate why the program exited. I use it to reset the USB port from time to time, when the libusb report a "stalled" status.


zuckschwerdt:
Very nicely done!

PR #1454, opened 2020-07-31
Fix unnecessary dereference

wintersteiger:
Seems odd to me that this one hasn't been noticed before. I think the intended behavior is the one without the `*` though.

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -640,7 +640,7 @@ static void print_kv_data(data_output_t *output, data_t *data, char const *forma
         }
 
         // print key
-        char *key = *data->pretty_key ? data->pretty_key : data->key;
+        char *key = data->pretty_key ? data->pretty_key : data->key;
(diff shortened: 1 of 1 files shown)


wintersteiger:
Nope, I was wrong. It seems that by convention this string should never be NULL? (It was in my tests, so couldn't test for first character of `pretty_key`.)

zuckschwerdt:
It looks strange, but there is a reason: data_make will default to the plain key if pretty_key is NULL, but an empty string is acceptable. https://github.com/merbanan/rtl_433/blob/master/src/data.c#L293 That's not what we want in the output though, we actually need to display something :)

Allowing the pretty key to be NULL could perhaps be nicer than strdup(""). I have some experimental code to use only static strings with key, pretty_key, and format. But that's a hard sell.

Anyway, that data_ stuff is somewhat tricky C, various checkers barf when looking at that code -- it's well tested and mature though.

PR #1455, opened 2020-08-04
Add Nexus-TE82s compatibility note

ondras:
Hi,

I got a Solight TE82S sensor and discovered that it uses the Nexus protocol. So I mentioned this device being compatible with the nexus decoder in the docs.


zuckschwerdt:
Appreciated!

Issue #1452, opened 2020-07-31
Support for CAME remote control (garage door in AM/ASK)


psa-jforestier:
Hello All.

I failed to create a decoder for my CAME remote control. The protocol seems easy, but I cant make a flex decoder work.

Pulse analysis :
```
rtl_433 -A -r Button1_433.83M_250k.cu8
(... skip frames)
Detected OOK package    @0.436348s
Analyzing pulses...
Total count:   13,  width: 12.40 ms             ( 3100 S)
Pulse width distribution:
 [ 0] count:    7,  width:  332 us [316;356]    (  83 S)
 [ 1] count:    6,  width:  660 us [652;664]    ( 165 S)
Gap width distribution:
 [ 0] count:    6,  width:  344 us [332;352]    (  86 S)
 [ 1] count:    6,  width:  672 us [668;692]    ( 168 S)
Pulse period distribution:
 [ 0] count:    3,  width:  684 us [680;688]    ( 171 S)
 [ 1] count:    3,  width: 1336 us [1332;1340]  ( 334 S)
 [ 2] count:    6,  width: 1000 us [988;1012]   ( 250 S)
Level estimates [high, low]:  15949,     31
RSSI: -0.1 dB SNR: 27.1 dB Noise: -27.2 dB
Frequency offsets [F1, F2]:    9701,      0     (+37.0 kHz, +0.0 kHz)
Guessing modulation: Manchester coding
Attempting demodulation... short_width: 332, long_width: 0, reset_limit: 696, sync_width: 0
Use a flex decoder with -X 'n=name,m=OOK_MC_ZEROBIT,s=332,l=0,r=696'
pulse_demod_manchester_zerobit(): Analyzer Device
bitbuffer:: Number of rows: 1
[00] {16} 27 0c     : 00100111 00001100

```

Seems simple : 2 kinds of pulse (332 and 660) and 2 kinds of gap (344 and 672). Cause the pulse and gap modulation are close, and a length of 2 is 2 times the length of 1, the coding seems simple. But how to create a Flex decoder with this ?
I upload a sample, and wrote all my guess  :  https://github.com/psa-jforestier/rtl_433_tests/blob/master/tests/Came/TOP432/README.md

zuckschwerdt:
That's not MC, just plain PWM:
![Button1_433 83M_250k](https://user-images.githubusercontent.com/12833/89031528-f2968e00-d332-11ea-8f29-cd432510b94a.png)

[...]

PR #1460, opened 2020-08-07
Added support for Came Top432 remote control

psa-jforestier:
This decoder is not enabled by default, because the protocol do not have any preamble/crc/checksum/parity, so it must be enabled by using "-R 164" .

This decoder resolve https://github.com/merbanan/rtl_433/issues/1452

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [16]  AlectoV1 Weather Sensor (Alecto WS3500 WS4500 Ventus W155/W044 Oregon)
     [17]  Cardin S466-TX2
     [18]  Fine Offset Electronics, WH2, WH5, Telldus Temperature/Humidity/Rain Sensor
-    [19]  Nexus, FreeTec NC-7345, NX-3980 temperature/humidity sensor
+    [19]  Nexus, FreeTec NC-7345, NX-3980, Solight TE82S temperature/humidity sensor
     [20]  Ambient Weather Temperature Sensor
     [21]  Calibeur RF-104 Sensor
     [22]* X10 RF
@@ -235,6 +235,8 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [159]  Insteon
     [160]  Interval Data Message (IDM)
     [161]  Interval Data Message (IDM) for Net Meters
+    [162]* ThermoPro-TX2 temperature sensor
+    [163]  Acurite 590TX Temperature with optional Humidity
 
 * Disabled by default, use -R n or -G
 
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [37]* Inovalley kw9015b, TFA Dostmann 30.3161 (Rain and temperature sensor)
     [38]  Generic temperature sensor 1
     [39]  WG-PB12V1 Temperature Sensor
-    [40]  Acurite 592TXR Temp/Humidity, 5n1 Weather Station, 6045 Lightning
+    [40]  Acurite 592TXR Temp/Humidity, 5n1 Weather Station, 6045 Lightning, 3N1, Atlas
     [41]  Acurite 986 Refrigerator / Freezer Thermometer
     [42]  HIDEKI TS04 Temperature, Humidity, Wind and Rain Sensor
     [43]  Watchman Sonic / Apollo Ultrasonic / Beckett Rocket oil tank monitor
@@ -237,6 +237,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [161]  Interval Data Message (IDM) for Net Meters
     [162]* ThermoPro-TX2 temperature sensor
     [163]  Acurite 590TX Temperature with optional Humidity
+    [164]* Came TOP432 remote control
 
 * Disabled by default, use -R n or -G
 
@@ -246,7 +247,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
   [-d <RTL-SDR USB device index>] (default: 0)
   [-d :<RTL-SDR USB device serial (can be set with rtl_eeprom -s)>]
 	To set gain for RTL-SDR use -g <gain> to set an overall gain in dB.
-	SoapySDR device driver is available.
+	SoapySDR device driver is not available.
   [-d ""] Open default SoapySDR device
   [-d driver=rtlsdr] Open e.g. specific SoapySDR device
 	To set gain for SoapySDR use -g ELEM=val,ELEM=val,... e.g. -g LNA=20,TIA=8,PGA=2 (for LimeSDR).
--- a/src/devices/came_top432.c
+++ b/src/devices/came_top432.c
@@ -79,12 +79,11 @@ static int came_top432_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         // no parity
 
     }
-	//bitbuffer_extract_bytes(bitbuffer, 0, 1, out, 12);
-	// Ignoring first bit of the bitbuffer, used as the start bit
-	// 
+    // Ignoring first bit of the bitbuffer, used as the start bit
+    //
 
     // packet is not repeated
-	
+
    // reconstruct the code with the 2 bytes of the bitbuffer, and keep the 12 bits (ignoring the start bit)
     code = (((b[0] << 8) | b[1]) >> 3) & 0x0FFF;
     sprintf(code_str, "%03x", code);
--- a/src/devices/came_top432.c
+++ b/src/devices/came_top432.c
@@ -10,24 +10,12 @@
 
  */
 
-/*
-    Use this as a starting point for a new decoder.
-
-    Keep the Doxygen (slash-star-star) comment above to document the file and copyright.
-
-    Keep the Doxygen (slash-star-star) comment below to describe the decoder.
-    See http://www.doxygen.nl/manual/markdown.html for the formating options.
-
-    Remove all other multiline (slash-star) comments.
-    Use single-line (slash-slash) comments to annontate important lines if needed.
-*/
-
 /**
 The device uses PPM encoding,
 - 0 is encoded as 320 us gap and 640 us pulse,
 - 1 is encoded as 640 us gap and 320 us pulse.
 The device sends a 4 times the packet when a button on the remote control is pressed.
-A transmission starts with a 320 us pulse. At the end of the packet, there is 36 periods of 320us between messages (11520us)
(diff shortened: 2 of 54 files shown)


README.md
```diff
@@ -246,7 +247,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
   [-d <RTL-SDR USB device index>] (default: 0)
   [-d :<RTL-SDR USB device serial (can be set with rtl_eeprom -s)>]
 	To set gain for RTL-SDR use -g <gain> to set an overall gain in dB.
-	SoapySDR device driver is available.
+	SoapySDR device driver is not available.
```
Mindavi, line 250: @zuckschwerdt it's unfortunate that the `maintainer_update.py` script creates these diffs. Should we do something about that? It think filtering this information from the man page / README is fine, as it depends on the build anyways.

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
+/** @file
+    Decode CAME remote control TOP-432EV, TOP-432NA, TOP-432EE.
+    This remote control is used for garage door and sliding gate. It transmits on 433.92 MHz (as it is written on the case), built since 2006.
```
Mindavi, line 3: How do you know these exist since 2006? Maybe good to add a source here (if possible).

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+
+    It works with CAME radio receiver cards "AF43S", capable of handling 4096 codes. CAME is an italian company. Theses remote controls are mainly sold in europ (France, Italy, Belgium). https://www.came.com and https://www.came-europe.com .
```
Mindavi, line 5: nit: `sold in Europe`

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+
+    This decoder is based on new_template.c
```
Mindavi, line 7: Please remove all template comments, only leave the text that is specifically applicable to this device. e.g. 'this decoder is based on that template' is not too useful.

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+
+    Copyright (C) 2016 Benjamin Larsson
```
Mindavi, line 9: Copyright (C) 2020 <your name>

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+ */
+
+/*
```
Mindavi, line 17: remove this block

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+
+    Keep the Doxygen (slash-star-star) comment below to describe the decoder.
+    See http://www.doxygen.nl/manual/markdown.html for the formating options.
+
+    Remove all other multiline (slash-star) comments.
+    Use single-line (slash-slash) comments to annontate important lines if needed.
+*/
+
+/**
+(this is a markdown formatted section to describe the decoder)
```
Mindavi, line 30: Remove these 2 lines

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+         * early as possible before attempting to parse data.
+         *
+         * Check "message envelope"
+         * - valid message length (use a minimum length to account
+         *   for stray bits appended or prepended by the demod)
+         * - valid preamble/device type/fixed bits if any
+         * - Data integrity checks (CRC/Checksum/Parity)
+         */
+
+        if (bitbuffer->bits_per_row[r] < CAMETOP432_BITLEN) {
```
Mindavi, line 87: Maybe you could use https://github.com/merbanan/rtl_433/blob/9c117521f0570c86eda651c1237c3d796bfbb62d/include/bitbuffer.h#L102-L104

with for example a minimum of 2 rows (or more, but keep in mind that signal quality may vary). This would reduce false positives a lot, I think. At least, the comment above mentions that the packet is sent 4 times.

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+    Remove all other multiline (slash-star) comments.
+    Use single-line (slash-slash) comments to annontate important lines if needed.
+*/
+
+/**
+(this is a markdown formatted section to describe the decoder)
+(describe the modulation, timing, and transmission, e.g.)
+The device uses PPM encoding,
+- 0 is encoded as 320 us gap and 640 us pulse,
+- 1 is encoded as 640 us gap and 320 us pulsep.
```
Mindavi, line 34: nit: pulsep -> pulse

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+        "brand",
+        "model",
+        "type",
+        "code",
+        "code_hex",
+        NULL,
+};
+
+/*
+ * r_device - registers device/callback. see rtl_433_devices.h
```
Mindavi, line 137: I'd just remove this whole part, only necessary when filling everything in.

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+#include "decoder.h"
+
+#define CAMETOP432_BITLEN      12
+
+static int came_top432_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    data_t *data;
+    int r; // a row index
+    uint8_t *b; // bits of a row
+    int parity;
```
Mindavi, line 62: Remove unused variables from here.

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+    // reconstruct the 12 bits code
+    code = ((b[0] << 4) | (b[1]>>4)) & 0xFFF;
+    sprintf(code_str, "%03x", code);
+    /* clang-format off */
+    data = data_make(
+            "brand",    "", DATA_STRING,   "CAME",
+            "model",    "", DATA_STRING,   "TOP432",
+            "type",     "", DATA_STRING,   "remote control",
+            "code",     "", DATA_INT,      code,
+            "code_hex", "", DATA_STRING,   code_str,
```
Mindavi, line 111: I'd only keep one of these, either keep the `code` or keep the `code_hex`. My preference goes to `code` (prevents string formatting issues), but whatever floats your boat. I don't think it's useful to keep both, as they mean the same thing.

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+    b = bitbuffer->bb[0];
+
+    // reconstruct the 12 bits code
+    code = ((b[0] << 4) | (b[1]>>4)) & 0xFFF;
+    sprintf(code_str, "%03x", code);
+    /* clang-format off */
+    data = data_make(
+            "brand",    "", DATA_STRING,   "CAME",
+            "model",    "", DATA_STRING,   "TOP432",
+            "type",     "", DATA_STRING,   "remote control",
```
Mindavi, line 109: I would remove this, but I'm not fully sure. @zuckschwerdt we don't typically not the device type here, right? I think this field is meant for the sub-type of the model, because that's applicable in some cases.

README.md
zuckschwerdt: it's just a lazy default because we really needed some up-to-date text. Any changes or improved workflow welcome. (would be great to also have the rtl_433.example.conf protocols in sync)

src/devices/came_top432.c
```diff
@@ -0,0 +1,169 @@
...
+ * and to src/CMakeLists.txt and src/Makefile.am or run ./maintainer_update.py
+ *
+ */
+r_device came_top432 = {
+        .name        = "Came TOP432 remote control",
+        .modulation  = OOK_PULSE_PPM,
+        .short_width = 320,  // 
+        .long_width  = 640,  // 
+        .sync_width  = 320,
+        //.gap_limit   = 0,  // dont know how to find this value
```
Mindavi, line 164: The amount of time in microseconds before we should consider a packet a new 'row', so should be something like 0.8 * `gap_size`, in which the gap size is the amount of 'dead time' between 2 packets.

src/devices/came_top432.c
zuckschwerdt: model should be "CAME-TOP432, no brand, no type.

src/devices/came_top432.c
zuckschwerdt: sync must never match short or long.

src/devices/came_top432.c
psa-jforestier: I tried this method at first, but it did not work. Will do an other try.

src/devices/came_top432.c
psa-jforestier: ok, I think I confused reset_limit and gap_limit

psa-jforestier: I'm really confused about values for `sync_width, gap_limit` and `reset_limit`. Here is a graphic representation of the signal.
came_analysis
- ok, `short` and `long` width are clear
- `sync` is the sync start bit, but it should be the length of the short pulse (when measuring with Audacity there is variation of +- 5% on each pulse length...)
- `gap_limit` : it depends of what a "packet" is. Is it all the bits of the message, or a single bit ? The only working value for this decoder is "0".
- `reset_limit` : should be time between two messages (they are repeated 4 times). I measure 15 000µs.

src/devices/came_top432.c
psa-jforestier: Dilema here ... hex code directly match with the dip switch inside the receiver and is easy to understant when doing bit-by-bit correlation.  I would prefer to let the two codes (hex and int).

zuckschwerdt:
You should really use the conf from #1452
Also it's better to use PWM. Using PPM is a hack here. See https://github.com/merbanan/rtl_433/blob/master/include/pulse_demod.h#L56

src/devices/came_top432.c
Mindavi: Yeah, since I tried to do the flex decoder and that failed, I don't think it's possible to get multiple rows.

zuckschwerdt: The packet gap is limited by "PD_MAX_GAP_RATIO times the longest pulse", i.e. gap_limit bigger than long_limit*10 won't work.

src/devices/came_top432.c
psa-jforestier: Flex decoder can be written with : 
`rtl_433 -r Button1_433.83M_250k.cu8 -X "n=CameTOP432,m=OOK_PWM,s=320,l=640,r=10000,g=830,preamble={1}8"`

psa-jforestier:
Closing PR to use a Flex decoder on a dedicated conf file.

Issue #1465, opened 2020-08-12
Abarth 124 Spider false trigger


evilpete:
getting false trigger for Abarth 124 Spider (also on wrong freq)

looking at code it seems the number of bit manchester decoded is not validated.

[g004_915M_2000k.cu8.gz](https://github.com/merbanan/rtl_433/files/5060598/g004_915M_2000k.cu8.gz)

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2020-08-10 21:53:32
model     : Abarth 124 Spider                      type      : TPMS          id        : 00000000
flags     : 00           Pressure  : 0 PSI         Temperature: -58 F        status    : 0             mic       : CHECKSUM
Modulation: FSK          Freq1     : 915.5 MHz     Freq2     : 914.9 MHz
RSSI      : -6.1 dB      SNR       : 24.0 dB       Noise     : -30.1 dB
```

evilpete:
fixed in  #1466 merge

PR #1466, opened 2020-08-12
Fix TPMS abarth124 False Positive

evilpete:
Fixes Bug described in #1465

amount of bits returned by `bitbuffer_manchester_decode()` was not being validated resulting in zeroed buffer being processed.

### DIFF
--- a/src/devices/tpms_abarth124.c
+++ b/src/devices/tpms_abarth124.c
@@ -49,8 +49,27 @@ static int tpms_abarth124_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsi
     int status;
     int checksum;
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Looks good. We shouldn't put too much manual debugging in various decoders if it can be avoided though.

evilpete:
I've been running a 915/433/314 Hop loop for a few days and am just addressing the false positives if they come up more then a few  times day.

Mindavi:
I'd rather not have the commented out lines either. If someone wants to debug it they'll put in their own debug logic, I'd assume. In my opinion adding all those commented out lines pollutes the code without really helping anyone.

PR #1469, opened 2020-08-12
FIX check manchester_decode result length

### LINKED ISSUE #1468: Bug in CurrentCost
in devices/current_cost.c the extracted data is checked to be 56 bit or  8 bytes in length

https://github.com/merbanan/rtl_433/blob/57c924d2315de6a0688ad65614cf838396ca946f/src/devices/current_cost.c#L55

but the 8th byte is used referenced 

https://github.com/merbanan/rtl_433/blob/57c924d2315de6a0688ad65614cf838396ca946f/src/devices/current_cost.c#L73

evilpete:
several decoders were either ignoring number of bits decoded via `bitbuffer_manchester_decode()` or where checking for the wrong number of bits

This fixes #1468 plus other bugs


zuckschwerdt:
Looks good.

src/devices/maverick_et73x.c
```diff
@@ -57,6 +59,11 @@ static int maverick_et73x_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     // decode the inner manchester encoding
     bitbuffer_manchester_decode(bitbuffer, 0, 0, &mc, 104);
 
+    // we require 7 bytes 13 nibble rounded up (  b[6] highest referance below )
+    if (mc.bits_per_row[0] < 56) {
```
Mindavi, line 63: I'm not sure why, but all maverick tests now fail (rtl_433_tests). So I'd guess this is incorrect?

src/devices/tpms_citroen.c
```diff
@@ -12,6 +12,8 @@
 Citroen FSK 10 byte Manchester encoded checksummed TPMS data
 also Peugeot and likely Fiat, Mitsubishi, VDO-types.
 
+Fcc-Id:
```
Mindavi, line 15: Is this meant as a hint for someone to go hunting for it? It doesn't really help if it's empty...

evilpete:
I've been trying to add then when ever I get the chance, I'm guessing I
started add the info but couldn't find a concise answer.

evilpete:
@Mindavi Done

Mindavi:
Thanks!

PR #1474, opened 2020-08-14
Feat came top432 flex

psa-jforestier:
Should solve https://github.com/merbanan/rtl_433/issues/1452


merbanan:
The config file is sort of code so if you want you can add your copyright.

merbanan:
I meant your name. I obviously didn't write this config file.

PR #1477, opened 2020-08-20
Fix bad conf for Fan-11t

psa-jforestier:
prevent from starting when using this conf, with an error message "Bad flex spec, unknown keyword (Fan-11t)!"


zuckschwerdt:
Oh, thanks for catching this!

Issue #1464, opened 2020-08-10
Garage Door - SecPlus & SecPlus 2.0+


gcormier:
~~I've found some old references (eg #652 ) which seem to indicate that rtl_433 can decode ternary to binary. Is this still the case? As I no longer see anything using `-X help`.~~

~~I'm looking at garage door opener decoding, specifically the Security+2.0 protocol - attached.~~

[sec2plus.pdf](https://github.com/merbanan/rtl_433/files/5052769/sec2plus.pdf)

EDIT: Update to general request for SecPlus/SecPlus 2.0+ support.

@argilo has done all the hard work and the code can be found here : https://github.com/argilo/secplus/blob/master/secplus.py I can provide samples and testing.

gcormier:
[oneclick.zip](https://github.com/merbanan/rtl_433/files/5052788/oneclick.zip)

zuckschwerdt:
On a glance it looks similar to Manchester or the x1521 encoder, i.e. we are only dealing with two symbols, that then can be decoded by grouping. Actual ternary with three symbols would only be possible if we could grab it as PCM.

The OOK looks like a very good fit for PCM plus custom decoding:
![oneclick](https://user-images.githubusercontent.com/12833/89819881-606d6180-db4c-11ea-969b-65f7e4f7805b.png)

gcormier:
So the original (old) Security+ used Ternary

https://github.com/argilo/secplus
`Transmissions use on-off keying, with an alphabet of three symbols (0, 1, 2) corresponding to three different pulse widths:`

- 0: 1.5ms off, 0.5ms on
- 1: 1ms off, 1ms on
- 2: 0.5ms off, 1.5ms on

Perhaps the new protocol does not. If you need any other captures let me know.

zuckschwerdt:
In PCM:
- 0: `0001`
- 1: `0011`
- 2: `0111`
That also comes down to a "double frequency" or 2-bit/4-bit encoding (Manchester is 1-bit/2-bit), with the nice property of a transition in the middle and at the end which can be read as PCM robustly.

The "new" protocol here doesn't show `111`or `000` sequences, the doc indicate codes of `00`, `01`, and `10`. So we should spot a `000` somewhere, which we don't, strange.

gcormier:
Right it mentions "11" being useful as an invalid sequence, so I do think it might use ternary. But perhaps the data stream is just binary?

This section seems relevant (you = "skilled in the art", not I :) )
[...]

PR #1480, opened 2020-08-24
Adding support for Secplus v2

evilpete:
Still needs cleanup,  but am generating a pull request for tracking and review... 

See also #1464

### DIFF
--- a/man/man1/rtl_433.1
+++ b/man/man1/rtl_433.1
@@ -612,7 +612,7 @@ Interval Data Message (IDM) for Net Meters
 Acurite 590TX Temperature with optional Humidity
 .TP
 [ \fB164\fI\fP ]
-Secplus v2
+Security+ 2.0 (Keyfob) 
 
 * Disabled by default, use \-R n or \-G
 .SS "Input device selection"
--- a/man/man1/rtl_433.1
+++ b/man/man1/rtl_433.1
@@ -246,7 +246,7 @@ Generic temperature sensor 1
 WG\-PB12V1 Temperature Sensor
 .TP
 [ \fB40\fI\fP ]
-Acurite 592TXR Temp/Humidity, 5n1 Weather Station, 6045 Lightning
+Acurite 592TXR Temp/Humidity, 5n1 Weather Station, 6045 Lightning, 3N1, Atlas
 .TP
 [ \fB41\fI\fP ]
 Acurite 986 Refrigerator / Freezer Thermometer
@@ -610,6 +610,9 @@ Interval Data Message (IDM) for Net Meters
 .TP
 [ \fB163\fI\fP ]
 Acurite 590TX Temperature with optional Humidity
+.TP
+[ \fB164\fI\fP ]
+Secplus v2
 
 * Disabled by default, use \-R n or \-G
 .SS "Input device selection"
--- a/src/devices/secplus_v2.c
+++ b/src/devices/secplus_v2.c
@@ -494,9 +494,6 @@ static int secplus_v2_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     fixed_total ^= ((uint64_t)bb[1]) << 4;
     fixed_total ^= (bb[2] >> 4) & 0x0f;
 
-    // ??
-    fixed_total ^= 0xff < 12;
-
     // fprintf(stderr, "fixed_total = %lu\n", fixed_total);
 
     // int button    = fixed_total >> 32;
@@ -542,13 +539,11 @@ static char *output_fields[] = {
 r_device secplus_v2 = {
         .name        = "Secplus v2",
         .modulation  = OOK_PULSE_PCM_RZ,
-        .short_width = 230,
-        .long_width  = 230,
+        .short_width = 250,
+        .long_width  = 250,
         .tolerance   = 50,
         .gap_limit   = 1500,
         .reset_limit = 9000,
-        // .gap_limit   = 2500,
-        // .reset_limit = 4000,
         .decode_fn = &secplus_v2_callback,
         .disabled  = 0,
         .fields    = output_fields,
--- a/src/devices/secplus_v2.c
+++ b/src/devices/secplus_v2.c
@@ -20,52 +20,34 @@ Security+ 2.0  is described in [US patent application US20110317835A1](https://p
 
 #include "decoder.h"
 
-// delete this soon
-void debug_10bit_bufs(uint16_t a, uint16_t b, uint16_t c)
-{
-    fprintf(stderr, "\na : ");
-    for (int j = 9; j >= 0; j--) {
-        fprintf(stderr, "%d ", (a >> j) & 0x01);
-    }
-    fprintf(stderr, "\t%d\n", a);
-    fprintf(stderr, "b : ");
-    for (int j = 9; j >= 0; j--) {
-        fprintf(stderr, "%d ", (b >> j) & 0x01);
-    }
-    fprintf(stderr, "\t%d\n", b);
-    fprintf(stderr, "c : ");
-    for (int j = 9; j >= 0; j--) {
-        fprintf(stderr, "%d ", (c >> j) & 0x01);
-    }
-    // fprintf(stderr, "\n%05X\n", c & 0x3f);
-    fprintf(stderr, "\t%d\n", c);
-    fprintf(stderr, "\n");
-}
-
 /**
 
 
 data comes in two bursts/packets
 
 Layout:
-bits = AA BB IIII OOOO X*30
+
+bits = `AA BB IIII OOOO X*30`
 
 AA = payload type  ( 2 bits 00 or 01 )
 BB = FrameID ( 2 bits always 00)
 IIII = inversion indicator ( 4 bits )
-OOOO = Order indicator ( 4 bits )
+OOOO = Order indicator ( 4 bits ).
 XXXX....  = data ( 30 bits )
 
---
+---
+
+
 
 data is broken up into 3 parts ( p0 p1 p2 )
 eg:
 
-data = ABCABCABCABCABCABCABCABCABCABC
+data = `ABCABCABCABCABCABCABCABCABCABC`
 becomes:
-    p0 = AAAAAAAAAA
-    p1 = BBBBBBBBBB
-    p2 = CCCCCCCCCC
+
+    `p0 = AAAAAAAAAA`
+    `p1 = BBBBBBBBBB`
+    `p2 = CCCCCCCCCC`
 
 these three parts are then inverted and reordered based on the 4bit Order and Inversion indicators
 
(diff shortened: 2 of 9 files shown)


src/devices/secplus_v2.c
```diff
@@ -0,0 +1,548 @@
...
+    Based on code by Clayton Smith https://github.com/argilo/secplus
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+/**
+Freq 310M
```
argilo, line 14: The transmitter I have actually cycles between 310, 315 and 390 MHz.

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,548 @@
...
+    // compute rolling_total from rolling_digits[]
+    uint32_t rolling_total = 0;
+    uint32_t rolling_temp  = 0;
+
+    for (int i = 0; i < 18; i++) {
+        rolling_temp = (rolling_temp * 3) + rolling_digits[i];
+        // fprintf(stderr, ">> %12d\t%d\n", rolling_temp, rolling_digits[i]);
+    }
+
+    rolling_total = reverse32(rolling_temp);
```
argilo, line 454:
```suggestion
    rolling_total = reverse32(rolling_temp) >> 4;
```

argilo: This is needed because the rolling code is a 28-bit integer, so it must be shifted right by four bits after going through a 32-bit reversal.

evilpete: that *Was* in there,  I guess It got deleted at some point, thanks...

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,555 @@
+/** @file
+    Security+ 2.0 rolling code
+
+    Copyright (C) 2020 Peter Shipley <peter.shipley@gmail.com>
+    Based on code by Clayton Smith https://github.com/argilo/secplus
```
Mindavi, line 5: License-wise, this is not fully compatible, as the code this is based on is GPL-3 instead of GPL-2 :(.

I didn't look into how 'copied' this looks, but keep that in mind. If it's all your own and loosely based upon the original, it shouldn't matter much (but IANAL).

argilo:
It looks to me like @evilpete's code is only very loosely based on mine. I'd be happy to release mine under GPLv2 if that makes things easier.

Mindavi:
I think the licensing thing is fine due to @argilo commenting that it's not a copy. No need to re-release under GPL-2 as far as I understand.

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,555 @@
...
+    fixed_total ^= ((uint64_t)bb[1]) << 24;
+    fixed_total ^= ((uint64_t)bb[2]) << 16;
+
+    bb = fixed_2.bb[0];
+    fixed_total ^= ((uint64_t)bb[0]) << 12;
+    fixed_total ^= ((uint64_t)bb[1]) << 4;
+    fixed_total ^= (bb[2] >> 4) & 0x0f;
+
+    // ??
+    fixed_total ^= 0xff < 12;
```
argilo, line 498: This can be removed, since it does nothing.

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,555 @@
...
+};
+
+//      Freq 310.01M
+//  -X "n=vI3,m=OOK_PCM,s=230,l=230,t=40,r=10000,g=7400,match={24}0xaaaa9560"
+
+r_device secplus_v2 = {
+        .name        = "Secplus v2",
+        .modulation  = OOK_PULSE_PCM_RZ,
+        .short_width = 230,
+        .long_width  = 230,
```
argilo, line 546: I presume these should be 250, since the nominal pulse width is 250us.

argilo:
I did some testing by generating signals with secplus, and so far this has correctly decoded everything I've thrown at it. It also correctly decodes the remote I have on hand. :+1:

man/man1/rtl_433.1
```diff
@@ -610,6 +610,9 @@ Interval Data Message (IDM) for Net Meters
 .TP
 [ \fB163\fI\fP ]
 Acurite 590TX Temperature with optional Humidity
+.TP
+[ \fB164\fI\fP ]
+Secplus v2
```
argilo, line 615: The actual name of the system is "Security+ 2.0". Secplus is just the name of my Python implementation.

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,459 @@
...
+
+    // compute rolling_total from rolling_digits[]
+    uint32_t rolling_total = 0;
+    uint32_t rolling_temp  = 0;
+
+    for (int i = 0; i < 18; i++) {
+        rolling_temp = (rolling_temp * 3) + rolling_digits[i];
+        // fprintf(stderr, ">> %12d\t%d\n", rolling_temp, rolling_digits[i]);
+    }
+
```
argilo, line 367: At this point, it's probably worth checking that `rolling_temp` is less than 2^28. If not, the code is invalid.

https://github.com/argilo/secplus/blob/dbe85c94372c8023c3334dcd7f112ed605c5a8de/secplus.py#L136-L137

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,459 @@
...
+    // rolling_total is a 28 bit unsigned number
+    // fixed_totals is 40 bit in a uint64_t
+    snprintf(fixed_str, sizeof(fixed_str), "%lu", fixed_total);
+    snprintf(rolling_str, sizeof(rolling_str), "%u", rolling_total);
+
+    /* clang-format off */
+    data_t *data;
+    data = data_make(
+            "id",       "ID",    DATA_STRING, "Secplus_v2",
```
argilo, line 421: Change this to `"Security+ 2.0"`?

argilo:
I noticed there's still a bunch of commented-out lines and a "delete this soon" section that could probably go now. Otherwise looks good and still works correctly.

src/devices/secplus_v2.c
zuckschwerdt: should be "model" and "Secplus-v2" would be prefered.

src/devices/secplus_v2.c
evilpete: done

src/devices/secplus_v2.c
evilpete: done

evilpete:
@zuckschwerdt 

Since this protocol has no identifying magic numbers nor CRC check, should I set it as disabled by default?

zuckschwerdt:
If there is a realistic chance of false positives: yes. Is the timing unique enough, maybe? Are there false positives when running against the rtl_433_tests files?

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,444 @@
...
+    uint32_t rolling_total = 0;
+    uint32_t rolling_temp  = 0;
+
+    for (int i = 0; i < 18; i++) {
+        rolling_temp = (rolling_temp * 3) + rolling_digits[i];
+        // fprintf(stderr, ">> %12d\t%d\n", rolling_temp, rolling_digits[i]);
+    }
+
+    // Max value = 2^28
+    if ( rolling_temp > 268435456 ) {
```
merbanan, line 369: I like the hex value better.

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,444 @@
...
+    bitbuffer_t bits_1    = {0};
+    bitbuffer_t fixed_1   = {0};
+    uint8_t rolling_1[16] = {0};
+
+    bitbuffer_t bits_2    = {0};
+    bitbuffer_t fixed_2   = {0};
+    uint8_t rolling_2[16] = {0};
+
+
+    // 280 is a cconservative guess
```
merbanan, line 276: s/cc/s/

merbanan:
@evilpete the preamble check will take care of lots of signals. Only signals with matching timing and matching preamble might collide. And then we have 2 more checks to filter out other signals. So we should enable the decoder.

src/devices/secplus_v2.c
argilo: This should also be `>=` instead of `>`.

merbanan:
I have no more comments, looks good to me.

src/devices/secplus_v2.c
evilpete: done

src/devices/secplus_v2.c
evilpete: done

argilo:
This branch fails to build now because `secplus_v1.c` is referenced in `CMakeLists.txt` and `Makefile.am` but is not checked in. Remove references to it and put v1 in a separate pull request?

evilpete:
> This branch fails to build now because `secplus_v1.c` is referenced in `CMakeLists.txt`, `Makefile.am`, `rtl_433.vcxproj` and `rtl_433.vcxproj.filters` but is not checked in. Remove references to it and put v1 in a separate pull request?

ahh sh*t sorry,   branch cross contamination ..   will fix immediately

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,444 @@
...
+        // break if we've recived both halfs
+        if (fixed_1.bits_per_row[0] > 1 && fixed_2.bits_per_row[0] > 1) {
+            break;
+        }
+
+    }
+
+    // Do was have what we need ??
+    if (fixed_1.bits_per_row[0] == 0 || fixed_2.bits_per_row[0] == 0) {
+        //  No?  Awww F'ck it then
```
Mindavi, line 332: :laughing:

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,444 @@
...
+            NULL);
+    /* clang-format on */
+
+    decoder_output_data(decoder, data);
+    return 1;
+}
+
+static char *output_fields[] = {
+
+        // Common fields
```
Mindavi, line 421: nit: I'd like to see this sorted in the same order as the `data_make`. That helps visually checking if everything is correct. (it is, so it's fine)

src/util.c
```diff
@@ -26,7 +26,7 @@ uint32_t reverse32(uint32_t x)
 {
     uint32_t ret;
     uint8_t* xp = (uint8_t*)&x;
-    ret = reverse8(xp[0]) << 24 | reverse8(xp[1]) << 16 | reverse8(xp[2]) << 8 | reverse8(xp[3]);
+    ret = (uint32_t) reverse8(xp[0]) << 24 | reverse8(xp[1]) << 16 | reverse8(xp[2]) << 8 | reverse8(xp[3]);
```
Mindavi, line 29: I think you need to merge master :)

src/devices/secplus_v2.c
```diff
@@ -0,0 +1,444 @@
...
+            fprintf(stderr, "%d ", roll_array[i]);
+        }
+        fprintf(stderr, "\n");
+    }
+
+    // SANITY check trinary valuse, 00/01/10 are valid,  11 is not 
+    for (int i = 0; i < 9; i++) {
+        if (roll_array[i] == 3) {
+            fprintf(stderr, "roll_array val  FAIL\n");
+            return 1; // DECODE_FAIL_SANITY;
```
Mindavi, line 241: You can just return DECODE_FAIL_SANITY here, nothing is done with the return value anyway ;)

Mindavi (approved):
All looks good to me. I did not check if it actually decodes the samples, but I'll assume it does. A check against rtl_433_tests for false positives would be good, but as merbanan said, there are enough sanity checks in here.

PR #1482, opened 2020-08-29
Acurite 6045 fix temperature 2.0F too low

### LINKED ISSUE #1401: Acurite 06045M Decoded Temperature Reading Low
I have a node set up in Node Red that displays several 06045M. I noticed the decoded value for the temperature is about 1.8 F to low compared to reading on my console. The humidity matches the console. As for distance of lightning strikes, from what  I saw is the distance data is being decoder closer then what the console reports.  I don't have enough data to compare.  The console I'm using is a 06047M.

rct:
Fixes #1401 

* 6045 documentation clean up
* exception check clean up
* deleted old debugging output that should no longer be necessary

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -224,7 +224,7 @@ Acurite 06045m Lightning Sensor decoding.
 
 Specs:
 - lightning strike count
-- estimated distance to front of storm, up to 25 miles / 40 km
+- estimated distance to front of storm, 1 to 25 miles / 1.6 to 40 km
 - Temperature -40 to 158 F / -40 to 70 C
 - Humidity 1 - 99% RH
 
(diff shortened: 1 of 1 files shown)


src/devices/acurite.c
```diff
@@ -376,26 +389,13 @@ static int acurite_6045_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsign
     *rawp = '\0';
 
     // Flag whether this message might need further analysis
-    if ((message_type != ACURITE_MSGTYPE_6045M) // 6045 message type is 0x2f
```
Mindavi, line 379: Why remove the `message_type` check? Is it incorrect?

Mindavi (approved):
Looks fine to me.

src/devices/acurite.c
rct: When that was originally written, devices were differentiated by message length.

I thought the casing for device had all been changed over to message_type, but I see now that I'm wrong. 

I'll fix that now.

PR #1483, opened 2020-09-02
Add support for Security plus v1

### LINKED ISSUE #1464: Garage Door - SecPlus & SecPlus 2.0+

evilpete:
Support for Security plus v1 

tested and functional but should  **not** be merged until @zuckschwerdt implements rows for PCM or a way or extend the reset_limit to receive  transmissions with longer gaps 

Based on code by Clayton Smith https://github.com/argilo/secplus

resolves #1464


evilpete:
tested it with  the commands:

Data in multiple rows

```
$HOME/Projects/rtl_433/build/src/rtl_433 -f 310.01M -s 250k  -R 164   
    -y "{96}113311731171371371177000   {96}733773171713177377317000"`
```

and data in single row

```
$HOME/Projects/rtl_433/build/src/rtl_433 -f 310.01M -s 250k  -R 164 \ 
    -y "{1645}89988b988b89b89b88bb80000000000000000000000000000073377317171317737731700000000000000000000000000000226622e622e26e26e22ee000000000000000000000000000001ccddcc5c5c4c5dcddcc5c0000000000000000000000000000089988b988b89b89b88bb80000000000000000000000000000073377317171317737731700000000000000000000000000000226622e622e26e26e22ee000000000000000000000000000001ccddcc5c5c4c5dcddcc5c000000000000000000000000000000000000000"
```

zuckschwerdt:
With c29f145 rows (gap_limit) are now available for PCM.

evilpete:
Added cache to secplus_v1

Added a patch for secplus_v2 as it does not deal with rows ( yet )

evilpete:
updates merged.

argilo:
This looks to be working. The only issue I noticed is that codes are printed multiple times (typically 3 times per button press). Remote send the pair of packets 4 times to overcome interference.

evilpete:
> This looks to be working. The only issue I noticed is that codes are printed multiple times (typically 3 times per button press). Remotes send the pair of packets 4 times to overcome interference.

This would be a little hard to fix since this decoder has to maintain state to waiting for the other (2nd?) half.

I guess we can add a second static buffer to drop repeat results within 1/2 second, but it is worth the effort ?

evilpete:
Updated to not look at the cache unless valid data was found in current packet
(Used to check the cache before looking at the packet)

src/devices/secplus_v1.c
```diff
@@ -0,0 +1,432 @@
...
+
+    search_index_2 = bitbuffer_search(bitbuffer, 0, cur_index, preamble_2, 8);
+    search_index_2 += 3;
+
+    // return first match in buffer
+    return (search_index_1 < search_index_2 ? search_index_1 : search_index_2);
+}
+
+// max age for cache in us
+#define CACHE_MAX_AGE 75000
```
argilo, line 132: 75ms is too short; the packets are actually spaced 100ms apart, so the only reason this currently works is that rtl_433 defaults to receiving samples from the RTL-SDR 524ms at a time. If the receive buffer in shortened (e.g. `-b 16384) then no data is received because the cache is always considered stale.

800000 might be a good value here, since the whole transmission lasts 800ms after a button press, and that value also happens to be larger than rtl_433's default RTL-SDR buffer size.

README.md
```diff
@@ -240,6 +240,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [164]  Security+ 2.0 (Keyfob)
     [165]  TFA Dostmann 30.3221.02 T/H Outdoor Sensor
     [166]  LaCrosse Technology View LTV-WSDTH01 Breeze Pro Wind Sensor
+    [167]  Security+ 1.0 (Keyfob)
```
argilo, line 243: These older fobs are just called "Security+".
```suggestion
    [167]  Security+ (Keyfob)
```

argilo:
> I guess we can add a second static buffer to drop repeat results within 1/2 second, but it is worth the effort ?

I agree it's not a critical feature, it just seemed like a "nice to have".

One way to do it would be to cache part 1 & part 2 separately, and return early if the received value is equal to the cached value.

src/devices/secplus_v1.c
evilpete: ok 800000 it is

README.md
evilpete: done

src/devices/secplus_v1.c
evilpete: done

zuckschwerdt:
Coalescing and dropping messages is a can of worms. We might one day offer some filter module to do that (likely we'll never e.g. derive values like rainfall though), for now you need a processor script.
rtl_433's basic job will always be to grab data of the air, demod, decode and output it with little added logic. Some user might really want to see each repeated message, maybe to gauge reliability.

merbanan:
One more point is that rtl_433 should be usable out of the box. I do think that a duplicate filter module is something that would make it easier to integrate with other consumers. But a duplicate filter will always keep state and that complicates things. I don't like complicated things.

src/devices/secplus_v1.c
```diff
@@ -0,0 +1,432 @@
...
+
+    // preformat unsigned int
+    char fixed_str[16]; // should be 10 chars max
+    snprintf(fixed_str, sizeof(fixed_str), "%u", fixed);
+
+    // fprintf(stderr,  "# Security+:  rolling=2320615320  fixed=1846948897  (id1=2 id0=0 switch=1 remote_id=68405514 button=left)\n");
+    /* clang-format off */
+    data_t *data;
+    data = data_make(
+            "Model",       "Model",    DATA_STRING, "Secplus_v1",
```
zuckschwerdt, line 373: should be "model" (lowercase). Is there some kind of (composite) "id" we can use?

evilpete: "Model" -> "model" : FIxed.

I have no idea about a composite ID (@argilo ?)

argilo: I'm not sure what a "composite id" means in this context.

argilo: The protocol itself is named Security+ and it it used in many devices sold under various brand names.

zuckschwerdt: Official names go in the r_device name field. We have a commitment to only output sanitized model names (of roughly `[A-Za-z][A-Za-z0-9]*(-[A-Za-z0-9]+)?`) -- esp. `+` is really bad for MQTT.

zuckschwerdt: I mean just any cobbled together id, as `id0`, `id1` are not good for that. Is `remote_id` a fixed unique id?

evilpete: I guess a ID _maybe_ can be derived from the  `fixed` data part  if all the flags bits were masked or truncated 

 @argilo what do you think ?

argilo: `remote_id` (for pushbutton remotes) and `pad_id` (for keypad remotes) are different for every remote. They don't seem to tell you anything about the model.

argilo: Looking at the other model names, it seems that spaces are allowed. If we wanted to be more descriptive, we could use "Security Plus remote", or even distinguish between the two types "Security Plus pushbutton remote" and "Security Plus PIN pad".

evilpete: I think what @zuckschwerdt  was asking for was a unique identifier for the `Id` field so  different remotes can be easily  distinguishable .

zuckschwerdt: Only alphanum and a single dash in the middle -- it's a short unique key, useable as database key, nothing descriptive -- e.g. we use the model as MQTT topic. If the same protocol has different sensors with different functions (not sub-function of the same sensor) that should be given as `subtype`, here maybe "pad" and "remote". S.a. https://triq.org/rtl_433/DATA_FORMAT.html#message-data
If remote_id and pad_id are mutually exclusive the output should just be `id` with the appropriate value.

evilpete: ok I added `id` field that is equal to `remote_id` or `pad_id` depending on the data.

argilo: > If remote_id and pad_id are mutually exclusive the output should just be id with the appropriate value.

They are, in the sense that you've either got a `remote_id` (for pushbutton remotes) or `pad_id` (for PIN pads). Agreed that they may as well both go in a single "id" field.

PR #1486, opened 2020-09-03
HTTP POST output

vhuk:
Proof of concept for REST (HTTP POST) output in JSON format. Introduces libcurl dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,8 @@ if(MINGW OR MSVC)
 list(APPEND NET_LIBRARIES ws2_32 mswsock)
 endif()
 
+list(APPEND NET_LIBRARIES curl)
+
 ########################################################################
 # Setup the include and linker paths
 ########################################################################
--- a/include/output_http.h
+++ b/include/output_http.h
@@ -1,0 +1,19 @@
+/** @file
+    REST (HTTP) output for rtl_433 events
+
+    Copyright (C) 2020 Ville Hukkamäki
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#ifndef INCLUDE_OUTPUT_HTTP_H_
+#define INCLUDE_OUTPUT_HTTP_H_
+
+#include "data.h"
+
+struct data_output *data_output_rest_create(char const *url, int header_count, char *headers[]);
+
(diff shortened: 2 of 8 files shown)


zuckschwerdt:
Interesting idea. The concept is "webhook", right? You are using this to push messages to an HTTP server.
Can you showcase your setup, what's on the side of the web server and how you process messages further?

I guess the actual HTTP headers here are mostly static and if we stick to basic HTTP/1 without anything advanced like compression or pipelining then we can use a a simple TCP socket instead of a lib dependency.

vhuk:
Primary driver to use curl is to implement TLS/SSL, support redirects from the server and URL parsing, e.g. provide basic authentication credentials in the URL. What you end up with is fairly flexible target URL where you can provide credentials and everything on the command line, for example -F rest:http://username:password@example.com/api?yourtoken=xyz. If there are any redirects, for example, curl will follow those.

HTTP headers are static; I use them to include authentication token into the headers to avoid having it in POST body or query string parameters.

Simple HTTP/1.x with no compression, ignoring redirects and nothing fancy could be implemented easily without external libraries. TLS support would in require something external.

My use case for this is to have separate rtl_433 instances running on the Raspberry Pis deliver their output to a centralized box for processing and logging. I'll put together a simplified version of the server side components at some point.

zuckschwerdt:
I see, so without some HTTP features this wouldn't be useful to you. TBH the chance of this project including additional deps is very slim, so it might still be worth it to check out if the Mongoose lib can do this: https://cesanta.com/docs/http/client-example.html -- there is TLS and following redirects shouldn't be too involved.

There is more to consider though: we can't block the main thread, so it's either non-blocking IO (Mongoose) or a worker thread with all the requirements for locking and sharing.
Also we would likely need a queue and batching as pushing to TCP (or worse HTTP) is unreliable (from the clients perspective): delay, timeout, slow or aborted transfer, errors (do we retry on 4xx or 5xx)? This means we could either end up with many pushes in-flight or we need control with e.g. a queue.

I still do like the idea to expand the feature set of rtl_433. In the meantime, did you look into using a bridge script (e.g. [rtl_433_custom.py](https://github.com/merbanan/rtl_433/blob/master/examples/rtl_433_custom.py))? We'll happily include an HTTP webhook push example.

The description of your setup sound much like mine. I use the SYSLOG output (simple UDP with some headers) to push from multiple Raspis to some event handling script on the (local) server. I also use the MQTT output to persist current state (and then display some topics on various screens). And finally I use the Influx output to archieve history (which goes to Grafana).

PR #1489, opened 2020-09-05
added support for LaCrosse LTV-WSDTH01

mbruski:
Code cleanup, CRC checking now works, minor field realign (ID, flags).


zuckschwerdt:
Thanks! I'm going to apply these updates manually as I already have other changes pending. Expect full working support for this tomorrow or so.

mbruski:
ID is confirmed fixed and found to match the bar code covering the program access port on the underside of unit.  It is actually 24h, not 19h or 20h as previously assumed.  Flags were previously assumed to be 5h but are actually 8h, five of which have an unknown function.  While playing with battery removal with unit switched on, I noticed that one bit (0?000000) was intermittently toggling on/off, this caused me to change my assumption about ID/flag length and  ID alignment.  It was then when I discovered that the resulting ID matched the bar code!.  I'm now thinking the toggled bit may be battery status but have yet to prove that.  As noted in the pull request, the CRC check now works correctly - my thanks to Christian for his help on that!

PR #1498, opened 2020-09-20
Fix Security+ 2.0 decoder for new gap_limit rows

argilo:
The change in c29f145a2628da4749f56a1dbbf1b330d845d9c1 broke the Security+ 2.0 decoder, because the data now arrives in multiple rows. To fix this, I've replaced the `while` loop that searched for packets in a single row with a `for` loop that iterates over the rows. The decoder now works with my remote once again.

@evilpete

### DIFF
--- a/src/devices/secplus_v2.c
+++ b/src/devices/secplus_v2.c
@@ -268,23 +268,19 @@ static int secplus_v2_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     bitbuffer_t fixed_2   = {0};
     uint8_t rolling_2[16] = {0};
 
+    for (uint16_t row = 0; row < bitbuffer->num_rows; ++row) {
+        if (bitbuffer->bits_per_row[row] < 110) {
+            continue;
+        }
 
-    // 280 is a conservative guess
-    if (bitbuffer->bits_per_row[0] <= 280) {
-        return DECODE_ABORT_LENGTH;
-    }
+        search_index = bitbuffer_search(bitbuffer, row, 0, _preamble, _preamble_len);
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Looks cleaner now. We should look for other PCM decoders where gap_limit now can help to clean up processing logic.

evilpete:
this removes the  ability handle false starts ( noise containing the preamble before the read data ). but that chances of that condition are very low.

I would not recommend removing the while search in Sec_v1 because there is no  preamble

argilo:
> but that chances of that condition are very low

Correct, since 28 bits are used for detection:

https://github.com/merbanan/rtl_433/blob/703d4fc1d1a535c1264b9e5b5a58e78f4996f706/src/devices/secplus_v2.c#L252-L253

Key fobs transmit three copies of the packets, so I don't expect false starts to be a problem.

zuckschwerdt:
Good general point about the preamble. I was toying with the idea that decoders (r_device) can declare a needed preamble, then the slicer will feed every matched preamble (basically that while search), but I'm not sure how to handle multiple rows, only some with matches?

evilpete:
> > but that chances of that condition are very low
> 
> Correct, since 28 bits are used for detection:
> 
> https://github.com/merbanan/rtl_433/blob/703d4fc1d1a535c1264b9e5b5a58e78f4996f706/src/devices/secplus_v2.c#L252-L253
> 
> Key fobs transmit three copies of the packets, so I don't expect false starts to be a problem.

Sorry,  I am a security person thus I have a habit of evaluating potential scenarios
no matter how  remote and it shows in my code

PR #1508, opened 2020-10-05
Added DATA_UINT DATA_INT64 DATA_UINT64,

evilpete:
renamed `DATA_DOUBLE` to `DATA_FLOAT` as per @merbanan 's request
(Note stdarg(3) does not differentiate between  types  double &  float )

included  # define for `DATA_DOUBLE` till all devices ate updated (code compatibly )

added new data types `DATA_UINT`,`DATA_INT64`,`DATA_UINT64`
with min/max tests added to tests/data-test.c

will update the 94 files that reference to `DATA_DOUBLE` in a day or two.


zuckschwerdt:
That's going to add a lot of extra cases and print functions in every output (influx and mqtt are missing here). Do we really need uint64? If we support int64 do we still need the int32/uint32?
Also `%ld` is not reliable to print in64_t, it should be `PRId64` -- but that's not supported on MSVC and `%j` also isn't. Perhaps cast to `long long` and use `%lld`.

Mindavi:
Is there any specific use case you have in mind for this? Please describe why this change is required, as that might help understand why this (rather big) change is needed. Maybe refer to the issue where this was brought up or create one to explain?

merbanan:
Using doubles is not needed, that is the main reason. Adding more types may or may not be needed.

zuckschwerdt:
Using "float" as term will make the intent clear but might be confusing as varargs only handle double -- at least in the data_make() arguments and processing it will be double.

evilpete:
no needed

PR #1511, opened 2020-10-06
Support for Schrader TPMS SMD3MA4 (Subaru)

RonNiles:
The latest rtl_433 could not read the TPMS from my new Subaru WRX, so I added some code and would like to contribute the changes to the project. Sample signals are in [merbanan/rtl_433_tests#367](https://github.com/merbanan/rtl_433_tests/pull/367)

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [165]  TFA Dostmann 30.3221.02 T/H Outdoor Sensor
     [166]  LaCrosse Technology View LTV-WSDTH01 Breeze Pro Wind Sensor
     [167]  Somfy RTS
+    [168]  Schrader TPMS SMD3MA4 (Subaru)
 
 * Disabled by default, use -R n or -G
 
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Looks really good.
- `bitbuffer_extract_bytes(bitbuffer, 0, 0, received_preamble, NUM_BITS_PREAMBLE);` is a logical no op, `bitbuffer->bb[0]` equals `received_preamble`. I guess you want the `0xE0` nibble masked of, either just compare "by ==" or use `bitbuffer_search()`
- `flags = get_next_bits(&decoded, &offset, NUM_BITS_FLAGS);` is `flags = (decoded[4] & 0x0e) >> 1;` -- I'll likely convert that after merge. I know we should offer something nicer, plans are already made :)
- No need to output `flags` as hex string. 3 bits is 0-7, so just DATA_INT looks the same

RonNiles:
Thanks for the suggestions. I have made changes consistent with your first and last suggestions. As for the `get_next_bits()` helper function, I left it as is for you to convert after the merge. I thought a helper function would increase readability and be less error prone than masking and shifting each byte individually. You are welcome to convert it to whatever suits the project,

zuckschwerdt:
Thanks. Yes such helper is really much nicer. Something like this will be added soon. The requirement right now is to have all 200 decoders as concise and straight forward as possible to minimize maintainance headache;)

RonNiles:
Sure, I think the conversion without the helper function looks like this
```
    /* FFFSSSSS SSSSSSSS SSSSSSSS SSSPPPPP PPPPPxxx */
    bits      = decoded.bb[0];
    flags     = bits[0] >> 5;
    serial_id = ((bits[0] & 0x1f) << 19) | (bits[1] << 11) | (bits[2] << 3) | (bits[3] >> 5);
    pressure  = ((bits[3] & 0x1f) <<  5) | (bits[4] >> 3);

```

PR #1513, opened 2020-10-07
Change width calc from r_device to slicers

zuckschwerdt:
This further modularizes the slicers (demods).
We can now run any pulse_data_t through the slicers, independent of a sample rate.

Fixed sample rate calculated values are removed from r_device and moved to the slicers.
This adds some setup with one float div and six float multiplies to each slicer call,
which should be negligible compared to looping through the actual pulse data.

Critical comments welcome.

### DIFF
--- a/include/r_device.h
+++ b/include/r_device.h
@@ -71,16 +71,6 @@ typedef struct r_device {
     /* private for flex decoder and output callback */
     void *decode_ctx;
     void *output_ctx;
-
-    /* private pulse limits (converted to count of samples) */
-    float f_short_width; ///< precision reciprocal for PCM.
-    float f_long_width;  ///< precision reciprocal for PCM.
-    int s_short_width;
-    int s_long_width;
-    int s_reset_limit;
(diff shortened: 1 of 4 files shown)


merbanan:
The code logic looks fine. But what is the reason for this? I think there is a feature beyond "We can now run any pulse_data_t through the slicers, independent of a sample rate.". What is that?

zuckschwerdt:
Yes :) We either need to copy the current active sample rate to rfraw inputs to match or (the saner option to me) we need to make slicers independent of any input rate.
In the future, if we choose to allow changing the sample rate on the fly (or implement decimating filters etc) this will be handy too. And this makes the r_device struct more "static" (i.e. better allows threaded processing and such).

merbanan:
This should also be one tiny step towards a decoder library.Just pass in a sample rate and a pulse table and you get back decoded data. Sample rate changes also came to mind. For example when you frequency hop. 1000k on 868MHz and 250k on 433MHz. Anyway looks good to me.

PR #1514, opened 2020-10-08
Add support for malformed frame sync for ERT SCM sensors

kennywakeland:


zuckschwerdt:
Interesting, thanks. Where does that alternative sync match, i.e. how is it related?
- "type" is a reserved key, keep "ert_type"
- have a look at the style check notes.

merbanan:
The real feature of this PR is the alternative sync pattern. With no source for this alternative sync or test files or follow up comments the PR is closed.

Issue #1510, opened 2020-10-05
Home Assistant MQTT Auto Discovery example state_topic missing subtype (message type)


rct:
Summary: The `state_topic` that is auto generated by the hass mqtt example python script doesn't include `subtype` for rtl_433 devices that have a subtype, causing Home Assistant to subscribe to the wrong topic.

The details:

The rtl_433 MQTT device topic includes `subtype` for devices like Acurite 5n1 and Acurite Atlas that send different message types with different payloads and have `subtype` field.  The topic bases are

* rtl_433/{host}/devices/Acurite-5n1/56/{channel}/{id}
* rtl_433/{host}/devices/Acurite-5n1/49/{channel}/{id}

(56 and 49 are the different message type IDs), so full topics are:

* rtl_433/{host}/devices/Acurite-5n1/**56/**{channel}/{id}/temperature_F
* rtl_433/{host}/devices/Acurite-5n1/**49/**{channel}/{id}/wind_dir_deg

The home assistant MQTT auto discovery example script doesn't capture/use the subtype if it is part of the topic.  So the `state_topic` in the generated config for Home Assistant is incorrect:

* rtl_433/{host}/devices/Acurite-5n1/{channel}/{id}/temperature_F
* rtl_433/{host}/devices/Acurite-5n1/{channel}/{id}/wind_dir_deg

At first glance it looks like instance naming rules need to be expanded. In theory this should probably include every possible permutation for the existence/non-existence of `channel`, `id`, and `subtype`. 

https://github.com/merbanan/rtl_433/blob/da77261dabaec8a92bc0d55270a16ceeb2d97301/examples/rtl_433_mqtt_hass.py#L386-L396

zuckschwerdt:
Yes, the default topic for MQTT is `devices[/type][/model][/subtype][/channel][/id]` a workaround would be to just request `devices[/type][/model][/channel][/id]` but your suggested change is way better (type might need to be there too). Do you want to PR this?

rct:
I want to mention a separate issue - This example script is a very nice starting point for Home Assistant users ONCE you figure out what the script does and it's purpose. There are almost no descriptions for many of the example scripts. Most are simpler than this one.

A basic description along the lines of  "rtl_433_mqtt_hass.py subscribes the rtl_433 event stream that is published to MQTT by rtl_433. The script publishes additional MQTT topics that can be used by Home Assistant to automatically discover and minimally configure new devices."

rct:
> 
[...]

PR #1520, opened 2020-10-15
Revert Acurite subtype to message_type

rct:
Per the discussion in #1510, `subtype` is used incorrectly for Acurite weather stations that send multiple message types. Per @zuckschwerdt `subtype` is intended to be used to identify somewhat distinct sub-devices. Message type isn't used that way for the Acurite 5n1 and Acurite Atlas which send 2 or 3 different message types respectively.

`subtype` is used in the naming of MQTT topics and possibly other output forms. The default topic is made using the following fields from the data.

    rtl_433/[host]/devices[/type][/model][/subtype][/channel][/id]

Identifying the Acurite `message_type` as `subtype` causes there to be 2 or 3 different topic prefixes for the one device. To get all of the wind speed updates, it would be required to subscribe to all 2 or 3 topics under the different subtypes.

This is potentially a breaking change for people who are using Acurite 5n1 or Atlas weather stations and `-F mqtt` output of rtl_33.

Note: I didn't change the use of subtype for the 275 sensor, which looks like it behaves differently depending on whether a probe is attached. I don't have enough info about that device to know what would be right for that device.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -544,7 +544,7 @@ static int acurite_atlas_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsig
             "channel",              NULL,           DATA_STRING, &channel_str,
             "sequence_num",         NULL,           DATA_INT,    sequence_num,
             "battery_ok",           NULL,           DATA_INT,    !battery_low,
-            "subtype",              NULL,           DATA_INT,    message_type,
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
LGTM. Let this sit for a week so others can comment, otherwise good for merge.

PR #1522, opened 2020-10-17
DSC revert subtype to fix MQTT topic naming

rct:
Per the discussion in #1510, `subtype` is used incorrectly for DSC sensors. Previously a `subtype` string was added in an attempt to identify the type of sensor. The DSC type code was supposed to indicate a *class of device*, not a specific model number. There are a half dozen different devices that use the same type code. As a result, the current `subtype` string is often incorrect.

Per @zuckschwerdt subtype is intended to be used to identify somewhat distinct sub-devices. This isn't the case for DSC sensors.

The presence of `subtype` affects naming, `subtype` is used in constructiong MQTT topics and possibly other output forms. The default topic is made using the following fields from the data:

    rtl_433/[host]/devices[/type][/model][/subtype][/channel][/id]

There might be some other devices that are using `subtype` incorrectly. A related PR #1520 for Acurite devices changes `subtype` to `message_type`.

I did not output the type code as a separate field because it is already part of the ID/ESN.  A consumer can easily extract the four bits from the ID.  A human reading the output can easily view the type code as the first hex digit of the ESN.

One place where the type code may be useful within rtl_433 is to output a different message type for the one-way keyfob which uses status bits differently.

### DIFF
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -57,9 +57,43 @@ Packet Decoding
 - CRC (cr)    = 8 bits, CRC, type/polynom to be determined
 
 The ESN in practice is 24 bits, The type + remaining 5 nybbles.
+The physical devices have all 6 digits printed in hex. Devices are enrolled
+by entering or recording the 6 hex digits.
 
 The CRC is 8 bit, reflected (lsb first), Polynomial 0xf5, Initial value 0x3d
+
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! LGTM. Let others interested have a look, otherwise good for merge.

PR #1528, opened 2020-10-21
Add support for WH31E RCC packet type

etihwnad:
The WH31E temperature and humidity sensor also includes a radio controlled clock (RCC) receiver.  It periodically stops transmitting T+H data to receive the radio clock signal (WWVB at 60kHz for me) and transmits successful reception events with the current time in UTC to the base station (Ambient Weather WS-3000).

This pull request adds the radio clock packet type sent by the WH31E to the base station.

Code comments document the inferred packet encoding along with a set of relevant timestamped packets recorded off the air.  The packet includes one byte of unknown meaning.

Two questions:

* Does the output data format in the commit match convention for the rest of rtl_433 ok?

* The packet timestamp is known to be in UTC.  I followed the string format used for the Oregon Scientific and Fine Offset WH1080 decoders, but it would be more appropriate to add the "Z" or "-00:00" time string suffix to properly identify the time zone according to RFC3339.  Happy to make another PR to change all of those.

### DIFF
--- a/src/devices/ambientweather_wh31e.c
+++ b/src/devices/ambientweather_wh31e.c
@@ -50,6 +50,61 @@ Some payloads:
     30 44 92 15 3d 07 5f 07 45 04 5f
     30 c3 81 d6 5b 90 35 08 35 44 2c
 
+
+Ambient Weather WH31E Radio Controlled Clock (RCC) packet WWVB
+
+These packets are sent with this schedule, according to the manual:
+    After the remote sensor is powered up, the sensor will transmit weather
+    data for 30 seconds, and then the sensor will begin radio controlled clock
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Nicely aligns with the RTC fields other sensors already have. I'll review and merge asap.

zuckschwerdt:
re time formats: any ISO8601 string would be fine, if you have info about a TZ even better.

etihwnad:
Thanks for the feedback!

I know for sure that a signal detected from WWVB will be in UTC.  It is not clear from documentation I've found about the others if the TZ info is UTC or local (both MSF and DCF broadcast in local time, for example).

Updated the PR to include the TZ.

PR #1532, opened 2020-10-23
Added output format option to flex getters

vidarino:
Added an option to specify an output format for flex getters, e.g. like so:

-X "(...),get=@2:{32}:code:%x"

to get the value in hexadecimal instead of the default int. (When showing values as integers I sometimes got big negative numbers, which were ugly at best.)

### DIFF
--- a/src/devices/flex.c
+++ b/src/devices/flex.c
@@ -71,6 +71,7 @@ struct flex_get {
     unsigned long mask;
     const char *name;
     struct flex_map map[GETTER_MAP_SLOTS];
+    const char *format;
 };
 
 #define GETTER_SLOTS 8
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Nice. Can you clean up the style warnings?

src/devices/flex.c
```diff
@@ -474,6 +481,11 @@ static void parse_getter(const char *arg, struct flex_get *getter)
             getter->bit_count = parse_bits(arg, bitrow);
             getter->mask = extract_number(bitrow, 0, getter->bit_count);
         }
+        else if (*arg == '%') {
+            getter->format = strdup(arg);
+            if (!getter->name)
```
zuckschwerdt, line 486: `if (!getter->format) ` :)

vidarino:
Oof. Fixed. :)

PR #1491, opened 2020-09-11
Tasmotized Sonoff RF-Bridge decoding support

halfbakery:
The [tasmotized Sonoff RF-Bridge](https://tasmota.github.io/docs/devices/Sonoff-RF-Bridge-433/#b1-to-b0-conversion-tools) could be a great tool for RF433 communication. The device contains a small EFM8BB1 MCU for RF433 communication, the firmware of which can be replaced by an open source alternative (Portisch firmware). The MCU is so resourceless, that it can support just a few RF433 protocols. But it has a hidden gem, called bucket sniffing.

The bucket sniffing output is a small pulse description message transmitted via the MQTT protocol. This addition listens for these messages and translates them into the standard PULSE_OOK format and lets the existing decoders do their jobs.

PR #1542, opened 2020-10-29
Skeleton implementation of the transport link abstraction layer

halfbakery:
Transport link is a thin abstraction layer interconnecting the output functions with the real transport protocols. The current implementation has some drawbacks:
* the output formatting functions are directly calling the fprintf/fputc and similar libc functions, which bind them to files
* every output creates a new instance, that can be problematic with MQTT
* hard to add input functionality

Links could be set up with the new `-L` command link option, and we can use them in the output formatting functions like the following:
```
-L mymqtt=mqtt://localhost:1883 -F mymqtt/csv:foo/bar/topic
```

Of course the current syntax is still supported, so no breaking change.

This is just a skeleton to show how I imagine the new layer, more works needed to add the real functionality. I will finish it if you like. Opinions?


zuckschwerdt:
Interesting stuff. If I understand correctly you propose to split outputs into "payload generator/consumer" and "message transport", right?
Seems a viable idea for MQTT, define the "link" then add different (-configured) "inputs" and outputs as needed. But I'm not sure if this could be applied to any other "output" we have?
(in our terminology "input" is a sample source so "output" currwould be a catch-all for incoming control data too, "link" paints a better picture though)

We are (slowly) moving towards an API, both for remote control and custom applications. I have some changes lined up for the remote control part: HTTP (ReST, RPC, WS) and maybe plain TCP if that's needed. It will be interesting to again look at that with your ideas in mind :)

zuckschwerdt:
Thinking about this, independant of the behind-the-scenes, it might be easier for users if we only offer the -F mqtt... option and then on each additional instance lookup if there is already the same host-port present and reuse that connection.
On the other hand it should be no problem to connect twice to the broker, once for pub and once sub.

halfbakery:
> Interesting stuff. If I understand correctly you propose to split outputs into "payload generator/consumer" and "message transport", right?

Definetely. The message transport parts are easily exendible, e.g. sending over a tcp connection or posting to an url. E.g.:
```
-L srv=http://foo.bar,token=XXXX -F srv/csv:/API/v1
```

> Seems a viable idea for MQTT, define the "link" then add different (-configured) "inputs" and outputs as needed. But I'm not sure if this could be applied to any other "output" we have?
> (in our terminology "input" is a sample source so "output" currwould be a catch-all for incoming control data too, "link" paints a better picture though)
> 
> We are (slowly) moving towards an API, both for remote control and custom applications. I have some changes lined up for the remote control part: HTTP (ReST, RPC, WS) and maybe plain TCP if that's needed. It will be interesting to again look at that with your ideas in mind :)
It was always in my mind:
```
-L mqtt://localhost:1883 -L my-tcp=tcp-listen://0.0.0.0:3333 -r mqtt/rfraw:foo/bar -r my-tcp/rfraw
```
I've seen you started to merge mongoose manager contexts which was a must to implement these. :-)

halfbakery:
> Thinking about this, independant of the behind-the-scenes, it might be easier for users if we only offer the -F mqtt... option and then on each additional instance lookup if there is already the same host-port present and reuse that connection.
> On the other hand it should be no problem to connect twice to the broker, once for pub and once sub.

Even my current code supports the following:
* output configuration without explicit link setup creates a link with name link0 .. linkN
* multiple output configurations without explicit link setup merges into one link

It hasn't worked yet, but I want to implement:
* link setup without name= creates a link with the name of the protocol (e.g. `-L mqtt://localhost:1883`)
* specifying outputs like `-F mqtt/rfraw` in that case

zuckschwerdt:
> I've seen you started to merge mongoose manager contexts which was a must to implement these. :-)

Thanks for testing and fixes, btw :)
I'm currently factoring out the SDR loop so the application can be simplified …and finally only run one happy (and fast) event loop.

Instead of putting e.g. an rfraw option on every output, just have it once and stack it on any output, nice!
Of course different outputs will have different requirements/rules, but I don't see a dangerous example yet. CSV on an MQTT topic? Stream KV over Syslog? why not ;)

halfbakery:
One of the reasons behind the -L option is to let users use mqtt input without mqtt output. It's nicer than the hack in my mqtt_rfraw implementation.

zuckschwerdt:
I'm still pondering if the explicit link option is needed or if e.g. a `-F rfraw+syslog` style could work for all cases.

halfbakery:
Setting up a link could require a lot of parameters (e.g. TLS has a lot of options), I think it's nicer and gives us an opportunity to reuse, e.g.:
```
-L srv1=http://foo.bar,tls=1,cafile=/etc/aaa.key=/etc/bbb -F srv1/csv:/API/v1 -F srv1/json:/API/v2
```
And a low hanging fruit for the future is some kind of filtering, e.g.:
```
... -F srv1/csv:/API/v1,if-device='wt1024'
```
in that case the explicit link setup is a big saver.

As I told earlier I want to keep the current syntax, so `-F mqtt://localhost:1883` will work if someone wants to use just one output with its default format.

zuckschwerdt:
The copyright on the outputs split out from data.c should be me -- everything except the original design of the data facility is my fault there ;)

halfbakery:
Almost done...

The MQTT driver has been split and refactored into a link driver and an output driver. The link driver is reuseable, so the
```
rtl_433 -L conn=mqtt: -F conn/json:foo/bar
```
will send the json messages into the `foo/bar` MQTT topic. Of course, the csv and kv outputs can be redirected too :-)

halfbakery:
Why the -L option is useful?

As I said earlier, transport link merging has been supported since the first version, so `-F csv:/tmp/a -F json:/tmp/a` will use the same transport link entry. An existing transport link connection will be reused, if all of the critical parameters are equal. The critical parameters are depending on the link type:
* for file is (filename)
* for mqtt is (host, port, user, password, client_id)
* for http (not yet implemented) could be (host, port, tls parameters, auth token etc.)

If we reduce the list of the critical parameters, we will exclude possible use cases (e.,g. sending to different MQTT brokers or using two connections with different authenticated credentials). Which is nicer:
```
rtl_433 -F mqtt://localhost:1883,u=foo,p=bar/json:a/b -F mqtt://localhost:1883,u=foo,p=bar/csv:a/c
```
or
```
rtl_433 -L conn=mqtt://localhost:1883,u=foo,p=bar -F conn/json:a/b -F conn/csv:a/c
```

How to specify input methods? With the new method it's possible to write similar things like that:
```
rtl_433 -L conn=mqtt://localhost:1883,u=foo,p=bar -r conn/rfraw:input/topic
```

I definietly open for a different notation, but the syntax of the -F and -r options are quite different and complicated. This implementation keeps compatibility with the current syntax and gives us a lot of opportunity to improve. Opinions?

zuckschwerdt:
We need a table with link types and output formats and see which combinations work and are useful. E.g. CSV won't work well with anything except file and stdout.

halfbakery:
output transport link / format combinations (with areas of improvement):
|         |  json | csv | kv | syslog | mqtt(nested) |urlencoded |
|--------|:------:|:----:|:---:|:---------:|:-----------------:|:---------------:|
| file    |   X    |  X  |  X  |            |                      |                    |
| mqtt |    X   |       |      |            |          X          |                    |
| tcp   |    X   |   X  |      |     X     |                      |        X          |
| udp  |         |        |     |     X     |                      |                     |
| http  |   X    |       |      |            |                      |         X         |
| websocket | X |  |      |            |                      |         X         |
| unix domain socket | X | X | |  X |                 |        X          |

The input case is easier, every combination is possible. Useful input formats:
* ook
* rfraw

valid input sources:
* mqtt topic
* http
* websocket
* tcp

halfbakery:
I've seen the new http server. Another good example for the -L option: :-)
* instead of a hard-coded outputs, you can configure individually, e.g.: `-L server=http://0.0.0.0:8433 -F server/json:/path-for-device1,filter-device=wt1024 -F server/json:/path-for-device2,filter-device=*`
* accepting rfraw input: `-r server/rfraw:/post/your/rfraw/here`

So you can configure the http-server with the -L option and the -r and -F options for input/output setup.

zuckschwerdt:
Do you mean an HTTP POST client? The server has a fixed API with different, known endpoints for RPC, JSON commands, Websocket, event- and plain (HTTP) stream, maybe later a ReSTful API and raw TCP API.

halfbakery:
My idea is to make the new http-server configurable (not now, just as a possibility). I know the current endpoint layout is static, but it needn't be:
* define event stream filtered by some user defined condition. e.g. create multiple event streams based on the transmitting device
* define input endpoints accepting POST request, e.g. processing rfraw input and send back the decoded result. Each defined endpoint could have different settings, e.g. list of the enabled decoders

On the other hand, a new output driver transmitting the data with HTTP POST would be useful. Both of them could be useful.

halfbakery:
From my point of view, this PR is ready to integrate. The code compiles and the tests are fine (except the maintainer_update). There are output modules which haven't been converted (influx and syslog), but it can be added later. The converted modules work like before, as far as I know there's no functionality loss. What is your opinion on integration?

Have I managed to convinced you about the -L argument? If it's your only concern, would you like to merge the PR without the -L command line option?

zuckschwerdt:
Thanks for thinking this through and implementing it in such short time. It will take some time (and likely debate) to review this.

The "link" abstraction is really interesting. But e.g. I need to ponder what this offers for the simple stuff like KV and CSV. Maybe this should just be JSON to/from everything. (rtl_433 can't be everything to everyone, pretty sure we won't e.g. offer BSON or MessagePack if there isn't a hard requirement by some very well used protocol.)

On the HTTP server topic: I really see no merit in changing the API dynamically. We fully expect users to commit to our fixed API. If there is solid case to conform with an existing API (Prometheus e.g.) we can just add that for everyone.

halfbakery:
> The "link" abstraction is really interesting. But e.g. I need to ponder what this offers for the simple stuff like KV and CSV. Maybe this should just be JSON to/from everything. (rtl_433 can't be everything to everyone, pretty sure we won't e.g. offer BSON or MessagePack if there isn't a hard requirement by some very well used protocol.)

The whole link abstraction was started with my [RfRaw  over MQTT PR](https://github.com/merbanan/rtl_433/pull/1491). Most of the ESP8266/ESP32/Arduino based projects have problematic RF433 support. Your project is the best and you have fantastic device support. The rtl_433 project couldn't be used as a library currently and it requires posix systems to run, but the IoT devices have only limited resources and has no posix compatibility. That's why I realized a client/server model would be ideal for everyone. The IoT devices should connect to a rtl_433 server, and send the captured OOK waveform for processing. The transport could be plain TCP, MQTT or HTTP. It can be even better than embedding an "rtl_433 lib" into an IoT application: easier to update a server on a linux system than reflash a device when a new RF device appears on the network.

I made a quick hack which put a new event loop polling for MQTT messages and processing it. You were right that this solution was a "little" weird, because:
* there are already two or three event loops and I had added another one
* the MQTT output has to be enabled because I couldn't find the place in the command line where I could configure input devices (the -r option has too many suboptions to extend it even more)

When I was trying to find a better place for my addition, I found some problems:
* the output device is a tightly coupled (format, transport) pair, it's not easy to extend (e.g. send JSON over HTTP or some key=value format over plain TCP). If we separate them, the possibilities would be multiplied and the code would be cleaner. Adding a new transport or format (e.g. BSON) would be easy and anyone could send them via any supported protocols (write to a file, send over plain TCP, HTTP POST or via MQTT).
* the input side is even worse, the input format processing (e.g. OOK reading) is tightly coupled to the libc stream API. Accepting OOK or RfRaw data over plain TCP or MQTT is not easy to implement. I had no time to check, but is the new http server able to accept and process RfRaw data?
* no transport connection sharing, adding an MQTT device requires a new MQTT connection too. Ok, it's not forbidden, but the typical pattern is to use just one MQTT connection per client.

The transport link layer means no more code, just a thin API between formats and transport protocols. First I implemented the output part to show you a demo about the possibilities, next I want to do the same with the input side: dissect the communication (currently file input) from input format processing. RfRaw over TCP? OOK over WebSocket? Why not?

> On the HTTP server topic: I really see no merit in changing the API dynamically. We fully expect users to commit to our fixed API. If there is solid case to conform with an existing API (Prometheus e.g.) we can just add that for everyone.

It was just a quick idea to add input where the rtl_433 could receive data to process (e.g. RfRaw, OOK etc.) from other sources or filter the decoded output based on some user defined condition. It could be easily implementable with a transport link enabled http server, but forget it first. :-) Let's focus on the transport link itself.

As I told you the current implementation works without the -L switch. I've found the -L option very useful which makes the configs easier to understand and shorter, but it's just optional.

merbanan:
@halfbakery and @zuckschwerdt can this be closed or is it still relevant ?

PR #1544, opened 2020-11-01
Add  support for ECODHOME smart socket

zuckschwerdt:
see #1525

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [171]  LaCrosse Technology View LTV-TH Thermo/Hygro Sensor
     [172]  Bresser Weather Center 6-in-1
     [173]  Bresser Weather Center 7-in-1
+    [174]  EcoDHOME Smart Socket and MCEE Solar monitor
 
 * Disabled by default, use -R n or -G
 
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -355,6 +355,7 @@ stop_after_successful_events false
   protocol 171 # LaCrosse Technology View LTV-TH Thermo/Hygro Sensor
   protocol 172 # Bresser Weather Center 6-in-1
   protocol 173 # Bresser Weather Center 7-in-1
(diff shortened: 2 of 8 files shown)


vannetta:
I'm testing now:
the 71b3 message is always 333363 for the consumption; as the sun is below horizon, there is only consumption: 8c74 12d6 (W) and 8c76 d4b9 (message 333363); so it could be a status.
8c74 d4b9 and 8c76 d4b9 don't appear at all.

The smart socket always shows 565a message; values are 00800, 007f00, 007e00, 008100, 008200. So it can't be a status because nothing has changed.
The 414b (power) message appears very few times, even though something is charging.

Tomorrow I'm going to test with the PV switched on.

vannetta:
I've done a complete test, everything matches. The smart socket (on without charge) always shows a 565a message, with different values. I dont'know what can be, but status (on) or total consumption in kWh (10,69 kWh).
These are the records:
[History_power2020-11-11 19-28-05.xlsx](https://github.com/merbanan/rtl_433/files/5529670/History_power2020-11-11.19-28-05.xlsx)
[ecodhome3.xlsx](https://github.com/merbanan/rtl_433/files/5529672/ecodhome3.xlsx)

merbanan:
@vannetta for viewing purposes you can use -F csv. It should be easy to filter out row and columns with that type of formatting.

zuckschwerdt:
10690 Wh should be sent as `29Cx` or (little endian) `Cx29` or with that 0x33 offset as `5CFx` / `Fx5C`. I haven't seen that in the data. You likely need to run this a few month (so 10690 goes to something higher) and watch what happens in the messages.

If that EcoDHOME-Transmitter-0x33-offset encoding is also used in the 71b3 messages then
- Raw data 333363 is 0 (W?)
- Raw data 863363 is 83 (W?)

vannetta:
I don't think that 333363 and 863363 are Watt. No other values are shown in MCEE for transmitters, neither is possible to switch them on or off. It would be possibile that those two message are a status?
About the sockets, I'll try in the to use different charges to move that 10,69 (the third decimal is not shown)

PR #1546, opened 2020-11-01
Fix Home Assistant MQTT auto discovery script naming

rct:
* Fixes issue with missing `subtype` component of MQTT topic per #1510 
* Added command line arguments for most settings
* MQTT username and password can be set in the environment
* Added first pass at documentation/usage info


zuckschwerdt:
This looks great!
You can remove mention of `-M newmodel`, it's default now.

rct:
@zuckschwerdt - was this waiting from something other than the correction to that comment?

PR #1548, opened 2020-11-03
Removed invalid "weather" device_class

soaringswine:
HASS does not have a device_class of "weather" and thus will fail to autodiscover any entities with that set: voluptuous.error.MultipleInvalid: value is not allowed for dictionary value @ data['device_class'] 
Please see related comment in PR #1390 and https://www.home-assistant.io/docs/configuration/customizing-devices/#device-class

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -118,7 +118,6 @@ mappings = {
         "device_type": "sensor",
         "object_suffix": "WS",
         "config": {
-            "device_class": "weather",
             "name": "Wind Speed",
             "unit_of_measurement": "km/h",
             "value_template": "{{ value|float }}"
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
@rct @sotatech

stephenhouser:
Looks like `examples/rtl_433_mqtt_hass.py` has changed since this PR. Would be nice to get this PR included as I have the same problem the `examples/rtl_433_mqtt_hass.py` throws errors about the unknown `device_class`.

Mindavi:
This is the list of (analog) device classes. It does indeed not include weather.

https://www.home-assistant.io/integrations/sensor

stephenhouser:
The commit https://github.com/merbanan/rtl_433/commit/539e2c9fd15a354ecc29f110169bd9296fe0d96e does not remove all the invalid weather class devices.

soaringswine:
> The commit [539e2c9](https://github.com/merbanan/rtl_433/commit/539e2c9fd15a354ecc29f110169bd9296fe0d96e) does not remove all the invalid weather class devices.

not sure how I missed those other instances, apologies!

PR #1549, opened 2020-11-03
Changed wind_km_h to wind_avg_km_h

soaringswine:
Changed wind_km_h to standard wind_speed_km_h for compatibility with scripts and uniformity with other devices.


zuckschwerdt:
Ok, this is only 8 days old, the breaking change won't be too hard on users.

zuckschwerdt:
But `wind_avg_km_h` and `wind_max_km_h` are the usual keys. Please use that.

soaringswine:
Ok, @zuckschwerdt I've updated it to "wind_avg_km_h".

zuckschwerdt:
@mbruski heads up: the wind speed key has changed to the default one used.

PR #1552, opened 2020-11-04
Added support for LaCrosse Technology View TH2 Thermo/Hygro sensor.

mbruski:
Modified src/devices/lacrosse_th3.c to support LTV-TH2.

TH2 differs from the TH3 in the following ways:
   pulse width is 107us (TH3 is 104us) so bitbuffer reports bit_per_row to be ~289 instead of ~256
   CRC for packets with sequence # 0,1,3,4,5 and 7 is CRC-8 poly 0x31 init 0xac
   CRC for packets with sequence # 2 and 6 is CRC-8 poly 0x31 init 0x00 (same as TH3 for all sequence numbers)

Output reflects correct device model number.

### DIFF
--- a/src/devices/lacrosse_th3.c
+++ b/src/devices/lacrosse_th3.c
@@ -1,5 +1,5 @@
 /** @file
-    LaCrosse Technology View LTV-TH3 Thermo/Hygro Sensor.
+    LaCrosse Technology View LTV-TH3 & LTV-TH2 Thermo/Hygro Sensor.
 
     Copyright (C) 2020 Mike Bruski (AJ9X) <michael.bruski@gmail.com>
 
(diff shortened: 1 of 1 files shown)


Mindavi:
Looks good to me. Works with the samples provided in https://github.com/merbanan/rtl_433_tests/pull/369 (if that PR is cleaned up we can merge that one too).

src/devices/lacrosse_th3.c
```diff
@@ -112,14 +134,14 @@ static int lacrosse_th3_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
     data = data_make(
-            "model",            "",                 DATA_STRING, "LaCrosse-TH3",
-            "id",               "Sensor ID",        DATA_FORMAT, "%06x", DATA_INT, id,
-            "seq",              "Sequence",         DATA_INT,     seq,
-            "flags",            "unknown",          DATA_INT,     flags,
-            "temperature_C",    "Temperature",      DATA_FORMAT, "%.1f C",  DATA_DOUBLE, temp_c,
-            "humidity",         "Humidity",         DATA_FORMAT, "%u %%", DATA_INT, humidity,
-            "mic",              "Integrity",        DATA_STRING, "CRC",
-            NULL);
+         "model",            "",                 DATA_FORMAT, "LaCrosse-TH%u", DATA_INT, model,
```
zuckschwerdt, line 137: This isn't great. The data extractor scripts look for fixed strings. It should be `model==3?"LaCrosse-TH3":"LaCrosse-TH2"`

Mindavi: Let's resolve that then. What do you mean with 'data extractor scripts'.

zuckschwerdt: The python script that gathers the json data for https://triq.org/explorer/ -- I guess I need to clean up and commit that then ;)

PR #1553, opened 2020-11-04
Added support for LaCrosse Technology View R1 Rainfall Gauge.

mbruski:
This decoder currently outputs the raw values as sent by the sensor (raw_rain1 and raw_rain2).
Work continues to determine how to calculate rain_mm from these values.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -181,6 +181,7 @@
     DECL(lacrosse_th3) \
     DECL(bresser_6in1) \
     DECL(bresser_7in1) \
+    DECL(lacrosse_r1) \
     /* Add new decoders here. */
 
 #define DECL(name) extern r_device name;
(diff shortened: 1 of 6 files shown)


Mindavi:
Do you think it would be good to merge this already or do you suggest we wait until the rain_mm encoding is figured out? Either way is fine by me, I don't think much people will start depending on the raw rain output (and we could consider keeping them for backwards compat if we really want to).

mbruski:
I setup this decoder for a member on Google groups.   Not having a sensor/console I'm dependent on someone offering me some console reading and corresponding raw values from the decoder so I can figure out how to compute rain_mm.   Figured I'd put the decoder out there and wait for someone to play with it.

Mindavi:
I think it would be good to merge it so that the next person will see the output and open a new issue with (hopefully) some decoded values. Leaving this open for a couple of days so other peope can give their opinion on this.

zuckschwerdt:
LGTM, rain_mm probably needs to be commented out for now, otherwise the compile checks go screaming ;)

PR #1555, opened 2020-11-06
Separate rtl_433_mqtt_hass example code & config, Add Dockerfile

maxlock:
This is a speculative request to make it easier to run the rtl_433_mqtt_hass example without changing the code itself. Additionally moving the mappings out into their own file to allow local customisation. This makes it easier to automate running the code as well and an example Dockerfile is added to show this too. The 2nd commit shows the actual changes better due to git incorrectly marking moves as new files etc.

You get a nice usage message too if required, and if none are provided it sticks to the original hardcoded defaults.

$ ./rtl_433_mqtt_hass.py -h
usage: rtl_433_mqtt_hass.py [-h] [--host HOST] [--port PORT] [--topic TOPIC]
                            [--username USERNAME] [--password PASSWORD]
                            [--prefix PREFIX] [--interval INTERVAL]
                            [--mapfile MAPFILE] [-e]

optional arguments:
  -h, --help           show this help message and exit
  --host HOST          MQTT host
  --port PORT          MQTT port
  --topic TOPIC        MQTT topic
  --username USERNAME  MQTT username
  --password PASSWORD  MQTT password
  --prefix PREFIX      Discovery prefix
  --interval INTERVAL  Discovery interval
  --mapfile MAPFILE    Mapping file
  -e                   configure from environment variables


maxlock:
rct, yes thats really strange, I can't see your code either? I guess we have the same idea though, I'll take a look at yours if/when it gets merged.

rct:
for reference here's the usage/help from PR #1546.  It shouldn't be too hard to merge these, though there are minor differences in option naming.

```
usage: rtl_433_mqtt_hass.py [-h] [-d] [-q] [-u USER] [-P PASSWORD] [-H HOST]
                            [-p PORT] [-r] [-R RTL_TOPIC]
                            [-D DISCOVERY_PREFIX] [-i DISCOVERY_INTERVAL]

Publish Home Assistant MQTT auto discovery topics for rtl_433 devices.

rtl_433_mqtt_hass.py connects to MQTT and subscribes to the rtl_433
event stream that is published to MQTT by rtl_433. The script publishes
additional MQTT topics that can be used by Home Assistant to automatically
discover and minimally configure new devices.

The configuration topics published by this script tell Home Assistant
what MQTT topics to subscribe to in order to receive the data published
as device topics by MQTT.

optional arguments:
  -h, --help            show this help message and exit
  -d, --debug
  -q, --quiet
  -u USER, --user USER  MQTT username
  -P PASSWORD, --password PASSWORD
                        MQTT password
  -H HOST, --host HOST  MQTT hostname to connect to (default: 127.0.0.1)
  -p PORT, --port PORT  MQTT port (default: 1883)
  -r, --retain
  -R RTL_TOPIC, --rtl-topic RTL_TOPIC
                        rtl_433 MQTT event topic to subscribe to (default:
                        rtl_433/+/events)
  -D DISCOVERY_PREFIX, --discovery-prefix DISCOVERY_PREFIX
                        Home Assistant MQTT topic prefix (default: hass433)
  -i DISCOVERY_INTERVAL, --interval DISCOVERY_INTERVAL
                        Interval to republish config topics in seconds
                        (default: 600)


It is strongly recommended to run rtl_433 with "-C si". This script requires
rtl_433 to publish both event messages and device messages.

MQTT Username and Password can be set via the cmdline or passed in the
environment: MQTT_USERNAME and MQTT_PASSWORD.

Prerequisites:

1. rtl_433 running separately publishing events and devices messages to MQTT.

2. Python installation
* Python 3.x preferred.
* Needs Paho-MQTT https://pypi.python.org/pypi/paho-mqtt

  Debian/raspbian:  apt install python3-paho-mqtt
  Or
  pip install paho-mqtt
* Optional for running as a daemon see PEP 3143 - Standard daemon process library
  (use Python 3.x or pip install python-daemon)

Running:

This script can run continually as a daemon, where it will publish
a configuration topic for the device events sent to MQTT by rtl_433
every 10 minutes.

Alternatively if the rtl_433 devices in your environment change infrequently
this script can use the MQTT retain flag to make the configuration topics
persistent. The script will only need to be run when things change or if
the MQTT server loses its retained messages.

Getting rtl_433 devices back after Home Assistant restarts will happen
more quickly if MQTT retain is enabled. Note however that definitions
for any transitient devices/false positives will retained indefinitely.

Suggestions:

Running this script will cause a number of Home Assistant entities (sensors
and binary sensors) to be created. These entities can linger for a while unless
the topic is republished with an empty config string.  To avoid having to
do a lot of clean up When running this initially or debugging, set this
script to publish to a topic other than the one Home Assistant users (homeassistant).

MQTT Explorer (http://http://mqtt-explorer.com/) is a very nice GUI for
working with MQTT. It is free, cross platform, and OSS. The structured
hierarchical view makes it easier to understand what rtl_433 is publishing
and how this script works with Home Assistant.

MQTT Explorer also makes it easy to publish an empty config topic to delete an
entity from Home Assistant.

Known Issues:

Currently there is no white or black lists, so any device that rtl_433 receives
including transients, false positives, will create a bunch of entities in
Home Assistant.

As of 2020-10, Home Assistant MQTT auto discovery doesn't currently support
supplying "friendly name", and "area" key, so some configuration must be
done in Home Assistant.

There is a single global set of field mappings to Home Assistant meta data.
```

examples/rtl_433_mqtt_hass/rtl_433_mqtt_hass.py
```diff
@@ -13,308 +13,54 @@
...
+parser.add_argument("--username", help="MQTT username", default=None, required=False)
+parser.add_argument("--password", help="MQTT password", default=None, required=False)
+parser.add_argument("--prefix", help="Discovery prefix", default="homeassistant", required=False)
+parser.add_argument("--interval", help="Discovery interval", default=600, type=int, required=False)
+parser.add_argument("--mapfile", help="Mapping file", default="example_mappings.json", required=False)
+parser.add_argument("-e", help="configure from environment variables", action="store_true", required=False)
+args = parser.parse_args()
+
+"""Override Arguments with Env vars if required"""
+if args.e:
```
rct, line 37: Processing environment variables after command line arguments means the environment variables take precedent, so they can't be overridden with from the command line.

maxlock:
I think I see whats happened, I wasn't tracking the correct upstream repo. I'm going to kill the pr and I'll base my changes on the correct one. Thanks for the feedback though @rct, that's a reasonable observation regarding config precedence.

zuckschwerdt:
If you kept the branch you can just reset it to new commits.

maxlock:
Yup I'm going to rebase my feature branch from master and sort the conflicts :)

PR #1556, opened 2020-11-08
Improve unit tests for bitbuffer with extra assertions

Mindavi:
This should ensure that the bitbuffer tests will actually return non-zero if anything fails. Previously this wasn't the case, because the assert.h header doesn't do anything when building in release mode.

Main discussion point for me: is this how we want to structure unit tests? I tried to use the same sort of method as used in `util.c`, `optparse.c`, but with some slight changes (e.g. I don't think it's very useful to log the amount of passed tests, just relying on the 'failed' count and the return code is enough).

### DIFF
--- a/src/bitbuffer.c
+++ b/src/bitbuffer.c
@@ -413,29 +413,46 @@ int bitbuffer_find_repeated_row(bitbuffer_t *bits, unsigned min_repeats, unsigne
 
 // Unit testing
 #ifdef _TEST
-#include <assert.h>
+
+#define ASSERT(expr) \
+    do { \
+        if (expr) { \
+            ++passed; \
+        } else { \
+            ++failed; \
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
All the test output is usually suppressed with `ctest`. Using `make CTEST_OUTPUT_ON_FAILURE=1 test` we can get those fail and summary lines:
```
4/8 Test #4: optparse_test ....................***Failed    0.06 sec
optparse:: atouint32_metric
optparse:: atoi_time
FAIL: 2 <> 99
optparse:: test (34/35) passed, (1) failed.
```
I think the group headers are nice, or we could put more info on each assert.
The detailed summary is also quite useful -- the ctest fail/pass summary isn't aware of individual tests.

Mindavi:
I've updated it with logging of pass/fail, I think it's fine to add it and it makes for better consistency between the tests.

zuckschwerdt:
Looks great!

PR #1559, opened 2020-11-11
rtl_433_mqtt_hass, take config from env vars & externalise mappings.

maxlock:
This is a successor to the previous pull I cancelled. It allows you to configure rtl_433_mqtt_hass from environment vars, which can be overridden by cli args. Additionally I've moved the mappings out to a file so a user can maintain their own custom mappings more easily. I've added a Dockerfile that I use as it's pretty useful, and takes advantage of the environment vars. Lastly I've moved everything into it's own subfolder for clarity.


maxlock:
I forgot to add that there is a small breaking change in that the --user argument is changed to --username to match --password which isn't abbreviated either.

klattimer:
This pull request seems to change the entire file - so I have a conflict with it here https://github.com/merbanan/rtl_433/pull/1665

pacc:
https://github.com/pbkhrv/rtl_433-hass-addons has a docker file that runs this as an addon in homeassistant / hassio
but I'm more interested in remote controls and alarms so I'll look into klattimers patch first.

merbanan:
Someone else will have to maintain this. I dont really know what this does and it impact.

PR #1562, opened 2020-11-14
Feature/rfraw testing mode

Mindavi:
Based on #1263.

I'd like this feature to also test if the correct values are decoded, but this already helps a lot with checking if the decoding still works properly (return code will be 1 when one of the decodes fails).

Let's discuss if this looks good and if so, I can add the remaining patterns.

I think we should in general add 1 rfraw buffer per device we support, unless we want to test some edge cases.


zuckschwerdt:
Looks good. Maybe long lines should be split into joined literals (`"foo" "bar"`), not sure.

We have more meta data on the decoders and it also might be nice to annotate the test data. I've tested to just parse the decoder doc comment and include tests there. Works fine to generate basically a header file with a big table of things. I would prefer that to get all-the-meta conveniently and extensible. But I did that in Python and I guess if this should be a resource that's generated on compile it would need to be C -- with RE libs and all that chore. Or maybe if there is no Python available skip that feature and assume the platform (embedded, Windows) won't use it anyway?

Mindavi:
@zuckschwerdt Could you open an issue with the idea(s)? That might be helpful to have a discussion about how we want to go about implementing some more structured data. If we want to implement something like that in pure C some strict guidance on how to write the comments might be good enough -- in that case regex is not required as the format is easy to parse. But let's have that discussion in an issue to form a good idea of how this could be done and what you envision with 'metadata'.

Is this good to go with some line length fixes otherwise? We can think and implement further metadata for this in the future, I think.

Mindavi:
Still not fully sure if we should add this to rtl_433 though, we could make a small script and a folder with files that does the same. That would also help keeping the rfraw and the expected output together for regression testing.

Mindavi:
I think using the free-text documentation in the files is a better way to go about this. Maybe it would be good to generate a header using the maintainer_update script? We could then check it in to this repository, in a 'generated' folder or something like that.

Creating it during the build is fine by me too. We could indeed decide to only generate it when python is installed / available. I'd be fine with that as well.

merbanan:
This should be in the rtl_433_tests repo. Just add text files with bit streams. In theory it could be just one huge file with lots of bitstreams. We should not add things to the actual decoders.

I think the way forward is to let loose a LLM to data mine the project for bitstreams and build a test suit based on that. If some one builds a test suite around that I am all for it as it can scale like a 1000 times more then the cu8 rf recordings.

PR #1575, opened 2020-12-05
Add missing UV index and light intensity readings for Cotech

andrewpile:
The PR adds UV and light readings to [Cotech](https://github.com/merbanan/rtl_433/blob/master/src/devices/cotech_36_7959.c)-based devices.

I have a [WeatherSense WeatherRack2](https://shop.switchdoc.com/products/wireless-weatherrack2) which is compatible with the Cotech. I derived the UV and lux fields from [this rather rough fork from SwitchLabs](https://github.com/switchdoclabs/rtl_433/commit/b8dfd38f6702f07363011cfb2ea575919f40ebef).

### DIFF
--- a/src/devices/cotech_36_7959.c
+++ b/src/devices/cotech_36_7959.c
@@ -20,7 +20,7 @@ Integrity check is done using CRC8 using poly=0x31  init=0xc0
 
 Message layout
 
-    AAAA BBBBBBBB C D E F GGGGGGGG HHHHHHHH IIIIIIII JJJJ KKKKKKKKKKKK LLLL MMMMMMMMMMMM NNNNNNNN OOOOOOOOOOOOOOOOOOOOOOOO XXXXXXXX
+    AAAA BBBBBBBB C D E F GGGGGGGG HHHHHHHH IIIIIIII JJJJ KKKKKKKKKKKK LLLL MMMMMMMMMMMM NNNNNNNN OOOOOOOOOOOOOOOO PPPPPPPP XXXXXXXX
 
 - A : 4 bit: ?? Type code? part of id?, never seems to change
 - B : 8 bit: Id, changes when reset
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Looks good, thanks! We'll test and merge soon.
Interesting about the fork, technically there is a need to GPL(compatible)-licence the addition (no linking of proprietary code to GPL allowed) -- but I guess a liberal licence was their implicit intention :)

andrewpile:
Yeah I also wasn't a fan of them recommending a fork that's clearly never going to get upstream fixes. I figured this PR was cleaner. Hopefully this doesn't break any other clones of this weather station-- based on what I can tell this sensor appears to exist on all of them, it was just undocumented.

Mindavi:
I had a quick look. As these are only additions, worst case could be that someone has a system that doesn't give meaningful values. This shouldn't break anything.

Have not compiled / tested this.

Legacyman:
I had a quick test (Cotech FT0205/FT020), and the light intensity and UV are to high (in the middle of the night)
rtl_433 version 20.11-5-g67ee1d69 branch master at 202012081410 inputs file rtl_tcp RTL-SDR
time      : 2020-12-12 01:51:00
model     : Cotech-367959 ID        : 233
Battery   : 1            Temperature: 2.4 C        Humidity  : 87 %          Rain      : 16.8 mm       Wind direction: 298       Wind      : 0.1 m/s       Gust      : 0.3 m/s
Light Intensity: 69627 lux                         UV Index  : 251           Integrity : CRC

I cannot provide more information at the moment, but if I can provide data, please let me know, and I'll find time for it.

Mindavi:
The linked website says 1 to 15+ for UV index, 0 to 200klux for the light intensity. Not sure about the lux, but it seems rather high. The UV index is very much out of the bounds that are set on the website.

andrewpile:
Hm, I'm in the middle of the night now and the Lux and UV are both 0. During the day the lux values seem pretty sane but the UV index does bounce around quite a lot and jump as high as 20, which I believe is wrong. Possibly the UV index is scaled 10x?

As I mentioned in the PR this is ported from a fork: https://github.com/switchdoclabs/rtl_433 Maybe someone with a sharper eye for bitwise ops can figure out what's going on.

FWIW, I got the 200klux range from here: https://www.switchdoc.com/wp-content/uploads/2020/10/WeatherRack2Installation1.2.pdf

Issue #1578, opened 2020-12-12
Acurite/Chaney 515 wireless refrigerator & freezer thermometer support


osenft:
I just got a new pair of Acurite 00515 fridge / freezer thermometers.

`rtl_433 -A` shows some data. I recorded the received packets and what the display shows and ended up with the following list (displayed temperature in F is appended at each line):

Freezer:
```
[00] {50} 39 64 36 f3 d2 9c c0 : 00111001 01100100 00110110 11110011 11010010 10011100 11 | 10.1
[00] {50} 39 64 36 f3 44 0e c0 : 00111001 01100100 00110110 11110011 01000100 00001110 11 | 11.5
[00] {50} 39 64 36 f3 3f 09 c0 : 00111001 01100100 00110110 11110011 00111111 00001001 11 | 12.0
[00] {50} 39 64 36 f3 3a 04 c0 : 00111001 01100100 00110110 11110011 00111010 00000100 11 | 12.5
[00] {50} 39 64 36 f3 b7 81 c0 : 00111001 01100100 00110110 11110011 10110111 10000001 11 | 12.8
[00] {50} 39 64 36 f3 39 03 c0 : 00111001 01100100 00110110 11110011 00111001 00000011 11 | 12.6
[00] {50} 39 64 36 f3 44 0e c0 : 00111001 01100100 00110110 11110011 01000100 00001110 11 | 11.5
[00] {50} 39 64 36 f3 d2 9c c0 : 00111001 01100100 00110110 11110011 11010010 10011100 11 | 10.1
[00] {50} 39 64 36 f3 e1 ab c0 : 00111001 01100100 00110110 11110011 11100001 10101011 11 |  8.6
[00] {50} 39 64 36 f3 ee b8 c0 : 00111001 01100100 00110110 11110011 11101110 10111000 11 |  7.3
[00] {50} 39 64 36 f3 7d 47 c0 : 00111001 01100100 00110110 11110011 01111101 01000111 11 |  5.8
[00] {50} 39 64 36 74 09 54 c0 : 00111001 01100100 00110110 01110100 00001001 01010100 11 |  4.6
[00] {50} 39 64 36 74 96 e1 c0 : 00111001 01100100 00110110 01110100 10010110 11100001 11 |  3.3
[00] {50} 39 64 36 74 1e 69 c0 : 00111001 01100100 00110110 01110100 00011110 01101001 11 |  2.5
[...]

PR #1579, opened 2020-12-12
Add support for Acurite 515 fridge/freezer sensors (#1578).

osenft:


src/devices/acurite.c
```diff
@@ -745,6 +748,40 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+            // 14 bits available after removing both parity bits.
+            int temp_raw = ((bb[3] & 0x7F) << 7) | (bb[4] & 0x7F);
+            tempf = (temp_raw - 1480) * 0.1f;
+            sprintf(channel_str, "%c%c", channel, sensor_type);
+            // Battery status is the 7th bit 0x40. 1 = normal, 0 = low
+            battery_low = (bb[2] & 0x40) == 0;
+
+            /* clang-format off */
+            data = data_make(
+                    "model",                "",             DATA_STRING, _X("Acurite-515","Acurite 515 sensor"),
```
zuckschwerdt, line 772: Don't use the legacy _X on new code.

src/devices/acurite.c
```diff
@@ -745,6 +748,40 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+                    "id",                   "",             DATA_INT,    sensor_id,
+                    "channel",              NULL,           DATA_STRING, &channel_str,
+                    _X("battery_ok","battery_low"), "",     DATA_INT,    _X(!battery_low,battery_low),
+                    "temperature_F",        "Temperature",  DATA_FORMAT, "%.1f F", DATA_DOUBLE, tempf,
+                    "mic",                  "Integrity",    DATA_STRING, "CHECKSUM",
```
zuckschwerdt, line 777: A check for the checksum is needed. Otherwise remove the mic indication.

zuckschwerdt: Ok, well the check is a few lines above this in the general case. Ignore my comment ;)

src/devices/acurite.c
```diff
@@ -745,6 +748,40 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             valid++;
         }
 
+        // 515 sensor messages are 6 bytes.
+        if (browlen == ACURITE_515_BITLEN / 8) {
+        // Sensor type is determined by bit 0 from the message_type.
+        // 0 = refrigerator, 1 = freezer
+        sensor_type = message_type & 1 ? 'F' : 'R';
```
rct, line 755: If it appears that message_types 8 and 9 have essentially been assigned to this senor, then I recommend testing all 6 bits of the message type to avoid false positives.

Add #defines with the other message types for 8 and 9.

src/devices/acurite.c
osenft: It's checked further up. So when it gets here, it's a valid checksum.

src/devices/acurite.c
osenft: Fixed.

src/devices/acurite.c
osenft: Fixed.

glenniana18:
Hi there!

When trying to troubleshoot my Acurite 515 sensors, I noticed that my latest version of rtl_433 (next) is outputting the temp in deg C.  This thread says it should be in deg F.

I am not sure if something got messed up, or if I'm not understanding something.  Or maybe this is the wrong place to post this. 
 Here's an output of my rtl_433:

image

zuckschwerdt:
The Acurite 515's temperature is read as Fahrenheit, but rtl_433 can convert units. You likely use the default conf file with `convert si` in it or a command line option of `-C si`.

glenniana18:
Thank you!  I did not know this, and you’re right, I removed “convert si”
and now it reports in degF!  I appreciate your help.

PR #1580, opened 2020-12-14
Add support for Blueline PowerCost Monitor

jbrzozoski:
This is still a work in progress, but it is nominally working and I would like other people's feedback.
Open questions and things to do:
- Do decoder callback functions in rtl_433 need to support multi-threaded re-entrancy?  I use some function-scoped static variables that could be problematic if so.
- Is the ID guessing algorithm too much processing?  It will trigger 16000 2-byte CRC checks any time a message comes in with a bad CRC, to help find what the proper ID should be.  There is a pre-check that it is looking for a row of the right length with the right first byte before that, so it shouldn't be hit too often, in my opinion.
- Why don't any of the temperature scaling equations others have provided match up with what I'm correlating between my radio data and LCD display?  I haven't put any in just yet and want to figure it out before I do.
- I need to validate the ID message and the battery status flag work as behaved.  I just copied them from old code but haven't tested them yet.

### DIFF
--- a/src/devices/blueline.c
+++ b/src/devices/blueline.c
@@ -1,0 +1,383 @@
+/** @file
+ *  BlueLine Innovations Power Cost Monitor, tested with BLI-28000
+
+    Copyright (C) 2020 Justin Brzozoski
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+ */
+
+/**
+ * Blueline PowerCost Monitor protocol
+ *
+ * Much of the groundwork for this implementation was based on reading the source and notes from older
+ * implementations, but this implementation was a fresh rewrite by Justin Brzozoski in 2020.  I would
+ * not have been able to figure this out without the other implementations to look at, but I wanted an
+ * implementation that didn't need to know the Kh factor or monitor ID ahead of time, which required
+ * changes.
+ *
+ * Some references used include:
+ *
+ * https://github.com/merbanan/rtl_433/pull/38 - an abandoned pull request on rtl_433 by radredgreen
+ *
+ * https://github.com/CapnBry/Powermon433 - a standalone Arduino-based Blueline monitor
+ *
+ * http://scruss.com/blog/2013/12/03/blueline-black-decker-power-monitor-rf-packets/ - the blog post
+ * where the other authors were trading notes in the comments
+ *
+ * The IR-reader/sensor will transmit 3 bursts every ~30 seconds.  The low-level encoding is on/off
+ * keyed pulse-position modulation (OOK_PPM).  The on pulses are always 0.5ms, while the off pulses
+ * are either 0.5ms for logic 1 or 1.0ms for logic 0.  Each burst is 32 bits long.  The pauses
+ * between the 3 grouped bursts is roughly 100ms.
+ *
+ * Data is sent less significant byte first for multi-byte fields.
+ *
+ * The basic layout of all bursts is as follows:
+ *
+ * - First is a 1 byte header, which is always the value 0xFE.
+ * - Second is a 2 byte payload, which is interpreted differently based on the two lowest bits of the first byte
+ * - Finally is a 1 byte CRC, calculated across the 2 payload bytes (not the header)
+ *
+ * The CRC is a CRC-8-ATM with polynomial 100000111, but it may be required to modify the payload bytes before
+ * calculating it depending on message type.
+ *
+ * There are 4 message types that can be indicated by the 2 lowest bits of the first payload byte:
+ * 0 - ID message (payload is not offset)
+ * 1 - power message (payload is offset)
+ * 2 - temperature/status message (payload is offset)
+ * 3 - energy message (payload is offset)
+ *
+ * For the ID message (0), the CRC can be calculated directly on the payload as sent, and when the payload is
+ * interpreted as a 16-bit integer it gives the ID of the transmitter.  This message is sent when the
+ * monitor is first powered on and if the button on the monitor is pressed briefly.  If the button on the
+ * monitor is held for >10 seconds, the monitor will change it's ID and report the new one.
+ *
+ * While the transmitter ID's are 16-bit, none of them can have any of the two lowest bits set or they
+ * would not be able to transmit their ID as message type 0, and when the offset is calculated (see below) they
+ * would also change the message type.
+ *
+ * For the 3 other message types, the payload must be offset before calculating the CRC or
+ * interpreting the data.  The offset is done by treating the whole payload as a single 16-bit integer and
+ * then subtracting the ID of the transmitter.  After the offset is done, then the CRC may be calculated
+ * and the payload may be interpreted.
+ *
+ * Note that if the transmitter's ID isn't known, the code can't easily determine if messages other than an
+ * ID payload are good or bad, and can't intepret their data correctly.  However, if the BLUELINE_USE_ID_TRANSMISSION
+ * or BLUELINE_USE_ID_AUTODETECT features are enabled, the system can learn the transmitter's ID by various methods.
+ *
+ * For the power message (1), the offset payload gives the number of milliseconds gap between impulses for the most
+ * recent impulses seen by the monitor.  To convert from this 'gap' to kilowatts, you will need your meter's
+ * Kh value. The Kh value is written obviously on the front of most meters, and 1.0 and 7.2 are very common.
+ *
+ * kW = (3600/gap) * Kh
+ *
+ * For the temperature message (2), the offset payload gives the temperature in an odd scaling in the last byte,
+ * and has some flag bits in the first byte.  The only known flag bit is the battery.  rtl_433 handles scaling back to
+ * degrees celsius automatically.
+ *
+ * For the energy message (3), the offset payload contains a continuously running power impulse accumulator. I'm
+ * not sure if there is a way to reset the accumulator.  The intended way to use it is to remember the accumulator
+ * value at the beginning of a time period, and then subtract that from the value at the end of the time period.  The
+ * accumulator will roll over to 0 after 65535.
+ *
+ * kWh = 0.001 * (accumulated pulses) * Kh
+ *
+ * Since the Kh value on all meters can vary, we do not handle it in rtl_433 and just report the raw millisecond
+ * gap and accumulated impulses as received directly from the monitor.
+ *
+ * USAGE HINTS:
+ *
+ * The requirement of knowing the ID before being able to receive a message means that this decoder
+ * will generally require a parameter to be useful.  When running in the default mode with no parameters,
+ * the only message it is able to decode is the one that announces a monitor's ID.  So, assuming you can get to
+ * the monitor to power cycle it or hit the button, this is the recommended method:
+ *
+ * 1) Start rtl_433
+ * 2) Tap the button or power cycle the monitor
+ * 3) Look for the rtl_433 output indicating the BlueLine monitor ID and note the ID field
+ * 4) Stop rtl_433
+ * 5) Restart rtl_433, explicitly passing the ID as a parameter to this decoder
+ *
+ * For example, if you see the ID 45364 in step 3, you would start the decoder with a command like:
+ * rtl_433 -R 176:45364
+ *
+ * If you are unable to access the monitor to have it send the ID message, you can also use the "auto" parameter:
+ * rtl_433 -R 176:auto -v
+ *
+ * Verbose mode is recommended to tell what the "auto" mode is doing.
+ *
+ * The auto parameter will try to brute-force the ID on any messages that look like they are from a
+ * BlueLine monitor.  This method usually succeeds within a few minutes, but is likely to get false positives
+ * if there is more than one monitor in range or the messages being received are all identical (i.e. if the
+ * meter is continuously reporting 0 watts).  If it succeeds, it will start reporting data with the new ID,
+ * which you should then use as a parameter when you re-run rtl_433 in the future.
+ *
+ */
+
+#include <stdlib.h>
+#include "fatal.h"
+#include "decoder.h"
+
+#define BLUELINE_MODEL "Blueline PowerCost Monitor"
+#define BLUELINE_BITLEN      32
+#define BLUELINE_STARTBYTE   0xFE
+#define BLUELINE_CRC_POLY    0x07
+#define BLUELINE_CRC_INIT    0x00
+#define BLUELINE_CRC_BYTELEN 2
+#define BLUELINE_TXID_MSG        0x00
+#define BLUELINE_POWER_MSG       0x01
+#define BLUELINE_TEMPERATURE_MSG 0x02
+#define BLUELINE_ENERGY_MSG      0x03
+
+#define BLUELINE_ID_STEP_SIZE 4
+#define MAX_POSSIBLE_BLUELINE_IDS (65536/BLUELINE_ID_STEP_SIZE)
+#define BLUELINE_ID_GUESS_THRESHOLD 10
+
+struct blueline_stateful_context {
+    unsigned id_guess_hits[MAX_POSSIBLE_BLUELINE_IDS];
+    uint16_t current_sensor_id;
+    unsigned searching_for_new_id;
+};
+
+static uint16_t guess_blueline_id(r_device *decoder, const uint8_t *current_row)
+{
+    struct blueline_stateful_context *const context = decoder->decode_ctx;
+    const uint16_t start_value = ((current_row[2] << 8) | current_row[1]);
+    const uint8_t recv_crc = current_row[3];
+    uint16_t working_value;
+    uint8_t working_buffer[2];
+    uint16_t best_id;
+    unsigned best_hits;
+    unsigned high_byte_steps;
+    unsigned low_byte_steps;
+
+    // TL;DR - Try all possible IDs against every incoming message, and count how many times each one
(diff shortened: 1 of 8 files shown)


bkenobi:
The battery corroded in the monitor and I'm having trouble getting it back up and running. If I can get it to connect to my display unit, I'll test any code you share.

zuckschwerdt:
Quick notes: yes, decoders must not be stateful. I need to look into this in detail later, but we really don't want any static state in decoders.

src/devices/blueline.c
```diff
@@ -0,0 +1,330 @@
...
+ * kW = (3600/gap) * Kh
+ *
+ * For the temperature message (2), the offset payload gives the temperature in an odd scaling in the last byte,
+ * and has some flag bits in the first byte.  The only known flag bit is the battery.
+ *
+ * For the energy message (3), the offset payload contains a continuously running power impulse accumulator. I'm
+ * not sure if there is a way to reset the accumulator.  The intended way to use it is to remember the accumulator
+ * value at the beginning of a time period, and then subtract that from the value at the end of the time period.
+ *
+ * kWh = 0.001 * (accumulated pulses) * Kh
```
merbanan, line 83: kh = kilo hour?

src/devices/blueline.c
```diff
@@ -0,0 +1,330 @@
...
+// If you know your ID and don't want this code to watch anything else,
+// set that and change both of the following defines to 0.
+#define BLUELINE_DEFAULT_ID 0xB134
+// If BLUELINE_USE_ID_TRANSMISSION is set to 1, this code will jump to monitoring a new ID
+// when someone hits the button on the transmitter.  The code will still REPORT the TXID
+// messages as data even if this is 0, it just won't switch to monitoring that ID.
+// If BLUELINE_ACTIVE_ID_AUTODETECT is set to 1, this code will try to autodetect IDs from
+// incoming messages which otherwise would fail CRC.  While there are a lot of false positives
+// on each individual message, only one ID will continuously match up, and it becomes
+// obvious which one is the right one within a few minutes.
```
merbanan, line 100: How many matches do you get? Is there a way to validate the payload parameters?

merbanan:
@jbrzozoski regarding recovering the correct id, if it still is almost real time on most platforms it is ok. But you must be as sure as possible before we try to recover the id. The sparsnas decoder uses something similar.

merbanan:
Regarding the temperature. If you post a set of readings Im sure we can figure it out.

merbanan:
I now remember what I thought initially. Each transmission contains 3 payloads that might be different. I think that if all payloads are used then it should be possible to recover a unique id. But that should not be the case based on comments I've read. So I think we need to pass the id by command line to the decoder. Or accept that this decoder needs a state.

The raw logic is that we got a 14 bit id and only 8 bits of crc to match it against. (And the crc is supposed to validate the payload, not to be used to recover the id, a broken payload will recover a bad id).

jbrzozoski:
Within any given burst, all 3 messages are often the same message repeated, so it's not practical to expect to be able to recover the ID from one burst.  Your summary that we need to either pass the ID from the command line or maintain a state was my conclusion as well, but I couldn't find any examples of other devices with custom command line options, so I tried to avoid that.

src/devices/blueline.c
jbrzozoski: That's just the standard name of the constant, and how it is labeled on most meters: https://ctlsys.com/support/kh_watthour_constant/

src/devices/blueline.c
jbrzozoski: Any individual message will get 2^6 matches.  The payload can unfortunately be any value and there's no way to narrow it down by looking for reserved or unused fields.

merbanan:
https://github.com/merbanan/rtl_433/blob/master/src/devices/fineoffset.c#L20

I suggest you add one option that sets the key and another option to brute force the key (maybe you can not reset the transmitter). That option will make the decoder keeping a state. I think that is an ok compromise. By default I prefer things to be stateless. @zuckschwerdt I think this is a good enough compromise.

A user could enable to option and things would just work. If anyone requests support the answer will be that they need to figure out the key for things to work. And for regression tests we can either add the key to the command line or just handle the message with a proper crc.

jbrzozoski:
I like the two options plan. A user can interactively use the option to brute force the key, and within a couple minutes rtl_433 should start reporting data including the ID key that it found.  Then the user can stop that interactive session and just pass in the discovered key on the command line for all future use.  

Not having the stateful key search running all the time avoids some weird issues that could crop up if there happens to be two Blueline transmitters within range, or a single rtl_433 session running so long that random noise eventually looks like a valid transmitter.  The brute force search mechanism has no concept of time. If enough random messages all match a key over several days without restarting rtl_433, it would consider that a match and switch to a bogus ID, which would then switch back quickly if a real transmitter is in range.

jbrzozoski:
Slight tweaking to behavior... 

If you don't pass any parameters to this decoder, all it will ever do is spit out the TXID message if it hears someone restart a monitor.  And I suppose if you just happened to have a monitor ID of 0 it would work, too, but that's unlikely.

You can pass in the special parameter "auto" which will put it in auto-detect mode *until* the first time it sees a TXID message or the brute-force passes the threshold, at which point it will switch to using that ID and turn off the autodetect. (To avoid any of the weird issues I mentioned earlier in this thread.)

Finally, any other parameter passed in is assumed to be an ID, and just passed through strtoul.

I like this since it actually can support multiple transmitters in range if you just pass multiple registrations on the command line like `-R 176:ABCD -R 176:5432`.  :+1: 

This kind of what you were thinking, @merbanan ?

src/devices/blueline.c
```diff
@@ -0,0 +1,356 @@
...
+                    best_hits = context->id_guess_hits[(working_value/BLUELINE_ID_STEP_SIZE)];
+                    best_id = working_value;
+                }
+            }
+            working_buffer[0] += BLUELINE_ID_STEP_SIZE;
+        }
+        working_buffer[1] += 1;
+    }
+
+    fprintf(stderr, "Attempting Blueline autodetect: best_hits=%u\n", best_hits);
```
merbanan, line 161: Needs to be behind verbosity check.

src/devices/blueline.c
```diff
@@ -0,0 +1,356 @@
...
+ * kWh = 0.001 * (accumulated pulses) * Kh
+ *
+ * Since the Kh value on all meters can vary, we do not handle it in rtl_433 and just report the raw millisecond
+ * gap and accumulated impulses as received over from the monitor values.
+ *
+ */
+
+// TODO - Add comments for the new parameters and expected usage
+
+#include <stdbool.h>
```
merbanan, line 92: No stdbool.h please.

src/devices/blueline.c
```diff
@@ -0,0 +1,356 @@
...
+#define BLUELINE_TXID_MSG        0x00
+#define BLUELINE_POWER_MSG       0x01
+#define BLUELINE_TEMPERATURE_MSG 0x02
+#define BLUELINE_ENERGY_MSG      0x03
+
+#define BLUELINE_ID_STEP_SIZE 4
+#define MAX_POSSIBLE_BLUELINE_IDS (65536/BLUELINE_ID_STEP_SIZE)
+#define BLUELINE_ID_GUESS_THRESHOLD 10
+
+struct blueline_context {
```
merbanan, line 116: Name it blueline_statefull_context, the it will be very clear what this is.

src/devices/blueline.c
```diff
@@ -0,0 +1,356 @@
...
+            data = data_make(
+                    "model", "", DATA_STRING, "Blueline",
+                    "id",    "", DATA_FORMAT, "%04x", DATA_INT, received_sensor_id,
+                    "mic",   "", DATA_STRING, "CRC",
+                    NULL);
+            /* clang-format on */
+            decoder_output_data(decoder, data);
+            payloads_decoded++;
+            if (context->searching_for_new_id) {
+                fprintf(stderr,"Switching to received Blueline ID 0x%04X\n", received_sensor_id);
```
merbanan, line 236: Verbose guard missing.

src/devices/blueline.c
```diff
@@ -0,0 +1,356 @@
...
+            offset_payload_u8[1] = (offset_payload_u16 >> 8);
+            calc_crc = crc8(&offset_payload_u8[0], BLUELINE_CRC_BYTELEN, BLUELINE_CRC_POLY, BLUELINE_CRC_INIT);
+        }
+
+        // If the CRC didn't match up, ignore this row!
+        if (calc_crc != recv_crc) {
+            if ((context->searching_for_new_id) && (message_type != BLUELINE_TXID_MSG)) {
+                uint16_t id_guess = guess_blueline_id(context, current_row);
+                if (id_guess != 0) {
+                    fprintf(stderr,"Switching to auto-detected Blueline ID 0x%04X\n", id_guess);
```
merbanan, line 212: Verbose guard missing.

src/devices/blueline.c
```diff
@@ -0,0 +1,356 @@
...
+        return NULL; // NOTE: returns NULL on alloc failure.
+    }
+    memset(context,0,sizeof(*context));
+    r_dev->decode_ctx = context;
+
+    if (arg != NULL) {
+        if (strcmp(arg, "auto") == 0) {
+            // Setup for auto identification
+            context->searching_for_new_id = true;
+            fprintf(stderr, "Blueline decoder will try to autodetect ID.\n");
```
merbanan, line 333: Verbose guard missing.

src/devices/blueline.c
```diff
@@ -0,0 +1,356 @@
...
+        } else {
+            // Assume user is trying to pass in hex ID
+            context->current_sensor_id = strtoul(arg, NULL, 16);
+            fprintf(stderr, "Blueline decoder using ID %04X\n", context->current_sensor_id);
```
merbanan, line 337: Verbose guard missing.

merbanan:
> 
> I like this since it actually can support multiple transmitters in range if you just pass multiple registrations on the command line like `-R 176:ABCD -R 176:5432`. 
> 
> All state is setup at creation time except when "auto" mode is enabled, and once that picks an ID it also stops changing state.

So the command -R 176:auto would enable the brute force mode?

bkenobi:
Since my Blueline meter is deadline been looking for a potential replacement. Will the Black & Decker EM100B work with this library? The device looks identical. I will look for the fccid to confirm it's the same.

bkenobi:
The best I can tell, the Black & Decker version of this unit is the EM100B with an FCCID of BLI-00239.  The Blue Line Innovations unit has a FCCID of SUE-PCMTX02 for this model but others have the BLI grantee code included as well.  It seems clear that Blue Line was the mfg of the B&D unit as well.  It is not confirmed that the transmission is compatible but it seems highly likely that Blue Line would not reinvent the protocol to sell to a different distributor.

bkenobi:
I received confirmation from BLI that the Black & Decker EM100B that looks identical to this unit i functionally identical and should be compatible with this library.  When this work is completed, it should be noted that it is also usable.

src/devices/blueline.c
jbrzozoski: I'm not seeing any way to determine if we're in verbose mode inside the creator functions.

jbrzozoski:
Just some minor tweaks in the right direction:
- added most of the verbosity guards as requested
- switched to displaying/expecting the ID in decimal (instead of hex) to be consistent with MQTT payloads/topics
- fixed a mismatch for the model displayed on TXID messages

And to answer your question, using `-R 176:auto` would enabled the autodetect mode.

jbrzozoski:
(Just pushing to fix a style check)

src/devices/blueline.c
jbrzozoski: If you'd like, I can add a commit into this pull request to add more parameters to the create_fn, either the whole of the r_cfg_t or just the verbosity value from it.  I'll adjust fineoffset.c to match.

merbanan: Remove the output for now. Lets focus on getting this patch in first. Then we can add it later.

src/devices/blueline.c
```diff
@@ -0,0 +1,362 @@
...
+            // TODO - Confirm battery flag is working properly
+
+            // These seem to be the best estimates from Powermon433,
+            // but they don't line up with what my official LCD display shows.
+            //
+            // I'm leaving the display raw while I figure this out better...
+            //
+            // A: deg_f = 0.823 * recvd_temp - 28.63
+            // B: deg_c = 0.457 * recvd_temp - 33.68
+
```
merbanan, line 269: We need to figure out the temperature encoding.

src/devices/blueline.c
```diff
@@ -0,0 +1,362 @@
...
+// TODO - Add comments for the new parameters and expected usage
+
+#include <stdlib.h>
+#include "fatal.h"
+#include "decoder.h"
+
+#ifndef UNUSED
+#define UNUSED(x) (void)(x)
+#endif
```
merbanan, line 98: This looks unused.

src/devices/blueline.c
```diff
@@ -0,0 +1,362 @@
...
+    return r_dev;
+}
+
+r_device blueline = {
+        .name        = "BlueLine Power Monitor",
+        .modulation  = OOK_PULSE_PPM,
+        .short_width = 500,
+        .long_width  = 1000,
+        .gap_limit   = 2000,
+        .reset_limit = 200000,
```
merbanan, line 356: I think the practical limit for this is much smaller.

merbanan:
Get a good set of sample signals and submit it to the rtl_433_tests repo. Then we try to figure out the temperature encoding and then I think we are set. The code looks fine to me. 

@zuckschwerdt ok to merge after those fixes?

src/devices/blueline.c
merbanan: Maybe 8000 would work?

src/devices/blueline.c
zuckschwerdt: Please don't access r_cfg from decoders -- we need some modularity.

src/devices/blueline.c
jbrzozoski: The one from Powerman433 is close enough.  I'm just being nit-picky perfectionist and trying to figure out why it's sometimes a degree or two off from the official LCD display.  I'll put it back in for now and continue experimenting on my own.

src/devices/blueline.c
jbrzozoski: Of course.  That's what it is.  :rofl: 
(Yes, I'll remove it.)

src/devices/blueline.c
jbrzozoski: It's almost a don't-care then, as long as it's bigger than gap_limit.  The pauses between rows in one burst is actually closer to 100000, so they always show up in the decoder one row at a time.  I'll bring this down to 8000.

jbrzozoski:
I'll submit some samples later today after I can head outside and hit the button a few times to make sure I have samples of the ID message.

jbrzozoski:
Samples are up:

https://github.com/merbanan/rtl_433_tests/pull/380

Mindavi:
I don't know why, but none of the verbose logging seems to be working. I guess this has something to do with rtl_433 internals? Otherwise this seems to work well with the submitted samples. Have not thoroughly reviewed it, but the code looks good in general.

jbrzozoski:
I noticed the verbose output wasn't working when I was testing the recorded
samples, but assumed it was somehow a side affect of the replay mechanism.
I just double checked on my receiver and it doesn't work there either. I'll
take a closer look in the next 5-7 hours.

jbrzozoski:
Verbosity and registrations are command line order dependent:

- Running `rtl_433 -vvv -R 23 -R 176` gives both decoders verbosity level 3.
- Running `rtl_433 -R 23 -vvv -R 176` gives decoder 23 verbosity level 0, and decoder 176 verbosity level 3.

@Mindavi do you know which order you had your command line setup?

jbrzozoski:
Is there no way to pass a parameter to one decoder without disabling all the other?
I had thought the options `-R 0 -R 176:12345` were the recommended way to do this, but now I'm seeing that just using `-R 176:12345` disables all the other decoders as well.  Is there a practical way to use this decoder with an option along with all the other default decoders?

zuckschwerdt:
Not really recommended but `-R -1 -R 1 -R 176:12345`does the trick.

Mindavi:
Yes, I had the verbose parameter at the end of the command. However, I can't get the verbose output working with the parameters anywhere on the line. Did you get it working?

jbrzozoski:
Not really the ideal way to test it, buy you can prove the "auto" mode works with the samples I provided.  You just have to pass each file multiple times to a single instance, so something like this will exercise the brute force method:
`rtl_433 -R 176:auto -r gaps.cu8 -r temp.cu8 -r impulse.cu8 -r gaps.cu8 -r temp.cu8 -r impulse.cu8`

I wasn't really checking for it when I made the samples, but the CRCs in those payloads hit enough orthogonal IDs that only the right one matches all of them and passes the threshold even if they're repeated.  (i.e. no false positives)

If you put it in "auto" mode and use the ID sample, it will immediately lock onto the ID payload and will decode everything after that, so a command line like this will decode all of the samples:
`rtl_433 -R 176:auto -r id.cu8 -r gaps.cu8 -r temp.cu8 -r impulse.cu8`

Mindavi:
Yeah, I figured it worked like that. Seems to work fine.

Are there any open comments here?

merbanan:
Config and manual additions are missing but besides that everything is ok by me.

jbrzozoski:
Did you need anything else from me on this?

And my main monitoring station now has these config lines:
```
protocol -176
protocol 176:12345
```
Which looks weird, but works fine setting an option on this decoder while leaving all default decoders enabled. :+1:

PR #1583, opened 2020-12-17
devices: add "LaCrosse Technology TX22U-IT". WIP.

CazYokoyama:
http://nikseresht.com/blog/?p=99 tells data format of TX22U-IT.
However, my 2 TX22U-IT are different than the one there.
As in the subject line, this is WIP. TX22U-IT can have rain sensor. In addition to rain sensor, there are several fields are not parsed.

It seems temperature report is the good timing to upload because it is reported every 1.5 - 2 minutes.

Signed-off-by: Caz Yokoyama <caz@caztech.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [173]  Bresser Weather Center 7-in-1
     [174]  EcoDHOME Smart Socket and MCEE Solar monitor
     [175]  LaCrosse Technology View LTV-R1 Rainfall Gauge
+    [176]  LaCrosse Technology TX22U-IT
 
 * Disabled by default, use -R n or -G
 
--- a/src/devices/lacrosse_tx22uit.c
+++ b/src/devices/lacrosse_tx22uit.c
@@ -1,0 +1,178 @@
+/** @file
+    LaCrosse Technology View TX22U-IT temperature, humidity, wind speed/direction
+    and rain sensor.
+
+    Copyright (C) 2020 Caz Yokoyama, caz at caztech dot com
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+/**
+LaCrosse Technology View TX22U-IT temperature, humidity, wind speed/direction
+and rain sensor
+
+Product pages:
+TBW
+
+Specifications:
+TODO: update for TX22U-IT
+- Wind Speed Range: 0 to 178 kmh
+- Degrees of Direction: 360 deg with 16 Cardinal Directions
+- Outdoor Temperature Range: -29 C to 60 C
+- Outdoor Humidity Range: 10 to 99 %RH
+- Update Interval: Every 31 Seconds
+
+http://nikseresht.com/blog/?p=99 tell protocol but my TX22U-IT is different protocol.
+*/
+
+#include "decoder.h"
+#define BIT_PER_BYTE 8
+
+static int
+decode_3bcd(uint8_t *p)
+{
+    return ((*p & 0x0f) * 100) + ((*(p + 1) >> 4) * 10) + (*(p + 1) & 0x0f);
+}
+
+static int
+decode_3nybble(uint8_t *p)
+{
+    return ((*p & 0x0f) << 8) | *(p + 1);
+}
+
+static int lacrosse_tx22uit_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    uint8_t const preamble_pattern[] = { 0xaa, 0x2d, 0xd4 };
+
+    data_t *data;
+    uint8_t b[13], *p;
+    uint32_t id;
+    int flags, offset, chk, size;
+    int raw_temp = -1, humidity = -1, raw_speed = -1, direction = -1;
+    float temp_c, rain_mm = -1.0, speed_kmh, wind_gust_kmh = -1.0;
+
+    offset = bitbuffer_search(bitbuffer, 0, 0,
+            preamble_pattern, sizeof(preamble_pattern) * 8);
+
+    if (offset >= bitbuffer->bits_per_row[0]) {
+        if (decoder->verbose)
+            fprintf(stderr, "%s: Sync word not found\n", __func__);
+        return DECODE_ABORT_EARLY;
+    }
+
+    offset += sizeof(preamble_pattern) * BIT_PER_BYTE;
+    size = bitbuffer->bits_per_row[0] - offset;
+    bitbuffer_extract_bytes(bitbuffer, 0, offset, b, size);
+    size /= BIT_PER_BYTE;
+
+    chk = crc8(b, size, 0x31, 0x00);
+    if (chk) {
+        if (decoder->verbose)
+           fprintf(stderr, "%s: CRC failed!\n", __func__);
+        return DECODE_FAIL_MIC;
+    }
+
+    if (decoder->verbose)
+        bitbuffer_print(bitbuffer);
+
+    /*
+      the first several hours since power-on
+          0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
+      aa aa 2d d4 a2 a5 05 72 10 58 20 00 38 00 40 00 fc 00 00 - 17.2C 58% 180 at 0.
+
+      thereafter,
+          0  1  2  3  4  5  6  7  8  9 10 11 12 13
+      aa aa 2d d4 a2 83 10 72 20 1c 38 00 33 00 00
+      aa aa 2d d4 a2 82 04 89 20 1c 70 00 00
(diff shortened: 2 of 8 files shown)


CazYokoyama:
Could anyone tell me why lacrosse_tx22uit.c is not built while src/CMakeLists.txt has it?

src/devices/lacrosse_tx22uit.c
```diff
@@ -0,0 +1,135 @@
+/** @file
+    LaCrosse Technology View TX22U-IT temperature, humidity, wind speed/direction
+    and rain sensor
```
zuckschwerdt, line 3: First lines in doc comments needs to end with a period.

src/devices/lacrosse_tx22uit.c
```diff
@@ -0,0 +1,135 @@
...
+- Wind Speed Range: 0 to 178 kmh
+- Degrees of Direction: 360 deg with 16 Cardinal Directions
+- Outdoor Temperature Range: -29 C to 60 C
+- Outdoor Humidity Range: 10 to 99 %RH
+- Update Interval: Every 31 Seconds
+
+http://nikseresht.com/blog/?p=99 tell protocol but my TX22U-IT is different protocol.
+*/
+
+#include <stdbool.h>
```
zuckschwerdt, line 30: plain `int` is prefered

src/devices/lacrosse_tx22uit.c
```diff
@@ -0,0 +1,135 @@
...
+            raw_speed = ((b[offset+1] >> 4) * 10) + (b[offset+1] & 0x0f);
+            break;
+        }
+    }
+    if (upload_data) {
+        temp_c = (raw_temp - 400) * 0.1f;
+        speed_kmh = raw_speed * 0.1f;
+
+        if (humidity < 0 || humidity > 100
+            || temp_c < -40 || temp_c > 70
```
zuckschwerdt, line 90: continuation lines need two indents (8 spaces)

src/devices/lacrosse_tx22uit.c
```diff
@@ -0,0 +1,135 @@
...
+            || direction < 0 || direction > 360
+            || speed_kmh < 0 || speed_kmh > 200)
+            return DECODE_FAIL_SANITY;
+
+        data = data_make(
```
zuckschwerdt, line 95: wrap data_make in `/* clang-format off */`

src/devices/lacrosse_tx22uit.c
CazYokoyama: will do.

src/devices/lacrosse_tx22uit.c
CazYokoyama: convert bool to int. and define FALSE and TRUE.

src/devices/lacrosse_tx22uit.c
CazYokoyama: will do.

src/devices/lacrosse_tx22uit.c
zuckschwerdt: Just use 0 and 1, we are dealing with bit-level stuff here anyway :)

src/devices/lacrosse_tx22uit.c
CazYokoyama: will be wrapped.
Thx.

CazYokoyama:
Could someone tell me what "WARNING: rtl_433 binary not found: skipping README/man generation!" mean? One of "Build Analyze Check" fails. What action I have to do?

src/devices/lacrosse_tx22uit.c
CazYokoyama: Done.

src/devices/lacrosse_tx22uit.c
```diff
@@ -0,0 +1,137 @@
...
+
+        if (humidity < 0 || humidity > 100
+                || temp_c < -40 || temp_c > 70
+                || direction < 0 || direction > 360
+                || speed_kmh < 0 || speed_kmh > 200)
+            return DECODE_FAIL_SANITY;
+
+        /* clang-format off */
+        data = data_make(
+            "model",            "",                 DATA_STRING, "LaCrosse-Tx22uit",
```
zuckschwerdt, line 96: Usually this would be "LaCrosse-TX22UIT"

src/devices/lacrosse_tx22uit.c
```diff
@@ -0,0 +1,137 @@
...
+#include "decoder.h"
+
+static int lacrosse_tx22uit_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    uint8_t const preamble_pattern[] = { 0xaa, 0xaa, 0x2d, 0xd4 };
+
+    data_t *data;
+    uint8_t *b = bitbuffer->bb[0];
+    int offset;
+    static int raw_temp = 0, humidity = -1, raw_speed = -1, direction = -1;
```
zuckschwerdt, line 39: You should output only the data that appears in a message. Don't use static to hold on to old data.
You can use `DATA_COND` to easily choose what fields to output in data_make.

src/devices/lacrosse_tx22uit.c
CazYokoyama: will change.

src/devices/lacrosse_tx22uit.c
CazYokoyama: done.

README.md
```diff
@@ -259,7 +260,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
   [-d <RTL-SDR USB device index>] (default: 0)
   [-d :<RTL-SDR USB device serial (can be set with rtl_eeprom -s)>]
 	To set gain for RTL-SDR use -g <gain> to set an overall gain in dB.
-	SoapySDR device driver is available.
+	SoapySDR device driver is not available.
```
merbanan, line 263: ?

merbanan:
The code looks fairly clean. Just fix the missing fields and it should be good to merge.

README.md
CazYokoyama: This is a one of update when I run maintainer_update.py. I have no idea. Caused by my environment?

src/devices/lacrosse_tx22uit.c
```diff
@@ -0,0 +1,133 @@
...
+    }
+#endif
+    /*
+       0  1  2  3  4  5  6  7  8  9 10 11 12 13 14
+      aa aa 2d d4 ad 03 05 92 20 01 3d 00 b0 00 00  19.2C  wind 13 at 0.0km/h
+      aa aa 2d d4 ad 02 20 01 40 00 1f 00 00
+      aa aa 2d d4 ad 01 20 01 a3 00 00
+      aa aa 2d d4 ad 02 10 56 20 01 86 00 00        56%
+   */
+    for (offset = 6; offset < bitbuffer->bits_per_row[0] / 8; offset += 2) {
```
zuckschwerdt, line 67: Shouldn't this offset be the one from bitbuffer_search? Maybe bitbuffer_extract_bytes is needed to align.
If aaaa2dd4 is the full preamble seen, then likely aa2dd4 is a better search, it will cope with missing initial sync preamble bits.
Also starting at byte 6 means skipping that `ad0?`, right? is that the ID and seq?

CazYokoyama: When I move TX22U-IT to garage which is 5m away through floor from side-by-side, I have the following bitbuffer in consistent and stable manner.

[00] {146} aa aa 2d d4 ac 25 05 38 10 60 20 00 3c 00 40 00 a2 00 00 

Its display shows "temperature_C" : 13.800, "humidity" : 61, "wind_avg_km_h" : 0.000, "wind_dir_deg" : 12, rain: 0.

I test with variety of preamble such as a single 0xaa. But "0xaa 0xaa 0x2d 0xd4 0xac, 0x25" gives the most stable data . The last 2 byte may be a device ID. But I can't judge because I only have one TX22U-IT.
Subsequent data seems 2 byte pack. I'll update my code according to this new data.

README.md
CazYokoyama: reverted.

src/devices/lacrosse_tx22uit.c
zuckschwerdt: the `aaaa` is a preamble, you should not require the full preamble, just a wide enough trailing portion. That way a few bits missing at the start will still be ok. The `2dd4` is a common sync-word, everything after that is the actual data.
A match on `aa2dd4` will be best. Then use bitbuffer_extract_bytes to get a bitrow with just the aligned data, i.e. b[0] will then be the byte after 2dd4.

CazYokoyama: Done on aa2dd4 bitbuffer_extract_bytes().

zuckschwerdt:
What is the state of this @CazYokoyama ? Is this still working and good for (manual) merge?

merbanan:
Merged in https://github.com/merbanan/rtl_433/commit/abcb6f444245d198015f697c4a4322e4f746342d

PR #1584, opened 2020-12-19
Remove additional invalid "weather" device class (#1548)

stephenhouser:
The previous merged commit does not remove all the invalid "weather" device class items from the templates.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -214,7 +214,6 @@ mappings = {
         "device_type": "sensor",
         "object_suffix": "WS",
         "config": {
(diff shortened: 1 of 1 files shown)


stephenhouser:
rtl_433 publishes both 'lux' and 'light_lux' across different devices. Reverted the change that renamed and added light_lux. Also added 'uvi' which is reported via fineoffset devices.

rct:
Thanks for these needed configuration data updates. I have some that I've made that I haven't contributed.

Unfortunately, these should probably get merged with  @maxlock's #1559, which moves the config data out of the python code.  I'm not sure what's going on with that PR. Need to check with @zuckschwerdt

PR #1588, opened 2020-12-23
Added FAN-53T decoder

phobozad:
Just a quick config file to support FAN-53T remotes.

https://fccid.io/2AAZPFAN-53T/User-Manual/User-manual-2228959


zuckschwerdt:
Thanks!

PR #1590, opened 2020-12-28
Blueline tweaks and improvements

jbrzozoski:
- Fix a signed/unsigned bug in temperature reception
- Use a new linear estimate that fits my logged temperature data better
- Speed up ID guessing method by 64x by using a reverse CRC
- Improve ID guessing logic to also ensure that the "winning" ID isn't tied with another possibility
- Improve comments and documentation

Signed-off-by: Justin Brzozoski <justin.brzozoski@gmail.com>

### DIFF
--- a/src/devices/blueline.c
+++ b/src/devices/blueline.c
@@ -64,14 +64,20 @@
  * and the payload may be interpreted.
  *
  * Note that if the transmitter's ID isn't known, the code can't easily determine if messages other than an
- * ID payload are good or bad, and can't intepret their data correctly.  However, if the BLUELINE_USE_ID_TRANSMISSION
- * or BLUELINE_USE_ID_AUTODETECT features are enabled, the system can learn the transmitter's ID by various methods.
+ * ID payload are good or bad, and can't intepret their data correctly.  However, if the "auto" mode is enabled,
+ * the system can try to learn the transmitter's ID by various methods. (See USAGE HINTS below)
  *
  * For the power message (1), the offset payload gives the number of milliseconds gap between impulses for the most
  * recent impulses seen by the monitor.  To convert from this 'gap' to kilowatts, you will need your meter's
  * Kh value. The Kh value is written obviously on the front of most meters, and 1.0 and 7.2 are very common.
  *
  * kW = (3600/gap) * Kh
+ *
+ * Note that the 'gap' value clamps to a maximum of 65533 (0xFFFD), so there is a non-zero floor when calculating the
+ * kW value using this report.  For example, with a Kh of 7.2, the lowest kW value you will ever see when monitoring
+ * the 'gap' value is (3600/65533)*7.2 = 0.395kW.  If you need power monitoring for impulse rates slower than every
+ * 65.533 seconds to do things like confirm that your power consumption is 0kW, you need to monitor the impulse
+ * counts and timing between energy messages (see 3 below).
  *
  * For the temperature message (2), the offset payload gives the temperature in an odd scaling in the last byte,
  * and has some flag bits in the first byte.  The only known flag bit is the battery.  rtl_433 handles scaling back to
@@ -104,15 +110,22 @@
  * rtl_433 -R 176:45364
  *
  * If you are unable to access the monitor to have it send the ID message, you can also use the "auto" parameter:
- * rtl_433 -R 176:auto -v
+ * rtl_433 -vv -R 176:auto
  *
- * Verbose mode is recommended to tell what the "auto" mode is doing.
+ * Verbose mode should be specified first on the command line to see what the "auto" mode is doing.
  *
  * The auto parameter will try to brute-force the ID on any messages that look like they are from a
  * BlueLine monitor.  This method usually succeeds within a few minutes, but is likely to get false positives
  * if there is more than one monitor in range or the messages being received are all identical (i.e. if the
  * meter is continuously reporting 0 watts).  If it succeeds, it will start reporting data with the new ID,
  * which you should then use as a parameter when you re-run rtl_433 in the future.
+ *
+ * Finally, passing a parameter to this decoder requires specifying it explicitly, which normally disables all
+ * other default decoders.  If you want to pass an option to this decoder without disabling all the other defaults,
+ * the simplest method is to explicity exclude this one decoder (which implicitly says to leave all other defaults
+ * enabled), then add this decoder back with a parameter.  The command line looks like this:
+ *
+ * rtl_433 -R -176 -R 176:45364
  *
  */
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
The temp conversion still looks a bit strange. I would expect something like `tF = (X - 28) / 1.3` or `tC = X / 2.333 - 30`.
Can you record a rather high temperature to better see the scaling? Not sure if using an oven or hot-water is feasible, maybe a hair-dryer?

jbrzozoski:
I got my hot air gun out and pushed it up to almost 50C.  I backed off there since I was worried the heat was going to ruin the waterproofing.  Still weirdly imperfect.  I agree that the true equation is most likely a whole-number fraction slope and offset.

The transmitter design is based on a MSP430F1132, so I looked up the datasheet to see how the internal temperature sensor works.  Internally, it's a 10-bit ADC which can use different voltage references, but the recommended one for temperature reading would give these ADC values:

adc_reading = 1023 * (3.55*X + 986) / 1500

(3.55mV per degC, offset of 986mV at 0C, 1500mV reference)

I'm still pondering to try and get from this equation back to the numbers they report on the radio, but it's not a simple right-shift or anything else immediately obvious.

Here's the full table of data I've correlated between the radio and display:

```
recvd_temp,deg_C
44,-13
45,-12
54,-7
55,-7
56,-6
58,-5
60,-4
62,-3
64,-2
65,-2
69,0
70,1
71,1
72,2
73,2
75,3
78,5
81,6
100,14
102,15
107,17
114,19
115,20
116,20
117,20
119,21
122,23
124,23
125,24
126,24
127,25
128,25
129,25
131,26
134,28
137,29
139,30
140,30
143,31
144,32
145,32
146,33
147,33
148,34
151,35
155,37
161,40
166,42
169,44
175,47
177,48
179,49
```

jbrzozoski:
I take back that "nothing immediately obvious" comment ... It's plausible it's just a simple offset of (Nadc - 591) or something like that ...

zuckschwerdt:
If the byte is a somewhat raw adc value then indeed a random slope and offset make sense. (And every hardware revision would now be in a tight spot to ensure protocol compatibility…)

A linear regression fits well, with an error-squares fit of 0.9984 (1=perfect) it is `tC = 0.436*X - 30.36`.
With 95% confidence the offset is -30.95 to -29.77, which would make 30 a good candidate, but the slope 0.4309 to 0.4409 has no obvious "round" number.

I guess your updated calculation fits the expected temperature range best then :)

merbanan:
The device might transmit the raw temperature reading from the temperature sensor. Then you need a lookup table/function because temperature sensors are nonlinear. But I've only seen one sensor with that design.

jbrzozoski:
> The device might transmit the raw temperature reading from the temperature sensor. Then you need a lookup table/function because temperature sensors are nonlinear. But I've only seen one sensor with that design.

Whoever programmed this thing for Blueline has already proven they think "outside the box" with that weird payload-offset trick. :)

src/devices/blueline.c
```diff
@@ -280,17 +331,33 @@ static int blueline_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+            // a full copy of my logged data is in the comments of this GitHub pull
+            // request:
+            //
+            // https://github.com/merbanan/rtl_433/pull/1590
+
+            const uint8_t temperature = offset_payload_u8[1];
             const uint8_t flags = offset_payload_u8[0] >> 2;
             const uint8_t battery = (flags & 0x20) >> 5;
-            const double temperature_C = (0.457 * temperature) - 33.68;
+            const double temperature_C = (0.436 * temperature) - 30.36;
```
merbanan, line 360: Use float here.

bkenobi:
I'm not planning on using this meter's temperature for anything since I have a weather sattion, but wanted to point out that it appears to be offset from my Acurite 5n1 weather station is reporting by approximately 3°F.  The outside temperature hasn't changed much in the last day so I don't know if this is always off by a constant.

Blueline vs Acurite

jbrzozoski:
I have no idea of the accuracy of the temperature sensor on this thing.  If it's actually using the built-in sensor on the MSP430, it's mediocre-to-poor and will vary from one unit to another.  All we can do is just replicate Blueline's math as closely as possible, and hope that they did a good job in designing the sensor and firmware on their side.

Anecdotally, my Blueline temperature readings line up within a degree of a LaCrosse outdoor thermometer I've been using for years.

jbrzozoski:
Do you guys need anything else from me on this one?

bkenobi:
I updated to the latest temperature correlation this morning.  The attached plot shows the temperature reported by the power meter against my weather station (Acurite 5in1) reporting through rtl_433.  I checked the temperatures earlier today when the module was reporting 47.98°F (8.9°C).  The display unit showed 49F and the Acurite shows 50F.  I assumed this was just an offset due to the different hardware in the weather station versus the power meter reader.  However, looking at the temps this evening I see that the offset goes away and the two match really well although the BL display currently shows 43F and the rtl_433 code is calculating 41.7.

I'm not intending on using this as a primary weather station reading.  This is more to help inform in case a future update is considered.

image

bkenobi:
I see this trend repeated daily where the power meter diverges from the weather station data starting in the morning ~10am until the afternoon ~2pm.  I have only seen peak temperatures in the high 40's to low 50's.  It is possible the difference is driven by location (weather station in sun, power meter in shade) but the rtl_433 data doesn't match the display by ~1.5°F currently.

bkenobi:
I think there may be 2 reasons why my weather station and power meter don't show the same temperature.  First, the weather station is located in the sun (as it should be) but the power meter is located on the north facing side of my house which gets no sun.  The shade will likely be cooler than in the open exposed to solar radiation.

Second, the peak temperatures I am seeing right now are exactly where the linear approximation breaks down.  As you can see from the plot of the recvd_tmp (in degF) versus output temp for both the raw converted data you posted and the equation, there is an error that happens to peak at 39°F of 2.4°F.  I will look at this going forward, but I'm wondering if a table lookup/interpolation would be more accurate if not too difficult.  As stated before, since this isn't a weather station, perhaps the accuracy is not something anyone will care about.

image

bkenobi:
This device was updated recently (Oct 29).  I had not updated my code since early this year (January or February) so when I updated I expected a few tweaks.  I don't know if it's documented and I missed it, but the device name that is published in MQTT has been altered so my Node-Red flow does not work anymore.  It should be an easy fix as it appears that the old name:
`Blueline_PowerCost_Monitor`

has been changed to:
`Blueline-PowerCost`

This is a subtle difference but might cause some issues for anyone who updates and finds their old HA setup no longer works as they expect.  Not a complaint by any means, just pointing out that if I missed the note in the documentation perhaps others will too.

PR #1592, opened 2021-01-02
Extend FineOffset WH1050 field widths

kw217:
I've had a WH1050 weather station for the last six years, and have been using a custom decoder. I recently discovered rtl_433 and am in the process of switching over.

I've noted one relevant difference: the rain field is actually wider than the rtl_433 decoder assumes. It currently assumes 12 bits (4095 = 1228.5mm), but I've seen a valid count of 6551 (1965.3mm) before changing the battery - which is 13 bits. Since there are 3 more unused bits beyond this which seem always to be zero, it seems reasonable to use them all, extending the rain field to 16 bits (19660.5mm).

The low battery bit is correct, so I've removed the uncertainty comment.

Please let me know if there's anything else I need to do for this PR.


zuckschwerdt:
Great, thanks! Can you also add information on the message layout, e.g. as seen here https://github.com/merbanan/rtl_433/blob/master/src/devices/cotech_36_7959.c#L21

kw217:
Thanks @zuckschwerdt , that's done now.

PR #1595, opened 2021-01-03
Change soil moisture to percent display

zuckschwerdt:
Changes Bresser 6in1 and Springfield soil moisture display to percent.
This is breaking (6in1 is new and probably not widely used).
See also #1492


merbanan:
Normalisation is always good.

src/devices/bresser_6in1.c
```diff
@@ -187,9 +187,9 @@ static int bresser_6in1_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             "channel",          "",             DATA_INT,    chan,
             "battery_ok",       "Battery OK",   DATA_INT,    batt,
             "temperature_C",    "Temperature",  DATA_COND, temp_ok, DATA_FORMAT, "%.1f C", DATA_DOUBLE, temp_c,
-            "humidity",         "Humidity",     DATA_COND, temp_ok, DATA_INT,    humidity,
+            "humidity",         "Humidity",     DATA_COND, temp_ok && moisture < 0, DATA_INT,    humidity,
```
Mindavi, line 190: Does this mean humidity is only shown when moisture < 0?

Mindavi:
Other than the one question, diff LGTM

src/devices/bresser_6in1.c
zuckschwerdt: `moisture >= 0` is valid and thus no humidity should be shown (the moisture level is transmitted in the humidity field)

Mindavi: Ah ok, clear. Good to know :)

PR #1596, opened 2021-01-04
Fix Insteon payload  buffer too small

evilpete:
used min packet instead of max packet size

this went unnoticed as full packets are rarely received. (there is a lot of padding in the protocal and only 23 or the 32 bytes are needed for extended packets )

### DIFF
--- a/src/devices/insteon.c
+++ b/src/devices/insteon.c
@@ -302,9 +302,9 @@ static int parse_insteon_pkt(r_device *decoder, bitbuffer_t *bits, unsigned int 
         cmd_array[cmd_array_len++] = (int)results[j];
     }
 
-    char payload[INSTEON_PACKET_MIN_EXT * 2 + 1] = {0};
+    char payload[INSTEON_PACKET_MAX_EXT * 2 + 2] = {0};
     p                = payload;
-    for (int j = 0; j < min_pkt_len; j++) {
+    for (int j = 0; j < results_len; j++) {
         p += sprintf(p, "%02X", results[j]);
     }
 


evilpete:
Note: this may (will)  change the payload field in the test data samples this the json may need to be  rebuilt

src/devices/insteon.c
```diff
@@ -302,9 +302,9 @@ static int parse_insteon_pkt(r_device *decoder, bitbuffer_t *bits, unsigned int
         cmd_array[cmd_array_len++] = (int)results[j];
     }
 
-    char payload[INSTEON_PACKET_MIN_EXT * 2 + 1] = {0};
+    char payload[INSTEON_PACKET_MAX_EXT * 2 + 2] = {0};
```
zuckschwerdt, line 305: What's at position `INSTEON_PACKET_MAX_EXT * 2 + 1` then?

evilpete: I added the extra char as extra.

Insteon has two packet types regular and "extended"

the layer 2  protocol allows for 32 char for extended packets although only 23 are used,  when bytes 24->32 may or may not be there.
(the code reads as many valid segments  that are received  (one byte per segment)) and only declares error if less then 23 were received (the crc is the 23rd )   

for regular/short packets only 10 bytes are used out of the  13 segments (the 10th being the CRC)

A bit flag is in the first byte/segments to indicate is the packet is extended or not,   this can also be derived from the 5 bit index number in each segments,  as the segments arrive the counter decrements down to zero.

Some Insteon devices will transmit the full layer 2 packet,  and other devices stop after the crc

zuckschwerdt: Making doubly sure there is a zero :) Usually with hex-code buffers they need to be of odd length 2-bytes groups of code plus 1 zero terminator -- thus my confusion ;) No need to change anything, just me being picky.

PR #1602, opened 2021-01-10
Fix: setting mqtt retain flag for hass example

culliard:
I used this script as a reference to add sensors to my Home assistant. However they were not persisted across hass restarts. With this fix they are not retained. (python 3.8.7 and paho-mqtt 1.5.1).


zuckschwerdt:
s.a. #1555 and #1559

rct:
@zuckschwerdt - this looks good and I believe should be merged for the current code.  It will need to be merged into #1559 when the conflicts are resolved.  LGTM

PR #1606, opened 2021-01-11
Add support for wmbus water meter Maddalena

jedi7:
* Add actual water measure
* Add measure of last months

Buf cols must be increased because the wmbus payload is quite long (over 2048 bits)

There are still some unknown data, which will need more reverse engineering.

### DIFF
--- a/include/bitbuffer.h
+++ b/include/bitbuffer.h
@@ -14,7 +14,7 @@
 
 #include <stdint.h>
 
-#define BITBUF_COLS 256 // Number of bytes in a column
+#define BITBUF_COLS 512 // Number of bytes in a column
 #define BITBUF_ROWS 25
 #define BITBUF_MAX_PRINT_BITS 50 // Maximum number of bits to print (in addition to hex values)
 
--- a/src/devices/m_bus.c
+++ b/src/devices/m_bus.c
@@ -252,13 +252,13 @@ static int m_bus_decode_records(data_t *data, const uint8_t *b, uint8_t dif_codi
                 case 0x4:
                     temp = (int16_t)((b[1]<<8)|b[0])*0.001;
 
-                    if(dif_sn == 0) {
+                    if (dif_sn == 0) {
                         data = data_append(data,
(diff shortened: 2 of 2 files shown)


jedi7:
Recorded data:
[wbus_maddalena.zip](https://github.com/merbanan/rtl_433/files/5796012/wbus_maddalena.zip)

merbanan:
Looks ok, can you fix the remaining style checks?

And can you also contribute this to our competition? https://github.com/weetmuts/wmbusmeters

jedi7:
I can :) but I do not know which, can you point me?
I will look at the wmbusmeters, maybe they are more far in the reverse engineering of the meters.

merbanan:
Style issues are fixed now. I just need to look through the increase of the columns. And yeah the wmbusmeters project has reverse-engineered lots of meters. They want to support your meter and they want the datagram.

include/bitbuffer.h
```diff
@@ -14,7 +14,7 @@
 
 #include <stdint.h>
 
-#define BITBUF_COLS 256 // Number of bytes in a column
+#define BITBUF_COLS 512 // Number of bytes in a column
```
zuckschwerdt, line 17: Can we make that 400 columns (if I check correctly you need 392) and add a comment that "wmbus needs at least 392"

jedi7: do we need so low limits? It is not running on embedded system, so 256B of memory is ok I hope :)

zuckschwerdt: actually +256B times BITBUF_ROWS, and we do aim to run on low-end embedded (e.g. old routers) ;)

jedi7: IMHO then we need at least ~437B  (( (256+16*2+3)*12)/8 ), when the data len will be 256 + 16*2 CRC + 3B

jedi7:
guys, I added also commit which fixes the length of the value. Is possible to test this change also on another wmbus, because the change can hit also the already present temperature meters.

jedi7:
this PR can be replaced by "better" PR https://github.com/merbanan/rtl_433/pull/1610

PR #1610, opened 2021-01-12
Feature/add metrics

jedi7:
Extending support for standard units on m-bus

### DIFF
--- a/include/bitbuffer.h
+++ b/include/bitbuffer.h
@@ -14,7 +14,8 @@
 
 #include <stdint.h>
 
-#define BITBUF_COLS 512 // Number of bytes in a column
+// NOTE: Wireless mbus protocol needs at least ((256+16*2+3)*12)/8 => 437 bytes
+#define BITBUF_COLS 450 // Number of bytes in a column
 #define BITBUF_ROWS 25
 #define BITBUF_MAX_PRINT_BITS 50 // Maximum number of bits to print (in addition to hex values)
 
--- a/include/bitbuffer.h
+++ b/include/bitbuffer.h
@@ -14,7 +14,7 @@
 
 #include <stdint.h>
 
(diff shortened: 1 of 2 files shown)


jedi7:
please just give me known how big the BITBUF_COLS can be, and I will update it.

jedi7:
I have prepared also branch which add timedate, and slightly modify the append function:
https://github.com/jedi7/rtl_433/pull/new/feature/add_mbus_datetime

include/bitbuffer.h
```diff
@@ -14,7 +14,7 @@
 
 #include <stdint.h>
 
-#define BITBUF_COLS 256 // Number of bytes in a column
+#define BITBUF_COLS 512 // Number of bytes in a column
```
zuckschwerdt, line 17: Make it 450 then and write a comment that m_bus needs this.

src/devices/m_bus.c
```diff
@@ -210,30 +196,265 @@ static char* oms_hum_el[4][4] = {
 {"Error 31","Error 32","Error 33","Error 34",}
 };
 
+static char* history_hours[4] = {
```
zuckschwerdt, line 199: type should be `char *x` instead of `char* x`.
indent these lists by 8 spaces.

jedi7: @zuckschwerdt  are you sure? I see the "static char* oms_hum" uses c++ asterix position (like type is char*, not *osm_hum). Indentation the same. I can change it every where to have it consistent. If needed :)

include/bitbuffer.h
jedi7: fixed

src/devices/m_bus.c
zuckschwerdt: Yes, but only fix your code for now to keep the diff readable. You can also check with clang-format for a rough guide.

PR #1614, opened 2021-01-17
update tech_433

marco402:


zuckschwerdt:
We still need to clean up and merge this. Keep it open and just push more commits if possible.

PR #1618, opened 2021-01-20
Support for Hyundai TPMS

teou1:
Hello,
during the last Lockdown winter months i played a little with a SDR Radio, among other stuff i decided to read the TPMS sensors of my car. These are VDO sensors on the winter set of tires, but bought from official Hyundai dealer, so i hope that is what comes with every EU Hyndai i30. I guess well see in summer.

The protocol was very similar to Citroen, Janste an Abarth, so only some minor corrections were needed, and here is the Hyundai protocol.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -184,6 +184,7 @@
     DECL(ecodhome) \
     DECL(lacrosse_r1) \
     DECL(blueline) \
+    DECL(tpms_hyundai_vdo) \
     /* Add new decoders here. */
 
 #define DECL(name) extern r_device name;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -167,6 +167,7 @@ add_library(r_433 STATIC
     devices/tpms_citroen.c
     devices/tpms_elantra2012.c
     devices/tpms_ford.c
+    devices/tpms_hyundai_vdo.c
     devices/tpms_jansite.c
     devices/tpms_pmv107j.c
     devices/tpms_renault.c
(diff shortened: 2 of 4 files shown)


zuckschwerdt:
Thanks! We'll need to clean up the Trailing whitespace errors and correct the indentation, then it's good for merge.

teou1:
ok, i tried my best :) I am not a programmer, actually i just learned how to create a pull request. I think i corrected the spaces and tabs with Mousepad, but i have no idea what maintainer_update is and why is it failing, sorry.

merbanan:
We need signal samples also for the rtl_433_tests repo.

merbanan:
Can you use post the raw message so that we can figure out the checksum? We need alot of recordings.

teou1:
What do you mean with raw sample, a cu8 file? I skipped that, the Manchester decoding is already done in citroen, abarth and janste files, i just adjusted the lenght and position of the parameters.

If you mean the full message in hex, i have a ton of samples from my testing files. Here is one with the longest messages. 
[tmps24-3all.txt](https://github.com/merbanan/rtl_433/files/5844545/tmps24-3all.txt)
Hyundai2 is with one extra byte at the end, i have tested several changes at a time

teou1:
My sensors are (in hex) 91772cdc, 91772fce, 917726dd, 91739ff3

This is verified through diagnostic on the can bus of the car, so that was the known data.

zuckschwerdt:
A CRC-8 with poly 0x07 init 0xaa seems to work (note: first byte included!), except for these codes:
```
138d4c317401943910 38
138d4c317402943910 3b
138d4c317403943910 3a
138d1a684001953913 01
138d1a684003953913 03
1f8d1a684000953913 00
```
The leading nibble is `1` here and we usually see `2`.

zuckschwerdt:
continued in #1643

PR #1620, opened 2021-01-21
Fix copy paste error in hass-example script.

conrad784:
I used this script to add sensors to my Homeassistant. However when running with -M level the noise level was not reported right, which is because of this little typo.


klattimer:
I've incorporated a bunch more features for hass which also includes the changes in this pull request here - https://github.com/merbanan/rtl_433/pull/1665

PR #1629, opened 2021-01-27
Update README.md

specialorange:
Site is no longer active


zuckschwerdt:
You might be experiencing an abuse protection of that server, it's still running. Let me know your IP range (remove the last one or two numbers) by mail.

PR #1630, opened 2021-01-27
Extend decoded values + HCA

jedi7:
* add support for BCD value decoding and 64bit values
* add support for H.C.A
* add support for OnTime and Operation Time

### DIFF
--- a/src/devices/m_bus.c
+++ b/src/devices/m_bus.c
@@ -178,8 +178,6 @@ typedef struct {
 
 static float humidity_factor[2] = { 0.1, 1 };
 
-static int consumed_bytes[8] = { -1, 1, 2, 3, 4, 4, 6, 8};
-
 
 static char* oms_hum[4][4] = {
 {"humidity","average_humidity_1h","average_humidity_24h","error_04", },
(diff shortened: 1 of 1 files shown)


src/devices/m_bus.c
```diff
@@ -310,7 +330,7 @@ static data_t *append_val(data_t *data, enum UnitType unit_type, uint8_t value_t
         fprintf(stderr, "M-Bus: Program error, exp (%d) is out of bounds", exp);
         return data;
     }
-    float fvalue = val * pow10_table[exp];
+    double fvalue = val * pow10_table[exp];
```
merbanan, line 333: Is double really needed?

jedi7: Not fully sure. But I increased it because we have here 64bit values. So to keep precision.

zuckschwerdt: This does not do what one might think. The calculation ist float, the float is then stored as double.

jedi7: you are right, I forgot to change also pow10_table

PR #1631, opened 2021-01-27
Increase precision of detected frequency

jedi7:
Increase precision of detected frequency to be able distinguish multiple devices (thermometers).
With precise frequency is simpler to localize them on freq. waterfall.


zuckschwerdt:
It's just a very rough estimation and there is not enough accuracy behind it.

PR #1636, opened 2021-01-30
Add GPSd tags option

### LINKED ISSUE #909: Add option for GPS coordinates in messages

zuckschwerdt:
Turns out addig GPSd support as simple TCP client is easy ;)

- supersedes  #1635
- supersedes  #1250
- closes #909

Needs a little more work with JSON filtering but is already quite useable.
An option to read other types of tag data from TCP (or UDP, or offer an inbound TCP/UDP tag server) is quick to add, which addresses #909

### DIFF
--- a/include/data_tag.h
+++ b/include/data_tag.h
@@ -12,16 +12,19 @@
 #ifndef INCLUDE_TAGS_H_
 #define INCLUDE_TAGS_H_
 
+struct gpsd_client;
+struct mg_mgr;
 struct data;
 
 typedef struct data_tag {
-    char *key;
-    char *val;
-    int prepend;
+    char const *key;
+    char const *val;
+    char const **includes;
+    struct gpsd_client *gpsd_client;
 } data_tag_t;
 
 /// Create a data tag. Might fail and return NULL.
-data_tag_t *data_tag_create(char *params);
+data_tag_t *data_tag_create(char *params, struct mg_mgr *mgr);
 
 /// Free a data tag.
 void data_tag_free(data_tag_t *tag);
--- a/include/optparse.h
+++ b/include/optparse.h
@@ -116,6 +116,14 @@ int atoi_time(char const *str, char const *error_hint);
 /// @return the original value of *stringp
 char *asepc(char **stringp, char delim);
 
+/// Similar to strsep but bounded by a stop character.
+///
+/// @param[in,out] stringp String to parse inplace
+/// @param delim the delimiter character
+/// @param stop the bounding character at which to stop
+/// @return the original value of *stringp
+char *asepcb(char **stringp, char delim, char stop);
+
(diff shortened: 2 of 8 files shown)


merbanan:
Looks good to me. Some fprintf might need a guard.

zuckschwerdt:
Thanks. Feels better not to include a lib for this (which might actually be on a remote device with the gps dongle and not on the host).
I still need to add JSON parsing and include-filters to support e.g. `-K gpsd://127.0.0.1:2947,lat,lon,alt` -- currently it's just JSON or NMEA dump…

zuckschwerdt:
Looking good now. The most likely use cases are given in the `-K` help output:

- `-K gpsd,lat,lon` (report lat and lon keys from local gpsd)
- `-K gpsd` (full json TPV report, default "gps" key)
- `-K foo=gpsd://127.0.0.1:2947` (with key and address)
- `-K bar=gpsd,nmea` (NMEA GPGGA report)
- `-K rmc=gpsd,nmea,filter='$GPRMC'` (NMEA GPRMC report)

Also <tag> can be a generic tcp address, e.g.
- `-K foo=tcp:localhost:4000` (read lines as TCP client)
- `-K bar=tcp://127.0.0.1:3000,init="subscribe tags\\r\\n"`
- `-K baz=tcp://127.0.0.1:5000,filter="a prefix to match"`

zuckschwerdt:
Ah well, we might need to add renaming of tag fields as the gps report includes "time" which you might want, to check if the fix is recent enough -- but obviously "time" is already taken. So either a `,gps_time=time` syntax or the whole tag inside a json object perhaps `-K gps_data=gpsd,lat,lon,time`. Not sure.

merbanan:
I think signal time should be time always. By default you get the gps_time key and with an option gps_time replaces time.

zuckschwerdt:
Sorry, my example was misleading. "time" is the signal time and should stay that way. But we copy keys from the gpsd json data, e.g. "lat", "lon". And there is a "time" key in the gpsd json data, which could be interesting as it is the time of the fix.
(It might be a very old fix if the gpsd gets disconnected, the rtl_433 tag reader doesn't care if you stop updating the tags and will the most recent one with not timeout.)
But we can't just allow the user to copy the gpsd "time" key, it needs to be named something else. Hence the idea of not just a `,key` syntax but optional `,renamed_key=key` -- but I think wrapping the whole gps tag data in an (sub-)object is nicer.

PR #1637, opened 2021-01-30
Added support for Holtek HT12E devices

socram8888:
This PR adds support for Holtek HT12E ([datasheet here](https://www.farnell.com/datasheets/1899539.pdf)), a very simple remote controller that sends 12 fixed bits.


zuckschwerdt:
Thanks for analyzing and developing a decoder. This (a push button remote control) should be easier to do as flex decoder conf file.
Keep the PR but commit instead a documented .conf file. You can find good examples for this kind of decoding in https://github.com/merbanan/rtl_433/tree/master/conf

PR #1639, opened 2021-01-31
fineoffset WH0290 minor improvements

### LINKED ISSUE #1638: WH0290 minor improvement suggestion
@harpchad   @peteakalad Thanks for your WH0290 work.

Minor suggestions:
  *  b[0] is not checked or printed.   It seems like it is family and is supposed to be 0x8d.   I would think that either it should be printed, or it should be checked and fail the match if not, but I may be missing something.  (In general I am a fan of for every bit, either insist it have some value to match, or if not decode it, even if it's just putting the bit in an unknown field.)
  * The upper bit of battery MSB is not checked or printed.   Would suggest "ub1".
  * The bitbench example shows "Bat_MSB:1d  Bat_LSB:2d" but that is 110 or 6 which is 1.2, and that seems odd and not matching what is supposed to be.
  * Something seems strange with the "PM10" field.  In my data (only a bit so far), it seems that PM10 = PM2.5 +1, and I see that in the examples.   I wonder if that's another form of coding and not PM10, or if people are seeing different values.   I don't have a crisp suggestion for figuring this out.

gdt:
This pull request has two not-really-related commits to the WH0290 decoder
  * improve comments
  * decode family and the unknown bit, to aid in future understanding

For what it's worth, I am seeing family=65 unknown1=0 with an EcoWitt WH41.

Resolves #1638.

### DIFF
--- a/src/devices/fineoffset.c
+++ b/src/devices/fineoffset.c
@@ -385,7 +385,9 @@ static int fineoffset_WH0290_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     }
 
     // Decode data
+    uint8_t family    = b[0];
     uint8_t id        = b[1];
+    uint8_t unknown1  = (b[2] & 0x80) ? 1 : 0;
     int pm25          = (b[2] & 0x3f) << 8 | b[3];
     int pm100         = (b[4] & 0x3f) << 8 | b[5];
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks. "family" and "unknown1" should be last (but before "mic"). That way we get a common display for "kv".

PR #1648, opened 2021-02-07
Support for Cavius alarms

SteveCooling:
Decoder for Cavius wireless smoke/heat/water alarms (www.cavius.com)

There seems to be some endianness issues (cavius.c:89), not quite sure what the preferred way to handle that is.


zuckschwerdt:
Some code style cleanup along the checks is needed ;)

Endianess: use shift and mask, like your `net_id_big` as you do know where the bits in the message are. You do not know where the bits in a `uint32_t` are.

src/devices/cavius.c
```diff
@@ -0,0 +1,163 @@
+/** @file
+*   Cavius smoke alarm decoder.
+*
```
zuckschwerdt, line 3: The `@file` comment should contain the copyright and licence. The next `/**` comment should document the decoder function.

src/devices/cavius.c
```diff
@@ -0,0 +1,163 @@
...
+*   CAVIUS_MESSAGE_MUTE     = 0x04
+*   CAVIUS_MESSAGE_UNKNOWN2 = 0x02
+*   CAVIUS_MESSAGE_UNKNOWN1 = 0x01
+*
+*   Sometimes the receiver has to be at 250ksps to decode. Don't know why.
+*/
+
+#include "decoder.h"
+
+const uint8_t CAVIUS_MESSAGE_PAIRING  = 0x80;
```
zuckschwerdt, line 32: you need an `enum` here.

src/devices/cavius.c
```diff
@@ -0,0 +1,163 @@
...
+    //printf("CRC_calced: %d OK: %d\n", crc_calced, crcok);
+
+    /* clang-format off */
+    data = data_make(
+            "netid",         "Net ID",      DATA_INT,    net_id_big,
+            "senderid",      "Sender ID",   DATA_INT,    sender_id_big,
+            "message",       "Message",     DATA_INT,    message,
+            "text",          "Description", DATA_STRING, text,
+            "crc",           "CRC-8",       DATA_INT,    crc,
+            "crcok",         "CRC OK",      DATA_INT,    crcok,
```
merbanan, line 134: Switch to MIC as in other decoders.

SteveCooling:
> Some code style cleanup along the checks is needed ;)
> 
> Endianess: use shift and mask, like your `net_id_big` as you do know where the bits in the message are. You do not know where the bits in a `uint32_t` are.

So, ditch the `bitbuffer_extract_bytes()` and get the parts I need directly from `databits.bb[0]` ?

zuckschwerdt:
Yes, add `uint8_t *b = databits.bb[0];` after `bitbuffer_invert()`, then just use e.g.
`int message = b[3];` and
`uint32t net_id = ((uint32_t)b[0] << 24) | (b[1] << 16) | (b[2] << 8) | (b[3]);`

Also the checksum for CRC is the remainder of the division, so
`crc8le(b, 7, 0x31, 0x0) == 0;` (note 7 instead of 6, and the poly is without the leading 1 bit).

SteveCooling:
> Also the checksum for CRC is the remainder of the division, so
> `crc8le(b, 7, 0x31, 0x0) == 0;` (note 7 instead of 6, and the poly is without the leading 1 bit).

Seems like I might have transplanted someone else's error :joy:

We did a python implementation while reversing the devices, and the CRC-settings were taken from `crc-8-maxim` here: http://crcmod.sourceforge.net/crcmod.predefined.html

Will do some testing and neccessary corrections as soon as time allows.

src/devices/cavius.c
SteveCooling: Should be OK now

src/devices/cavius.c
SteveCooling: Fixed

src/devices/cavius.c
SteveCooling: Done.

zuckschwerdt:
Note that we use the common CRC poly notation of "lower bits", not the full n+1 bits, these are the same:
- 0x131 (9 bit)
- 0x31 (8 bit, lower)
- 0x98 (Koopman notation, 8 bit upper)

src/devices/cavius.c
zuckschwerdt: `/** @file` not `/*** @file`
and there can be only one comment block (`/**`) for the decoder function, indent is 0 spaces.

SteveCooling: As far as I can tell I have complied with this now

SteveCooling:
I hope it's all good now ;-)

zuckschwerdt:
Looks mostly ok. The doc comment is markdown, i.e. `NNNNMMCSSSS` needs an indent of four spaces, the two lists need a dash in front of each item.

The `static const uint8_t SYNC[]` should be `uint8_t const preamble[]`.

`.name        = "Cavius",` should be "Cavius smoke, heat, and water detector"

We need to bring the output keys to our standard set:
- Needs a "model" key, e.g. "Cavius-Security"
- Which of "netid" "senderid" is the fixed "id" of the sending device?
- Esp. Battery low should report a `"battery_ok" 0` output. Search for DATA_COND to see how this works.

SteveCooling:
> * Which of "netid" "senderid" is the fixed "id" of the sending device?

* Net ID is the ID of the master device (The one others are paired to when they are connected together)
* Sender ID is the actual sending device

zuckschwerdt:
Ok, then "sender_id" should be just "id". "netid" should probably be "net_id" as that's more in line with other keys we have.

SteveCooling:
> * Esp. Battery low should report a `"battery_ok" 0` output. Search for DATA_COND to see how this works.

Not exactly sure how the devices treat the battery low condition. They might set the `battlow` in all messages when the condition is true. I have assumed that this is true, and thus I treat the message field as a bitmap and set the `battery_ok` field accordingly.

zuckschwerdt:
We exclude the battery low bit on message matching now. Should be fixed for that case, not sure if multiple bits could really be set otherwise?

Issue #781, opened 2018-07-30
Ford TPMS: Values missing


UhtredTheBold:
Hi,

I'm trying to read the tire pressure from my Ford and I have picked up the TPMS signal with the built in decoder [89] but I'm not seeing any values being returned:

-r -F json

`{"time" : "@0.524288s", "model" : "Ford", "type" : "TPMS", "id" : "452f111f", "code" : "8d4c05", "mic" : "CHECKSUM"}`

Is this something I'm doing wrong or is the decoder not able to read this particular model of the sensor?

example -a -t capture - [https://1drv.ms/u/s!AilDCtYZODMhisQpypCco1IH_LKpFQ](url)

merbanan:
https://github.com/merbanan/rtl_433/blob/master/src/devices/tpms_ford.c

2 bytes in the message payload are unknown. That part has not been decoded because we don't know how to do it.

zuckschwerdt:
Perhaps you could take samples under known conditions and annotate the captures? Although that is quite an untertaking to get right…

zuckschwerdt:
If you get a chance, please also compare the ID. (could be that we got it backwards or inverted…) Maybe the car unit can display the connected IDs.

mioke77:
Any results?
I have these values read:
2018-11-24 16:29:39 :   Ford    :       TPMS    :       45b955ab        :       904606  :       CHECKSUM
2018-11-24 16:29:39 :   Ford    :       TPMS    :       45b955ab        :       904606  :       CHECKSUM
2018-11-24 16:29:40 :   Ford    :       TPMS    :       45b955ab        :       904606  :       CHECKSUM
2018-11-24 16:29:40 :   Ford    :       TPMS    :       45b955ab        :       904606  :       CHECKSUM
2018-11-24 16:31:43 :   Ford    :       TPMS    :       45b9ad88        :       974506  :       CHECKSUM
2018-11-24 16:31:43 :   Ford    :       TPMS    :       45b9ad88        :       974506  :       CHECKSUM
2018-11-24 16:31:43 :   Ford    :       TPMS    :       45b9ad88        :       974506  :       CHECKSUM
2018-11-24 16:31:43 :   Ford    :       TPMS    :       45b9ad88        :       974506  :       CHECKSUM
2018-11-24 16:41:27 :   Ford    :       TPMS    :       45b906f2        :       944506  :       CHECKSUM
[...]

PR #1654, opened 2021-02-19
add pressure and temperature value

Krezalis:
https://github.com/merbanan/rtl_433/issues/781

### DIFF
--- a/src/devices/tpms_ford.c
+++ b/src/devices/tpms_ford.c
@@ -10,7 +10,7 @@
 */
 /**
 FSK 8 byte Manchester encoded TPMS with simple checksum.
-Seen on Ford Fiesta, Focus, ...
+Seen on Ford Fiesta, Focus, Kuga, Escape ...
 
 Packet nibbles:
 
@@ -34,6 +34,11 @@ static int tpms_ford_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsigned 
     char id_str[9];
     int code;
     char code_str[7];
+    float pressure;
+    float pressure_bar;
+    char pressure_str[20]; //64.05 psi ~4.40 bar
+    int temperature;
+    char temperature_str[6]; //125°C
 
     bitbuffer_manchester_decode(bitbuffer, row, bitpos, &packet_bits, 160);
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! I'll test in the next few days.

src/devices/tpms_ford.c
```diff
@@ -52,12 +57,21 @@ static int tpms_ford_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsigned
...
+    temperature = b[5]-64;
+    sprintf(temperature_str, "%d%s", temperature, "°C");
 
     data = data_make(
         "model",        "",     DATA_STRING, "Ford",
         "type",         "",     DATA_STRING, "TPMS",
         "id",           "",     DATA_STRING, id_str,
         "code",         "",     DATA_STRING, code_str,
+        "pressure",         "",     DATA_STRING, pressure_str,
+        "temperature",         "",     DATA_STRING, temperature_str,
```
zuckschwerdt, line 74: make this `"temperature_C", "", DATA_INT, temperature,` please. remove "temperature_str"

src/devices/tpms_ford.c
```diff
@@ -52,12 +57,21 @@ static int tpms_ford_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsigned
...
+    
+    temperature = b[5]-64;
+    sprintf(temperature_str, "%d%s", temperature, "°C");
 
     data = data_make(
         "model",        "",     DATA_STRING, "Ford",
         "type",         "",     DATA_STRING, "TPMS",
         "id",           "",     DATA_STRING, id_str,
         "code",         "",     DATA_STRING, code_str,
+        "pressure",         "",     DATA_STRING, pressure_str,
```
zuckschwerdt, line 73: make this `"pressure_kPa",`... please. remove "pressure_str"

src/devices/tpms_ford.c
```diff
@@ -52,12 +57,21 @@ static int tpms_ford_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsigned
 
     code = b[4]<<16 | b[5]<<8 | b[6];
     sprintf(code_str, "%06x", code);
+    
+    pressure = 0.3 + b[4] * 0.25;
```
zuckschwerdt, line 61: is this perhaps `(b[4] + 1) * 0.25f` ?

zuckschwerdt:
Manually merged. The output is very likely wrong. We really need a table of infaltion/deflation codes and a range of temperatures.

src/devices/tpms_ford.c
Krezalis: No. Tested: when 00 forscan read from BdyCM value 2.1kPa. This is not correct because no pressure and outbut must be 0. And there are different step and start value.

Krezalis:
> 
> 
> Manually merged. The output is very likely wrong. We really need a table of infaltion/deflation codes and a range of temperatures.

Pressure value is correct: received on air packed and related with information which vehicle send over CAN. Temperature is not tested because module not send temperature information, but this formula look good: tested in house temp ~24C, outdoor ~0, and value from comment. Distance is related to temperature change in C, F is distance 1.8.

PR #1655, opened 2021-02-20
Update README.md MacOS install

drfruct:
I created a [formula](https://formulae.brew.sh/formula/rtl_433#default) to install rtl_433 on MacOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ Official [binary builds for Windows](https://bintray.com/chzu/dist/rtl_433) (32 
 
 On Debian (sid) or Ubuntu (19.10+), `apt-get install rtl-433` for other distros check https://repology.org/project/rtl-433/versions
 
+On MacOS, `brew install rtl_433`.
+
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Is it possible to add optional dependencies or to just add SoapySDR and OpenSSL (TLS) support (not sure what the usually policy is: minimal/optimal deps?)
CMake flags should be: `-DENABLE_RTLSDR=ON -DENABLE_SOAPYSDR=ON -DENABLE_OPENSSL=ON`

drfruct:
There are no Homebrew formula for SoapySDR and I'm not sure what I can add `-DENABLE_SOAPYSDR=ON` flag.
I'll read how I can add the rest of the CMake flags.

lawendel:
Actually, now you can find a BREW soapySDR formula (https://formulae.brew.sh/formula/soapysdr). Would it be possible to add support for a soapysdr compatible device like Airspy R2? Many thanks for your input!

Andy

PR #1661, opened 2021-02-27
Change Ford TPMS pressure calculation

Krezalis:
fix pressure formula


zuckschwerdt:
Are you sure the sensor can not report `0 PSI`? Lowest report is `0.3 PSI`?

Krezalis:
Yes, i bind sensors and read value from BdyCM

Issue #1534, opened 2020-10-25
Decoder protocol 123 not working for my Jansite external TPMS sensors


riegelbrau:
Hallo team, many thanks for this exciting project! I am using it already for several temperature and humidity sensors from Klimalogg and others. That is working well!
For my home brewing equipement I'm trying to build a fermentation pressure monitoring setup, like Andreas Spiess showed in his videos (very cool project!). So I bought the same Jansite TPMS sensors that Andreas referenced in issue #1007. They display their values on the included car unit, but I am not able to receive any signals with rtl_433 to get the decoded information. So I appreciate any help. 
I put one TPMS sensor in a pressure chamber to force sensor value changes, I gathered some data using the parameters from #1007 and uploaded the files here. Please respond if I should run special commands.

Thanks in advance.

[TPMS_out.txt](https://github.com/merbanan/rtl_433/files/5435468/TPMS_out.txt)
[Jansite_different.OOK.txt](https://github.com/merbanan/rtl_433/files/5435469/Jansite_different.OOK.txt)

zuckschwerdt:
Your collection of codes looks good, the OOK shows the same 52µs timing with MC data. But the codes seem longer: 82 bit instead of the expected 56.
[...]

PR #1663, opened 2021-02-28
Add decoder for jansite solar tpms

merbanan:
Jansite solar tpms #1534 

Try with -R 180 -s 1000k.


enavarro222:
Thanks a lot for that work !
I got a small project on a shelve with that hardware for a while with no time to spend on it... I just try : it works ! 🍾

PR #1665, opened 2021-03-01
Update mqtt hass to include triggers

klattimer:
Improved the example for the hass bridge. The problems here for integration included
 - No triggers to create actions on hass, only sensor readings were sent
 - The paths and schema were a bit messed up, I think hass might have changed in the past, so this is catching up


examples/rtl_433_mqtt_hass.py
```diff
@@ -111,7 +111,7 @@
 
 # Fields that get ignored when publishing to Home Assistant
 # (reduces noise to help spot missing field mappings)
-SKIP_KEYS = NAMING_KEYS + [ "time", "mic", "mod", "freq", "sequence_num",
+SKIP_KEYS = NAMING_KEYS + [ "mic", "mod", "freq", "sequence_num",
```
deviantintegral, line 114: I haven't personally found the time sensors to be useful. Are there types of readings where they are?

examples/rtl_433_mqtt_hass.py
```diff
@@ -514,6 +522,60 @@
...
+    },
+    "button": {
+        "device_type": "device_automation",
+        "object_suffix": "BTN",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
```
deviantintegral, line 542: This is great! I will test is specifically soon.

examples/rtl_433_mqtt_hass.py
```diff
@@ -560,17 +622,19 @@ def sanitize(text):
 def rtl_433_device_topic(data):
     """Return rtl_433 device topic to subscribe to for a data element"""
 
-    path_elements = []
+    #path_elements = []
```
deviantintegral, line 625: This would need to be removed entirely before merging. What exactly does this fix, any suggestions on how to test it?

examples/rtl_433_mqtt_hass.py
```diff
@@ -560,17 +622,19 @@ def sanitize(text):
...
-        if key in data:
-            element = sanitize(str(data[key]))
-            path_elements.append(element)
+    #for key in NAMING_KEYS:
+    #    if key in data:
+    #        element = sanitize(str(data[key]))
+    #        path_elements.append(element)
 
-    return '/'.join(path_elements)
+    uid = data.get('id')
```
deviantintegral, line 632: Is this to mean "unique id"?

examples/rtl_433_mqtt_hass.py
```diff
@@ -560,17 +622,19 @@ def sanitize(text):
...
+    if uid is None: uid = 0
+    return 'ID'+str(uid) # '/'.join(path_elements)
```
deviantintegral, line 634: To be cleaned up before merging.

deviantintegral (commented):
Looks like this conflicts with master. I left some notes, but what is is needed so we can get this merged soon?

zuckschwerdt:
Someone needs to test this and likely strip it somewhat down, e.g. I don't think the path change from model to protocol is viable as the protocol isn't usually visible.

zuckschwerdt:
To get "protocol" in the output you need to start with `-M protocol`, which is suppose is not something users normally enable.
It would be interesting to know which protocols don't have a "model" output? It should really be everywhere.

deviantintegral:
This looks to error out as is when not specifying `-M protocol`:

```
Starting rtl_433_mqtt_hass.py...
INFO:root:MQTT connected: Connection Accepted.
INFO:root:Subscribing to: rtl_433/+/events
Exception in thread Thread-1:
  File "/usr/local/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 3591, in _thread_main
    self.loop_forever(retry_first_connection=True)
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 1756, in loop_forever
    rc = self._loop(timeout)
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 1164, in _loop
    rc = self.loop_read()
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 1556, in loop_read
    rc = self._packet_read()
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 2439, in _packet_read
    rc = self._packet_handle()
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 3033, in _packet_handle
    return self._handle_publish()
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 3327, in _handle_publish
    self._handle_on_message(message)
  File "/usr/local/lib/python3.9/site-packages/paho/mqtt/client.py", line 3570, in _handle_on_message
    on_message(self, self._userdata, message)
  File "/rtl_433_mqtt_hass.py", line 633, in mqtt_message
    bridge_event_to_hass(client, topicprefix, data)
  File "/rtl_433_mqtt_hass.py", line 709, in bridge_event_to_hass
    model = 'P' + str(data["protocol"])
KeyError: 'protocol'
```

deviantintegral:
I can't push to this branch, but I pared this back to just the device triggers and I was able to see my doorbell register 🎉 .

Here's a diff against `master`. Should I open a new PR for it?

```diff
diff --git a/examples/rtl_433_mqtt_hass.py b/examples/rtl_433_mqtt_hass.py
index 706d9437..bccf3ebf 100755
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -111,7 +111,7 @@ NAMING_KEYS = [ "type", "model", "subtype", "channel", "id" ]
 
 # Fields that get ignored when publishing to Home Assistant
 # (reduces noise to help spot missing field mappings)
-SKIP_KEYS = NAMING_KEYS + [ "time", "mic", "mod", "freq", "sequence_num",
+SKIP_KEYS = NAMING_KEYS + [ "mic", "mod", "freq", "sequence_num",
                             "message_type", "exception", "raw_msg" ]
 
 
@@ -164,7 +164,15 @@ mappings = {
             "state_class": "measurement"
         }
     },
-
+    "time": {
+        "device_type": "sensor",
+        "object_suffix": "UTC",
+        "config": {
+            "device_class": "timestamp",
+            "name": "Timestamp",
+            "icon": "mdi:clock-in"
+        }
+    },
     "battery_ok": {
         "device_type": "sensor",
         "object_suffix": "B",
@@ -537,6 +545,61 @@ mappings = {
         }
     },
 
+    "channel": {
+        "device_type": "device_automation",
+        "object_suffix": "CH",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+    "button": {
+        "device_type": "device_automation",
+        "object_suffix": "BTN",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+    "code": {
+        "device_type": "device_automation",
+        "object_suffix": "CODE",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+    "unit": {
+        "device_type": "device_automation",
+        "object_suffix": "unit",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+    "tristate": {
+        "device_type": "device_automation",
+        "object_suffix": "tristate",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+    "cmd": {
+        "device_type": "device_automation",
+        "object_suffix": "CMD",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+
 }
 
 
@@ -592,8 +655,7 @@ def rtl_433_device_topic(data):
 
     return '/'.join(path_elements)
 
-
-def publish_config(mqttc, topic, model, instance, mapping):
+def publish_config(mqttc, topic, model, instance, mapping, value=None):
     """Publish Home Assistant auto discovery data."""
     global discovery_timeouts
 
@@ -615,10 +677,16 @@ def publish_config(mqttc, topic, model, instance, mapping):
     discovery_timeouts[path] = now + args.discovery_interval
 
     config = mapping["config"].copy()
-    config["name"] = object_name
-    config["state_topic"] = topic
-    config["unique_id"] = object_name
-    config["device"] = { "identifiers": object_id, "name": object_id, "model": model, "manufacturer": "rtl_433" }
+
+    if device_type == 'device_automation':
+        config["topic"] = topic
+        config["platform"] = 'mqtt'
+        config["payload"] = value
+    else:
+        config["state_topic"] = topic
+        config["unique_id"] = object_name
+        config["name"] = object_name
+    config["device"] = { "identifiers": [object_id], "name": object_id, "model": model, "manufacturer": "rtl_433" }
 
     if args.force_update:
         config["force_update"] = "true"
```

zuckschwerdt:
If @klattimer wants to trim down this PR that would be preferred. But I'm really not sure about some things here.

Is "time" a valueable source of information? It's just infrequent meta data on other events, right?
I'm not sure keys like "code" or "tristate" will stay around, there should be a note "experimental" or such.
Looking up where/why we use "unit" a saw that "channel" is maybe abused in Nexa/Proove -- it should be a fixed part of the sender identification, not a target selection, not a variable input. We need to address that someday.
What does it mean that publish_config sends a data value now? Isn't this script meant to just output config?

klattimer:
Sorry guys I was out of the loop here for a bit. If my PR needs updating, please let me know what needs doing and I'll try to rattle through it today. 

Incidentally, my script for launching hass integration and rtl433 is this, there's no protocol included, so not sure what's going on with that.

#!/bin/bash
python3 /usr/local/bin/rtl_433_mqtt_hass.py -d -H localhost -R "sensors/rtl_433/events" -i 10 &
/usr/local/bin/rtl_433

klattimer:
The final line in my rtl433 config is this - which might be why 

`output mqtt://192.168.1.2:1883,retain=0,devices=sensors/rtl_433/P[protocol:255]/ID[id:0],events=sensors/rtl_433/events,states=sensors/rtl_433/states`

zuckschwerdt:
Most interesting points would be: why/how are you using "time"?
Are you using channel with a specific device as an input, as mentioned it should (usually) be a fixed part of the ID?

I guess "code", "unit", "tristate", "cmd" are used like "button" as "something was pressed on a remote"-kind of input, or is there more to it? Is this where "channel" is used? We should then change it to maybe "group"?

What is the "value" in publish_config used for? Isn't this all "static" config?

It's suspect using "protocol" as "model" won't generally work well. A protocol is an internal abstraction and might have multiple models as output. The "model" keys were somewhat messy in the past, but we have them all normlized, short and safe now.

klattimer:
OK, first question. time comes through as an entity in HASS - and I wanted to know how long ago door sensors/motion sensors were triggered, the button events are basically, any and every other value sent to MQTT because I didn't bother to look too hard at it. It was "I need these things to work so, send em all and sort it later". 

I do remember (as this was ages ago now) that model was causing the HASS integration to crash pretty hard so that's why I swapped out protocol. Might not be the right thing to do, but it works so I went with it.

Other than that, I'm not 100% sure what the rest of the code does exactly, I submitted this PR almost a year ago, if someone had picked it up sooner I would probably be able to work on it with them, as it goes I'm kinda busy with other stuff at the minute.

klattimer:
For model it might have been that HASS was unhappy with special characters or spaces or something... Not sure.

zuckschwerdt:
Yeah, pretty sure the old model keys were a mess. You could find slashes, ampersand and such in there. Not anymore!

I will take a look to find out what is needed to support button-remotes. And perhaps change some things there.

Basically you want "all the keys that signal a button press". Can you list the devices (or protocols) that (at some time) where relevant to this? I guess Nexa, Proove, ...?

klattimer:
You're stretching my memory beyond human limits. 

For buttons you need to use the automation type trigger, that was originally the most important missing thing for motion sensors and door open sensors to work. 

What I'd like is the correct remote button push value to be received by RTL433 as with what I currently have it's a mess, on/off codes are identical and a whole section of my RF remote just produces the same output.

zuckschwerdt:
> on/off codes are identical and a whole section of my RF remote just produces the same output.

Is that a hardware issue, a decoder issue, or an issue with the MQTT HASS discovery?
I guess that last one? We perhaps need a compound single output, so instead of
`channel: 2, unit: 3, button: 4` something like `button: 234`, right?

klattimer:
what about the state? 2340 for off, 2341 for on... following your example?

zuckschwerdt:
I wasn't sure about that either. Are both different triggers or do they toggle a state? I guess I need to read up on Binary Sensors, Automation Trigger and https://www.home-assistant.io/integrations/device_trigger.mqtt/

> Note that only one trigger may be defined per unique discovery topic.

But

> Also note that the combination of type and subtype should be unique for a device.

So maybe should emit two discovery topics and use "payload" and "subtype" turn_on, turn_off

zuckschwerdt:
If I read that right we need:
- for every protocol that might be a trigger
- each trigger needs to be a MQTT topic on it's own
- all triggers needs to be found by the script

I.e. we need to find the protocols that could be a trigger and normalize the output, e.g. maybe add a type "trigger".

The MQTT output needs to transform the output from trigger protocols to form a unique topic per trigger, e.g.
`.../trigger/[group]/[unit]/[button]` or `.../[group]/[unit]/[button]/trigger`

The script can then find those and hand out config for a `button_short_release`. That would work for e.g. PIR, smoke detectors and simple remotes. But not for (binary) switches. We would need something more to autoconfigure e.g. `turn_on`/`turn_off` instead of `button_short_release`. We might use e.g. "switch" instead of "trigger" in the topic, but generally protocols won't know what buttons to expect e.g. from a remote.

In any case: every new trigger randomly appearing will add auto config to HASS, right? Somebody using a RF-remote nearby would add new items.

klattimer:
"In any case: every new trigger randomly appearing will add auto config to HASS, right? Somebody using a RF-remote nearby would add new items." - Yes, actually this is important. I added a tweak on my hass script which only sends the config to MQTT if the device has sent a signal more than 5 times. I get a lot of stray signals from temperature sensors near by and I wanted to limit that. 

In reality it's probably better to have a threshold within a timeframe. It's easy enough to trigger your own sensors 5 times someone else's, not so much.

examples/rtl_433_mqtt_hass.py
deviantintegral: Since I posted this, I've now seen the use. There are sensors that may send a message periodically even if they haven't changed. Having an entity for this in Home Assistant allows us to easily know when the message was last received, and not just changed value.

deviantintegral:
Lots of discussion here, but it's been a few months so let's see if we can get a todo list:

> Is "time" a valueable source of information? It's just infrequent meta data on other events, right?

- [ ] As @klattimer and I mentioned, it is! Let's roll back those changes in this PR.

> Yeah, pretty sure the old model keys were a mess. You could find slashes, ampersand and such in there. Not anymore!

- [ ] Seems like this bit was off the need of this PR, and likely fixed now. Make sure there's no protocol / model changes in this PR.

> Yes, actually this is important. I added a tweak on my hass script which only sends the config to MQTT if the device has sent a signal more than 5 times.

No action item, but a note; we changed the docs for the HA addon to only suggest running the autodiscovery script when actively trying to add devices. That should limit adding transient devices, while not requiring some perhaps tricky logic for "seen X messages".

deviantintegral:
I've pushed a new PR which contains all commits in this branch to https://github.com/merbanan/rtl_433/pull/2102.

gdt:
If this PR is really replaced by #2102 perhaps best to close this one.

PR #1666, opened 2021-03-07
Added website and sensor specifications for Jansite TPMS Model Solar

freundm:


zuckschwerdt:
Thanks. First line should be a title and end with a dot and two newlines. (just put the link in an extra line).
We want to stick to 7bit ascii in source files, "±" and "°" need to be replaced.

freundm:
Fixed.

I pulled the ° from [src/devices/tpms_jansite.c](https://github.com/merbanan/rtl_433/blob/master/src/devices/tpms_jansite.c#L14), I can remove that as well if you want.

merbanan:
I have now merged the decoder, can you rebase your PR ?

AutomGuy:
Hi Guys,
i have a question. what is the send intervall of the jansite solar TPMS?

Many thx in advance

PR #1671, opened 2021-03-12
Add Amazon Basics Meat Thermometer decoder

merbanan:


merbanan:
@zuckschwerdt do you have a suggestion for a good model name?

zuckschwerdt:
Maybe `Basics-Meat` as the distributor and kind are easily implied.

PR #1674, opened 2021-03-13
Auriol AFT 77 B2

Heimetli:


Heimetli:
What is "Error: STICKY-PAREN error"?

zuckschwerdt:
E.g.
```
for( int row = 0; row < bitbuffer->num_rows; row++ )
```
should be
```
for (int row = 0; row < bitbuffer->num_rows; row++) {
```
Try clang-format if you are unsure about formatting ;)

src/devices/auriol_aft77b2.c
```diff
@@ -0,0 +1,151 @@
...
+of 1900us length.
+The packets are ppm modulated (distance coding) with a pulse of ~488 us
+followed by a short gap of ~488 us for a 0 bit or a long ~976 us gap for a
+1 bit, the sync gap is ~1170 us.
+
+The data is grouped in 17 nibbles
+
+    [prefix] [0x05] [0x0C] [id0] [id1] [0x00] [flags] [sign] [temp0] [temp1] [temp2]
+    [0x00] [0x00] [sum] [sum] [lsrc] [lsrc]
+
```
merbanan, line 24: Can you add a row as an example?

src/devices/auriol_aft77b2.c
```diff
@@ -0,0 +1,151 @@
...
+        frame[i] = (ptr[i] << 4) | (ptr[i+1] >> 4) ;
+
+    // Check the sum
+    if( sum(frame,6) != frame[6] )
+       return DECODE_FAIL_SANITY;
+
+    // Check the lsrc
+    if (lsrc(frame,6) != frame[7])
+       return DECODE_FAIL_SANITY;
+
```
merbanan, line 129: I think we are missing 2 types of lfsr_digest8 class function and that this needs one of them. We should add the support for that later. And revdgst should output the command options needed for implementation.

zuckschwerdt:
- we have `sum()` as `add_bytes()`.
- don't pad statement termination `;` with spaces
- remove `.disabled = 0,`
- don't use `_X()`, model should be `Auriol-AFT77B2`
- put space after commas and around operators please.

(what styleguide did you refer to when sticking parens to control flow keywords and space padding arguments and semicolon but not operators? We mostly adhere to https://en.wikipedia.org/wiki/Indentation_style#Variant:_Stroustrup)

Heimetli:
It doesn't work any more, because add_bytes returns an int

src/devices/auriol_aft77b2.c
```diff
@@ -0,0 +1,154 @@
...
+            NULL);
+    /* clang-format on */
+
+    decoder_output_data(decoder, data);
+    return 1;
+}
+
+static char *output_fields[] = {
+        "model",
+        "temperature_C",
```
merbanan, line 141: id is missing here.

zuckschwerdt:
> It doesn't work any more, because add_bytes returns an int

truncating the result with `(uint8_t)add_sum()` is fine, other decoders commonly do that.

merbanan:
Merged in 11b510e. Some cleanups remain.

Issue #1640, opened 2021-01-31
FineOffset WH0290: PM10 is apparently not really PM10


gdt:
The WH0290 decoder produces two 14-bit fields, PM2.5 and PM10.  So far, it seems like it is always true that
 * PM10 = PM2.5 + 1 (PM2.5 is 0 to 14)
 * PM10 = PM2.5 + 2 (28 to ?)
 * PM10 = PM2.5 + 3 (>=110?)

This doesn't make sense from a physical viewpoint of the definitions of PM10 and PM2.5, and the device is advertised only as a PM2.5 monitor.  However, EcoWitt says that it has a "Honeywell HPM Series Particle Sensor" and that is documented to do both PM2.5 and PM10.

Obviously more understanding is needed, and this issue is just to note that the situation is unclear.

peteakalad:
I did some digging on these honeywell HPM series particle sensors, and their datasheet does state the following (my emphasis):

> 2 PM1.0 in μg/m3, PM4.0 in μg/m3, and PM10 in μg/m3 **are calculated from PM 2.5 readings.**

Source:
https://sensing.honeywell.com/honeywell-sensing-particulate-hpm-series-datasheet-32322550.pdf

So the PM2.5 is a proper read reading and PM10 is as we suspected simply derived/calculated from the PM2.5 value.

gdt:
That's useful info.   That shows that the PM10 output from the module is bogus, and it's still unclear to me where that second 14 bits comes from.   But It's not pretty clear that it cannot be useful.   The datasheet lacks a reference to a scientific paper that shows there is merit to the concept of estimating PM10 from PM2.5 :-(

peteakalad:
The other bits will just be passed back from the sensor.

merbanan:
I added a comment in the code about this. Improvements are welcome. Closing issue for now.

gdt:
I don't think we should be outputting data that we know is wrong.    Would you be amenable to a PR that relabels PM10 to some sort of "unknown" label?

peteakalad:
Not sure 'unknown' would be appropriate, its known - as confirmed above to be a PM10 value calculated/derived from the PM2.5 value.

The PM10 value is what the honeywell sensor outputs, rtl433 is not calculating it.   There must be some science behind honeywell's decision to derive it like it does.  Its not necessarily wrong all the time.  It may not be of value to us, but to someone a PM2.5 calculated PM10 value may be better than no PM10 value at all.

[...]

PR #1678, opened 2021-03-16
fineoffset: Add explanation of supposed PM10 value, and call it estimated

gdt:
After discussion in https://github.com/merbanan/rtl_433/issues/1640
this commit adjusts the comment about the WH0290 particulate sensor,
mostly surrounding the nature of the field labeled PM10.

The measurement of the sensor serial port was done by @peteakalad, and
the determination of the pm2.5/pm10 conversion was done by @gdt.

Following consensus from #1640, change description to call it "estimated", as a compromise betwen calculated and alleged.

### DIFF
--- a/src/devices/fineoffset.c
+++ b/src/devices/fineoffset.c
@@ -331,19 +331,51 @@ static int fineoffset_WH24_callback(r_device *decoder, bitbuffer_t *bitbuffer)
 Fine Offset Electronics WH0290 Wireless Air Quality Monitor
 Also: Ambient Weather PM25
 Also: Misol PM25
-Also: EcoWitt WH41
+Also: EcoWitt WH0290, EcoWitt WH41
 
-The sensor sends a package each ~10m. The bits are PCM modulated with Frequency Shift Keying.
+The sensor sends a data burst every 10 minutes.  The bits are PCM
+modulated with Frequency Shift Keying.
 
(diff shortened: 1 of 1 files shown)


src/devices/fineoffset.c
```diff
@@ -411,7 +443,7 @@ static int fineoffset_WH0290_callback(r_device *decoder, bitbuffer_t *bitbuffer)
             "id",               "ID",           DATA_INT,    id,
             "battery_ok",          "Battery Level",  DATA_FORMAT, "%.1f", DATA_DOUBLE, battery_ok,
             "pm2_5_ug_m3",      "2.5um Fine Particulate Matter",  DATA_FORMAT, "%i ug/m3", DATA_INT, pm25/10,
-            "pm10_0_ug_m3",     "10um Coarse Particulate Matter",  DATA_FORMAT, "%i ug/m3", DATA_INT, pm100/10,
+            "estimated_pm10_0_ug_m3",     "Estimate of 10um Coarse Particulate Matter",  DATA_FORMAT, "%i ug/m3", DATA_INT, pm100/10,
```
zuckschwerdt, line 446: needs same change in output_fields_WH25

gdt: Fixed, I think - but I didn't understand until just now.

PR #1686, opened 2021-04-07
Update holman_ws5029.c

grahamf72:
:fix
Line 98 - the most significant 4 bits of the rain should only be shifted left by 8 bits, not by 12 bits as per the original. The incorrect shift by 12 bits, causes it to incorrectly report rainfall of ~3000mm every time the 8th bit overflows.


zuckschwerdt:
Good catch. This has always been wrong, since this was added! Can you remove the comment?

grahamf72:
> Good catch. This has always been wrong, since this was added! Can you remove the comment?

I hope that worked - I am only a beginner with git.

PR #1687, opened 2021-04-08
Add X10 Dim, Bright, All Lights ON, and All Off commands

dfleck:
Adds remaining X10 commands. Adds debugging. Cleaned up data verification. Adjusted timing to meet specs.

### DIFF
--- a/src/devices/lacrosse_tx35.c
+++ b/src/devices/lacrosse_tx35.c
@@ -15,6 +15,8 @@ LaCrosse/StarMétéo/Conrad TX35DTH-IT, TFA Dostmann 30.3155     Temperature/Hum
 LaCrosse/StarMétéo/Conrad TX29-IT, TFA Dostmann 30.3159.IT     Temperature Sensors.
 Tune to 868240000Hz
 
+LaCrosse TX25U Temperature/Temperature Probe at 915 MHz
+
 Protocol
 ========
 Example data : https://github.com/merbanan/rtl_433_tests/tree/master/tests/lacrosse/06/gfile-tx29.cu8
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
Looks good, thanks!
- line 11 (mods) should be line 5
- can you put `clang-format on/off` around the data_make (see other decoders)
- should "ALL LTS ON" be "ALL ON"?

dfleck:
> Looks good, thanks!
> 
> * line 11 (mods) should be line 5
> * can you put `clang-format on/off` around the data_make (see other decoders)
> * should "ALL LTS ON" be "ALL ON"?

First two changes made.
There's two types of modules, Light and Appliance. ALL OFF turns all of both off. ALL LTS ON turns all Light modules on but leave Appliance modules alone.

dfleck:
Sorry, I'm a bit new to this. I intended the addition of the LaCrosse TX25U to be a separate pull request to keep things clean but the changes ended up being added to this request.

zuckschwerdt:
Ok :) Looks good too. We'll fix this on merge.

dfleck:
What are the "rules" regarding enabling a device? The RF protocol for X10 is almost identical to X10_Security which is enabled, so I don't think there are any particularly bad interactions with other devices at the signal level. I have tested X10 with several controllers I have and they seem to work as expected. Should I have enabled the X10 device?

zuckschwerdt:
> What are the "rules" regarding enabling a device?

Checksum. There has to be a check to distinguish random bits from a proper transmission. X10_sec has 13 parity bits. Not sure why there is mention of "CRC" (maybe because the notion of xor?), the `mic` should actually be "SUM", i.e. the "strength" of a sum, mor than just "PARITY" but not "CRC".

merbanan:
> What are the "rules" regarding enabling a device?

That there is enough logic to reject bogus/random transmissions.

dfleck:
In that case I've gone ahead and enabled the X10 device. There are three quality checks:
1) The packet is the proper length
2) The data is two bytes but sent as four. The extra two bytes are the complement of the "regular" two bytes. There is a check that byte 0 is the complement of byte 1 and byte 2 is the complement of byte 3.
3) Some bits are always zero or always one. These are validated.

PR #1695, opened 2021-04-16
Add option to set force_update for all sensors

joegross:
If your sensor values change infrequently and you prefer to write the most recent value even if not changed.

This is useful if you're graphing the sensor data or want to alert on missing data.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -55,6 +55,10 @@ Getting rtl_433 devices back after Home Assistant restarts will happen
 more quickly if MQTT retain is enabled. Note however that definitions
 for any transitient devices/false positives will retained indefinitely.
 
+If your sensor values change infrequently and you prefer to write the most
+recent value even if not changed set -f to append "force_update = true" to
+all configs. This is useful if you're graphing the sensor data or want to
+alert on missing data.
 
 Suggestions:
 
@@ -535,6 +539,9 @@ def publish_config(mqttc, topic, model, instance, mapping):
     config["unique_id"] = object_name
     config["device"] = { "identifiers": object_id, "name": object_id, "model": model, "manufacturer": "rtl_433" }
 
+    if args.force_update:
+        config["force_update"] = "true"
+
     if args.debug:
         print(path,":",json.dumps(config))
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks. Looks good. Unfortunatly there is kind of a pile up of other changes, not sure what the implications are.
@rct and @maxlock are more familiar with the script.

rct:
This looks fine to me.  I still haven't had time to pull the various PRs and other changes people have submitted into one clean PR that can be applied to the current base.   

@zuckschwerdt I think this is self contained enough that you could merge this.  There is merging that has to be done anyway. This won't complicate that too much.

@joegross - Just curious what you are using this on in Hass?  I've seen the option, but I don't understand the implications from the docs on the MQTT page. I guess I haven't noticed a need for it so far.

I suspect this is also one of those things that if you have enough devices, you'll probably want to selectively enable this option. I've got to imagine this causes extra rows to be added to the recorder database, where the default behavior for hass is to try to suppress those if the state hasn't changed.

joegross:
> @joegross - Just curious what you are using this for in Hass? I've seen the option, but I don't understand the implications from the docs on the MQTT page. I guess I haven't noticed a need for it so far.

I have a collection of Acurite temperature sensors throughout the house as well as a bunch of 986 sensors in frdges/freezers. I use hassio to send the data to influx with a grafana.

The 986 sensor, in particular, measures temperature in 1F increments, and in a deep freezer it can stay within a 1F range for hours at a time.

I have grafana alerts on temperature alert (did the freezer lose power or did I leave the door open?) but there's no way to differentiate between the sensor losing power or signal and the temperature being stable. My solution is to force_update and then grafana can alert on "no data".

> I suspect this is also one of those things that if you have enough devices, you'll probably want to selectively enable this option. I've got to imagine this causes extra rows to be added to the recorder database, where the default behavior for hass is to try to suppress those if the state hasn't changed.

I agree with that, though influx is extremely efficient in storing data so in my case there's no harm in storing duplicate data points. I saw force_update was enabled on a seemingly random selection of sensors, so short of a larger refactor I didn't see a good way of selecting this option without imposing a new default on others.

rct:
@joegross - thanks for the info.

I have 986 sensors - I wrote the decoder for them, but I haven't used them in alerts yet (on my list). I can usually see the freezer's on/off and defrost cycles as long as the sensor isn't sitting on top of some big solid frozen thing.

I was thinking more about the Hass recorder database than InfluxDB. I've had issues before with the recorder database (especially when it was on an SD card.)

Trying to refactor the auto discovery script to do things in a model specific way is on my list, but it's not very high up at the moment.

PR #1699, opened 2021-04-21
normalize use of .long_width = 0 with OOK_PULSE_MANCHESTER_ZEROBIT

nunojpg:


zuckschwerdt:
I guess the idea was to have MANCHESTER match PCM with short=long indicating NRZ. I.e. "pulse length" is short and "symbol length" is long (but note that symbol here is a half-bit only after decoding).

This looks better though :)

PR #1704, opened 2021-04-25
Add Tesla charge port opener decoder

nunojpg:
Sample: https://github.com/merbanan/rtl_433_tests/pull/392

### DIFF
--- a/conf/tesla_charge-port-opener.conf
+++ b/conf/tesla_charge-port-opener.conf
@@ -1,0 +1,19 @@
+# Tesla charge port opener
+#
+#  * When the button on the charge handle is pressed the signal is repeated at 0.15s intervals for 10 times.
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Can you add some comments to the top (similar to other conf files)? From your notes, like:
- r=1200 and t=20 are minimum values confirmed to work with this hardware, but larger values should be used for tolerance.
- The relevant signal is repeated 10 times, while there is in fact a different and unique leading signal. This might be nuisance on circuit start.
- There are no unique codes or transmissions variants. The code is the same for all Tesla charge port handles and there is a single button.
- The transmitter hardware is reported to be using a Si4010.
- There is always 2 rows = 094aa9b38da19 and 1 row = 094aa9b38da18.

nunojpg:
@zuckschwerdt thanks, pushed.

PR #1705, opened 2021-04-27
Roja Flex support

Hofyyy:
Supported Device:
- Remote
- Window Shutter

Extra feature:
- Signal Cloner / Generator
   Could be activated in source.
   Take a real message, extract HOMEID and generate all possible command for all 15 channels:
   This is helpful if you what use an own bridge and you need the commands.
   
Review Feedback:

> - Copyright line missing?
> - Decoder doc should be just Markdown (no leading // ), first line the device name.
> - data_make() should be wrapped with /* clang-format off/on */

done

> - those short functions are preferred to be inline code (nothing in the decoder is reuseable...)

done. Two big functions are still there. Which is think helps.

> preamble is best shortened to match truncated start, i.e. 0xaa 0xaa 0xd3...

Not done, because this will change all offsets, and in my tests this works great.
I understand the idea behind this point. Maybe I can rewrite it in a second round with
changed offsets.

> - you would usually check for the required minimum (bit)length, then truncate -- we don't mind if there is garbage at the end.

Question: I only read 19 bytes (which is the message) what should I truncate?

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,49 +223,49 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [142]  Fine Offset Electronics/ECOWITT WH51 Soil Moisture Sensor
     [143]  Holman Industries iWeather WS5029 weather station (older PWM)
     [144]  TBH weather sensor
-    [145]  Rojaflex
-    [146]  WS2032 weather station
-    [147]  Auriol AFW2A1 temperature/humidity sensor
-    [148]  TFA Drop Rain Gauge 30.3233.01
-    [149]  DSC Security Contact (WS4945)
-    [150]  ERT Standard Consumption Message (SCM)
-    [151]* Klimalogg
-    [152]  Visonic powercode
-    [153]  Eurochron EFTH-800 temperature and humidity sensor
-    [154]  Cotech 36-7959 wireless weather station with USB
-    [155]  Standard Consumption Message Plus (SCMplus)
-    [156]  Fine Offset Electronics WH1080/WH3080 Weather Station (FSK)
-    [157]  Abarth 124 Spider TPMS
-    [158]  Missil ML0757 weather station
-    [159]  Sharp SPC775 weather station
-    [160]  Insteon
-    [161]  ERT Interval Data Message (IDM)
-    [162]  ERT Interval Data Message (IDM) for Net Meters
-    [163]* ThermoPro-TX2 temperature sensor
-    [164]  Acurite 590TX Temperature with optional Humidity
-    [165]  Security+ 2.0 (Keyfob)
-    [166]  TFA Dostmann 30.3221.02 T/H Outdoor Sensor
-    [167]  LaCrosse Technology View LTV-WSDTH01 Breeze Pro Wind Sensor
-    [168]  Somfy RTS
-    [169]  Schrader TPMS SMD3MA4 (Subaru)
-    [170]* Nice Flor-s remote control for gates
-    [171]  LaCrosse Technology View LTV-WR1 Multi Sensor
-    [172]  LaCrosse Technology View LTV-TH Thermo/Hygro Sensor
-    [173]  Bresser Weather Center 6-in-1, 7-in-1 indoor, new 5-in-1, 3-in-1 wind gauge, Froggit WH6000, Ventus C8488A
-    [174]  Bresser Weather Center 7-in-1
-    [175]  EcoDHOME Smart Socket and MCEE Solar monitor
-    [176]  LaCrosse Technology View LTV-R1 Rainfall Gauge
-    [177]  BlueLine Power Monitor
-    [178]  Burnhard BBQ thermometer
-    [179]  Security+ (Keyfob)
-    [180]  Cavius smoke, heat and water detector
-    [181]  Jansite TPMS Model Solar
-    [182]  Amazon Basics Meat Thermometer
-    [183]  TFA Marbella Pool Thermometer
-    [184]  Auriol AHFL temperature/humidity sensor
-    [185]  Auriol AFT 77 B2 temperature sensor
-    [186]  Honeywell CM921 Wireless Programmable Room Thermostat
-    [187]  Hyundai TPMS (VDO)
+    [145]  WS2032 weather station
+    [146]  Auriol AFW2A1 temperature/humidity sensor
+    [147]  TFA Drop Rain Gauge 30.3233.01
+    [148]  DSC Security Contact (WS4945)
+    [149]  ERT Standard Consumption Message (SCM)
+    [150]* Klimalogg
+    [151]  Visonic powercode
+    [152]  Eurochron EFTH-800 temperature and humidity sensor
+    [153]  Cotech 36-7959 wireless weather station with USB
+    [154]  Standard Consumption Message Plus (SCMplus)
+    [155]  Fine Offset Electronics WH1080/WH3080 Weather Station (FSK)
+    [156]  Abarth 124 Spider TPMS
+    [157]  Missil ML0757 weather station
+    [158]  Sharp SPC775 weather station
+    [159]  Insteon
+    [160]  ERT Interval Data Message (IDM)
+    [161]  ERT Interval Data Message (IDM) for Net Meters
+    [162]* ThermoPro-TX2 temperature sensor
+    [163]  Acurite 590TX Temperature with optional Humidity
+    [164]  Security+ 2.0 (Keyfob)
+    [165]  TFA Dostmann 30.3221.02 T/H Outdoor Sensor
+    [166]  LaCrosse Technology View LTV-WSDTH01 Breeze Pro Wind Sensor
+    [167]  Somfy RTS
+    [168]  Schrader TPMS SMD3MA4 (Subaru)
+    [169]* Nice Flor-s remote control for gates
+    [170]  LaCrosse Technology View LTV-WR1 Multi Sensor
+    [171]  LaCrosse Technology View LTV-TH Thermo/Hygro Sensor
+    [172]  Bresser Weather Center 6-in-1, 7-in-1 indoor, new 5-in-1, 3-in-1 wind gauge, Froggit WH6000, Ventus C8488A
+    [173]  Bresser Weather Center 7-in-1
+    [174]  EcoDHOME Smart Socket and MCEE Solar monitor
+    [175]  LaCrosse Technology View LTV-R1 Rainfall Gauge
+    [176]  BlueLine Power Monitor
+    [177]  Burnhard BBQ thermometer
+    [178]  Security+ (Keyfob)
+    [179]  Cavius smoke, heat and water detector
+    [180]  Jansite TPMS Model Solar
+    [181]  Amazon Basics Meat Thermometer
+    [182]  TFA Marbella Pool Thermometer
+    [183]  Auriol AHFL temperature/humidity sensor
+    [184]  Auriol AFT 77 B2 temperature sensor
+    [185]  Honeywell CM921 Wireless Programmable Room Thermostat
+    [186]  Hyundai TPMS (VDO)
+    [187]  Rojaflex
 
 * Disabled by default, use -R n or -G
 
--- a/README.md
+++ b/README.md
@@ -223,48 +223,49 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [142]  Fine Offset Electronics/ECOWITT WH51 Soil Moisture Sensor
     [143]  Holman Industries iWeather WS5029 weather station (older PWM)
     [144]  TBH weather sensor
-    [145]  WS2032 weather station
-    [146]  Auriol AFW2A1 temperature/humidity sensor
-    [147]  TFA Drop Rain Gauge 30.3233.01
-    [148]  DSC Security Contact (WS4945)
-    [149]  ERT Standard Consumption Message (SCM)
-    [150]* Klimalogg
-    [151]  Visonic powercode
-    [152]  Eurochron EFTH-800 temperature and humidity sensor
-    [153]  Cotech 36-7959 wireless weather station with USB
-    [154]  Standard Consumption Message Plus (SCMplus)
-    [155]  Fine Offset Electronics WH1080/WH3080 Weather Station (FSK)
-    [156]  Abarth 124 Spider TPMS
-    [157]  Missil ML0757 weather station
-    [158]  Sharp SPC775 weather station
-    [159]  Insteon
-    [160]  ERT Interval Data Message (IDM)
-    [161]  ERT Interval Data Message (IDM) for Net Meters
-    [162]* ThermoPro-TX2 temperature sensor
-    [163]  Acurite 590TX Temperature with optional Humidity
-    [164]  Security+ 2.0 (Keyfob)
-    [165]  TFA Dostmann 30.3221.02 T/H Outdoor Sensor
-    [166]  LaCrosse Technology View LTV-WSDTH01 Breeze Pro Wind Sensor
-    [167]  Somfy RTS
-    [168]  Schrader TPMS SMD3MA4 (Subaru)
-    [169]* Nice Flor-s remote control for gates
-    [170]  LaCrosse Technology View LTV-WR1 Multi Sensor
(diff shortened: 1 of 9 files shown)


zuckschwerdt:
We usually want the fields "id" and "channel". Are "homeid", "radiochannel" these? Can we rename them?

> this will change all offsets

usually we add the length of the preamble to `start_pos` so `bitbuffer_extract_bytes()` puts the first data byte a `0`.

> only read 19 bytes (which is the message) what should I truncate?

I thought `msg_bitcount` is used to treat the messages as variable length. I now see it's just for with/without CRC. Ok.

Try `clang-format` if you can. Otherwise at least: `switch( command )`-> `switch (command)` etc.

Hofyyy:
> We usually want the fields "id" and "channel". Are "homeid", "radiochannel" these? Can we rename them?

Yes.
If "channel" means a radio channel normally that fits. It can bis 1-15 or 0 for all.
Id seems to be ok too, if that means a static id which is used for one "remote" in one home installation.
It seems to be a second id, which I named token, which is different per message.

So long story short, I think the rename is fine.

> Try clang-format if you can. Otherwise at least: switch( command )-> switch (command) etc.

I will try. but I have no clue how to update this magic push request.
I will do my best.

zuckschwerdt:
No worries then, I'll just merge manually and do the style updates and id/channel myself.

Hofyyy:
So next round. Renamed and reformatted.

Hofyyy:
Passt alles so? :-)

Hofyyy:
Ping :-)

zuckschwerdt:
There are a few things to change still, I need to get around to merge and change this.
- E.g. it should be `"mic", "Integrity", DATA_COND, has_crc, DATA_STRING, "CRC",`
- Key order should be model, id, channel, ...
- `model` needs to be a fixed string with no whitespace
- if the same id can be multiple deviceType then that (model) info should be `subtype`, otherwise multiple `model` lines with `DATA_COND`

Can you run maintainer_update.py and commit those changes?

Hofyyy:
Danke für das Feedback. Ich gucke mal ob ich die maintriner_update.py finde.

zuckschwerdt:
Thanks. Oops, put the device **last** (above `/* Add new decoders here. */`) in ` include/rtl_433_devices.h` and rerun please.

Hofyyy:
I did the move. Sorry I cloned one decoder, and used same position to know what I have to change. So I did not see the "right" pos. But the maintainer_update.py seems to not do any further change now.

Hofyyy:
No problem. Looks good now.

Hofyyy:
Hello, because of the reason that you have knowledge in the 433 area. Maybe you have a hint for us in the following topic. https://github.com/RFD-FHEM/RFFHEM/issues/955. With the help of your tool and the new plugin I try to do the same for the FHEM server in receive and transmit direction. But until now I was not able to receive a message from the remote. Do you know the 1011 chip? Is it easy to transfer information from a rtl422 analyse into settings for the ti1011 chip?

merbanan:
Not automatically but you can use the analyzer output values as a starting point for the CC1101 chip configuration.

Hofyyy:
> Not automatically but you can use the analyzer output values as a starting point for the CC1101 chip configuration.
I hope the questions are not to stupid, but my knowledge starts when I receive the hex values.
- I wrote the decoder, so I know the frequency, the sync word (0xD391D391), Deviation 20KHz (433,90 -> 433,94), 

I do not know the modulation (2FSK, GFSK, MSK)
I am not sure in case of bandwidth and Datarate (Assume, 110KHz and 10 KHZ)

merbanan:
@Hofyyy well if you are trying to replicate the signal then just record a sample with the signal grabber (-S) then run it through the analyzer (-A). Then compare the output with your replicated signal. When they look the same you are done. The GFSK vs 2FSK parameter can be found by looking at the signal spectra of a recorded signal. If the frequency instantly goes from the low to the high frequency it is a 2FSK signal, if it smoothly transitions it is GFSK. My bet is on GFSK.

zuckschwerdt:
With MSK the data rate is tied to the deviation for narrow bandwidth, so these parameters look good. https://en.wikipedia.org/wiki/Minimum-shift_keying
Perhaps record your CC1101 signal the same way you record the original, then compare the spectrum (https://triq.org/pdv/) and actual bits decoded with `-w file.ook` and also pdv.

Hofyyy:
Thanks for the support. I will try to learn more about it. At the moment this is like a foreign language for me. Could that be easy answered from your side with an UHR capture in complex16s? :-P

merbanan:
I prefer rtl_sdr 8-bit output. rtl_433 -S unknown should record unknown signals for you automatically.

Hofyyy:
> I prefer rtl_sdr 8-bit output. rtl_433 -S unknown should record unknown signals for you automatically.

To get that right. rtf_433 with the Roja Flex decoder plugin is perfectly able to decode each bit. But I am not able to decode the same signal with another board which would be used in forte for TX too.

I did attache a capture and a picture. Thanks in advanced.

[g002_433.92M_250k.cu8.zip](https://github.com/merbanan/rtl_433/files/6446534/g002_433.92M_250k.cu8.zip)
115456720-c655d100-a223-11eb-80f3-ed447cc8d22a

Hofyyy:
> I prefer rtl_sdr 8-bit output. rtl_433 -S unknown should record unknown signals for you automatically.

Sorry, for the questions. I flashed a second 1011 and it worked. So all questions are obsolete. 

Can I do something to help in regards to this pull request?

zuckschwerdt:
- it should be `"mic", "Integrity", DATA_COND, has_crc, DATA_STRING, "CRC",`
- Key order should be model, id, channel, ...
- `model` needs to be a fixed string with no whitespace
- if the same id can be multiple deviceType then that (model) info should be `subtype`, otherwise multiple `model` lines with `DATA_COND`

if you're up for that ;)

zuckschwerdt:
>     * it should be "mic", "Integrity", DATA_COND, has_crc, DATA_STRING, "CRC",
>       That I do not understand, because I know the type of the CRC ... so why not use my implementation?

Those are fixed keys: https://triq.org/rtl_433/DATA_FORMAT.html#message-data

>     * if the same id can be multiple deviceType then that (model) info should be subtype, otherwise multiple model lines with DATA_COND
>       That I do not understand, as I understand the protocol correct, there is no device ID. It is more a global ID (key for the remote), which is valid for the whole installation. So the remote has the ID + 15 channels. And a shutter will be trained on a ID + CH.

A single remote should have a fixed "model" and "id". You can't switch the model on each message, also never use model=unknown.

Hofyyy:
I did the CRC renaming to have the same string like the other plugins.

It would be great if you could do the ID change like you want.
At the moment I see a static key from the Remote and then I decide
in regards do the command of it is an answer from the shutter or not.
And I can see if it a bridge (clone from the remote) because the bridge 
is not sending an CRC ... Line 234 to 246. 

I think you are much faster then me to make the adjustment.

Hofyyy:
One more question, because my impression is you are an expert in this area.

Right in front of the 16 Bit CRC is some magic 16 bit token.
Do you see a chance to check if you see how to calculate it?

I generated all possible messages I know, see attached file.
[all_remote_commands_v2.docx](https://github.com/merbanan/rtl_433/files/6468666/all_remote_commands_v2.docx)

zuckschwerdt:
Well, the last one is just a sum, starting one byte after the sync words (0x31 in your examples).
The one before that could be a parity check or xor-sum, but the one row with the `ea0102` obviously does not fit :/

Hofyyy:
> Well, the last one is just a sum, starting one byte after the sync words (0x31 in your examples).
> The one before that could be a parity check or xor-sum, but the one row with the `ea0102` obviously does not fit :/

Sorry I did not get it. Could you give me one more detailed example?
08 31 22 fd 20 0a 01      0a 85
How is the first stuff a sum which gives a 85?

zuckschwerdt:
Just add the bytes, like [this BitBench](https://triq.net/bitbench?c=d391d391083122fd200a010a85b37a&c=d391d391083122fd201a011aa593bc&c=d391d391083122fd208a018a85b34c&c=d391d391083122fd209a019aa5938a&c=d391d391083122fd20da01da251092&c=d391d391083122fd20ea01025d8187&c=d391d391083122fd210a010a86330b&c=d391d391083122fd211a011aa613cd&c=d391d391083122fd218a018a86333d&c=d391d391083122fd219a019aa613fb&c=d391d391083122fd21da01da2690e3&c=d391d391083122fd21ea01025e01f6&c=d391d391083122fd220a010a87b386&c=d391d391083122fd221a011aa79340&c=d391d391083122fd228a018a87b3b0&c=d391d391083122fd229a019aa79376&c=d391d391083122fd22da01da27106e&c=d391d391083122fd22ea01025f817b&c=d391d391083122fd230a010a8833df&c=d391d391083122fd231a011aa81319&c=d391d391083122fd238a018a8833e9&c=d391d391083122fd239a019aa8132f&c=d391d391083122fd23da01da289037&c=d391d391083122fd23ea0102600182&c=d391d391083122fd240a010a8932b1&c=d391d391083122fd241a011aa91277&c=d391d391083122fd248a018a893287&c=d391d391083122fd249a019aa91241&c=d391d391083122fd24da01da299159&c=d391d391083122fd24ea01026100ec&f=SYNC8h8h8h8h%208h%208h%2B8h%2B8h%2B8h%2B8h%2B8h%2B8h%3D8h%20CRC8h8h&cf=ADD&co=40&cl=56&cw=8).

README.md
```diff
@@ -274,7 +275,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
   [-d <RTL-SDR USB device index>] (default: 0)
   [-d :<RTL-SDR USB device serial (can be set with rtl_eeprom -s)>]
 	To set gain for RTL-SDR use -g <gain> to set an overall gain in dB.
-	SoapySDR device driver is available.
+	SoapySDR device driver is not available.
```
merbanan, line 278: Stray change from the maintainer update script.

man/man1/rtl_433.1
```diff
@@ -151,7 +151,7 @@ RTL\-SDR device driver is available.
 To set gain for RTL\-SDR use \-g <gain> to set an overall gain in dB.
 .RE
 .RS
-SoapySDR device driver is available.
+SoapySDR device driver is not available.
```
merbanan, line 154: Same here.

src/devices/rojaflex.c
```diff
@@ -0,0 +1,317 @@
...
+
+        /* clang-format off */
+        data = data_make(
+                    "model",        "Model",     DATA_STRING, deviceType,
+                    "id",           "ID",        DATA_STRING, id,
+                    "channel",      "Channel",   DATA_INT,    msg[CHANNEL_OFFSET] & 0xF,
+                    "token",        "Msg Token", DATA_STRING, tokenString,
+                    "commandtype",  "Command ",  DATA_STRING, getCommandString(&msg[0]),
+                    "commandvalue", "Value",     DATA_INT,    msg[COMMAND_VALUE_OFFSET],
+                    "mic",          "Integrity", DATA_STRING, (dataframe_bitcount == DATAFRAME_BITCOUNT_INCL_CRC) ? "CRC" : "",
```
merbanan, line 206: Only include mic if there is a message integrity check. Add it conditionally.

Look here how to do it:
https://github.com/merbanan/rtl_433/blob/master/src/devices/ambientweather_tx8300.c#L114

src/devices/rojaflex.c
```diff
@@ -0,0 +1,317 @@
...
+                COMMAND_ID_REQUESTSTATUS};
+
+        uint8_t channel = GENERATE_COMMANDS_FOR_CURRENT_CHANNEL ? msg[CHANNEL_OFFSET] & 0xF : 0;
+        uint8_t command = 0;
+        uint8_t msg_new[19];
+        uint16_t sum = 0;
+
+        fprintf(stderr, "\n");
+        fprintf(stderr, "%s: Signal cloner\n", __func__);
+        fprintf(stderr, "%s: \n", __func__);
```
merbanan, line 233: printouts need to be around verbose guards.

Here you can see how to do it:
https://github.com/merbanan/rtl_433/blob/master/src/devices/ambientweather_tx8300.c#L69

src/devices/rojaflex.c
```diff
@@ -0,0 +1,317 @@
...
+*/
+
+/**
+Signal documentation
+
+Default signal layout
+0xaaaaaaaa d391d391 SS KKKKKK ?CDDDD TTTT CCCC
+
+4 Bytes Preamble
+4 Bytes Sync Word
```
merbanan, line 29: Can you add it as a comment? I assume it is d391d391.

Hofyyy: done

src/devices/rojaflex.c
Hofyyy: done

I did not do this before, because this is special code, which you need to activate by a define. This will not happen during normal usage.

src/devices/rojaflex.c
Hofyyy: Done

man/man1/rtl_433.1
Hofyyy: It seems to be that the *.1 is a local generated file? 
And its coming from #ifdef SOAPYSDR
            "\tSoapySDR device driver is available.\n"
#else
            "\tSoapySDR device driver is not available.\n"
#endif

I can correct the Readme.md or what should I do?

Hofyyy:
@merbanan Could you bring this through the door? I will not be able to provide much during this week.

zuckschwerdt:
I'll be merging this, there are some final touches needed. Maybe on the weekend.

Hofyyy:
Sorry for pushing, but the RTL software is great and I would like to have official Rojaflex support :-)

zuckschwerdt:
My bad. I'll get on it.

zuckschwerdt:
Merged, but note that there are some changes in the output to better represent the fields.

zuckschwerdt:
@merbanan I've updated the code style and merged this, but I feel the model should be fixed to the protocol as "RojaFlex-Remote" with an additonal key of the decoded "device" as "Remote"/"Bridge"/"Shutter"/"Unknown". What do you think?

Hofyyy:
Only for information, at the moment it can decode messages from the remote and from the shutter. and you can detect that from the message type. I do not know if that fits to your question.

zuckschwerdt:
Yes, we do want a key to show "Remote"/"Bridge"/"Shutter"/"Unknown". But best case that should not be the "model" key, we use that more like a protocol-type information.

Hofyyy:
Ok, I do not know the original Protocol specification, so I only can assume it.
It seems to be that the only key which shows it is the following:

e.g.
if ((msg[COMMAND_ID_OFFSET] & 0xF) == 0x5) {
            deviceType = "RojaFlex-Shutter";
}

zuckschwerdt:
With "keys" I'm referring to our common set of output fields, s.a. https://triq.org/rtl_433/DATA_FORMAT.html
We currently use e.g. `"model", "RojaFlex-Shutter"` but it should rather be
```
"model", "RojaFlex"
"subtype", "Shutter"
```
that's my thought anyway ;)

Hofyyy:
Its fine for me :)

zuckschwerdt:
Great! The convention for "model" values is "Xyz-Abc". We usually call advanced "remote control"-type protocols "Xy-Security". Some like the Nexa/Proove are generic "command to group-unit", so this would fit. Model values are not meant to convey a specific use-case anyway. But we could perhaps call this "RojaFlex-Automation", maybe. Not sure.

Hofyyy:
I am not sure how to implement this in a quick way. I will check it next weekend. If you would like to change it, because you know how to do it. Feel free.

zuckschwerdt:
No worries, I'll change that as needed. Just waiting for @merbanan to help out with:
- split in "model" and "subtype"?
- model of "RojaFlex-Security" (as usual) or "RojaFlex-Automation" or maybe just "RojaFlex"?

merbanan:
The main aim in the output was always to make it descriptive and as unique as possible. Adding subtype makes sense. There might be other cases where this can be useful.

Issue #1701, opened 2021-04-21
Question Sgooway Door Sensors


rourke750:
I recently purchased https://www.aliexpress.com/i/486746257.html door sensors and noticed strange behavior from rtl_433.  So it seems to match the [86]  Wireless Smoke and Heat Detector GS 558 decoder but I noticed that when it first starts im able to trigger the sensor and it correctly states that it received the signal.  What is weird though is after a couple tries it will no longer detect the sensor being triggered.  The only way it will then start seeing the signal is to use a second door sensor and then activate it and then immediately activate the second one.  Going forward this is the only way that rtl picks up the signal being fired.  Any ideas about this?

zuckschwerdt:
The GS558 decoder is prone to false positives. The protocol matches simple "doorbell"-style transmissions.

To debug the problems with your senders it's best to capture the transmissions (`-S all`) or a defined period (`-w myfile_433.92M_250k.cu8 -T 30` for 30 seconds) and then review them on https://triq.org/pdv/

rourke750:
After doing that, is there a way to locally create decoders or do I need to submit a pull request to the test repo and have someone else make one?

zuckschwerdt:
For simple sensors like this a flex decoder is the best way to go. Analyze the timing with `-A` then look in the conf folder for similar examples to copy. Submit the .conf file as PR or text here if you can, we always want more decoders ;)

rourke750:
So I got this https://triq.org/pdv/#AAB104011803E4017427148181818181928192928192928181928181928181818192928355 is this all I need?

zuckschwerdt:
There should have been a `-X` line. Something like `-R 0 -X 'n=sgooway,m=OOK_PWM,s=280,l=990,r=2000'` should work.

rourke750:
Is this for analyzing the timings or for generating a decoder?

zuckschwerdt:
That's already a flex decoder, the `-A` analyzer should show that suggestion. It will give you the signal as byte data. You can add `match=SOMECODE` and other things, see the [conf](https://github.com/merbanan/rtl_433/tree/master/conf) directory for examples.

rourke750:
Ah I see.  So I can use that data to make a flex decoder that I can add to my config so that it only picks up those devices.

I have two devices so far one being 
model     : sgooway      count     : 1             num_rows  : 1
rows      : 
len       : 24           data      : fa4dbc
codes     : {24}fa4dbc

[...]

PR #1707, opened 2021-05-03
added sgooway door detector

rourke750:


rourke750:
In reference to https://github.com/merbanan/rtl_433/issues/1701

zuckschwerdt:
Thanks! I'd say this is a "x1527-like OTP encoder" device, but not an actual x1527. The sync pulse is missing (bits=24, but should be bits=25 for x1527).
Can you confirm it works with bits=24 but does not work with bits=25, and if so change the info?

rourke750:
So it does work with 25 bits but I noticed it would send a bunch of requests where at 24 it only gets one.  Based on the page it says the ic code is 1527.  Is the sync bit important, should I change my config to 25 and see what happens?

zuckschwerdt:
I see. It is x1527 then :)
The single code with 24 bits is bad/truncated, the sync is missing. All other codes are proper 25 bits.

You should use bits=25 to avoid problems and try out `,unique`(suppress duplicate row output) or `,countonly` (suppress detailed row output).

rourke750:
Switched to 25 bits and added unique:
Looks like this
decoder {
  name=sgooway,
  modulation=OOK_PWM,
  short=280,
  long=990,
  reset=2000,
  bits=25,
  get=@0:{25}:code,
  unique
}
still seeing 20ish messages per broadcast

zuckschwerdt:
The `get` should have a length of `{24}` only. The last bit is the sync.
The reason for so many messages could be the distance between messages, is it longer than reset?

rourke750:
Set it to 24 bits getting significantly fewer.  All messages coming in are within 1 second.

PR #1708, opened 2021-05-03
Add smoke detector subtype to Interlogix

Sunoo:
I had an Interlogix device coming up with the subtype "unknown", after some trial and error I was able to confirm it was a smoke detector.


Sunoo:
Closing this for now, I spotted a second device reporting that same ID, and I only am aware of one Interlogix smoke detector in this house, so maybe it’s something else? I’ll reopen it if I figure out what the other device is.

PR #1712, opened 2021-05-06
Added kPa value transformation

linuxrrze:
While running rtl433 for some time now I noticed lots of `TPMS` devices popping up.

Seems like those are "Tire Pressure Monitoring Systems" values reported as `kPa` instead of `hPa`.

Maybe you'd like to include this.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -210,8 +210,8 @@ mappings = {
         "config": {
             "device_class": "pressure",
             "name": "Pressure",
-            "unit_of_measurement": "hPa",
-            "value_template": "{{ float(value|float) * 10.0 | round(2) }}"
+            "unit_of_measurement": "kPa",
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Is this something to workaround a UI bug in HA? For this application (tire pressure) `kPa` is the correct (scaled) unit.

linuxrrze:
> Is this something to workaround a UI bug in HA? For this application (tire pressure) `kPa` is the correct (scaled) unit.

My first change really was to just pass that to homeassistant (originally they were skipped because of the unknown measurement).
I'll gladly change that back an re-submit the PR.

zuckschwerdt:
Thanks. Yes, if kPa works use that. Just add a commit or rebase this PR.

linuxrrze:
> Thanks. Yes, if kPa works use that. Just add a commit or rebase this PR.

I'll collect some data overnight (to see that homeassistant really can handle kPa) and update this PR tomorrow.

Issue #1644, opened 2021-02-03
Remove Autotools build


zuckschwerdt:
The autoconf/automake build was broken since 14c5b9a (7. October 2019) and nobody complained.
(If RTL-SDR input was enabled, libusb-1.0 was needed but not configured.)
I've fixed it with 9514e16, but I guess there isn't anybody still using this build option.
(also using libusb-1.0 for error messages is now optional with 9514e16)

The Autotools build isn't well maintained and features are missing.
The CMake build also supports traditional Unix Makefiles, as well as e.g. Ninja (which I use, very fast) and more generators like Visual Studio.

Aside: try Ninja with
```
mkdir build ; cd build ; cmake -DFORCE_COLORED_BUILD:BOOL=ON -GNinja .. && cmake --build . -j 10
```

Let's keep this issue open for a while to discuss.

zuckschwerdt:
Also note that the `librtlsdr.pc` pkg-config file from `librtlsdr-dev` on Debian Buster is broken. Automake trips on that. CMake can deal with it. This patch would be needed
```
--- /usr/lib/x86_64-linux-gnu/pkgconfig/librtlsdr.pc	2021-02-14 13:41:07.857851107 +0100
+++ /usr/lib/x86_64-linux-gnu/pkgconfig/librtlsdr.pc	2021-02-14 13:41:32.101618910 +0100
@@ -1,4 +1,4 @@
-prefix=
-exec_prefix=
-libdir=
-includedir=
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${prefix}/lib/x86_64-linux-gnu
+includedir=${prefix}/include
```
Maybe someone wants to report that and get it fixed?

merbanan:
Kill it when you see fit.

Mindavi:
Closed by #1714

PR #1714, opened 2021-05-09
build: Remove Autotools (autoconf, automake)

zuckschwerdt:
Announced in 21.05 release notes. S.a. #1644

Apply in around 2 weeks, if nobody complains.


Mindavi:
I had a quick look at package definitions on repology. Most of them seem to be using cmake, and the ones that aren't should be able to easily update.

+1 from me.

PR #1718, opened 2021-05-21
Fix: bitbuffer_clear() was not clearing syncs_before_row

obones:
This is most useful when allocating a `bitbuffer_t` on heap which requires using `bitbuffer_clear()` to initialize it.

### DIFF
--- a/src/bitbuffer.c
+++ b/src/bitbuffer.c
@@ -18,6 +18,7 @@ void bitbuffer_clear(bitbuffer_t *bits)
 {
     bits->num_rows = 0;
     memset(bits->bits_per_row, 0, BITBUF_ROWS * 2);
+    memset(bits->syncs_before_row, 0, BITBUF_ROWS * sizeof(bits->syncs_before_row[0]));
     memset(bits->bb, 0, BITBUF_ROWS * BITBUF_COLS);
 }
 


zuckschwerdt:
Thanks!

zuckschwerdt:
We don't actually have bitbuffer_t on heap though, or did you notice that somewhere?

obones:
You are right, there is no such usage in this repository.
I noticed it while experimenting general heap allocation for `bitbufer_t` on low memory devices (ESP32) which were running out of stack space.
I pushed this experiment in my own [branch](https://github.com/obones/rtl_433/tree/bitbuffer_on_heap) to avoid loosing this work, but I won't submit a pull request for it as it's not a good idea anyway, the heap is also severely constrained on those devices.
I'll keep experimenting and should I find a viable solution, I'll create a pull request then.

But in the meantime, I'm creating pull requests for items that I stumble across and that help having better all around compatibility.

zuckschwerdt:
The bitbuffer_t could be smaller, if memory serves (pun) only m_bus needs those very long rows.
And it's too small for one protocol that has an excessive preamble. We are thinking about some kind of compression: remove leading 0xaa runs and only count the length. For m_bus we could break into multiple rows and flag an overrun. All that requires quite some work though :/

PR #1719, opened 2021-05-21
fix: Visual Studio does not support VLA

obones:
Such arrays must be allocated with malloc

### DIFF
--- a/src/devices/honeywell_cm921.c
+++ b/src/devices/honeywell_cm921.c
@@ -14,9 +14,7 @@ Honeywell CM921 Thermostat (subset of Evohome).
 868Mhz FSK, PCM, Start/Stop bits, reversed, Manchester.
 */
 
-#include <malloc.h>
 #include "decoder.h"
-#include "fatal.h"
 
 // #define _DEBUG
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Yes and no. We really don't want alloc's in decoders. And it's debug stuff anyway, which should not exists in finished decoders.
Perhaps some kind of `DATA_BINARY` for the data_* functions would be better.

obones:
With the current source code, VS does not compile, so I submitted that PR that helped on my build.
But if you feel that it should be done differently, I'm all for it, it's just that I do not really know how to do any better here

zuckschwerdt:
Are you using the included `vs15` project? Then select the "Release" configuration.
If you are building with CMake "Visual Studio 15 2017" or "Visual Studio 16 2019" generator then use `--config Release`.

gvanem:
> Such arrays must be allocated with malloc.

`alloca()` is a lot faster:
```c
char *tstr = alloca((buf_sz * 2) + 1);
```

But why not simply: `char tstr [256];`?

zuckschwerdt:
We don't know how big a message might get, I guess the limit now is a whole bitrow_t. But 127 bytes (x2+1 hex chars) seems safe.

Can you change the PR to that (just commit or rebase) and add to `i < buf_sz` a check of `&& i < 127` or truncate buf_sz beforehand.

obones:
Just to be clear, you want me to change the existing code to use a fixed length of 127 for `tstr` and add a check that `i` does not go above 127.
This means there is no longer any `malloc`/`alloca` calls. Is that correct?

zuckschwerdt:
Yes, that's what I meant. In the long run I'm considering to add something like DATA_BYTES and DATA_BITS, similar to string or array to automatically print hex or "{n}...". Adds much boilerplate code to all outputs, but it's a common and useful case I guess.

obones:
Ok, this is done, but I made the array 256 bytes long because it contains 2 hex nibbles per input byte.

PR #1721, opened 2021-05-22
[WIN32] emulate alarm()

gvanem:
For the sake of reducing the `#ifdef`-soup around around all the calls to `alarm()` in the main code, 
I've added this simple `win_alarm()` using the [CreateTimerQueueTimer()](https://docs.microsoft.com/en-us/windows/win32/api/threadpoollegacyapiset/nf-threadpoollegacyapiset-createtimerqueuetimer) function in `kernel32.dll`.

This function has been there since Win-XP and works just fine.

### DIFF
--- a/include/compat_alarm.h
+++ b/include/compat_alarm.h
@@ -1,0 +1,30 @@
+/** @file
+    @brief compat_alarm adds an alarm() function for Windows.
+
+    Except for MinGW-w64 when `_POSIX` and/or `__USE_MINGW_ALARM`
+    is defined
+ */
+
+#ifndef INCLUDE_COMPAT_ALARM_H_
+#define INCLUDE_COMPAT_ALARM_H_
+
+#ifdef _WIN32
+#include <windows.h>
+#include <signal.h>
+#include <io.h>    /* alarm() for MinGW is possibly here */
+
+#if !defined(_POSIX) && !defined(__USE_MINGW_ALARM)
+int win_alarm(unsigned seconds);
+#define alarm(sec)  win_alarm(sec)
+#define HAVE_win_alarm
+#endif
+
+/* No SIGUSRx on Windows. Use this unless MinGW-w64
+ * has support for it (untested by me).
+ */
+#if !defined(__USE_MINGW_ALARM)
+#define SIGALRM  SIGBREAK
+#endif
+
+#endif  /* _WIN32 */
+#endif  /* INCLUDE_COMPAT_ALARM_H_ */
--- a/include/compat_time.h
+++ b/include/compat_time.h
@@ -1,5 +1,5 @@
 /** @file
-    compat_time addresses compatibility time and alarm() functions.
+    compat_time addresses compatibility time functions.
 
     topic: high-resolution timestamps
     issue: <sys/time.h> is not available on Windows systems
@@ -12,7 +12,6 @@
 // ensure struct timeval is known
 #ifdef _WIN32
 #include <winsock2.h>
-#include <signal.h>
 #else
 #include <sys/time.h>
 #endif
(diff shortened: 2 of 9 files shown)


zuckschwerdt:
Great stuff! Should this perhaps be compat_alarm.c/h to separate the concerns?

gvanem:
> Should this perhaps be compat_alarm.c/h to separate the concerns?

I can do that. But this requires changes of `CMakeLists.txt`, `Makefile.am`, `*.vcxproj` files and more (?)

gvanem:
Now what? *Needs maintainer_update* !?

zuckschwerdt:
Maybe something about the sort order in the build files? You can try ./maintainer_update.py

zuckschwerdt:
For the compat_alarm.c we might need an exported empty dummy function so the compilation unit isn't empty on non _WIN32.

zuckschwerdt:
Like we had here: https://github.com/merbanan/rtl_433/blob/01a21b2c36115975e7b76df24896f6a835a80b7c/src/compat_time.c

gvanem:
> For the compat_alarm.c we might need an exported empty dummy function

Okay.

gvanem:
> Maybe something about the sort order in the build files? You can try ./maintainer_update.py

I looked at it. I understand nothing. What is this all about:
```
 ./maintainer_update.py
  [ -z "$(git status --untracked-files=no --porcelain)" ]
  shell: /usr/bin/bash -e {0}
```
Only 1 `exit(1)` in the script.

zuckschwerdt:
maintainer_update.py is run to check if it would change anything. If so the test fails. We PRs to be unchanged on maintainer_update.py
If there are faulty changes that script makes we need to fix that.

gvanem:
> If so the test fails.

Which test?

> Can you run maintainer_update.py and does it apply any changes?

Locally or what? How?

Maybe you ought to explain the error-situations in that script a tad bit better?

zuckschwerdt:
> Which test?

The Github Action showing "Build Analyze Check / Needs maintainer_update (pull_request)"

> Locally or what? How?

Do you have Python installed? Then just run the script from the rtl_433 directory.

> Maybe you ought to explain the error-situations in that script a tad bit better?

There is just one possible error: The build files are not what maintainer_update.py would have created.

But that check/error isn't fatal. If this PR is ready and works for you I'll merge anyway.

gvanem:
> Do you have Python installed?

Sure. But I did my PR at Github only.

All checks but this *Need maintainer_update* fails. If this indicates a problem with my changes to `rtl_433.vcxproj*`, 
that could be another CI build option (hint, hint). AFAICS, the VStudio solution isn't part of these jobs.

zuckschwerdt:
I'm thinking that we can get rid of the vs15 build files. The CMake generator for Visual Studio is working fine to create these files as needed. Not sure if Windows developers have CMake around, but then again you'll need to install librtlsdr and other libs, so why require CMake as well.

zuckschwerdt:
Ha, found it: trailing whitespace in vs15/rtl_433.vcxproj:
trailing

gvanem:
> Ha, found it: trailing whitespace in vs15/rtl_433.vcxproj

Aha! As I wrote, I did all this in the Github editor. Why it leaves trailing white-space is beyond me.

zuckschwerdt:
> As I wrote, I did all this in the Github editor.

I never tried that, except for fixing very simple merge conflicts. Is the workflow easier for you or is there a lack of good tools for Windows? I sometimes use the Git integration in Visual Studio, but the buttons are not intuitive to me :/

zuckschwerdt:
One last thing: isn't `#include "compat_alarm.h"` missing from rtl_433.c? It's needed for `#define alarm(sec)  win_alarm(sec)`, right?

gvanem:
>  It's needed for #define alarm(sec) win_alarm(sec), right?

Oh. Forgot that.

PR #1724, opened 2021-05-27
Field names must be comma separated 

obones:
If not, it would create a single item with the two strings concatenated


Mindavi:
LGTM, but hold off merging until https://github.com/merbanan/rtl_433/pull/1735 is merged, since that one would otherwise have to fix the conflicts.

zuckschwerdt:
I'll rebase, no problem.

PR #1730, opened 2021-05-31
Add FreeBSD installation instructions

rosorio:
Now FreeBSD has pre-compiled binaries for rtl_443 :)


zuckschwerdt:
That's great news! Thanks.

PR #1735, opened 2021-06-07
Remove oldmodel macro use

zuckschwerdt:
This removes the deprecated "oldmodel" workaround macro `_X()`.

Regression checks pass, but if anyone can spare some time then please review.

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -144,9 +144,9 @@ static int acurite_rain_896_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
     /* clang-format off */
     data = data_make(
-            "model",                "",             DATA_STRING, _X("Acurite-Rain","Acurite Rain Gauge"),
+            "model",                "",             DATA_STRING, "Acurite-Rain",
             "id",                   "",             DATA_INT,    id,
-            _X("rain_mm","rain"),   "Total Rain",   DATA_FORMAT, "%.1f mm", DATA_DOUBLE, total_rain,
+            "rain_mm",              "Total Rain",   DATA_FORMAT, "%.1f mm", DATA_DOUBLE, total_rain,
             NULL);
     /* clang-format on */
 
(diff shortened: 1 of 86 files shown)


src/devices/acurite.c
```diff
@@ -1364,24 +1363,18 @@ r_device acurite_th = {
  */
 static char *acurite_txr_output_fields[] = {
         "model",
-        "subtype",      // TODO: remove this
         "message_type", // TODO: remove this
```
Mindavi, line 1366: This one also has 'TODO: remove this'

src/devices/bt_rain.c
```diff
@@ -75,7 +75,7 @@ static int bt_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             "battery",          "Battery",          DATA_STRING, battery ? "LOW" : "OK",
             "transmit",         "Transmit",         DATA_STRING, button ? "MANUAL" : "AUTO", // TODO: delete this
```
Mindavi, line 76: Has a TODO: delete this, replaced by the 'button' value, I guess.

src/devices/bt_rain.c
```diff
@@ -91,7 +91,6 @@ static char *output_fields[] = {
         "battery",
         "transmit", // TODO: delete this
```
Mindavi, line 92: Can be removed as well if it's removed above.

Mindavi (approved):
Thanks for doing this cleanup. I put some comments on some missed TODOs, I think those would be good to check before merging. Looks good in general.

I did look at most of this, but not everything. I don't see anything surprising though.

obones:
Shouldn't this be the occasion to also make sure all the "battery" string fields are replaced by "battery_ok" int fields?
Or should it be done in another pull request?

zuckschwerdt:
Thanks. Yes, since there is no oldmodel switch anymore there's no reason to have old battery fields. I'll make a second PR for a clean history without merge commits.
I'll look into the remaining "remove this" notes afterwards.

PR #1738, opened 2021-06-12
Add low battery flag decode

bennydiamond:
Through experimentation I figured Esperanza-EWS temperature/humidity sensor actually has a working "Low Battery" flag.
I used cheap weather station sensor replacement units (marketed with product number JYWDJ-009) that can be bought on eBay/Aliexpress.

PXL_20210611_214316881

Flag is raised when voltage gets below around 2.75V. Over 2.75V flag is set to '0'.

Just for posterity. The sensor I used will continue to transmit mostly reliably until voltage drops below 1.7V.

### DIFF
--- a/src/devices/esperanza_ews.c
+++ b/src/devices/esperanza_ews.c
@@ -13,6 +13,11 @@
 Largely the same as kedsum, s3318p.
 @sa kedsum.c s3318p.c
 
+List of known supported devices:
+- JYWDJ-009
+      * Known voltage operating range 1.7V - 3.8V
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Can you add your voltage findings to the comment section (line 30 "Flags:\n ... low-bat...)
And can you make it
```
            "battery_ok",          "Battery",             DATA_INT,     !battery_low,
```

bennydiamond:
Resulting output:
image

PR #1743, opened 2021-06-21
Fix 5n1 battery

ManuelMcLure:
Commit ef2d37cf51e3264d11cde9149ef87de2f0a4d37a reversed the battery on indicator on the Acurite 5n1/3n1. This fixes it.


zuckschwerdt:
Thanks. This affects three data_make's that sneakily had `battery_low ? "OK" : "LOW",` -- I only caught one.
Can you also change line 907 `!!battery_low, // NOTE: is this really flipped?` to `!battery_low,`

ManuelMcLure:
Done.

PR #1765, opened 2021-07-17
Add basic logger

zuckschwerdt:
Adds basic structured logging.
- allows (e.g. HTTP output) to capture log messages
- allows library use and different apps with custom log handling
- allows coloring of logs (`key: value`), highlighting of errors, ...
- `fprintf`'s are re-defined to use logging for now, until calls are converted

### DIFF
--- a/include/decoder.h
+++ b/include/decoder.h
@@ -12,5 +12,7 @@
 #include "data.h"
 #include "util.h"
 #include "decoder_util.h"
+#include "log.h"
+#define LOG_MODULE __func__
 
 #endif /* INCLUDE_DECODER_H_ */
--- a/include/log.h
+++ b/include/log.h
@@ -1,0 +1,48 @@
+/** @file
+    Basic logging, internal.
+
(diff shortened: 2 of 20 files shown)


gvanem:
This is a good idea. I've for some time liked to have a console-log like I've done in my [ADSB-decoder](https://github.com/gvanem/Dump1090):
dump1090-win

I.e. not a scrolling text-window, but *one* line per unique detected device. Not sure how to detect such uniqueness.

zuckschwerdt:
Great idea! The HTTP output app already does something like this. Using e.g. ncurses we could bring that to the console: a section with unique devices and last reading (csv table like), and another section with running log output.

The current setup is to funnel some messages to the data output, which should be replaced with structured logging for a clear distinction: device readings, log messages.

gvanem:
>  Using e.g. ncurses we could bring that to the console: 

Some more modern than that? *notcurses* look awesome:  https://youtu.be/mOmMcFXdd6k?t=115

zuckschwerdt:
Closing in favor of #2254

PR #1775, opened 2021-08-07
Add support for Markisol curtain remote

dstahlke:
Many curtains share this same protocol.  I tested with two BF-415 remotes.  The protocol seems to match the description here, tested on BF-101, BF-301, BF-305: https://github.com/akirjavainen/markisol/blob/master/Markisol.ino

### DIFF
--- a/src/devices/markisol.c
+++ b/src/devices/markisol.c
@@ -1,5 +1,5 @@
 /** @file
-    Markisol (a.k.a E-Motion, BOFU, BF-30x, BF-415) curtains remote.
+    Markisol (a.k.a E-Motion, BOFU, Rollerhouse, BF-30x, BF-415) curtains remote.
 
     Copyright (C) 2021 Dan Stahlke <dan@stahlke.org>
 
@@ -53,7 +53,8 @@ static int markisol_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         if (bitbuffer->bits_per_row[i] == 41 || bitbuffer->bits_per_row[i] == 42) {
             uint8_t *b = bitbuffer->bb[i];
             for (int j = 0; j < 5; ++j) {
-                buf[j] = ~reverse8(((b[j] << 8) + b[j+1]) >> 7);
+                buf[j] = (b[j] << 1) + (b[j+1] >> 7); // shift stream to discard spurious first bit
+                buf[j] = ~reverse8(buf[j]);
                 cksum += buf[j];
             }
             got_proper_row_length = 1;
@@ -103,6 +104,7 @@ static int markisol_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             "control",        "Control",        DATA_STRING, control_strs[control],
             "channel",        "Channel",        DATA_INT,    channel,
             "zone",           "Zone",           DATA_INT,    zone,
+            "mic",            "Integrity",      DATA_STRING, "CHECKSUM",
             NULL);
     /* clang-format on */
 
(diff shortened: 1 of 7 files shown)


zuckschwerdt:
Run maintainer_update.py to sort out to build files. If there is no checksum set disabled to 1.

dstahlke:
Ran maintainer_update.py and pushed result.  The device does have a one-byte checksum.  I return DECODE_FAIL_MIC if checksum doesn't match.

src/devices/markisol.c
```diff
@@ -0,0 +1,136 @@
...
+    uint8_t buf[5];
+    uint8_t cksum = 0;
+    int got_proper_row_length = 0;
+    for (int i = 0; i < bitbuffer->num_rows; i++) {
+        if (decoder->verbose > 0)
+            fprintf(stderr, "%s: bits_per_row[%d] = %d\n", __func__, i, bitbuffer->bits_per_row[i]);
+        if (bitbuffer->bits_per_row[i] == 41 || bitbuffer->bits_per_row[i] == 42) {
+            uint8_t *b = bitbuffer->bb[i];
+            for (int j = 0; j < 5; ++j) {
+                buf[j] = ~reverse8(((b[j] << 8) + b[j+1]) >> 7);
```
zuckschwerdt, line 56: This should have been `<< 1` not `<< 8`, right?

src/devices/markisol.c
```diff
@@ -0,0 +1,136 @@
...
+    };
+
+    /* clang-format off */
+    data_t *data = data_make(
+            "model",          "Model",          DATA_STRING, "Markisol",
+            "id",             "",               DATA_FORMAT, "%04X", DATA_INT, address,
+            "control",        "Control",        DATA_STRING, control_strs[control],
+            "channel",        "Channel",        DATA_INT,    channel,
+            "zone",           "Zone",           DATA_INT,    zone,
+            NULL);
```
zuckschwerdt, line 106: Add "mic" of "SUM" (see other decoders)

src/devices/markisol.c
```diff
@@ -0,0 +1,136 @@
...
+        "control",
+        "channel",
+        "zone",
+        NULL,
+};
+
+// rtl_433 -f 433900000 -X 'n=name,m=OOK_PWM,s=368,l=704,r=10000,g=10000,t=0,y=5628'
+
+r_device markisol = {
+        .name           = "Markisol", // E-Motion, BOFU, BF-30x, BF-415, etc.
```
zuckschwerdt, line 125: This can be a list of supported devices, e.g. "Markisol, E-Motion, BOFU, BF-30x, BF-415 curtain remote"

src/devices/markisol.c
dstahlke: I was joining consecutive bytes into a 16-bit value, then shifting the whole thing.  But <<1 is probably clearer.  I rewrote it and added a comment.

src/devices/markisol.c
dstahlke: Done.

src/devices/markisol.c
dstahlke: Done.  Also added "Rollerhouse".

src/devices/markisol.c
zuckschwerdt: Oh, I messed up counting the parens, you are right, either way works.

dstahlke:
Is further work needed on my end, or is it ready for merge?

zuckschwerdt:
No, it's good. We are just always busy and backlogged, sorry.

dstahlke:
Thanks!  Backlog is understandable.  I just wanted to make sure I wasn't the holdup.

Issue #1462, opened 2020-08-08
Optional preamble compression


ehagan:
It would be useful for some devices to be able to optionally compress or discard the preamble. 

Implementing a decoder for the Inkbird ITH-20R temperature and humidity sensor has required increasing BITBUF_COLS and PD_MAX_PULSES substantially which increases the memory footprint. The preamble from this device is 14400 on-off bits which pushes the actual payload out of the available data window for decoding.

Please see this conversation in the group for the full context
https://groups.google.com/g/rtl_433/c/oeExmwoBI0w

Dmitriy Kozyrev wrote a decoder and for the device and posted a link to a zip of his work. I've merged the changes into my working copy and have pushed to my fork for convenience.

This is the message format the decoder reads as copied from Dmitriy's post:
```
Total packet length 14563 bits:
Preamble: aa aa aa ... aa aa (14400 on-off sync bits)
Sync Word (16 bits): 2DD4
Data (147 bits):
Byte    Sample      Comment
0-2     D3910F      Always the same across devices, a device type?
3       00          00 - normal work , 40 - unlink sensor (button pressed 5s), 80 - battery replaced
4       01          Changes from 1 to 2 if external sensor present
5-6     0301        Unknown (also seen 0201), sw version?
7       58          Battery % 0-100
8-9     A221        Device id, always the same for a sensor but each sensor is different
10-11   D600        Temperature in °C * 10, little endian, so 0xD200 is 210, 21.0°C or 69.8°F
12-13   F400        Temperature °C * 10 for the external sensor,  0x1405 if not connected
14-15   D301        Relative humidity %  * 10, little endian, so 0xC501 is 453 or 45.3%
16-17   38FB        CRC16
18      0           Unknown 3 bits (seen 0 and 2)

CRC16 (bytes 0-15), without sync word):
poly=0x8005  init=0x2f61  refin=true  refout=true  xorout=0x0000  check=0x3583  residue=0x0000
```
And the full set of changes to get this thing to work are here:
https://github.com/ehagan/rtl_433/tree/feat-inkbird_ith20r

I ran ` rtl_433 -F json -M utc -R 40 ` under memusage with and without the changes. 
feat-inkbird_ith20r branch:
```
 Memory usage summary: heap total: 40818814, heap peak: 38142376, stack peak: 92400
         total calls   total memory   failed calls
 malloc|       9285        2055942              0
realloc|       1110         461800              0  (nomove:483, dec:0, free:0)
[...]

PR #1782, opened 2021-08-28
Feat inkbird ith20r

heffer86:


zuckschwerdt:
Note #1462 -- we do have experimental rowspill now but still need pulse_data compression or circular-buffer for the massive preamble from Inkbird ITH20R.

PR #1783, opened 2021-08-29
fs20: Improve bitstream decoding, add some FHT support

StefanBruens:
The bitstream decoding is improved significantly:
- The encoding is corrected to OOK_PULSE_PWM
- Preamble scan is added, as sometimes a random leading bit is added, or
  the initial bit is dropped, both due to noise.
- MIC/sanity check is added, i.e. parity is checked for each individual
  data byte, and the 'sum' byte is checked for a sane value.

Also report the flags from the command byte, and the optional extension
byte.

The FHT protocol uses the same bitstream encoding as FS20, but can be
detected by a different range for the sum byte. Add decode tables for
cmd and flags.

### DIFF
--- a/src/devices/fs20.c
+++ b/src/devices/fs20.c
@@ -29,6 +29,53 @@ Command extensions are also not decoded. feel free to improve!
 
 #include "decoder.h"
 
+static int fs20_find_preamble(bitbuffer_t *bitbuffer, int bitpos)
+{
+    // Preamble is 12 x '0' | '1', but we ignore the first preamble bit
+    // Last bit ('1') is at position (pattern[1] >> 4 & 1)
+    uint8_t const preamble_pattern[2] = {0x00, 0x10};
+    uint8_t const min_packet_length = 4 * (8 + 1);
+
+    // fast scan for 8 consecutive '0' bits
+    uint8_t *bits = bitbuffer->bb[0];
+    while ((bitpos + 12 + min_packet_length < bitbuffer->bits_per_row[0])
+        && ((bits[(bitpos / 8) + 1] == 0) || (bits[(bitpos / 8)] != 0))) {
+        bitpos += 8;
+    }
+    if (bitpos) {
+        bitpos--;
+        bitpos &= ~0x3;
+    }
+
+    while ((bitpos = bitbuffer_search(bitbuffer, 0, bitpos, preamble_pattern, 12)) < bitbuffer->bits_per_row[0]) {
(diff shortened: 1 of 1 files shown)


src/devices/fs20.c
```diff
@@ -29,6 +29,53 @@ Command extensions are also not decoded. feel free to improve!
...
+    // preamble not found
+    return DECODE_FAIL_SANITY;
+}
+
+struct parity_byte {
+    uint8_t data;
+    uint8_t err;
+};
+
+struct parity_byte get_byte(uint8_t* bits, unsigned pos)
```
zuckschwerdt, line 66: this should be `static`?

zuckschwerdt:
I guess you could add
`                "mic",              "Integrity",    DATA_STRING, "PARITY",`
as there are now some (weak) checks.

Regarding `.disabled=0`, did you test on rtl_433_tests to see if there are false positives?

src/devices/fs20.c
```diff
@@ -65,35 +112,125 @@ static int fs20_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+            continue;
+        cmd = res.data;
+
+        res = get_byte(bits, bitpos + 36);
+        if (res.err)
+            continue;
+
+        if (cmd & 0x20) {
+            ext = res.data;
+        if (bitpos + 45 + 9 > bitbuffer->bits_per_row[0])
```
zuckschwerdt, line 203: these two lines need one more indent

src/devices/fs20.c
```diff
@@ -65,35 +112,125 @@ static int fs20_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
-    if (bitbuffer->bits_per_row[0] != 58) {
-        // check extended length (never tested!)
-        if (bitbuffer->bits_per_row[0] != 67)
-            return DECODE_ABORT_LENGTH;
+    int rc = DECODE_FAIL_MIC;
+    int bitpos = 0;
+
+    while ((bitpos = fs20_find_preamble(bitbuffer, bitpos)) >= 0) {
+        if (decoder->verbose) {
+            printf("Found preamble at %d\n", bitpos);
```
zuckschwerdt, line 172: stdout is reserved for data (e.g. JSON), use fprintf(stderr, ...)

merbanan:
@zuckschwerdt  How about merging this but keeping it disabled ?

zuckschwerdt:
I guess this dropped off waiting for feedback. I'll merge and correct the issues on top.

PR #1800, opened 2021-09-10
Fix "-Wextra-semi-stmt" warning

gvanem:
Another clang-cl `-Wextra-semi-stmt` fix.


merbanan:
LGTM

gvanem:
BTW. there are many such warnings in `src/mongoose.c`.
Time to update it? And perhaps patch it to fix those warnings too?

zuckschwerdt:
Mongoose did a rewrite. We should change to the new version. But that will take some meticulous work.

PR #1802, opened 2021-09-10
Fix mixing SoapySDR allocator

gvanem:
Using MSVC debug-mode (`-MDd`) to build `rtl_433.exe` with a Soapy library built for release-mode (`-MD`)
does not work. It will crash inside MSVC's CRT on a call to `free_dbg (ptr_from_soapy)`.

Hence let Soapy free it's own allocated memory using the `SoapySDR_free (ptr_from_soapy)` function.


merbanan:
LGTM

PR #1814, opened 2021-09-23
Add support for Clipsal cent-a-meter CMR113 power meter

mikey:
Device only measures current but is only sold in Australia so driver
gives power output assuming voltage is 240V.

### DIFF
--- a/src/devices/cmr113.c
+++ b/src/devices/cmr113.c
@@ -1,0 +1,127 @@
+/** @file
+    Clipsal CMR113 cent-a-meter power meter
+
+    Copyright (C) 2021 Michael Neuling <mikey@neuling.org>
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+/**
+    The demodulation comes in a few stages:
+
+     A) Firstly we look at the pulse lengths both high and low. These
+        are demodulated using OOK_PULSE_PIWM_DC before we hit this
+        driver. Any short pulse (high or low) is assigned a 1 and a
+        long pulse (high or low) is assigned a 0. ie every pulse is a
+        bit.
+
+     B) We then look for two patterns in this new bitstream:
+         - 0b00 (ie long long from stream A)
+         - 0b011 (ie long short short from stream A)
+
+     C) We start off with an output bit of '0'.  When we see a 0b00
+        (from B), the next output bit is the same as the last
+        bit. When we see a 0b011 (from B), the next output is
+        toggled. If we don't see ether of these patterns, we fail.
+
+     D) The output from C represents the final bitstream. This is 83
+        bits repeated twice. There are some timestamps, transmitter
+        IDs and CRC but all we decode below are the 3 current values
+        which are 10 bits each representing AMPS/10. We do check the
+        two 83 bit are identical and fail if not.
+
+     Kudos to Jon Oxer for decoding this stream and putting it here:
+          https://github.com/jonoxer/CentAReceiver
+
+**/
+
+#include "decoder.h"
+
+#define COMPARE_BITS 83
+#define COMPARE_BYTES (COMPARE_BITS/8)
+
+static int cmr113_decoder(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    int start, bit;
+    uint8_t buf[4];
+    uint8_t b1[COMPARE_BYTES], b2[COMPARE_BYTES];
+    bitbuffer_t b = {0};
+    double current[3];
+    data_t *data;
+
+    if ((bitbuffer->bits_per_row[0] < 350) || (bitbuffer->bits_per_row[0] > 450))
+        return DECODE_ABORT_LENGTH;
+
+    bitbuffer_extract_bytes(bitbuffer, 0, 0, buf, 32);
+    if ((buf[0] != 0xb0) || (buf[1] != 0x00) || (buf[2] != 0x00))
+        return DECODE_ABORT_EARLY;
+
+    start = 0;
+    bit = 0;
+    bitbuffer_clear(&b);
+    while ((start + 3) < bitbuffer->bits_per_row[0]) {
+        bitbuffer_extract_bytes(bitbuffer, 0, start, buf, 3);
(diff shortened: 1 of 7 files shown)


src/devices/cmr113.c
```diff
@@ -0,0 +1,110 @@
...
+	    reflect_bytes(buf, 2);
+	    current += ((float)buf[0] + ((buf[1] & 0x3) << 8))/10.0;
+    }
+    /* These seem to be exclusively in Australia, so assume 240V */
+    power = 240 * current;
+
+    data = data_make(
+	    "model",   "",         DATA_STRING, "CMR112",
+	    "power_W", "Power",    DATA_FORMAT, "%.0f W", DATA_DOUBLE,  power,
+	    "current", "Current",  DATA_FORMAT, "%.1f A", DATA_DOUBLE,  current,
```
merbanan, line 81: SI unit missing.

merbanan, line 81: SI unit missing.

examples/rtl_433_mqtt_relay.py
```diff
@@ -18,7 +18,7 @@
 UDP_IP = "127.0.0.1"
 UDP_PORT = 1433
 
-MQTT_HOST = "127.0.0.1"
+MQTT_HOST = "192.168.86.37"
 MQTT_PORT = 1883
```
merbanan, line 22: Stray hunk

merbanan:
Can you add a bit stream description also? And then we need some sample signals also.

examples/rtl_433_mqtt_relay.py
mikey: oops, yes. Sorry

mikey:
> Can you add a bit stream description also? And then we need some sample signals also.

Yes will do. I have some good samples. I just need to take some photos of the device.

merbanan:
Not sure we should output the power. The power depends on the voltage. In regular grids the voltage can vary up to 5-10% and still be ok. But the actual power will then change depending on that. So I prefer to only relay data that is actually coming from the transmitter.

src/devices/cmr113.c
mikey: SI missing on the current? should be "current_A" ?

merbanan: I think that is the correct unit.

mikey:
> Not sure we should output the power. The power depends on the voltage. In regular grids the voltage can vary up to 5-10% and still be ok. But the actual power will then change depending on that. So I prefer to only relay data that is actually coming from the transmitter.

Fair enough. It would be nice to be able to add data about your configuration so we could do this. 

I'm also using the TS-FT002. Being able to add information about your water tank would enable us to broadcast the tank %. 

But maybe that's out of scope for this (already awesome) project :-)

mikey:
> Not sure we should output the power. The power depends on the voltage. In regular grids the voltage can vary up to 5-10% and still be ok. But the actual power will then change depending on that. So I prefer to only relay data that is actually coming from the transmitter.

Would you also like me to keep the three current sensors separate, rather than aggregating them?

zuckschwerdt:
Convert tabs to 4 spaces, model should likely be `Clipsal-CMR112`, run `maintainer_update.py` to get e.g. example conf right.

merbanan:
> Would you also like me to keep the three current sensors separate, rather than aggregating them into one?

Not sure I follow you. Each transmitters data should be translated to a message. As direct translation as possible.

mikey:
> > Would you also like me to keep the three current sensors separate, rather than aggregating them into one?
> 
> Not sure I follow you. Each transmitters data should be translated to a message. As direct translation as possible.

OK. There is only 1 transmitter but it contains data for the 3 attached current sensors in each transmission. 

I'll output 3 separate current values (I'm just output the sum of the 3 currently).

zuckschwerdt:
> I'll output 3 separate current values (I'm just output the sum of the 3 currently).

Preferred keys then are: `current_1_A`, ...

mikey:
Pushed some updates:
1. 4 space tabs
2. Ran maintainer_update.py (BTW SoapySDR needs some updates according to it)
3. Remove stray hunk
4. Updated bitstream description 
5. Added SI unit to output
6. Removed power output and split current in 3 (current called current_1_A, current_3_A, current_3_A)
7. Updated model number

mikey:
@merbanan I've added some samples in https://github.com/merbanan/rtl_433_tests/pull/401

zuckschwerdt:
> samples in [merbanan/rtl_433_tests#401](https://github.com/merbanan/rtl_433_tests/pull/401)

Can you add back the freq/rate info to the filenames, it's the only metadata we have.

zuckschwerdt:
The `if` branch L55 needs more indent. (Clang warns about those things). `){` needs a space.
Indent on continued line, e.g. 97 107 117 is 8 spaces. (Try clang-format if you have Clang, it *mostly* gets things right.)

mikey:
> The `if` branch L55 needs more indent. (Clang warns about those things). `){` needs a space.
I understand these two and have fixed

> Indent on continued line, e.g. 97 107 117 is 8 spaces. (Try clang-format if you have Clang, it _mostly_ gets things right.)

I think I've fixed this.

zuckschwerdt:
Looks good, one thing I just notice: the "cmr112" should all be "cmr113", right?

mikey:
yes, sorry. I've updated the PR title now. I can't update the branch name without closing this PR. 

The code is correct and only refers to cmr113.

mikey:
> Branch name isn't an issue. But check the "Files changed" tab for the contents of the PR, there's 112 in multiple places.

Sorry, you're right. My bad, I 'git grepped' badly. Fixed now.

mikey:
I've updated the tests as well as that had the same mistake.

zuckschwerdt:
No worries. There is still one instance in example conf.
Also "model" needs to be "Clipsal-CMR113" (note the dash),
and output_fields needs to have: "model", "current_1_A", "current_2_A", "current_3_A", NULL.

mikey:
> No worries. There is still one instance in example conf.
:facepalm: Fixed

> Also "model" needs to be "Clipsal-CMR113" (note the dash),

OK

> and output_fields needs to have: "model", "current_1_A", "current_2_A", "current_3_A", NULL.

OK

I've repushed and also fixed up the tests to add the dash. Sorry for the noise.

PR #1829, opened 2021-10-10
Add EnOcean ERP1 decoder

wintersteiger:
This adds an EnOcean ERP1 decoder. It just decodes the radio frames into byte strings to be further interpreted by (existing) higher level decoders. I don't necessarily plan to extend this any further; it just serves as a quick test to see whether EnOcean devices are active and operating correctly. Together with higher level decoders, this can help to remove the need for an EnOcean-specific radio module in a home automation gateway, but I don't necessarily envisage that to be based on an RTL-SDR.


zuckschwerdt:
Christoph, what is your (recommended) software setup? I'm using an EnOcean UART-Module connected to a Raspi running the "Python EnOcean" lib with added Secure PTM mode https://github.com/kipe/enocean/pull/116 -- but that lib seems stale.

PR #1834, opened 2021-10-15
Add Linear Megacode Support

aaronsp777:


zuckschwerdt:
Ah, better :)
Can you also surround the `data_make()` block with `/* clang-format off */` and `/* clang-format on */`, then you can align the fields nicely. Have a look in our other decoder source files.
Field ordering should start with "model", "id", ...
`.name` should be "Linear Megacode Garage/Gate Remotes".
`model` should likely be "Megacode-Remote".
The last field in the initializers in line 89 and 100 needs a trailing `,` and the closing `};` then goes on the next line.

The `8388608` looks magic, better written as `0x800000`, also the 128, 15 and 65535 are easier to grasp as 0x7f, 0xf and 0xffff. The `ones` is actually a `bits` counter, right?
Otherwise looks straight forward and good to merge.

aaronsp777:
Fixed.   
This is my first PR on any project I've ever done.
Thank you for the review.

aaronsp777:
Let me know what I need to do next, as I am still new to PR's.

PR #1835, opened 2021-10-15
Add support for HASS long term statistics

kdknigga:
This adds "state_class" metadata so that HASS can track [long-term statistics](https://developers.home-assistant.io/blog/2021/05/25/sensor_attributes/).

I kind of guessed for the state class of `rain_mm`, `rain_in`, and `strike_count` as total_increasing, as I don't have any devices that report those things so I'm not 100% sure how they work.


zuckschwerdt:
Thanks! `strike_count` will reset for each new event, AFAIK.

PR #1836, opened 2021-10-16
Added OpenSSL development library to build guide

Mhowser:
I ran into this issue a few minutes ago, hopefully this will help someone else.
```
-- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR) 

-- OpenSSL development files not found, TLS won't be possible.
```

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -38,23 +38,33 @@ Depending on your system, you may need to install the following libraries.
 
 Debian:
 
-    sudo apt-get install libtool libusb-1.0-0-dev librtlsdr-dev libssl-dev rtl-sdr build-essential cmake pkg-config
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! The OpenSSL dependency is optional and only needed for e.g. mqtts.
Do you think splitting the e.g. apt-get into a second line and hint "if you need TLS connections" would be helpful?

Mhowser:
Yeah that would be better as it's not a hard requirement.

zuckschwerdt:
Do you want to write that hint and update this PR? Otherwise I'll merge and do that.

PR #1850, opened 2021-11-07
fixing leading zero printout

Gollam:
If the first digit of ID is zero, it wasn't printed. With this change, leading zeroes are also displayed correctly.


Gollam:
@zuckschwerdt the change:

image

image

PR #1851, opened 2021-11-08
Change HASS logging to use logging levels

deviantintegral:
This PR:

- Uses the `logging` module instead of prints. This allows for filtering based on logging levels.
- Wires in `--debug` and `--quiet` to trigger `DEBUG` and `WARNING` logs respectively.
- Enables mqttc debug logging when `DEBUG` is set.
- Adds some additional debug log messages I found useful troubleshooting what turned out to be an auth issue (which is added in 23b2df8).
- Fixes allowing both quiet and debug to be specified at the same time.


zuckschwerdt:
Thanks.

Issue #1726, opened 2021-05-28
Running decoders and devices on memory limited architecture


obones:
As already hinted in other discussions, I'm trying to call the `run_ook_demods` and `run_fsk_demods` methods on an ESP32 WROOM32D board. This is for the [RFLink32](https://github.com/cpainchaud/RFLink32) project where we believe `rtl_433` plugins would be much useful.
The ESP32 architecture has very limited memory available per default, about 320k out of which only 160k are available to the end user.
`rtl_433` uses very little heap memory but has a somewhat large stack usage around the 20 to 40kb range. 
While this is fine under Windows and Linux, it is not appropriate for ESP32, where the default stack size is a mere 8kb which should not be pushed any higher than 16 to 20kb to keep the board stable.

Using Visual Studio and its `/analyze:stacksize` command line option, I was able to find the functions that consume the most stack and this issue will present them with action suggestions. For most of them, I have a branch ready to be merged into this project but I have not yet created the pull requests to avoid unnecessary clutter before we agree on them.

**pulse_demod**
`pulse_demod_dmc`, `pulse_demod_piwm_raw` and `pulse_demod_piwm_dc` create a local array of twice the `PD_MAX_PULSES` value to store the pulses and gaps from `pulses` all at once. This consumes `1200 * 2 * 4 = 9600` bytes when it would be easy enough to pick either a pulse or a gap depending on the index.
I have made the appropriate changes in this branch: https://github.com/obones/rtl_433/commits/pulse_demod_stack_usage

**bitbuffer_manchester_decode and bitbuffer_differential_manchester_decode**
These methods require a `bitbuffer_t` to place the result of their decoding, but only ever use the first row in that buffer. This means that 24 out of the 25 rows are using stack space for nothing, along with `num_rows` and the entire `syncs_before_row` array that is not used. That's `24 * 450 + 24 * 16 + 25 * 16 + 16 = 11600` bytes that are unused on stack.
Changing `bitbuffer_manchester_decode` and `bitbuffer_differential_manchester_decode`to require a `bitrow_t` and a `uint16_t *` as output parameters saves all that space.
This impacts quite a few devices, among which almost all the TPMS ones. It also required introducing a few `bitrow` based methods that allowed to adjust *GE Color Effects*, *SecPlus v2* and *Vaillant VRT 340f* to stop requiring `bitbuffer_t` storage but only bit rows to do their decoding, giving a huge stack space save as well.
[...]

PR #1863, opened 2021-11-12
Introduce RTL_433_REDUCE_STACK_USE which, when defined, drastically reduces the size of bitrow_t

obones:
This saves a lot of stack space, as explained in #1726

### DIFF
--- a/include/bitbuffer.h
+++ b/include/bitbuffer.h
@@ -15,8 +15,14 @@
 #include <stdint.h>
 
 // NOTE: Wireless mbus protocol needs at least ((256+16*2+3)*12)/8 => 437 bytes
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Thanks! Good catch on the `BITBUF_COLS`. There are still two instances left as loops that also need to be `44` ;)

zuckschwerdt:
You can also fix lacrossews.c form `BITBUF_ROWS` to `bitbuffer->num_rows` if you like. flex.c is fine as is.

obones:
Thanks, I had not noticed those as the Github actions do not send email when it's just warnings.
This is now fixed in this branch

Issue #1866, opened 2021-11-12
Lots of warnings when compiling with Visual Studio 2019


obones:
I know my situation is a bit unusual, but I'm using Visual Studio 2019 to build this project. I'm using the solution file placed in the `vs15` folder that was updated by the IDE. 
As a base principle, I want zero messages from the compiler but when building the master branch, I get more than 300 warnings which I have placed in the following CSV file: [warnings.csv](https://github.com/merbanan/rtl_433/files/7529129/warnings.csv)

Here is a summary of the situation:

- 250 C4244/C4267/C4305 warnings: "conversion from XX to YY, possible loss of data" 
Most are from `double` to `float` and come from floating point literals
Then there are conversions from `double`/`float` to an integer type
And finally a dozen are from a large integer type to a narrower (one `int64_t` to `int32_t` for instance)

- 20 C4996 warnings: "The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name"

- 13 C6340 warnings: "Mismatch on sign: 'unsigned XX' passed as _Param_(N) when some signed type is required in call to 'fprintf'"
Those ones are fixable via a change to the %d marker, but this might be Microsoft specific

- 12 C28251 warnings: "Inconsistent annotation for 'XXX': this instance has no annotations"
I'm not sure what this is about

- 4 C6246 warnings: "Local declaration of 'XXX' hides declaration of the same name in outer scope"
Not a big problem in and of itself, but I believe it would make the code clearer if this is fixed.

- 3 C4090 warnings: "different const qualifiers"
These are oversights that I believe should be fixed for clarity's sake

The remaining warnings all revolve around static analysis which suggest that there might be some out of bounds reading, null de-referencing and the like.

What I have done on my local copy is to completely disable the following warnings: 4244; 4267; 4305; 4996
This removes most of them, but I'm not too happy with this, especially the ones related to data loss.

I can spend some time looking around for fixes on most of those warnings, but I would like your opinion before I start. For instance, for the ones related to floating points literals, should I change them all to use the `f` suffix, or should I change the variable types to all use `double`?
For conversions from floating point to integer types, should I use `trunc` or `round`?

Any opinions are most welcome.

zuckschwerdt:
We have Visual Studio 2017 and Visual Studio 2019 in the GiHub Actions and see the same mess. We are using the CMake targets though.
[...]

PR #1872, opened 2021-11-15
Fix: make sure fprintf uses an unsigned marker when an unsigned value is given

obones:
Fixes C6340 as reported in #1866

### DIFF
--- a/src/bitbuffer.c
+++ b/src/bitbuffer.c
@@ -29,7 +29,7 @@ void bitbuffer_add_bit(bitbuffer_t *bits, int bit)
         return;
     }
     if (bits->bits_per_row[bits->num_rows - 1] == UINT16_MAX - 1) {
-        fprintf(stderr, "%s: Warning: row length limit (%d bits) reached\n", __func__, UINT16_MAX);
+        fprintf(stderr, "%s: Warning: row length limit (%u bits) reached\n", __func__, UINT16_MAX);
     }
 
     uint16_t col_index = bits->bits_per_row[bits->num_rows - 1] / 8;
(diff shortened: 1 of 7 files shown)


zuckschwerdt:
Not sure if works at intended, needs more research.
E.g. there is `#define UINT16_MAX (65535)` (or maybe `#define UINT16_MAX  0xffff`) which acording to literal rules will be `int`.
And in the printf's the (unsigned) shorts will be values will be coerced to integer according to promotion rules.

merbanan:
Well the iterators could be changed to int. But int_t types should keep their size IIRC.

obones:
Under Visual Studio, it tells me that `UINT16_MAX` is defined like this in  `stdint.h`

    #define UINT16_MAX       0xffffui16

So in that instance, it clearly is a `uint16_t`

zuckschwerdt:
Then let's start with the observation that all those instances were uint16_t. And as printf args they were not in there "native" size but all were coerced to `int`. This change now coerces to `unsigned int`. Doesn't matter anyway and both share the same conversion rank. But note that this likely changed the emitted code, not just tidied up some specifiers ;)

merbanan:
But do we have different definitions of UINT16_MAX on different platforms ? That would be annoying.

zuckschwerdt:
I don't think that matters, the literal would be at least uint16_t anyway and literal rules use the smallest type that fits the number. And any larger integral type will also fit. In the printf's we chose `int`, which can represent uint16_t. now we chose `unsigned` and the same applies ;) It should only matter to the compilers that we don't use `hd`, but `hu` and any higher rank should be fine.

zuckschwerdt:
aside: it's easier for me to compile that changelog if commit messages follow https://triq.org/rtl_433/CONTRIBUTING.html#commit-messages
I.e. the verb is not followed by a colon, and the area-of-work would preferably be "minor:" here -- I can just filter those out as not relevant to the changelog :)

obones:
> aside: it's easier for me to compile that changelog if commit messages follow https://triq.org/rtl_433/CONTRIBUTING.html#commit-messages

Bummer, I thought I did, but I now see that I missed the first part altogether. I'll keep that in mind for the next ones

PR #1873, opened 2021-11-15
Fix: floating point literals must use the f suffix when working with float variables

obones:
This fixes C4244/C4267/C4305 warnings related to `double` to `float` conversions when a literal is involved in the warning, as reported in #1866


merbanan:
LGTM, but I'll wait abit before committing.

src/devices/acurite.c
```diff
@@ -729,7 +729,7 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             // 11 bits needed for specified range -40 C to 70 C (-40 F - 158 F)
             // range -100 C to 1538.4 C
             int temp_raw = ((bb[4] & 0x7F) << 7) | (bb[5] & 0x7F);
-            tempc = temp_raw * 0.1 - 100;
+            tempc = temp_raw * 0.1f - 100;
```
zuckschwerdt, line 732: there is still a double lurking here. prefered would be `(temp_raw - 1000) * 0.1f`

zuckschwerdt: or maybe it's an implicit float…

obones: According to VC++, it gets indeed converted to a `float` not a `double`, there is no longer a warning on that line.

obones: Note that I hesitated changing `100` to `100.0f` but I'm not sure it makes it any clearer...

merbanan: temp_raw should be converted to float and then -100 should be converted to float and then everything is float

zuckschwerdt: my point was an integer substraction, then a float multiply instead of two float ops.

merbanan: The rules give that it should be float. But I see your point if the int is large.

obones: The initial goal of that PR was to address the warnings by simply adding the suffix wherever required. 
That being said, I agree that there are optimization opportunities at various places (`value / 3600.0f / 1000.0f` comes to mind) but shouldn't this be addressed in another PR?

zuckschwerdt: True, and greping for `f - ` afterwards will find most of those instances.

src/devices/acurite.c
```diff
@@ -588,7 +588,7 @@ static int acurite_atlas_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsig
         /* clang-format off */
         data = data_append(data,
                 "wind_dir_deg",     NULL,           DATA_FORMAT,    "%.1f",         DATA_DOUBLE, wind_dir,
-                "rain_in",          "Rainfall Accumulation", DATA_FORMAT, "%.2f in", DATA_DOUBLE, raincounter * 0.01f,
+                "rain_in",          "Rainfall Accumulation", DATA_FORMAT, "%.2f in", DATA_DOUBLE, raincounter * 0.01,
```
merbanan, line 591: In this case I actually wanted it to be a float but it seems to be missed from the C-standard. Or something like that.

obones:
I have just rebased this branch on the latest master, please let me know if this is still of interest

src/devices/acurite.c
gdt: I'm going to interpret this as 1) you would have liked to have DATA_FLOAT as a format and used it  2) given that it says DOUBLE, this change is ok.   Of course speak up if that second point is a  misinterpretation; the first point doesn't really bear on "should we merge this".

gdt:
This looks ready to merge to me, by code review.

PR #1874, opened 2021-11-16
Annotations must be the same in the header and source files

obones:
This fixes C28251 warnings as mentioned in #1866


zuckschwerdt:
Looks good. If this gets out of hand, and I understand right, we can switch to `_Use_decl_annotations_`.

PR #1875, opened 2021-11-16
Fix warnings about size_t being cast to a narrower type

obones:
This fixes a few C4267 warning as reported in #1866 because the x64 target defines `size_t` as a 64 bit unsigned integer.

### DIFF
--- a/src/data_tag.c
+++ b/src/data_tag.c
@@ -63,7 +63,7 @@ static void gpsd_client_event(struct mg_connection *nc, int ev, void *ev_data)
             // Success
             fprintf(stderr, "GPSd Connected...\n");
             if (ctx->init_str && *ctx->init_str) {
-                mg_send(nc, ctx->init_str, strlen(ctx->init_str));
+                mg_send(nc, ctx->init_str, (int)strlen(ctx->init_str));
             }
         }
         else {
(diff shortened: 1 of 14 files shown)


src/data_tag.c
```diff
@@ -63,7 +63,7 @@ static void gpsd_client_event(struct mg_connection *nc, int ev, void *ev_data)
             // Success
             fprintf(stderr, "GPSd Connected...\n");
             if (ctx->init_str && *ctx->init_str) {
-                mg_send(nc, ctx->init_str, strlen(ctx->init_str));
+                mg_send(nc, ctx->init_str, (int)strlen(ctx->init_str));
```
obones, line 66: This one should not be required if mongoose was consistent in its methods declarations: `mg_send_http_chunk`, `mg_send_websocket_frame` and almost all others take a `size_t`, but for some reason `mg_send` does not...

src/http_server.c
```diff
@@ -1090,7 +1090,7 @@ static void http_broadcast_send(struct http_server_context *ctx, char const *msg
             mg_set_timer(nc, mg_time() + KEEP_ALIVE); // reset keep alive timer
         }
         else if (cctx && !cctx->is_chunked) {
-            mg_send(nc, msg, len);
+            mg_send(nc, msg, (int)len);
```
obones, line 1093: This one should not be required if mongoose was consistent in its methods declarations: `mg_send_http_chunk`, `mg_send_websocket_frame` and almost all others take a `size_t`, but for some reason `mg_send` does not...

merbanan:
Most of the changes merged.

PR #1876, opened 2021-11-16
Add file meta data as watermark

zuckschwerdt:
Here is an idea to add critical file meta data as watermark in first few samples.

It's at the absolute lowest dB (LSB in the stronger of an I/Q pair) and will be invisible (we don't expect data in the first few samples anyway though). Fully compatible with all other applications, as the data is somewhat steganographically hidden.

We should then test files for a watermark, otherwise default to the filename info, and after that override with arguments if present.

Just a demo for now to get feedback.

### DIFF
--- a/include/fileformat.h
+++ b/include/fileformat.h
@@ -117,6 +117,20 @@ void file_info_clear(file_info_t *info);
 /// @return the detected file format, 0 otherwise
 int file_info_parse_filename(file_info_t *info, const char *filename);
 
+/// Detect a watermark from cu8, cs8, cs16, cf32 files and set file info.
+///
+/// @param[in,out] info the file info to parse into
+/// @param buf the IQ sample data, needs to be at least 1024 bytes length
+/// @param verbose the verbosity level for reports to stderr
+/// @return the detected file format, 0 otherwise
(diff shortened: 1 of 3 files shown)


merbanan:
LGTM, maybe a text specification for this also. And maybe add a version also if we want to extend the logic.

zuckschwerdt:
A spec to get this adopted elsewhere is a good idea. Versioning can be done with the fixed tag, currently the prefix is "SDR:", it can be e.g. "SDR+" or something like that later.

This patch now also remedies the long standing nuisance that -f and -s are needed if input files are not at the defaults. The logic for reading files now is:

1. use the defaults
2. parse the filename tags (e.g. `_1024k`)
3. parse a watermark
4. use overrides (e.g. `250k:`)

I renamed some files in the tests repo and all samples now check out without applying any `-s` (from the `samplerate` files).
Also did a test run by recoding all samples with a watermark and no regressions appeared.

merbanan:
Ok, to merge then.

merbanan:
I think we are good enough with encoding the stuff we need in the file name. Can we close this ?

merbanan:
Closing, reopen if needed.

PR #1877, opened 2021-11-16
Fix arithmetics overflow warnings with VS2019

obones:
This fixes warnings of this kind, seen via #1866:

> Warning C26451 Arithmetic overflow: Using operator '*' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '*' to avoid overflow (io.2).

### DIFF
--- a/src/baseband.c
+++ b/src/baseband.c
@@ -86,7 +86,7 @@ float magnitude_true_cu8(uint8_t const *iq_buf, uint16_t *y_buf, uint32_t len)
     for (i = 0; i < len; i++) {
         int16_t x = iq_buf[2 * i] - 128;
         int16_t y = iq_buf[2 * i + 1] - 128;
-        y_buf[i]  = (uint16_t)(sqrt(x * x + y * y) * 128.0); // max 181, scaled 23170, fs 16384
+        y_buf[i]  = (uint16_t)(sqrt((double)x * x + (double)y * y) * 128.0); // max 181, scaled 23170, fs 16384
         sum += y_buf[i];
     }
     return len > 0 && sum >= len ? MAG_TO_DB((float)sum / len) : MAG_TO_DB(1);
@@ -117,7 +117,7 @@ float magnitude_true_cs16(int16_t const *iq_buf, uint16_t *y_buf, uint32_t len)
     for (i = 0; i < len; i++) {
         int32_t x = iq_buf[2 * i];
         int32_t y = iq_buf[2 * i + 1];
-        y_buf[i]  = (int)sqrt(x * x + y * y) >> 1; // max 46341, scaled 23170, fs 16384
+        y_buf[i]  = (int)sqrt((double)x * x + (double)y * y) >> 1; // max 46341, scaled 23170, fs 16384
         sum += y_buf[i];
     }
     return len > 0 && sum >= len ? MAG_TO_DB((float)sum / len) : MAG_TO_DB(1);
@@ -126,7 +126,7 @@ float magnitude_true_cs16(int16_t const *iq_buf, uint16_t *y_buf, uint32_t len)
 
 // Fixed-point arithmetic on Q0.15
 #define F_SCALE 15
-#define S_CONST (1 << F_SCALE)
+#define S_CONST ((int)(1 << F_SCALE))
 #define FIX(x) ((int)(x * S_CONST))
 
 /** Something that might look like a IIR lowpass filter.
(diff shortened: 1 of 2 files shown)


src/baseband.c
```diff
@@ -116,7 +116,7 @@ float magnitude_true_cs16(int16_t const *iq_buf, uint16_t *y_buf, uint32_t len)
     for (i = 0; i < len; i++) {
         int32_t x = iq_buf[2 * i];
         int32_t y = iq_buf[2 * i + 1];
-        y_buf[i]  = (int)sqrt(x * x + y * y) >> 1; // max 46341, scaled 23170, fs 16384
+        y_buf[i]  = (int)sqrt((double)x * x + (double)y * y) >> 1; // max 46341, scaled 23170, fs 16384
```
merbanan, line 119: The code is fixed point by design.

src/baseband.c
```diff
@@ -85,7 +85,7 @@ float magnitude_true_cu8(uint8_t const *iq_buf, uint16_t *y_buf, uint32_t len)
     for (i = 0; i < len; i++) {
         int16_t x = iq_buf[2 * i] - 128;
         int16_t y = iq_buf[2 * i + 1] - 128;
-        y_buf[i]  = (uint16_t)(sqrt(x * x + y * y) * 128.0); // max 181, scaled 23170, fs 16384
+        y_buf[i]  = (uint16_t)(sqrt((double)x * x + (double)y * y) * 128.0); // max 181, scaled 23170, fs 16384
```
merbanan, line 88: This should be fixed point but I guess there is a reason for the 128.0.

src/baseband.c
obones: Yes, but `sqrt` takes a `double` which is why there is the warning.

src/baseband.c
obones: I believe the reason is more about `sqrt` taking a `double` instead of a `float`

zuckschwerdt: This is just a test function, still turning int muls into double muls isn't great.

src/pulse_analyzer.c
```diff
@@ -273,40 +273,40 @@ void pulse_analyzer(pulse_data_t *data, int package_type)
         device.modulation  = OOK_PULSE_PPM; // TODO: there is not FSK_PULSE_PPM
         device.short_width = to_us * hist_gaps.bins[0].mean;
         device.long_width  = to_us * hist_gaps.bins[1].mean;
-        device.gap_limit   = to_us * (hist_gaps.bins[1].max + 1);                        // Set limit above next lower gap
-        device.reset_limit = to_us * (hist_gaps.bins[hist_gaps.bins_count - 1].max + 1); // Set limit above biggest gap
+        device.gap_limit   = to_us * (hist_gaps.bins[1].max + 1.0);                        // Set limit above next lower gap
+        device.reset_limit = to_us * (hist_gaps.bins[hist_gaps.bins_count - 1].max + 1.0); // Set limit above biggest gap
```
zuckschwerdt, line 277: Not great, it was int addition before, now it's double addition.

obones: Well, yes, the warning is about doing a 4 byte addition that then gets promoted to an 8 byte type for the multiplication, potentially leading to an overflow.
Using the floating point constant forces `double` computation which is not great, I agree. But how do I get rid of the warning then?

src/baseband.c
merbanan: I think that sqrtf() should work and is slightly better type wise. If it is better accuracy wise I don't know.

src/baseband.c
merbanan: And same here with sqrtf().

src/pulse_analyzer.c
merbanan: This is non performance critical code. I'm fine with this change here but maybe it should be a cast? gap_limit is (float) while to_us is (double).

obones:
I have just rebased this branch on the latest master, please let me know if this is still of interest

zuckschwerdt:
Yes, these clean-ups are of interest. I really need to find quiet time to contemplate the effects (e.g. changing some double calcs to float only).

src/baseband.c
```diff
@@ -126,7 +126,7 @@ float magnitude_true_cs16(int16_t const *iq_buf, uint16_t *y_buf, uint32_t len)
 
 // Fixed-point arithmetic on Q0.15
 #define F_SCALE 15
-#define S_CONST (1 << F_SCALE)
+#define S_CONST ((int)(1 << F_SCALE))
```
gdt, line 129: This surprises me.  What type is 1<< F_SCALE without a cast.  unsigned int?

src/baseband.c
gdt: I guess the point is that x and y are limited to [-128,127] because of iq_buf being bytes, but the compiler is not figuring it out?   Is there in fact an overflow possibility, as determined by a human reviewing?  Or is it a false positive warning.   The position I have seen elsewhere, and tend to agree with, is that code should be fixed for valid warnings, and not modified much if any for invalid warnings, except that restructuring it slightly to be easier to reason about, withtout hurting the generated code is ok.

merbanan:
Most of the changes merged.

PR #1878, opened 2021-11-16
add _CRT_NONSTDC_NO_DEPRECATE to avoid the C4996 warning

obones:
Add _CRT_NONSTDC_NO_DEPRECATE to avoid the C4996 warning POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name) which is more of an annoyance than anything else.

This fixes warnings discussed in #1866 

Note that I tried various "tricks" to get rid of it only for the few functions that were reported, none worked (`#define`, `inline` methods...)


zuckschwerdt:
Yes, we have that in cmake too.

PR #1881, opened 2021-11-17
Remove strdup from key, pretty_key, format in data_t

zuckschwerdt:
This is a quick proof of concept that we can remove the strdup from key, pretty_key, and format in data_t.

All three strings should be static lifetime anyway. With these exceptions:
- Tags assign keys at runtime, but the tag key is kept with static lifetime
- Flex assigns keys at runtime, the lifetime is bound to the flex decoder instance
- The unit conversion will change keys and format strings, but we can keep them around

A free on exit for replaced strings needs to be added still.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -54,10 +54,10 @@ typedef union data_value {
 } data_value_t;
 
 typedef struct data {
-    char        *key;
-    char        *pretty_key; /**< the name used for displaying data to user in with a nicer name */
+    char const  *key;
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
@obones I had this around for some time, might be interesting to test for memory limited platforms.

merbanan:
@zuckschwerdt what is the status of this ?

zuckschwerdt:
It is still a viable idea, but we don't really need it.

PR #1884, opened 2021-11-20
klimalogg: Fix format of temperature_C

palstek:
It should be number format (float) instead of string.


zuckschwerdt:
A `* 0.1f` is prefered over `/ 10.0`, also @merbanan might have insight why it's a string? To insure against bad BCD values?

zuckschwerdt:
And you can unroll and remove `bcd_decode8()`. The second call (tenth value) isn't doing anything really.

merbanan:
Well I think I did it that way just because I had the exact data that was presented on the receiver station. What does this fix? The current code should match the receiver. But going the float way should make it susceptible to rounding errors.

zuckschwerdt:
Consumers might expect a number to process, store, and aggregate. The `%.1f` precission should always work out to the same display value.

merbanan:
So the json output is different? If so the patch is ok.

PR #1653, opened 2021-02-17
Add support for the Govee Water Leak Detector H5054 and Govee Contact Sensor B5023

skilau:
Add support for the Govee Water Leak Detector H5054 and the Govee Contact Sensor B5023.

This is a basic driver for what is a basic (but quite nice) water leak sensor, as well as their contact sensor.

https://www.govee.com/products/110/water-leak-detector
https://www.govee.com/products/27/govee-door-contact-sensor

Pulled most of the info from this running thread:
https://github.com/merbanan/rtl_433/issues/1518

I have 5 of these leak sensors, and was able to test:
1) Each sensor has a unique ID value.
2) Button Press.
3) Water Leak.
4) Battery 5 Bars.
5) Battery 4 Bars.
6) Battery 1 Bar.

Unable to test Battery bars 2 and 3.
(But took the Event data for both of those from the thread mentioned above)

I also have 2 of the Contact Sensors, and was able to test both for:
1) Open.
NOTE: The sensor does not support sending a Close event.  This unfortunately makes this sensor far more limited in its actual uses in the real world.

PR #1887, opened 2021-11-24
update value template in rtl_433_mqtt_hass.py for `battery_ok` to allow for float and double battery values

snicker:
This was spurred by the fact that I am working tonight on finally having my Govee leak sensors report events in Homeassistant (thanks to the excellent work in #1653, thanks everyone), and relying on [rtl_433_mqtt_hass.py](https://github.com/merbanan/rtl_433/blob/master/examples/rtl_433_mqtt_hass.py) to get me there.

The battery value was being reported as 1% due to the fact that `0.79` was being converted to int (0), then float (0.0), then multiplied by 99, adding 1. For devices reporting binary values of 0 and 1 for OK, this is still functional. For devices reporting integer values from 0-100, this will break down.

Architecturally, it may be a halfway decent solution to start reporting `battery_level` (float) and `battery_ok` (binary) as two distinct output fields for device drivers, but that's probably a different discussion altogether.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -172,7 +172,7 @@ mappings = {
             "device_class": "battery",
             "name": "Battery",
             "unit_of_measurement": "%",
-            "value_template": "{{ float(value|int) * 99 + 1 }}",
+            "value_template": "{{ float(value) * 99 + 1 }}",
             "state_class": "measurement"
         }
     },


zuckschwerdt:
Thanks. Can you deduce a reason why it's `1->100%, 0->1%`  and not simply `* 100`?

snicker:
> Thanks. Can you deduce a reason why it's `1->100%, 0->1%` and not simply `* 100`?

I'm not certain I understand the question. I actually think there is a broader architectural problem with the drivers and consistency in how battery is reported by the device manufacturer with respect to how rtl_433 reports to downstream consumers. Most device drivers seem to simply regurgitate whatever data is coming in on the protocol straight to a `battery_ok` field. However, there are a variety of ways different manufacturers have chosen to report battery status:

1. Binary - either the battery is good or dying 
2. Percentage as a floating point number between 0 and 1 (as is the case with the Govee meters)
3. Percentage as an integer between 0-100
4. Discrete stepwise values as integers (eg, levels 1-5)

I don't know if you think it's within the scope of rtl_433's drivers to mutate the values from the manufacturers into a consistent format, or if it makes sense for the drivers to have different output fields depending on the type of reporting 1-4 above (eg, `battery_ok` is binary, `battery_percentage` is mutated to always be float between 0-1, `battery_level` is 4 above)

this change will work for 1. and 2. above, but not 3. or 4.

zuckschwerdt:
We already normalize, see https://triq.org/rtl_433/DATA_FORMAT.html#common-device-data
The `battery_ok` is always a percentage value in as many steps as the device reports. Mostly 0 and 1 only.

My question was if HASS needs 1% as the `0` value`? Otherwise just the plain battery_ok value (as float and converted to percentage) will do.

snicker:
Ah, I had not seen that, that means 1&2 above should be correct values. HASS doesn't "need" anything, it can accept 1% as the zero value or 0%. 

The issue (which is corrected by my PR) is that the current Jinja value template would take any floating point value less than 1.0 (eg `0.79`) and return 0.0. Therefore, even if a battery were being reported by a device as 0.99, it would show as 1% in HASS.

Try this for yourself if you have an instance of HASS running in the "Template" section of Developer Tools:
```
{% set value = 1.0 %}
Value = 1.0: {{ float(value|int) * 99 + 1 }}
{% set value = 0.5 %}
Value = 0.5: {{ float(value|int) * 99 + 1 }}
{% set value = 0.49 %}
Value = 0.49: {{ float(value|int) * 99 + 1 }}
{% set value = 0.0 %}
Value = 0.0: {{ float(value|int) * 99 + 1 }}

After PR:
{% set value = 1.0 %}
Value = 1.0: {{ float(value) * 99 + 1 }}
{% set value = 0.5 %}
Value = 0.5: {{ float(value) * 99 + 1 }}
{% set value = 0.49 %}
Value = 0.49: {{ float(value) * 99 + 1 }}
{% set value = 0.0 %}
Value = 0.0: {{ float(value) * 99 + 1 }}
```

Including an image if you do not:
image

snicker:
and to answer your original question, this could simply be `* 100`. Agree that would be a simpler implementation, but I figured there were design decisions that I am not aware of for this approach of having 1% as the zero value

zuckschwerdt:
Yes, the patch is good, thank you for the details. I was just wondering if you can give more insight into a sane mapping.
That `*99+1` was just a random example I wrote, no deeper design decision there ;) I might have noted a value of `0%` not showing in the HASS dashboard, maybe. I never really use this code. Safer to just use this PR as is and not go over board with more changes :)

PR #1895, opened 2021-12-04
Add decoder priority stages

zuckschwerdt:
Adds a priority to each decoder. This allows to define execution order with later stages only running as fallback.

The default priority is `0`. All decoders at a priority will run, the next priority will run only if no events have been produced so far. In other words, priorities will run in turn until an event is produced.

E.g. if AlectoV1, which has a strong checksum, decodes a frame, then Springfield-Soil, Prologue, ThermoPro-TX2 will not run, without the need to explicitly check for that false positive in each later decoder.

Another example is TPMS, if no decoder finds a known decoding we could fallback to a generic decoder.

### DIFF
--- a/include/r_device.h
+++ b/include/r_device.h
@@ -53,7 +53,8 @@ typedef struct r_device {
     float tolerance;
     int (*decode_fn)(struct r_device *decoder, struct bitbuffer *bitbuffer);
     struct r_device *(*create_fn)(char *args);
-    unsigned disabled;
+    unsigned priority; ///< Run later and only if no previous events were produced
+    unsigned disabled; ///< 0: default enabled, 1: default disabled, 2: disabled, 3: disabled and hidden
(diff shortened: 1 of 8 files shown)


merbanan:
I would prefer to have a 8 bit priority range. And that we by default fill the range suitably so that we don't need to re-scale the priorities if new devices are found.

zuckschwerdt:
Yes, I started with just a `fallback` bit, but it's `unsigned` now. Priorities would e.g. be:
- 0: default
- 1: normal delayed run for the examples above
- 2..: even after that
- 8: TPMS unknown but with e.g. detected checksum
- 9: TPMS fallback universal decoder, just raw bits (with well-known TPMS framing, e.g. MC and certain length).

zuckschwerdt:
Arbitrary priorities done.

merbanan:
LGTM

PR #1909, opened 2021-12-18
Add support AVE TPMS

charestp:
Add support for AVE TPMS device. Device information can be found at:
http://www.avetechnology.com/index.php?unit=products&lang=en&act=view&id=23

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [205]  Telldus weather station FT0385R sensors
     [206]  LaCrosse TX34-IT rain gauge
     [207]  SmartFire Proflame 2 remote control
+    [208]  AVE TPMS
 
 * Disabled by default, use -R n or -G
 
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -215,6 +215,7 @@
     DECL(telldus_ft0385r) \
     DECL(lacrosse_tx34) \
     DECL(proflame2) \
+    DECL(tpms_ave) \
 
     /* Add new decoders here. */
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -195,6 +195,7 @@ add_library(r_433 STATIC
     devices/thermopro_tp12.c
     devices/thermopro_tx2.c
     devices/tpms_abarth124.c
+    devices/tpms_ave.c
     devices/tpms_citroen.c
     devices/tpms_elantra2012.c
     devices/tpms_ford.c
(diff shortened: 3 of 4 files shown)


zuckschwerdt:
This looks great! Very interesting about the different scaling mode. I wonder if other sensors have something like this too.
- There is one stray `fprintf`.
- The `battery_ok` needs to be an int (0/1) or float (0-1). Do all values `< 6` really map to `1` (100%) or is it a gauge?

charestp:
About the scaling mode, that comes from the specs in manual. There is a car
mode and a truck mode. Former has a range of 700 kPa and resolution of
2.35, while the latter has a range of 1400 kPa and resolution of 5.5. The
manual also mentions 4 modes, which when I change the identified mode bits,
have a different 0 offset for each of car and truck mode.

Concerning the battery, I've used observation on the receiver. There is an
icon with all bars filled with value < 6. One bar disappears when value is
6 and all bars but one disappear and the icon flashes when value is 7. I've
not been able to display any other values. If boolean is required for
battery_ok, then I can set 1 for values <= 6 and 0 otherwise.

Sorry for the fprintf, I missed it. I also observed that some tabs were
left in the files and some clang warnings were there. I will try to fix
these. That's my first github pull request and I'm still learning the
process. Do I simply modify the code and push it on my branch? Is a git
amend ok?

I created a small software to push any signal to the receiver with a hackrf
and noted down observed results on the receiver. I've also had a few pulse
grabs from the sensor. That's how I came with all different values. I can
share those files, but I don't know how.

zuckschwerdt:
Yes, add more commits (we squash on merge) if you like. Amend and force push is also fine.

Either the receiver does not care to distinguish high battery levels, or it less bits (e.g 1 bit for weak, 1 bit for low, otherwise good)? 
Best not to throw information away. If you'd say battery is 3 bits then maybe a float with `(8-batt) * 0.125f` (so we ge a linear 0-1 range).

The last flag bit could be an idle update vs running-mode, perhaps?
If you are still looking for the other two bits: we expect to see a rapid deflation warning bit and maybe a general warning or over-heated bit.

charestp:
I decided to change to double value and use 1.0, 0.75 and 0.25 for battery. That would be aligned with my observation of 4/4 (<6), 3/4 (6) and 1/4 (7) bars on the battery icon. I've not been able to come with a linear interpretation of those bits.

For last flags bit, I've always observed a value of 010 from the sensor. When I generated my own signal to transmit to the receiver, I saw no change on the screen for all values. That's why I kept the information as flags.

zuckschwerdt:
Thanks. battery_level is fine.

I just noticed the kPa to PSI conversion factor. We always try to output the native value and encourage user to use e.g. `-C cutomary` to get all sensors in F/PSI/Mi.
With the conversion formula there is no true native value. But heads-up: we might remove that conversion and change to pressure_kPa on some future cleanup.

PR #1911, opened 2021-12-22
Add support for DeltaDore X3D

mr-sven:
The Payload is not fully clear due the fact I have only thermostat's and switch actors.


zuckschwerdt:
Looks really good at a quick glance, thanks! (and I did laugh at the length thing and "data from previous larger messages")

src/devices/deltadore_x3d.c
```diff
@@ -0,0 +1,191 @@
...
+
+    data_t *data;
+
+    if (bitbuffer->num_rows != 1) {
+        return DECODE_ABORT_EARLY;
+    }
+
+    int row = 0;
+    // Validate message and reject it as fast as possible : check for preamble
+    unsigned start_pos = bitbuffer_search(bitbuffer, row, 0, preamble, sizeof (preamble) * 8);
```
zuckschwerdt, line 109: maybe add `+ sizeof (preamble) * 8)` to set start_pos to first data byte and clean up all subsequent references. Also `>=` in the next `if`

src/devices/deltadore_x3d.c
```diff
@@ -0,0 +1,191 @@
...
+
+    if ((frame[len - 2] << 8 | frame[len - 1]) != crc) {
+        if (decoder->verbose) {
+            fprintf(stderr, "%s: CRC invalid %04x != %04x\n", __func__, frame[len - 2] << 8 | frame[len - 1], crc);
+        }
+        return DECODE_FAIL_MIC;
+    }
+
+    char frame_str[64 * 2 + 1] = {0};
+    // do not put lenght and crc to raw data
```
zuckschwerdt, line 159: leng"th" ;)

src/devices/deltadore_x3d.c
```diff
@@ -0,0 +1,191 @@
...
+    for (int i = 0; i < len - 3; ++i) {
+        sprintf(&frame_str[i * 2], "%02x", frame[i + 1]);
+    }
+
+    /* clang-format off */
+    data = data_make(
+            "model",        "",                 DATA_STRING, "DeltaDore X3D",
```
zuckschwerdt, line 166: "DeltaDore-X3D"

src/devices/deltadore_x3d.c
```diff
@@ -0,0 +1,191 @@
...
+        "mic",
+        NULL,
+};
+
+r_device deltadore_x3d = {
+        .name        = "DeltaDore X3D devices",
+        .modulation  = FSK_PULSE_PCM,
+        .short_width = 25,
+        .long_width  = 25,
+        .reset_limit = 25600,
```
zuckschwerdt, line 188: this seems excessive, with whitening how many bit-width gaps do you expect worst-case?
E.g. 5 -> 25*5 µs, round up -> 160

mr-sven: I don't know, I took this value from the analyzer.

mr-sven: so 160 should enough?

zuckschwerdt: That was just a random example. How many consecutive 0's (before removing the whitening) are there in valid data?

mr-sven: Hm, ok, in a very bad case, the whitened data gets all zero. 
The maximum packet length is 64. The first byte (length) is xored with 0xff, so it will never be zero, the second byte is always 0xff. 
Left 62. 
The chance that the CRC gets whitened to zero is also very low, so left 60.
The message id and retry count is always around position 15, so 45 left.
So my idea is 45 * 8 * 25 = 9000.
What do you think?

zuckschwerdt: That's a worst^2  case ;) Our detector will pick up the timing from the preamble, but even with that I guess that 5-10% margin/jitter is the maximum to work still. So no more than 10 to 20 zero bits will work reliably.
Maybe make it 9000, but it will very likely not work with that many zeros. (The reason Manchester code and similar are used).

mr-sven: I'll make some tests.

mr-sven: I drilled it down to 800, I think more than 4 bytes of zero should not happen.

src/devices/deltadore_x3d.c
```diff
@@ -50,6 +50,24 @@ Known Data:
...
+Payload standard message 0x01:
+- Unknown        {8}  always 0x00
+- Response       {8}  0x00 from Thermostat, 0x01 answer from actor
+- Unknown       {16}  0x0001 or 0x0000
+- Command?      {16}
+   - 0x0001        read register
+   - 0x0008        status?
+   - 0x0009        write register
+- Register No   {16}
+- Command resp  {8}  0x01 successfull
```
zuckschwerdt, line 62: "successful"

mr-sven: Done

mr-sven:
At the moment I've no more additional info, I'm working on decoding the messages. I'll continue on this if I have more valid infos.

zuckschwerdt:
@mr-sven what's the state of this? Did you find out the proper payload structure?

mr-sven:
the only valid item I would say at the moment is the room temperature, maybe the payload in a separate value. The rest is at the moment to complex and undefined.

mr-sven:
I spend a bit of time analyzing the data and I added some source from my other analyze project.
Here is an output of the current known data.
x3d

src/devices/deltadore_x3d.c
```diff
@@ -0,0 +1,474 @@
...
+        "wnd_stat",
+        "retry",
+        "actor",
+        "response",
+        "raw",
+        "mic",
+        NULL,
+};
+
+r_device deltadore_x3d = {
```
zuckschwerdt, line 466: This needs to be `r_device const deltadore_x3d` now

src/devices/deltadore_x3d.c
```diff
@@ -0,0 +1,474 @@
...
+        if (decoder->verbose) {
+             decoder_logf(decoder, 0, __func__, "CRC invalid %04x != %04x\n", frame[len - 2] << 8 | frame[len - 1], crc);
+        }
+        return DECODE_FAIL_MIC;
+    }
+
+    struct deltadore_x3d_message_header head = {0};
+    uint8_t bytes_read = 2; // step over length and FF field
+    bytes_read += deltadore_x3d_parse_message_header(&frame[bytes_read], &head);
+    char *class, *wnd_stat;
```
zuckschwerdt, line 337: `char *` variable should be `char const *` when being given a constant value

mr-sven:
I now added most of the known protocol, for my point of view it's done and can be merged

mr-sven:
Waiting for merge

zuckschwerdt:
It seems some output keys are missing from output_fields, e.g. `temperature_type`, can you check that?
And the doc-comment should start (line 11) with `/** @fn int deltadore_x3d_decode(r_device *decoder, bitbuffer_t *bitbuffer)` to properly attach to the decoder function.
Otherwise good for merge.

mr-sven:
I don't know where that NOLINT came from, maybe c&p.

zuckschwerdt:
I've refactored the CCITT data whitening to bit_util.c with 11d2c7a. Can you check it still works, and do we have any test samples?

mr-sven:
Is working, added a valid sample: #3116

zuckschwerdt:
Great, thanks!
I should have said: "sample files", i.e. a .cu8 capture for regression tests. But a code sample is useful too.
Using `rtl_433 -y aaaaaaaa8169967ede78db58b8f7f836575bd3940e71ed2636425a2bb7b05cf1e79aeb45fd2d53` I get a CRC error though?

mr-sven:
ah, yes, I removed CRC from that data because I also modified the device id, so this data was only for ccitt tests.
I've a capture file, ahm should I upload it here or put it in any folder?

mr-sven:
Ah, I've seen the tests repo, I will make an pull request

mr-sven:
Here https://github.com/merbanan/rtl_433_tests/pull/472 .

zuckschwerdt:
Thanks! Is that a manual recording (e.g. `-w filename.cu8`) or an automatic sample capture (e.g.`-S all`)? It seems like an automatic capture but I've never seen an automatic sample that long, nearly a whole second and full 15 sample buffers long. Interesting!

mr-sven:
I used `rtl_433 -f 868.95M -R 269 -S known` and triggered an register read. The protocol is very complex and there is still some guessed data left. If you are interested in you can take a look in my X3D controller repo https://github.com/mr-sven/x3d-rfm-esp32/blob/main/X3D-Protocol.md

PR #1915, opened 2021-12-23
Add rtl_tcp pass-through output 

zuckschwerdt:
This adds an rtl_tcp pass-through output for received I/Q data.
Use with `-F rtl_tcp` / `-F rtl_tcp:127.0.0.1:1234` and Gqrx to see the live data as rtl_433 sees it.

### DIFF
--- a/include/output_rtltcp.h
+++ b/include/output_rtltcp.h
@@ -1,0 +1,31 @@
+/** @file
+    rtl_tcp output for rtl_433 raw data.
+
+    Copyright (C) 2022 Christian Zuckschwerdt
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#ifndef INCLUDE_OUTPUT_RTLTCP_H_
+#define INCLUDE_OUTPUT_RTLTCP_H_
+
+#include "raw_output.h"
+
+#include <stdint.h>
+
+struct r_cfg;
+
+/** Construct rtl_tcp data output.
+
+    @param host the server host to bind
+    @param port the server port to bind
+    @param cfg the r_api config to use
+    @return The initialized rtltcp output instance.
+            You must release this object with raw_output_free once you're done with it.
+*/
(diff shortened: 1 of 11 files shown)


zuckschwerdt:
This is just a work-in-progress or proof-of-concept and not finished.

Be sure to match frequency and sample rate in Gqrx. Setting different parameters from Gqrx is not supported for now.

bkenobi:
I understand this isn't ready for release but how would I install this to test?  I just cloned the main branch and recompiled but then realized as a testing feature it won't be part of the main branch yet.

Mindavi:
You can checkout the feat-rtsrv branch in git.

zuckschwerdt:
I have some changes for a more performant ring buffer and sendfile api, but this should be working well now and is ready for tests and merge.

I guess it would be rather useful to just connect to any running rtl_433 and check the live data coming in.
Be sure to match frequency and sample rate in the UI. Setting different parameters from a UI is not supported for now.

Note that there is https://github.com/pothosware/SoapyRTLTCP to make this work on any SoapySDR-enabled GUI (CubicSDR, SDRangel, SigDigger, ...), Gqrx has support for rtl_tcp built-in.

zuckschwerdt:
I'd like to merge this in the next days, final call for testing and comments :)

Dinosaur-Ceiling:
Not sure if this has been closed or not, I cannot get the -F rtl_tcp to function. I am using the RTL-SDR driver application on Android which is a port of rtl_tcp to the platform which allows other applicatoions to start rtl_tcp and use the raw I/Q samples that it reads from USB DVB-T dongles based on the RTL2832U chip. The end goal being to use the RTL with an OTG cable to survey TPMS data and visualize it with the Kismet GUI. 

I attempted the following command with the intent of taking rtl_433 output and making it visible in Kismet: `rtl_433  -f 315000000 -d rtl_tcp:127.0.0.1:14423 -F rtl_tcp:127.0.0.1:2501` Kismet can see the RTL_SDR USB driver, but from what I can tell, since I am running Kali Nethunter, it is not seeing any of the data being generated and would required a TCP passthrough; however, this command does not work and generates an error of " invalid output format".

zuckschwerdt:
That command looks ok, but you need a recent version from git master and compile yourself. The last release is older that the rtl_tcp output support.

PR #1918, opened 2021-12-26
src/optparse.c: fix build without __has_attribute

ffontaine:
Fix build failure without `__has_attribute` (e.g. gcc 4.8) which is raised since https://github.com/merbanan/rtl_433/commit/6c8af75c757712bd58b169317795484a72e9a16c

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/optparse.c
+++ b/src/optparse.c
@@ -231,7 +231,7 @@ int atoi_time(char const *str, char const *error_hint)
                 break;
             }
             // intentional fallthrough
-#if defined(__GNUC__) || defined(__clang__)
+#if defined __has_attribute
 #if __has_attribute(fallthrough)
             __attribute__((fallthrough));
 #endif


zuckschwerdt:
You can remove the outer #if (gnuc, clang): some gnuc/clang have the feature but not __has_attribute, ok to ignore those imho.

ffontaine:
Thanks for your review, I updated the PR

zuckschwerdt:
@ffontaine we aim to support older and restricted/embedded platforms with rtl_433, we don't have much feedback on that though.
Can you share some use-case or story where you are stuck with gcc 4.8. Also what version (if at all) are available of: cmake, doxygen, OpenSSL, rtlsdr, libusb, SoapySDR?
(esp. cmake is interesting, we require 2.6 now, but support is running out and we might want to switch to at least 3.0)

ffontaine:
rtl_433 was added to [buildroot](https://buildroot.org) a few months ago. I sent those patches because they were needed to fix a build failure on the autobuilders. This failure was raised by an arm toolchain from 2014.05-29. You can find the different versions of the cmake/openssl/etc here: http://autobuild.buildroot.org/stats/master.html.

PR #1924, opened 2021-12-30
Renault TPMS 0435R

ebikt:
FSK 9 byte Manchester encoded TPMS with xor checksum, Renault 0435R.
Values are guessed. Note that messages also trigger tpms_jansite (protocol 123) decoder, as they are structurally very similar.

List of compatible Renault vehicles (from: https://www.vdo.com/media/190553/vdo-2017-tpms_2017-05-03.pdf)
 FLUENCE (L30_)
 LAGUNA III (BT0/1)
 LAGUNA III Grandtour (KT0/1)
 LATITUDE (L70_)
 MEGANE III Coupe (DZ0/1_)
 MEGANE III Grandtour (KZ0/1)
 MEGANE III Hatchback(BZ0_)
 SCÉNIC III (JZ0/1_)
 ZOE (BFM_)

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -219,6 +219,7 @@
     DECL(simplisafe_gen3) \
     DECL(yale_hsa) \
     DECL(regency_fan) \
+    DECL(tpms_renault_0435r) \
 
     /* Add new decoders here. */
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -208,6 +208,7 @@ add_library(r_433 STATIC
     devices/tpms_pmv107j.c
     devices/tpms_porsche.c
     devices/tpms_renault.c
+    devices/tpms_renault_0435r.c
     devices/tpms_toyota.c
     devices/tpms_truck.c
     devices/ts_ft002.c
--- a/src/devices/tpms_renault_0435r.c
+++ b/src/devices/tpms_renault_0435r.c
@@ -1,0 +1,191 @@
+/** @file
+    FSK 9 byte Manchester encoded TPMS with xor checksum, 0435R.
+
+    Copyright (C) 2021 Tomas Ebenlendr <ebik@ucw.cz>
+
+    This program is free software; you can redistribute it and/or modify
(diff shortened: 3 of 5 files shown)


zuckschwerdt:
Great stuff.
Please indent structs by 8 spaces. wrap data_make with clang-format off, output_fields don't match (rad vs speed).
Random thoughts, cmHg sound too wild, it's a meteorologic unit. Could be just an arbitrary scale of 4/3 to cover more kPa with a single byte.

ebikt:
As for cmHg, the dashboard displays 1/100 of that value (with with precision of 1/10, i.e., I see typically values like `2.2` for each tire.) Thus I guess it is a unit. Maybe https://en.wikipedia.org/wiki/Torr which is practically equal to mmHg, i.e. the unit will be 10Torr for the sensor and 1000Torr for drivers dashboad. I will update the description to Torr instead of cmHg if there is no other objection to the unit.

zuckschwerdt:
Seeing `2.2` for a car tire I'd assume "bar", so 220 kPa. But your tire is really at 290 kPa? Somewhat high, is it a truck?
I don't object to Torr or mmHG, but would not assume to see anything other that PSI or kPa (=bar, atü) in this domain.
(also wouldn't 760 Torr = 1 atm = 1013 hPa be the "zero" pressure, i.e. = 0 bar? With ambient pressure units the zero value is vacuum, right?)

ebikt:
Ok, my bad, I mixed the value before and after conversion in my head. The dashboard indeed displays bar. I checked my last logs and the 2.2 matches the value in kPa.

zuckschwerdt:
Can you rebase and replace the Torr references with e.g. "scale is 4/3 kPa" (just so we don't confuse anyone later).
The "model" should be `Renault-0435R`. Indentation for continuation lines, the lines below `data_make`, should be 8 spaces.
Good for merge then.

ebikt:
Ok, I created branch `renault_tpms_0435R-pre_rebase` in my repository. It has added commit `c0867ab` with requested changes.
Then I did rebase and squash of my commits into new commit `c689a96`. This way it can be checked that the source code did not change during rebase.

PR #1935, opened 2022-01-08
Funkbus operates on 433.42MHz not 433.92

evgeni:


zuckschwerdt:
Is there a spec for this, or is it found for one device? I guess "433.92Mhz" refers to the band name.

evgeni:
https://www.manualslib.de/manual/276073/Insta-5-7922-004.html?page=8#manual says 433,42
and the Gira (:smile:) device I have here operate on .42

PR #1948, opened 2022-01-20
Add support for Regency ceiling fans

dtiller:
This decoder handles Regency brand ceiling fan commands. I will upload test data soon.

### DIFF
--- a/src/devices/regency_fan.c
+++ b/src/devices/regency_fan.c
@@ -225,6 +225,6 @@ r_device regency_fan = {
     .gap_limit   = 8000,
     .reset_limit = 14000,
     .decode_fn   = &regency_fan_decode,
-    .disabled    = 1, // disabled and hidden, use 0 if there is a MIC, 1 otherwise
+    .disabled    = 0, // disabled and hidden, use 0 if there is a MIC, 1 otherwise
     .fields      = output_fields,
 };
--- a/src/devices/regency_fan.c
+++ b/src/devices/regency_fan.c
@@ -1,0 +1,230 @@
+/** @file
+    Decoder for Regency fan remotes
+
+    Copyright (C) 2020-2022 David E. Tiller
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+ */
+
+/**
+ The device uses OOK_PULSE_PPM encoding.
+ The packet starts with 576 uS start pulse.
+ - 0 is defined as a 375 uS gap followed by a 970 uS pulse.
+ - 1 is defined as a 880 uS gap followed by a 450 uS pulse.
+
+ Transmissions consist of the start bit followed by bursts of 20 bits.
+ These packets ar repeated up to 11 times.
+ 
+ As written, the PPM code always interpets a narrow gap as a 1 and a
+ long gap as a 0, however the actual data over the air is inverted,
+ i.e. a short gap is a 0 and a long gap is a 1. In addition, the data
+ is 5 nibbles long and is represented in Little-Endian format. In the
+ code I invert the bits and also reflect the bytes. Reflection introduces
+ an additional nibble at bit offsets 16-19, so the data is expressed a 3
+ complete bytes.
+
+ The examples below are _after_ inversion and reflection (MSB's are on 
+ the left).
+
+ Packet layout
+ Bit number
+ 0  1  2  3  4  5  6  7  8  9  10 11 12 13 14 15 16 17 18 19 20 21 22 23
+  CHANNEL  |  COMMAND  |            VALUE       | 0  0  0  0| 4 bit checksum
+
+ CHANNEL is determined by the bit switches in the battery compartment. All
+ switches in the 'off' position result in a channel of 15, implying that the
+ switches pull the address lines down when in the on position.
+ 
+ COMMAND is one of the following:
+
+ CMD_STOP	 0x01
+	value: (0xc0, unused).
+
+ CMD_FAN_SPEED	 0x02
+	value: 0x01-0x07. On my remote, the speeds are shown as 8 - value.
+
+ CMD_LIGHT_INT	 0x04
+	value: 0x00-0xc3. The value is the intensity percentage.
+		0x00 id off, 0xc3 is 99% (full).
+
+ CMD_LIGHT_DELAY 0x05
+	value: 0x00 is 'off', 0x01 is on.
+
+ CMD_FAN_DIR	 0x06
+	value: 0x07 is one way, 0x83 is the other.
+
+ The CHECKSUM is calculated by adding the nibbles of the first two bytes
+ and ANDing the result with 0x0f. 
+
+ */
+
+#include <stdlib.h>
+#include "decoder.h"
+
+#define NUM_BITS        21
+#define NUM_BYTES       3
+#define BYTE_START      1
+
+#define CMD_CHAN_BYTE   0
+#define VALUE_BYTE      1
+#define SUM_BYTE        2
+
+#define CMD_STOP        1
+#define CMD_FAN_SPEED   2
+#define CMD_LIGHT_INT   4
+#define CMD_LIGHT_DELAY 5
+#define CMD_FAN_DIR     6
+
+static char *command_names[] = {
+    /* 0  */ "invalid",
+    /* 1  */ "fan_speed",
+    /* 2  */ "fan_speed",
+    /* 3  */ "invalid",
+    /* 4  */ "light_intensity",
+    /* 5  */ "light_delay",
+    /* 6  */ "fan_direction",
+    /* 7  */ "invalid",
+    /* 8  */ "invalid",
+    /* 9  */ "invalid",
+    /* 10 */ "invalid",
+    /* 11 */ "invalid",
+    /* 12 */ "invalid",
+    /* 13 */ "invalid",
+    /* 14 */ "invalid",
+    /* 15 */ "invalid"
+};
+
+static int regency_fan_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+
+    data_t *data = NULL;
+    int index = 0; // a row index
+    int num_bits = 0;
+    int debug_output = decoder->verbose;
+    int return_code = 0;
+
+    if (debug_output > 1) {
+        bitbuffer_printf(bitbuffer, "%s: ", __func__);
+    }
+
+    if (bitbuffer->num_rows < 1) {
+        if (debug_output > 1) {
+            fprintf(stderr, "No rows.\n");
+        }
+
+        return 0;
+    }
+
+    bitbuffer_invert(bitbuffer);
+
+    for (index = 0; index < bitbuffer->num_rows; index++) {
+        num_bits = bitbuffer->bits_per_row[index];
+
+        if (num_bits != NUM_BITS) {
+            if (debug_output > 1) {
+                fprintf(stderr, "Expected %d bits, got %d.\n", NUM_BITS, num_bits);
+            }
+
+            continue;
+        }
+
+        uint8_t bytes[NUM_BYTES];
+        bitbuffer_extract_bytes(bitbuffer, index, BYTE_START, bytes, NUM_BITS);
+        reflect_bytes(bytes, NUM_BYTES);
+
+        // Calculate nibble sum and compare
+        int checksum = add_nibbles(bytes, 2) & 0x0f;
+        if (checksum != bytes[SUM_BYTE]) {
+            if (debug_output > 1) {
+                fprintf(stderr, "Checksum failure: expected %0x, got %0x\n", bytes[SUM_BYTE], checksum);
+            }
+
+            continue;
+        }
+
+        /*
+         * Now that message "envelope" has been validated, start parsing data.
+         */
+        uint8_t command = bytes[CMD_CHAN_BYTE] >> 4;
+        uint8_t channel = ~bytes[CMD_CHAN_BYTE] & 0x0f;
+	uint32_t value = bytes[VALUE_BYTE];
+        char value_string[64] = {0};
+
+        switch(command) {
+            case CMD_STOP:
+                sprintf(value_string, "stop");
+                break;
+
+            case CMD_FAN_SPEED:
+                sprintf(value_string, "speed %d", value);
+                break;
+
+            case CMD_LIGHT_INT:
+                sprintf(value_string, "%d %%", value);
+                break;
+
+            case CMD_LIGHT_DELAY:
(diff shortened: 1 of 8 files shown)


src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
+/** @file
+    Decoder for Regency fan remotes
```
zuckschwerdt, line 2: end with a full stop (.)

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+    Copyright (C) 2020-2022 David E. Tiller
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+ */
+
+/**
+ The device uses OOK_PULSE_PPM encoding.
```
zuckschwerdt, line 13: Repeat the device name.

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+ is 5 nibbles long and is represented in Little-Endian format. In the
+ code I invert the bits and also reflect the bytes. Reflection introduces
+ an additional nibble at bit offsets 16-19, so the data is expressed a 3
+ complete bytes.
+
+ The examples below are _after_ inversion and reflection (MSB's are on
+ the left).
+
+ Packet layout
+ Bit number
```
zuckschwerdt, line 33: indent code blocks by four spaces

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+
+ CMD_FAN_DIR     0x06
+        value: 0x07 is one way, 0x83 is the other.
+
+ The CHECKSUM is calculated by adding the nibbles of the first two bytes
+ and ANDing the result with 0x0f.
+
+ */
+
+#include <stdlib.h>
```
zuckschwerdt, line 64: remove stdlib

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+#include "decoder.h"
+
+#define NUM_BITS        21
```
zuckschwerdt, line 67: expand all defines into the code.

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+static int regency_fan_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+
+    data_t *data = NULL;
+    int index = 0; // a row index
+    int num_bits = 0;
+    int debug_output = decoder->verbose;
+    int return_code = 0;
+
+    if (debug_output > 1) {
```
zuckschwerdt, line 109: remove this

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+        bitbuffer_printf(bitbuffer, "%s: ", __func__);
+    }
+
+    if (bitbuffer->num_rows < 1) {
```
zuckschwerdt, line 113: remove this

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+        if (debug_output > 1) {
+            fprintf(stderr, "No rows.\n");
+        }
+
+        return 0;
+    }
+
+    bitbuffer_invert(bitbuffer);
+
+    for (index = 0; index < bitbuffer->num_rows; index++) {
```
zuckschwerdt, line 123: use row or row_num instead of index

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+                fprintf(stderr, "Checksum failure: expected %0x, got %0x\n", bytes[SUM_BYTE], checksum);
+            }
+
+            continue;
+        }
+
+        /*
+         * Now that message "envelope" has been validated, start parsing data.
+         */
+        uint8_t command = bytes[CMD_CHAN_BYTE] >> 4;
```
zuckschwerdt, line 151: use plain `int`

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+
+            default:
+                if (debug_output > 1) {
+                    fprintf(stderr, "Unknown command: %d\n", command);
+                    continue;
+                }
+                break;
+        }
+
+        return_code = 1;
```
zuckschwerdt, line 185: `        return_code += 1;`

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+
+        data = data_make(
```
zuckschwerdt, line 187: wrap in `/* clang-format off */`
indent

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+        "model",            "",     DATA_STRING,    "Regency-compatible Remote",
+        "type",             "",     DATA_STRING,    "Ceiling Fan",
```
zuckschwerdt, line 189: remove `type`

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+                    fprintf(stderr, "Unknown command: %d\n", command);
+                    continue;
+                }
+                break;
+        }
+
+        return_code = 1;
+
+        data = data_make(
+        "model",            "",     DATA_STRING,    "Regency-compatible Remote",
```
zuckschwerdt, line 188: Model should be "Regency-Remote"

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+ * List of fields that may appear in the output
+ *
+ * Used to determine what fields will be output in what
+ * order for this device when using -F csv.
+ *
+ */
+static char *output_fields[] = {
+    "model",
+    "type",
+    "channel",
```
zuckschwerdt, line 213: indent cont' lines by 8

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+
+r_device regency_fan = {
+    .name        = "Regency Ceiling Fan Remote (-f 303.75M to 303.96M)",
+    .modulation  = OOK_PULSE_PWM,
+    .short_width = 580,
+    .long_width  = 976,
+    .gap_limit   = 8000,
+    .reset_limit = 14000,
+    .decode_fn   = &regency_fan_decode,
+    .disabled    = 0, // disabled and hidden, use 0 if there is a MIC, 1 otherwise
```
zuckschwerdt, line 228: remove

src/devices/regency_fan.c
```diff
@@ -0,0 +1,230 @@
...
+        NULL);
+
+        decoder_output_data(decoder, data);
+    }
+
+    // Return 1 if message successfully decoded
+    return return_code;
+}
+
+/*
```
zuckschwerdt, line 203: remove this comment

src/devices/regency_fan.c
dtiller: What is your rationale for not using #define constants? They give an awful lot of context in the code as to what the constant _means_, not just it's value.

int value = MAX_VALUE;

is a lot clearer than

int value = 37;

src/devices/regency_fan.c
dtiller: Why would you add the possibility of returning a non- 0 or 1 value here? If the return_code is (erroneously) set in the code in a leter edit, you run the risk of returning something other than 0 or 1.

zuckschwerdt: return is the *number of* events generated, or an error code.

src/devices/regency_fan.c
zuckschwerdt: what's `MAX_VALUE` then? which data does it apply to? does it change? ;)
`int value = 37; // maximum expected row length, shorter for "learn" packet`

Explain bits, nibbles, bytes in the comment text above (already looks good), that's much better than `BYTE_START`(is it a start/sync byte? or counted from where?).
The decoder code then can't avoid bit manipulation and numbers, embrace that ;) and if needed comment better than a #define could, e.g. ` case 1: // stop command`

but TL;DR, it's easier to reason about and maintain 200+ decoders without a preprocessor involved.

src/devices/regency_fan.c
```diff
@@ -0,0 +1,218 @@
...
+
+            continue;
+        }
+
+        /*
+         * Now that message "envelope" has been validated, start parsing data.
+         */
+        int command = bytes[CMD_CHAN_BYTE] >> 4;
+        int channel = ~bytes[CMD_CHAN_BYTE] & 0x0f;
+        uint32_t value = bytes[VALUE_BYTE];
```
zuckschwerdt, line 140: should be plain `int`

dtiller:
Not all architectures default to 32 bits for an Int, which this variable requires.  It is not definitive in the C spec. Using these typedefs guarantee the decoder will work on all platforms.

zuckschwerdt:
> 32 bits which this variable requires

you lost me, I see a single byte? (and then `%d` is wrong for `uint32_t`)

src/devices/regency_fan.c
dtiller: In the end it's your code, but I disagree about #defines and using explicit width types. As I said there are several hundred #defines in the existing device code, not to mention the rtl_433 code itself. You explicitly use uint8_t in your function definitions, so it's not like I'm adding something that's otherwise unused.

merbanan:
uint8_t bytes[NUM_BYTES];
uint32_t value = bytes[VALUE_BYTE];

sprintf(value_string, "speed %d", value);

So we have u8->u32 and then %d in the sprintf. When reading about printf %d needs int. And int will hold u8 so I don't understand what you are referring to. In practice this means nothing but I aim to keep things as simple as possible if possilbe.

dtiller:
Using the standard uint8_t-like typedefs in stdint.h is only part of the solution. For printf/scanf there are the constants defined in inttypes.h - these really are part of the C99 spec.

https://debrouxl.github.io/gcc4ti/inttypes.html <https://debrouxl.github.io/gcc4ti/inttypes.html>

Examples:

PRId8 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRId8>
Format specifier for printing an integer of type int8_t or uint8_t as a signed decimal integer.
PRId16 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRId16>
Format specifier for printing an integer of type int16_t or uint16_t as a signed decimal integer.
PRId32 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRId32>
Format specifier for printing an integer of type int32_t or uint32_t as a signed decimal integer.
PRIdFAST8 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRIdFAST8>
Format specifier for printing an integer of type int_fast8_t or uint_fast8_t as a signed decimal integer.
PRIdFAST16 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRIdFAST16>
Format specifier for printing an integer of type int_fast16_t or uint_fast16_t as a signed decimal integer.
PRIdFAST32 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRIdFAST32>
Format specifier for printing an integer of type int_fast32_t or uint_fast32_t as a signed decimal integer.
PRIdLEAST8 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRIdLEAST8>
Format specifier for printing an integer of type int_least8_t or uint_least8_t as a signed decimal integer.
PRIdLEAST16 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRIdLEAST16>
Format specifier for printing an integer of type int_least16_t or uint_least16_t as a signed decimal integer.
PRIdLEAST32 <https://debrouxl.github.io/gcc4ti/inttypes.html#PRIdLEAST32>
Format specifier for printing an integer of type int_least32_t or uint_least32_t as a signed decimal integer.

zuckschwerdt:
bottom line: we have a lot of code to manage, we need it simple, plain and uniform; all decoders should roughly look the same, the odd ones are not great and need to get fewer. And the PR is likely the only chance to get help by the author.

Thanks for sticking with this and working through all the requests. Looks good to merge.

merbanan:
LGTM

sheilbronn:
Hi there,
I seem to have a regency remote control in my suburban environment and I noticed some invalid percentages (192%,  195%) being emitted. Are these valid values or should the corresponding signal be suppressed before output?

```
26 06:25:38 {"protocol":211,"model":"Regency-Remote","channel":15,"command":"light_intensity","value":"192 %","mod":"ASK","freq":433.85,"rssi":-9.729,"snr":6.362,"noise":-16.091}
20 17:18:52 {"protocol":211,"model":"Regency-Remote","channel":15,"command":"light_intensity","value":"0 %","mod":"ASK","freq":433.883,"rssi":-2.178,"snr":4.2,"noise":-6.377}
18 19:44:25 {"protocol":211,"model":"Regency-Remote","channel":11,"command":"light_intensity","value":"0 %","mod":"ASK","freq":433.854,"rssi":-1.383,"snr":12.218,"noise":-13.601}
21 19:49:24 {"protocol":211,"model":"Regency-Remote","channel":1,"command":"light_intensity","value":"195 %","mod":"ASK","freq":433.927,"rssi":-6.751,"snr":12.841,"noise":-19.591}
23 20:57:12 {"protocol":211,"model":"Regency-Remote","channel":7,"command":"light_intensity","value":"64 %","mod":"ASK","freq":433.898,"rssi":-4.097,"snr":4.336,"noise":-8.434}
```

There are also fan_speed command on the same channel but I don't know whether this makes sense at all...
I attached the Regency readings that got picked in my logs recently as a file attachment: [regency.log](https://github.com/merbanan/rtl_433/files/8560632/regency.log)

Let me know if I can help with more data, but I don't know the owner of the device....
Shall I open a separate issue?

dtiller:
Sheilbronn,

My first guess is that what you're seeing is not a regency remote but something else that uses a similar-enough protocol to be decoded. Unless the owner of the remote has 4 fans, you won't see the channel jumping around, and the fan speed values would increase/decrease monotonically, not go from 'stopped' to 'two' in one go.

If you want to capture more data I can see if I can tighten the code to more accurately detect a regency remote. Use the -S option with either 'known' or 'all'.

sheilbronn:
> If you want to capture more data I can see if I can tighten the code to more accurately detect a regency remote. 

I'm happy to give it a try. These signals are quite rare, though.

> Use the -S option with either 'known' or 'all'.

I restarted rtl_433 with -S known.
Now my log dir is filling up with files named like "g0XX_433.9M_1024k.cu8".
Am I supposed to watch out for the Regency remote signal in my rtl_433 log file and then find out the corresponding cu8 file that caused the signal ?!

dtiller:
Ouch! You might want to enable only the regency decoder and select known. I think that would be rtl_433 -R 211 -S known etc

See if that doesn’t record everything.

Issue #1312, opened 2020-02-10
Acurite TXR, 5n1, 6045 message integrity, add parity check.


rct:
Acurite TXR, 5n1, 6045 and possibly others like the new Atlas weather station use parity in addition to checksums for integrity checking. 

Currently the parity isn't verified, it's only looked at in verbose mode.

I've noticed I seem to get messages for non-existent devices from time to time. I suspect the message is mangled but the checksum still passes.  The parity check in addition to the checksum might reduce these.

Only the data bytes.  in the Acurite message have parity checks.  The ID bytes, and checksum byte are 8 bits of data with no parity.

(Note: I'm opening this as an issue to let others know. I will eventually fix this, but it will probably be a few weeks before I get some time and am near my Acurite gear for testing.)

ChrisM510:
Let me know if I can help... I might need detailed instructions on how to gather the info, but I have RTL_433 + SDR + WeeWX running on my Pi4.

jcwren:
If you need any data from an Acurite 6045M, I just acquired one. Just tell me what you need.

jcwren:
I see that the 6045M decoder has two unknown bits. One of these might indicate that the remote has gone into silent mode for 2 hours after 10 strike detections. Manual doesn't say anything about a time window on those 10 strikes. I'd hope they'd have to occur within a 5 or 10 minute period.

rct:
I haven't touched this in a while, but I need to get back to it.

I think I was down to 1 unknown bit - `ussb1`, and I believe I finally figured out is it really the LSB of the strike count bit, which changes that to 8 bits - wraps at 255 strikes, and matches what people were occasionally reporting about their consoles. 

I haven't seen a thunderstorm in months.

The 10 strikes count is just for stopping the unit from beeping.  I haven't seen anything that would indicate it needs to send that in the message.  There really isn't a reason to send that as any receiver could make that determination itself.

I've noticed via the `active` bit, that my sensor seems to cycle into active mode very regularly about once an hour.

jcwren:
Hrm, not sure about USSB1. I don't have any hits, but it's set. I do see the active bit flipping periodically. Here's the log to date where active has flipped (obviously I deleted all the lines where it was the same for some duration).

```
[...]

PR #1953, opened 2022-01-23
Add support for AcuRite 01190 Leak Detector

qwindelzorf:
This PR adds support for the AcuRite 01190M leak detector module (instruction manual [here](https://www.acurite.com/media/manuals/01190M-instructions.pdf)). 

As a side effect of adding support for this sensor, the integrity checks for a few other similar sensors have been improved, most specifically the AcuRite Wireless Temperature and Humidity Sensor (called a tower sensor in the existing code base).

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -852,8 +852,6 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             // value if they successfully decoded a message.
             if (decoded > 0) {
                 valid++;
-            } else {
-                continue;
             }
         }
 
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -717,20 +717,42 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         // TODO: - see if there is a type in the message that
         // can be used instead of length to determine type
         if (browlen == ACURITE_TXR_BITLEN / 8) {
+            // TODO: Verify parity bits
+
+            // Verify checksum, add with carry
+            int chk = add_bytes(bb, 6);
+            if ((chk & 0xff) != bb[6]) {
+                if (decoder->verbose)
+                    bitrow_printf(bb, 7 * 8, "%s: bad checksum: ", __func__);
+                continue; // return DECODE_FAIL_MIC;
+            }
+
             char const *channel_str = acurite_getChannel(bb[0]);
             // Tower sensor ID is the last 14 bits of byte 0 and 1
             // CCII IIII | IIII IIII
             sensor_id = ((bb[0] & 0x3f) << 8) | bb[1];
+
             //sensor_status = bb[2]; // TODO:, uses parity? & 0x07f
             humidity = (bb[3] & 0x7f); // 1-99 %rH
+
             // temperature encoding used by "tower" sensors 592txr
             // 14 bits available after removing both parity bits.
             // 11 bits needed for specified range -40 C to 70 C (-40 F - 158 F)
-            // range -100 C to 1538.4 C
+            // Possible ranges are -100 C to 1538.4 C, but most of that range
+            // is not possible on Earth.
             int temp_raw = ((bb[4] & 0x7F) << 7) | (bb[5] & 0x7F);
             tempc = temp_raw * 0.1 - 100;
+            if (tempc < -40 || tempc > 70) {
+                if(decoder->verbose) {
+                    fprintf(stderr, "%s: Acurite TXR sensor 0x%04X Ch %s -- Impossible temperature: %0.2f C\n",
+                            __func__, sensor_id, channel_str, tempc);
+                }
+                continue;
+            }
+
             // Battery status is the 7th bit 0x40. 1 = normal, 0 = low
             battery_low = (bb[2] & 0x40) == 0;
+
 
             /* clang-format off */
             data = data_make(
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -752,7 +752,7 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             battery_low = (bb[2] & 0x40) == 0;
 
             if (message_type == ACURITE_MSGTYPE_LEAK_DETECTOR) {
-                int is_dry = (bb[3] & 0x10) >> 4;
+                int is_wet = (bb[3] & 0x10) >> 4;
 
                 /* clang-format off */
                 data = data_make(
@@ -760,7 +760,7 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
                         "id",                   "",             DATA_INT,    sensor_id,
                         "channel",              NULL,           DATA_STRING, channel_str,
                         "battery_ok",           "Battery",      DATA_INT,    !battery_low,
-                        "leak_detected",        "Leak",         DATA_INT,    !is_dry,
+                        "leak_detected",        "Leak",         DATA_INT,    is_wet,
                         "mic",                  "Integrity",    DATA_STRING, "CHECKSUM",
                         NULL);
                 /* clang-format on */
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -637,6 +637,136 @@ static int acurite_atlas_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsig
     return 1; // one valid message decoded
 }
 
+int acurite_tower_sensor_decode(r_device*, uint8_t*);
+int acurite_leak_detector_decode(r_device*, uint8_t*);
+
+int acurite_tower_sensor_decode(r_device* decoder, uint8_t* bb)
+{
+    // checksum in the last byte has been validated in the calling function
+
+    // Verify parity bits
+    // Bytes 2, 3, 4, and 5 should all have a parity bit in their MSB
+    int parity = parity_bytes(&bb[2], 4);
+    if (parity) {
+        if (decoder->verbose)
+            bitrow_printf(bb, 7 * 8, "%s: bad parity: ", __func__);
+        return DECODE_FAIL_MIC;
+    }
+
+    // Channel is the first two bits of the 0th byte
+    // but only 3 of the 4 possible values are valid
+    char const* channel_str = acurite_getChannel(bb[0]);
+    if (*channel_str == 'E') {
+        if (decoder->verbose)
+            fprintf(stderr, "%s: Acurite TXR sensor : bad channel Ch %s\n", __func__, channel_str);
+        return DECODE_FAIL_SANITY;
+    }
+
+    // Tower sensor ID is the last 14 bits of byte 0 and 1
+    // CCII IIII | IIII IIII
+    int sensor_id = ((bb[0] & 0x3f) << 8) | bb[1];
+
+    // Battery status is the 7th bit 0x40. 1 = normal, 0 = low
+    // pxxx xxxB
+    int battery_low = (bb[2] & 0x40) == 0;
+
+    // Humidity is stored in byte 3
+    // The value is directly encoded as %rH
+    // The possible values here are 0-128, but the manufacturer specifies that valid values
+    // are only 1-99 %rH
+    // pIII IIII
+    int humidity = (bb[3] & 0x7f);
+    if (humidity < 0 || humidity > 100) {
+        if (decoder->verbose) {
+            fprintf(stderr, "%s: Acurite TXR sensor 0x%04X Ch %s : Impossible humidity: %d %%rH\n",
+                __func__, sensor_id, channel_str, humidity);
+        }
+        return DECODE_FAIL_SANITY;
+    }
+
+    // temperature encoding used by "tower" sensors 592txr
+    // 14 bits available after removing both parity bits.
+    // 11 bits needed for specified range -40 C to 70 C (-40 F - 158 F)
+    // Possible ranges are -100 C to 1538.4 C, but most of that range
+    // is not possible on Earth.
+    // pIII IIII pIII IIII
+    int temp_raw = ((bb[4] & 0x7F) << 7) | (bb[5] & 0x7F);
+    float tempc = temp_raw * 0.1 - 100;
+    if (tempc < -40 || tempc > 70) {
+        if (decoder->verbose) {
+            fprintf(stderr, "%s: Acurite TXR sensor 0x%04X Ch %s : Impossible temperature: %0.2f C\n",
+                __func__, sensor_id, channel_str, tempc);
+        }
+        return DECODE_FAIL_SANITY;
+    }
+
+    data_t* data;
+    /* clang-format off */
+    data = data_make(
+            "model",                "",             DATA_STRING, "Acurite-Tower",
+            "id",                   "",             DATA_INT,    sensor_id,
+            "channel",              NULL,           DATA_STRING, channel_str,
+            "battery_ok",           "Battery",      DATA_INT,    !battery_low,
+            "temperature_C",        "Temperature",  DATA_FORMAT, "%.1f C", DATA_DOUBLE, tempc,
+            "humidity",             "Humidity",     DATA_FORMAT, "%u %%", DATA_INT,    humidity,
+            "mic",                  "Integrity",    DATA_STRING, "CHECKSUM",
+            NULL);
+    /* clang-format on */
+
+    decoder_output_data(decoder, data);
+
+    return 1;
+}
+
+int acurite_leak_detector_decode(r_device* decoder, uint8_t* bb)
+{
+    // checksum in the last byte has been validated in the calling function
+
+    // Verify parity bits
+    // Bytes 2, 3, 4, and 5 should all have a parity bit in their MSB
+    int parity = parity_bytes(&bb[2], 4);
+    if (parity) {
+        if (decoder->verbose)
+            bitrow_printf(bb, 7 * 8, "%s: bad parity: ", __func__);
+        return DECODE_FAIL_MIC;
+    }
+
+    // Channel is the first two bits of the 0th byte
+    // but only 3 of the 4 possible values are valid
+    char const* channel_str = acurite_getChannel(bb[0]);
+    if (*channel_str == 'E') {
+        if (decoder->verbose)
+            fprintf(stderr, "%s: Acurite TXR sensor : bad channel Ch %s\n", __func__, channel_str);
+        return DECODE_FAIL_SANITY;
+    }
+
+    // Tower sensor ID is the last 14 bits of byte 0 and 1
+    // CCII IIII | IIII IIII
+    int sensor_id = ((bb[0] & 0x3f) << 8) | bb[1];
+
+    // Battery status is the 7th bit 0x40. 1 = normal, 0 = low
+    int battery_low = (bb[2] & 0x40) == 0;
+
+    // Leak indicator bit is the 5th bit of byte 3. 1 = wet, 0 = dry
+    int is_wet = (bb[3] & 0x10) >> 4;
+
+    data_t* data;
+    /* clang-format off */
+    data = data_make(
+            "model",                "",             DATA_STRING, "Acurite-Leak",
+            "id",                   "",             DATA_INT,    sensor_id,
+            "channel",              NULL,           DATA_STRING, channel_str,
+            "battery_ok",           "Battery",      DATA_INT,    !battery_low,
+            "leak_detected",        "Leak",         DATA_INT,    is_wet,
+            "mic",                  "Integrity",    DATA_STRING, "CHECKSUM",
+            NULL);
+    /* clang-format on */
+
+    decoder_output_data(decoder, data);
+
+    return 1;
+}
+
 /**
 This callback handles several Acurite devices that use a very
 similar RF encoding and data format:
@@ -658,7 +788,7 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 {
     int browlen, valid = 0;
     uint8_t *bb;
-    float tempc, tempf, wind_dir, wind_speed_kph, wind_speed_mph;
+    float tempf, wind_dir, wind_speed_kph, wind_speed_mph;
     uint8_t humidity, sequence_num, message_type;
     // uint8_t sensor_status;
     uint16_t sensor_id;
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Looks great!

Best to avoid library functions, `if(strcmp(channel_str, "E") == 0)` could be `if (*channel_str == 'E')`

qwindelzorf:
> Best to avoid library functions, `if(strcmp(channel_str, "E") == 0)` could be `if (*channel_str == 'E')`

I opted for `strcmp` because the originating function returns a string-like thing. At the moment, it's a single character, but it could easily be changed to a multi-character value, and this code has no control over that. There's already precedent for this, as there's another AcuRite channel decode function that returns 2-character strings.

qwindelzorf:
Looking through the contribution docs, I see that clang-format is being used. However, I have not run it over the changed file in this PR because it ends up changing a bunch of whitespace in unrelated code. Instead, i tried to match the existing code style as best I could. I'm not sure how that is looked upon in this kind of a PR. I can add that cleanup if it is desired.

zuckschwerdt:
> have not run it over the changed file in this PR because it ends up changing a bunch of whitespace

yes, limit format changes to only the new code. Looks good.

zuckschwerdt:
But fix the sticky-paren and trailing whitespace lines ;)

zuckschwerdt:
> I opted for `strcmp` because the originating function returns a string-like thing.

True. We try to avoid intricate code in the decoders, esp. string and mem function, where possible.

I just looked at `acurite_getChannelAndType()` and it might be said that checking the first char is better to detect the error condition ;)

qwindelzorf:
> > I opted for `strcmp` because the originating function returns a string-like thing.
> 
> True. We try to avoid intricate code in the decoders, esp. string and mem function, where possible.
> 
> I just looked at `acurite_getChannelAndType()` and it might be said that checking the first char is better to detect the error condition ;)

Yeah, I'm not convinced of the validity of that function anyway. It looks like it's probably decoding the same channel data as the normal function, but also picking up unrelated bits, which are actually a different field. But, again, not gonna' touch it since its unrelated.

I'll switch to just comparing the byte directly. It's cheaper and safer.

zuckschwerdt:
> not convinced of the validity of that function anyway.

Ideally we want a single straight flow of decoding in every decoder, small helper functions might be good coding style but are a distraction here. A single block of this-is-where-all-the-bits-go is more obvious.

qwindelzorf:
> > not convinced of the validity of that function anyway.
> 
> Ideally we want a single straight flow of decoding in every decoder, small helper functions might be good coding style but are a distraction here. A single block of this-is-where-all-the-bits-go is more obvious.

Seems reasonable to me.  Fixed.

qwindelzorf:
> But fix the sticky-paren and trailing whitespace lines ;)

Fixed

qwindelzorf:
Bump?

zuckschwerdt:
I just made a mess at #1955, we need to sort that first.

qwindelzorf:
> I just made a mess at #1955, we need to sort that first.

No worries. I just wanted to make sure you weren't waiting on me for something I had missed.

rct:
Thanks for doing this.  There is a bunch of stuff in here that needed to be rationalized but I’ve never gotten around to it.

note that the “tower” code is shared by a family of Acurite devices, Atlas, 5n1, 3n1, 6045m lightning,etc.  so it would be good to make sure all of those still work.

off the top of my head, I have a recollection of some discrepancy in which bytes are parity checked on a per device basis. I believe the bytes containing bits of the ID are never parity checked, but the number of bits used for ID varies between devices.  I doubt we have samples with ID bytes representing even/odd parity.  If we ever add a test suite of bit buffers, this would be one of those cases to addd unit tests for.

After this PR gets implemented, can we tell how many decodes are failing due to the MIC checks?

It is interesting to me that you’ve also added a check for values that are out of the specified range, like below -40 or above 158.  This has been on my mind for a while. I have a problem with bad data getting through with the Ambient FT007H sensors on a regular basis. I believe I saw a small discussion of filtering bad data in a recent PR for a new decoder.  I think maybe the discussion of filtering should take place in a separate issue.

rct (commented):
I think the humidity check should be widened based on observations: to include at least 0 and 100.

For example, Humidity will be 0 if the battery is low enough or the humidity sensor fails.  While the humidity might be invalid, there is useful data in the rest of the message like he battery low indicator.

there is also the diagnostic value of being able to determine that the station is being received but one sensor isn’t working.  The 5n1 will send humidity 0 and temp -40 when the battery is low enough.

Also a more general and compact data filtering construct might be needed to avoid many decoders winding up with these nested if bocks with debug output.

DAVe3283:
I have about 50 GiB of RF dumps from a number of AcuRite sensors, including 3 of the 5n1 weather stations and 9 of the tower temp/humidity sensors. I would be happy to share however many samples are desired.

And much more interesting, I managed to capture a sample that _should_ be failed out. It currently decodes as over 700°C: [g16356_433.92M_250k.zip](https://github.com/merbanan/rtl_433/files/7934613/g16356_433.92M_250k.zip)

qwindelzorf:
> note that the “tower” code is shared by a family of Acurite devices, Atlas, 5n1, 3n1, 6045m lightning,etc. so it would be good to make sure all of those still work

I've verified the tower. leak sensor, and 5n1 all still work, since that's what I have access to. Though, as implemented, this should only impact the Tower and Leak sensors, since the changes are all enclosed by a check based on number of bits. The changes here probably should be rolled out more broadly for the other AcuRite devices, but that ought to be part of a different PR which refactors a bunch of this code to deduplicate stuff. As-is, I wanted this PR to be minimally invasive.

> off the top of my head, I have a recollection of some discrepancy in which bytes are parity checked on a per device basis. I believe the bytes containing bits of the ID are never parity checked

You are correct. The first two bytes (the header) and the last byte (the checksum) don't have parity. This PR only checks parity on bytes 2-5.

qwindelzorf:
> I think the humidity check should be widened based on observations: to include at least 0 and 100.

I can change that, but I don't think the tower sensor can ever legitimately report either 0 or 100. I've seen them in both incredibly dry (10-12%rH) environments, and very humid environments, and at the the reported values become somewhat unreliable. There's definitely interesting results there, but those should not be reported as passing and sent on upstream.

> Also a more general and compact data filtering construct might be needed to avoid many decoders winding up with these nested if bocks with debug output.

Agreed. But that would need to be part of a bigger refactor PR. The whole AcuRite decoder needs to be reworked to share a lot of the logic that is constant across the whole family of sensors.

DAVe3283:
I can go either way on the sanity checks. The sensor documentation says it will never go outside the ranges used in this PR, so filtering them out seems reasonable (it implies a failed sensor), but at the same time, it should be redundant with the parity & checksum. I suspect the parity & checksum will capture the vast majority of bad decodes.

I personally lean slightly towards allowing any valid message (parity + checksum are good) through, even if the values are bogus (like the examples above with low battery or failed humidity sensor).

qwindelzorf:
> > For example, Humidity will be 0 if the battery is low enough or the humidity sensor fails. While the humidity might be invalid, there is useful data in the rest of the message like he battery low indicator.
> 
> Yes, please don't filter on what is reasonable but only on what the device would never send. There is always still some interesting information ;)

In this specific case, the [manufacturer calls out](https://www.acurite.com/shop-all/weather-instruments/weather-sensors-and-parts/replacement-parts/indoor-outdoor-temperature-humidity-sensor.html#attributespecifications) that the valid humidity values are 1% to 99%. So, unless folks feel very strongly about this, I'm very hesitant to add those specific values. I can't think of a case in which a 0% or 100% decoded value would be "real" data. It might be *interesting* data, but should not be part of the decoded data stream. Maybe just log those cases for debug?

src/devices/acurite.c
```diff
@@ -713,36 +714,103 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+                continue; // return DECODE_FAIL_MIC;
+            }
+
+            // Verify checksum, add with carry
+            int chk = add_bytes(bb, 6);
+            if ((chk & 0xff) != bb[6]) {
+                if (decoder->verbose)
+                    bitrow_printf(bb, 7 * 8, "%s: bad checksum: ", __func__);
+                continue; // return DECODE_FAIL_MIC;
+            }
```
DAVe3283, line 736: Isn't this checksum check redundant with the one on line 694? If I understand the code correctly, we will never get this far if the first checksum fails, and they appear to calculate over the exact same range and compare the exact same byte.

rct:
I've opened an issue #1964 with some of my thoughts on data sanity checking.

I'm glad to see progress, but I think a PR that adds support for a specific device shouldn't potentially be making changes to other devices.

> I can't think of a case in which a 0% or 100% decoded value would be "real" data. It might be interesting data, but should not be part of the decoded data stream. Maybe just log those cases for debug?

There are a number of problems with this view though. I touch on them in #1964.  

You are discarding the entire messages because of one bad data element.  n some cases, this might be good.  Temperature might also be bad if humidity is bad.  But I think you are making the assumption that every one of these is going to be a message integrity problem. 

For example, if the humidity sensor goes bad and outputs 0, should all data from that device be discarded?  I had a bad humidity sensor in my 5n1 weather station for a while until I could get a replacement board. 

If the humidity is 0 because of a low battery (yes I've also seen this happen) should I not be able to get messages with the battery low flag?

rct:
Speaking of refactoring:

> Though, as implemented, this should only impact the Tower and Leak sensors, since the changes are all enclosed by a check based on number of bits.

I did device detection based on message length a very long time ago when all I had seen so far as the 5n1 and the tower. Now it seems to be clear that acurite developers have been defining message types that can be used to differentiate messages.  

A historical note, at the time I was implementing the tower, there was an existing 5n1 decoder written by others but it had problems. So I created the new protocol for the tower and the 5n1s. Then I added the 6045m lighting later. 

Initially I had gotten the number of bits used in the 6045m ID wrong. I have to go back and look at the various messages based on current understanding and figure out if the bytes that can be parity checked are consistent so that adding parity to the MIC can be done consistently and DRY.

src/devices/acurite.c
```diff
@@ -717,21 +717,43 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+            // is not possible on Earth.
             int temp_raw = ((bb[4] & 0x7F) << 7) | (bb[5] & 0x7F);
             tempc = temp_raw * 0.1 - 100;
+            if (tempc < -40 || tempc > 70) {
+                if(decoder->verbose) {
+                    fprintf(stderr, "%s: Acurite TXR sensor 0x%04X Ch %s -- Impossible temperature: %0.2f C\n",
+                            __func__, sensor_id, channel_str, tempc);
+                }
+                continue;
+            }
```
rct: Note: It we start implementing a lot of sanity checks, these will be contributing to a DRY problem for the decoders.

src/devices/acurite.c
qwindelzorf: good point. I had seen it (and the previously extant parity check on 693) when first reading through this code, but assumed they were non-enforcing. The existing one is more robust anyways. I'll remove the one I added.

rct:
> 
> I guess I didn't make clear in the comments or PR description why the sanity checks were necessary. They are not just "nice to have" in this case. The problem is that the leak detector and tower use the exact same message format, but repurpose bits in the middle of the message, relying upon a single bit to differentiate the messages between them (bit 1 of byte 2, 0-based indexes). 

Ah Ok, now I understand better.

So for readability/clarity both the tower and the leak sensor should check the message type. 

In order to get the leak detector PR merged with minimal changes, would the following be enough to differentiate reliability between the two devices in the absence of an error in the message type bit that still passes the checksum?

- Checking the message type bits for both, 
- Checking the parity on the byte with the message type 

Also you mentioned the checksum wasn't being checked for the tower. I can believe that, but I don't see how that is happening. Maybe I'm just being blind, but I think any row that doesn't have a valid checksum should be skipped by the continue, so will never get processed.  In any case where the checksum isn't being checked is something I'd like to get addressed ASAP.

I do think the whole message should be parity checked. There is an open issue #1312 but it never caused me enough pain.  I have many more problems with bad data from my Ambient FT007H sensors making my Home Assistant graphs occasionally useless.  I think as long as we make sure that any bytes that contain ID bits aren't parity checked, we should be good to implement parity checking for these devices that share the same decoder.

src/devices/acurite.c
DAVe3283: The existing parity code is non-enforcing, really just prints the parity, and I am not sure it is that meaningful. But the checksum check was enforcing/correct from what I can tell.

qwindelzorf:
> So for readability/clarity both the tower and the leak sensor should check the message type.

Agreed. Right now, the tower and leak sensors are split based on an `if (message_type == ...)` check. So, either way, the `message_type` variable is checked. If it is nonsense, the message will not decode as either type. 

> Also you mentioned the checksum wasn't being checked for the tower. I can believe that, but I don't see how that is happening.

Checksum is, and always has been, checked. Parity was being computed, but not checked. This PR adds parity checking for the Tower and Leak sensors only. The existing parity computation was left as-is, not doing anything other than logging. That should probably be cleaned up, but seems like its outside the scope of this PR.

qwindelzorf:
>  I meant in any refactoring, ...

I agree. The AcuRite stuff has evolved to a point where it probably needs a rewrite to DRY things up a bunch.

> Ah Ok thought you added code to check the checksum, that @DAVe3283 flagged, but maybe that was a formatting change.

Yeah, originally, when I was reading through the code the first time, I saw the checksum code but didn't catch that it was enforcing, so I added another checksum validation stage, which was redundant. I just pushed a commit fixing that.

-------

So far, I've only got the one leak detector, and the ID seems to be static. I left it powered off for a week, and it came back up with the same ID. I've got 3 more coming, so I'll have a better idea of their differences after I get the rest online.

The leak detection seems really robust from my initial testing. It picks up quickly, and doesn't seem to false-alarm easily.

I also like that the sensor and transmitter are separable, so you can put the sensor down in some dark hole, and have the transmitter stuck somewhere higher where it won't be blocked. And, you can turn the local alarm off. I embarked on this whole endeavor because I would far prefer to get a notification on my phone when a leak occurs (via `rtl_433`  -> some python scripts -> home assistant) than to have something beeping in a crawlspace under my house where I may not hear it. This also gives a nice way to tell if/when the battery is dying. The AcuRite remote that they sell doesn't seem to have any provision for that.

src/devices/acurite.c
```diff
@@ -713,36 +714,95 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+        // as the last 3 bits of the 3rd byte (bb[2]).
         if (browlen == ACURITE_TXR_BITLEN / 8) {
+            // Verify parity bits
+            // Bytes 2, 3, 4, and 5 should all have a parity bit in their MSB
+            int parity = parity_bytes(&bb[2], 4);
+            if (parity) {
+                if (decoder->verbose)
+                    bitrow_printf(bb, 7 * 8, "%s: bad parity: ", __func__);
+                continue; // return DECODE_FAIL_MIC;
+            }
```
rct, line 728: This whole function badly needs to be refactored because all of these if/continues are a little bit of a minefield.

Unless I'm not reading this right, because of the above, anything below this that has a message of the same length, but doesn't use the parity on bytes 2,3,4 will no long receive these rows when the values happen to fail the parity check.

rct (changes requested):
Because of pre-existing structure problems with this code (mixed if and continues), the parity check could cause other devices to not receive their rows if they use the same message length.

src/devices/acurite.c
DAVe3283: In this specific case, there are only 2 message types with this length (Tower and Leak Detector), and we have verified they both work correctly with the parity verification. If other messages/sensors were to be added in the future, they would get the parity.

If we refactor to switch on the message type before the parity, we end up repeating the parity check over the same bytes in two places. Since there are no other (currently know) sensors/messages of this length, I think it makes sense to calculate the parity first and only once (DRY).

It does make sense as a separate PR to evaluate if the parity check can be applied to ALL AcuRite sensors, like the checksum already was.

rct: > In this specific case, there are only 2 message types with this length (Tower and Leak Detector), and we have verified they both work correctly

Just to make sure we are on the same page, do you agree that the use of the `continue` makes it so any code below that check that wants to decode a 7 byte message might not get messages if they don't implement the same parity on bytes 2,3,4?   Can you see how that also might not be clear to the person implementing the next Acurite device?

The use of those if's and continue's were most likely my fault originally, but a bunch of devices have been added since I originally did the tower, 5n1, and 6045m. 

Also yes I agree, there are only two sensors that have declared that they use the 7 byte (56 bit) message length. But the 899 rain sensor which doesn't `#define` it's message length, the Atlas has two messages lengths, with/without lightning detector, but only one is `#define`d.  

>  I think it makes sense to calculate the parity first and only once (DRY).

Yes, that's what kept me from implementing the parity check for all, and it was a dumb move on my part. At this point I don't know whether every device covered by this decoder implements parity on all bytes except for 0,1, and N.

So I'd rather just see the parity check repeated even if it isn't DRY. 

And most importantly, if the tower and leak sensors get refactored into their own functions the way the Atlas and 6045 are we can avoid the spaghetti and unintended consequences.

DAVe3283: Yes, we are on the same page. The AcuRite decode really should be refactored and cleaned up a lot, but that would make this PR a lot more involved and touch devices we don't strictly need to.

I am not sure adding one more `continue` is a meaningful barrier to someone writing another future AcuRite device, since the whole routine is pretty convoluted already. (IMO it is clearer with these changes, if only just.)

I am not opposed to re-writing the AcuRite decoder, but we don't have test .cu8 files from all of the AcuRite sensors; just the 5n1, tower, and leak detector. If you have samples of the other sensors, I would be happy to rework the whole file once we know how similar/unique the protocol actually is. But I think that is best served as a separate PR, since the goal of this one is to add the leak detector + fix the tower, and based on my testing, it does both, without affecting the other sensor decodes.

src/devices/acurite.c
```diff
@@ -713,36 +714,95 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+                /* clang-format on */
+            }
+            else if (message_type == ACURITE_MSGTYPE_TOWER_SENSOR) {
+
+                // Humidity is stored in byte 3
+                // The value is directly encoded as %rH
+                // The possible values here are 0-128, but real values are only 1-99 %rH
+                // pIII IIII
+                humidity = (bb[3] & 0x7f);
+                if (humidity < 1 || humidity > 99) {
```
rct, line 767: If this check allows the values 0 and 100, I'm fine with having this sanity check here until we can do something better.

src/devices/acurite.c
```diff
@@ -713,36 +714,95 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+                }
+
+                // temperature encoding used by "tower" sensors 592txr
+                // 14 bits available after removing both parity bits.
+                // 11 bits needed for specified range -40 C to 70 C (-40 F - 158 F)
+                // Possible ranges are -100 C to 1538.4 C, but most of that range
+                // is not possible on Earth.
+                int temp_raw = ((bb[4] & 0x7F) << 7) | (bb[5] & 0x7F);
+                tempc        = temp_raw * 0.1 - 100;
+                if (tempc < -40 || tempc > 70) {
```
rct, line 782: as long as this allows messages with -40 through, I think this should be ok.

rct (changes requested):

src/devices/acurite.c
rct: I'm not advocating refactoring the whole acurite decoder in this PR.

I've mentioned what I think needs to change in order to get this merged and it's not hard. If you want to wait, I'll add the message type check for the tower sensor and factor it out, and then the changes for this can be rebased and resubmitted.

As you've said there are other devices you haven't tested that are covered by the same decoder. Ordering of those if's and continue's is significant.  It does seem to be the case where changes to the tower sensor are necessary to avoid the false positives from the new leak sensor. But let's not make a bad situation worse.

src/devices/acurite.c
DAVe3283: -40 should still come through as is here.

src/devices/acurite.c
DAVe3283: The changes you propose seem reasonable. I am pretty confident the changes in this PR only affect the leak detector and the tower. And, as you pointed out, any future AcuRite decoders for 7-byte messages. But we can refactor to be more explicit at the expense of some repetition.

src/devices/acurite.c
```diff
@@ -713,36 +714,95 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         // M = Message type
         message_type = bb[2] & 0x3f;
 
-        // tower sensor messages are 7 bytes.
-        // TODO: - see if there is a type in the message that
-        // can be used instead of length to determine type
+        // Multiple AcuRite sensors use the same basic message format,
+        // with a shared header and integrity scheme. message type is packed
+        // as the last 3 bits of the 3rd byte (bb[2]).
         if (browlen == ACURITE_TXR_BITLEN / 8) {
```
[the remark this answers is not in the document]
qwindelzorf: Ok. I'll take a swing at that this evening.

src/devices/acurite.c
qwindelzorf: I've just pushed commits to break out the tower and leak detector decode into their own functions. It makes things more readable, but does have a larger overall footprint.

src/devices/acurite.c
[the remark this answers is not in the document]
qwindelzorf: Just pushed a commit to split those sub-decoders (is that even a thing??) out into their own functions.

src/devices/acurite.c
qwindelzorf: Changed to allow 0 and 100.

src/devices/acurite.c
qwindelzorf: No change here. -40 should still pass through, but -41 won't.

qwindelzorf:
I think I've addressed all the open feedback with the most recent commits.

src/devices/acurite.c
zuckschwerdt: But note that the value is from an unsigned source, the compiler will know that and look upon `< 0` in disgust (issue a compiler warning)

qwindelzorf: I dealt with that by decoding it into a signed int. Since it's a 7 bit value, if anything <0 ever occurs, then something has gone __terribly__ wrong, and failing the decode is the least of our problems.

zuckschwerdt: At least Clang still knows where it came from and the possible range (a good thing). Checking for `< 0` will produce a warning.

qwindelzorf: clang 12 seems quite happy to compile the refactored version of this code. No warnings.

zuckschwerdt: Ah ok, I remember seeing this on some other code but it's not detected here. Though my CppCheck plugin is not satisfied still.

qwindelzorf: Interesting.... cpcheck has no complaints for me. It is mad about other pre-existing functions for which variable scope could be reduced, but that's it. Perhaps you have some additional checks turned on somewhere that I don't?

zuckschwerdt: I'm not aware of setting any custom options, hmm.
cppchk

qwindelzorf: Very odd. I'm running CppCheck via an extension in VSCode, and not getting any squigglies:
image
Clearly we have something configured differently, I just don't know what/where.

src/devices/acurite.c
```diff
@@ -713,39 +836,25 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
 
-            decoder_output_data(decoder, data);
-            valid++;
+            // The decoder attempts for this size message will return a positive
+            // value if they successfully decoded a message.
+            if (decoded > 0) {
+                valid++;
+            } else {
+                continue;
+            }
```
rct, line 857: The changes look very nice. Honwever, one nit -- Please delete this continue. 

This is what I was trying to get rid of.  Unless I haven't had enough coffee yet, It shouldn't be necessary and creates some subtle complexity for maintenance (if you want to handle any 7 byte messages you need to do it here).

rct (changes requested):

src/devices/acurite.c
qwindelzorf: Done

DAVe3283: I agree, this continue isn't necessary. If not removed, it could be placed after `valid++;` inside the if (if we decoded a message, move to the next one without evaluating the other decoders (515, 5n1)). But just nuking it is fine, probably premature optimization to try and avoid checking other decoders.

qwindelzorf:
Bump?

zuckschwerdt:
Both acurite_tower_sensor_decode(), acurite_leak_detector_decode() should be static, with no forward declaration, right?

qwindelzorf:
> Both acurite_tower_sensor_decode(), acurite_leak_detector_decode() should be static, with no forward declaration, right?

You're right. I originally was thinking to make these full decoders, but that would be veering into the "refactor it all" territory. So, for now, they should just be file-static helper functions.

Fixed.

rct:
> I originally was thinking to make these full decoders,

I'm not suggesting any changes, I think this is ready to merge.  But out of curiousity:  @zuckschwerdt - In the current code base, how heavyweight are decoders?   My recollection is they used to be pretty heavy in terms of at least memory foot print, so it was worth combining devices that could be handled under the same decoder.

zuckschwerdt:
> how heavyweight are decoders? My recollection is they used to be pretty heavy in terms of at least memory foot print, 

The heavyweight is FSK decoding, but after we got the pulses nothing is any real workload. Bitbuffers are small now (row spill 999387e) and slicing is pretty efficient. We don't even need to worry about precedence in decoders anymore (like one needs to reject before the next one is candidate, common reason to combine decoders), we now got `priority` #1895 for that.

DAVe3283:
I have a couple .cu8 captures that decode wrong prior to this merge, one that fails parity only and sanity, and one that only fails sanity, in both cases due to interference (visible on spectrogram). Should I submit a PR against rtl_433_tests to contribute them to the test suite?

rct:
Are there any good samples available for the leak detector? (@DAVe3283 ?)

I'd like to test some changes to acurite.c and I don't see any in rtl_433_tests.

DAVe3283:
> Are there any good samples available for the leak detector? (@DAVe3283 ?)
> 
> I'd like to test some changes to acurite.c and I don't see any in rtl_433_tests.

I have several GB of RF captures for various AcuRite - sensors. I could look into how to add proper tests, or would happily share some samples with you and you could add the tests.

What kind of samples do you want? I have a few samples where multiple sensors overlap and cause bad decodes, and a ton of clean samples.

rct:
Right now I could use a known good sample or a `-y  xxxx` string just to verify that I didn't break anything.

The rtl_433_tests repo seems to be in the process of some rearranging and there is old/new stuff in there, probably to support the explorer and other stuff at https://triq.org/explorer/   Probably need some direction from @zuckschwerdt on what he's currently looking for there. 

I've been using `-y xxx` strings to come up with and test some of the edge cases.

zuckschwerdt:
The notes at https://github.com/merbanan/rtl_433_tests#organization are the wishlist. The repo should mostly be okay, but there are lots of PRs pending which are not in the right shape. Since the size of the repo is always a concern we would need to merge each PR and immediately force-push a cleaned up amend-commit. (rationale: adding then removing or changing files, e.g. shrinking pictures, increases the repo size.)

DAVe3283:
@rct here are some normal samples. I can provide the raw capture files for any of these if you want.

- Acurite-5n1_659_A
  - g004_433.92M_250k.cu8 `3D6C87FF5A3AAC75FF/ 2D6C87FF5A3AAC65FF/ 1D6C87FF5A3AAC55FF`
  - g008_433.92M_250k.cu8 `3D6C8EFF747290B2FF/ 2D6C8EFF747290A2FF/ 1D6C8EFF74729092FF`
  - g020_433.92M_250k.cu8 `3D6C87FF3A3AAC55FF/ 2D6C87FF3A3AAC45FF/ 1D6C87FF3A3AAC35FF`
- Acurite-Leak_8803
  - Channel A
    - dry_cha_21v_001.cu8 `1D9C7E78177E49FF/ 1D9C7ED8177EA9FF/ 1D9C7EB8177E89FF` (low battery)
    - dry_cha_30v_003.cu8 `1D9CBE78177E89FF/ 1D9CBED8177EE9FF/ 1D9CBEB8177EC9FF`
    - wet_cha_bat_001.cu8 `1D9CBEE8177EF9FF/ 1D9CBE48177E59FF/ 1D9CBE28177E39FF` (leak detected)
  - Channel B
    - dry_chb_30v_002.cu8 `5D9CBE78177EC9FF/ 5D9CBED8177E29FF/ 5D9CBEB8177E09FF`
  - Channel C
    - dry_chc_30v_001.cu8 `DD9CBE78177E49FF/ DD9CBED8177EA9FF/ DD9CBEB8177E89FF`
- Acurite-Tower_2388_A
  - g003_433.92M_250k.cu8 `36ABBBD177634CFF/ 36ABBBD177634CFF/ 36ABBBD177634CFF`
- Acurite-Tower_3321_A
  - g001_433.92M_250k.cu8 `3306BB357818BEFF/ 3306BB357818BEFF/ 3306BB357818BEFF`
- Acurite-Tower_6478_A
  - g002_433.92M_250k.cu8 `26B1BB35782165FF/ 26B1BB35782165FF/ 26B1BB35782165FF`
- Acurite-Tower_16225_A
  - g005_433.92M_250k.cu8 `009EBB3C782436FF/ 009EBB3C782436FF/ 009EBB3C782436FF`

And here are some that should at least partially fail, due to overlap/corruption. Some of these only fail the sanity checks currently, and are the ones we had the most trouble handling without outputting garbage decodes IIRC.

- [Decode failures (mixed).zip](https://github.com/merbanan/rtl_433/files/9421355/Decode.failures.mixed.zip)

NorthernMan54:
@qwindelzorf @rct I recently updated to the latest release ( 22.11 ) and found that a couple of my acurite tower sensors stopped being reported. After double clicking on the issue, and digging into it found that the signal validation checks are rejecting my valid / real Acurite Sensors.  When running in verbose mode they are being reported as having an invalid humidity. 

```
acurite_txr_callback: row 1 bits 57, bytes 7, extra bits 1, msg type 0x04
acurite_tower_decode: 0x1765 Ch A : Impossible humidity: 127 %rH
acurite_txr_callback: stats: row 1, msg type 0x04, bytes 7, decoded 0, error -4
acurite_txr_callback: row 2 bits 57, bytes 7, extra bits 1, msg type 0x04
acurite_tower_decode: 0x1765 Ch A : Impossible humidity: 127 %rH
acurite_txr_callback: stats: row 2, msg type 0x04, bytes 7, decoded 0, error -4
acurite_txr_callback: row 3 bits 57, bytes 7, extra bits 1, msg type 0x04
acurite_tower_decode: 0x1765 Ch A : Impossible humidity: 127 %rH
acurite_txr_callback: stats: row 3, msg type 0x04, bytes 7, decoded 0, error -4
```

I know that this is a valid sensor as I had pulled one part to look at why it was reporting 127, as I was looking to trigger my bathroom fan based on a humidity delta, as was puzzled as to why it was reporting humidity of 127 with an older version of rtl_433.  And they had not included a humidity sensor in the device.

From an older version of rtl_433.

```
{"time":"2023-03-26 17:41:01","model":"Acurite-Tower","id":5989,"channel":"A","battery_ok":1,"temperature_C":25.4,"humidity":127,"mic":"CHECKSUM"}
```

The issue is here 

https://github.com/merbanan/rtl_433/blob/0ce05017205a30a10033f89ae4171984d47c3629/src/devices/acurite.c#L1018

And I have changed my environment with this if change, and can report that it works successfully.

`((humidity < 0 || humidity > 100) && humidity != 127)`

DAVe3283:
@NorthernMan54 what part number are your sensors? Interesting that they are reporting as temperature+humidity sensors if they physically don't have humidity sensors hardware.

Can you get some RF dumps of the messages?

rct:
@NorthernMan54 - Can you please move this to a new issue, rather than this merged PR?

Good work tracking this down.  As @DAVe3283 asked, can you provide the device part number(s)?

I think I saw that Acurite had a temperature only device but I guessed it would have a different message format. 

While your fix would work, if a humidity value of 127 (0x7f after removing parity)  indicates the type of device that has no humidity sensor, it would be better to not output humidity at all.

PR #1955, opened 2022-01-23
ADD support for Honeywell ceiling fans

dtiller:
This decoder handles the Honeywell model Salermo ceiling fan. The fan is actually made by Intertek but sold under Honeywell's name.

### DIFF
--- a/src/devices/honeywell_fan.c
+++ b/src/devices/honeywell_fan.c
@@ -40,14 +40,14 @@ there's no such command/value distinction. It looks very suspicious that the fan
 speed commands all share command 000 and the speed value (bit-reversed) appears in the
 value area.
 
-Button  Fixed Other Bits       Function
-ONE     16CD  1 0 0 0 0 1 !d d  Low speed fan
-TWO     16CD  0 1 0 0 0 1 !d d  Medium speed fan
-THREE   16CD  1 1 0 0 0 1 !d d  High speed fan
-OFF-M   16CD  0 0 0 1 0 1 !d d  Fan off (momentary press)
-OFF-C   16CD  0 0 1 0 1 1 !d d  Light off delay (continuous press)
-STAR-M  16CD  1 1 0 1 0 1 !d d  Light on/off (momentary press)
-STAR-C  16CD  0 1 1 1 0 1 !d d  Light dim/brighten (continuous press)
+    Button  Fixed Other Bits       Function
+    ONE     16CD  1 0 0 0 0 1 !d d  Low speed fan
+    TWO     16CD  0 1 0 0 0 1 !d d  Medium speed fan
+    THREE   16CD  1 1 0 0 0 1 !d d  High speed fan
+    OFF-M   16CD  0 0 0 1 0 1 !d d  Fan off (momentary press)
+    OFF-C   16CD  0 0 1 0 1 1 !d d  Light off delay (continuous press)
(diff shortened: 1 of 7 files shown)


zuckschwerdt:
Looks perfect!

zuckschwerdt:
The "mic" [should indicate](https://triq.org/rtl_433/DATA_FORMAT.html#message-data) the "strengh" of a check (if it exists).
A preamble of 16 bits vs. 8 data bits is pretty good, but I would rate this as `PARITY` (we do know it passed some tests, but does not guarantee valid data). Or even remove it as there is no actual check on the data portion, not sure.

src/devices/honeywell_fan.c
```diff
@@ -0,0 +1,171 @@
...
+            }
+            break;
+        }
+
+        /* clang-format off */
+        data_t *data = data_make(
+                "model",            "",     DATA_STRING,    "Honeywell-Remote",
+                "command",          "",     DATA_STRING,    command_string,
+                "dimmable",         "",     DATA_INT   ,    dimmable,
+                "mic",              "",     DATA_STRING,    "PARITY",
```
merbanan, line 143: MIC stands for message integrity check. Right now we have checksum, crc and parity. From the description I see neither of those. The point of the tag is how much can we trust the message. I always wanted to add repeats as a mic, and here it would fit well with some changes (bitbuffer_find_repeated_row). Then we have the question how many repeats should be required for a mic tag and here I think 3-5 should a good range depending on the length of the message.

zuckschwerdt:
This awefully akward: I wasn't careful enough to really understand the protocol. This should be reverted and implemented as flex decoder spec. The fixed code is a good indication of a per-unit "house code". A flex decoder is simpler and gives users an easy way to tweak a "match" (the fixed code) and perhaps timings if needed. The command and dimmable should be extracted with "get" and a mapping. Sorry!

dtiller:
What’s the procedure for backing those out of your repo? Do you want me to submit another pull request with those removed then ones for each conf-based ones, or are you going to handle the undo?

zuckschwerdt:
We could just reset the commit. But an automatic revert commit would leave your code visible, so your work isn't lost and can be recovered if needed.
Since the documentation is already excellent you can can copy the .c file to a .conf and then PR with only the comments, the timings and bits>=, bits <=, repeats, match, get.

dtiller:
Ok, go ahead and reverse what you think can be re-done as a ‘conf’. When’s that’s done I’ll re-pull master and start new branches.

zuckschwerdt:
Done. Thanks for being so patient, apologies again for not catching this before all your work went into this.

dtiller:
I've never written a conf style decoder, but I bet most of the work was in figuring out the protocol and writing the documentation. None of that was wasted.

Do you have any advice on whether or not there are 12 addr bits/8 tri-state bits or 20 address bits? It's confusing - it seems like there are both of those formats in the wild.

zuckschwerdt:
It's a mess with these encoders and the many clones. A giveaway for the x1527 (EV1527) is the sync pulse at the end. The
PT2260/PT2262 SC2260/SC2262 have sync at the front (the sync usually merges with the next transmission, so will likely appear exactly opposite...). Both styles have different ternary encoding (and documents that claim things we don't observe...). And there are more modern senders using all bits (no ternary) but still 24/25 bits.

I usually just look at the patterns at hand and decide what's likely. But the flex conf can't do ternary decoding anyway, so your current logic and values should already work.

PR #1956, opened 2022-01-23
minor: Fix variable size

robho:


zuckschwerdt:
Thanks, but the type doesn't really matter here and should be just `int`.

robho:
I agree that it doesn't matter. Maybe I should have used some other verb than "Fix" because the intention was just to make the code consistent since the other variables used uint8_t and there was no reason for uint16_t for this specific variable.

Anyway.. It's not important to me, you may leave it closed.

zuckschwerdt:
Reopen with change to int and shuffle them down to declare on first use, if you like. Cleanup is always good.

PR #1959, opened 2022-01-23
Change wt0124 to read and use additional checksum

robho:
I have used samples from rtl_433_tests (directories PoolWT0122 and wt0124)
to get the calculations correct.


zuckschwerdt:
This is great, thanks!

PR #1961, opened 2022-01-24
ADD support for Heatilator gas log remotes

dtiller:
This adds support for Heatilator (and possibly Heat&Go) gas logs with the thermo-control remote. It is very similar in packet structure to the generic_remote, and both decoders attempt to decode these packets. I set it to 'disabled' since there may be a manufacturer's bit pattern in the serial_number but I don't know what it is.


zuckschwerdt:
I just realize that this (and the Honeywell Fan before) are likely simple EV1527 encoders. (PWM with 24/25 bits, lots of repeats).
The flex decoder might work better. We prefer to support simple devices with spec files.

dtiller:
Deleted - added decoder as a conf decoder.

PR #1962, opened 2022-01-25
Converted C-style decoder to conf style.

dtiller:
This PR adds back honeywell fan support as a conf-style decoder.


zuckschwerdt:
Thanks!

PR #1963, opened 2022-01-25
ADD heatilator gas log support

dtiller:
This PR adds back heatilator gas log support as a conf-style decoder.


dtiller:
Thanks for the merges.

Question: Is there a concept of test data for conf-style decoders? If so, I'll convert the data I already PR'ed for the test repo.

zuckschwerdt:
cu8 are always good to have. I'll need to check if (and possibly add that) the `protocol` meta-file in the tests repo can support .conf decoder files.

zuckschwerdt:
This is now supported with https://github.com/merbanan/rtl_433_tests/commit/f1b9c0f by placing e.g. `honeywell-fan.conf` in a file `protocol`.

PR #1966, opened 2022-01-25
ci: Fix Maverick ET73 temperature_1 and recognize negative temps

hdtodd:
The Maverick ET73 decoder for temperature_1 doesn't align the right-most nibble of the 12-bit
raw temp integer correctly, so that value is always incorrect.  temperature_2 is fine in that regard.

However, neither decoding accounts for possible negative temperatures.  The ET73 has a range of -10˚C to +210˚C.  Seems unlikely that a smoker thermometer would need to report below freezing, but the second edit to the .c file accounts correctly for negative temps.

Test files, and a replay of a prior test file, are coming in a subsequent pull request for the test directory.

Sorry if I got any of this process wrong ... I'm new to some of these tools and your (well-organized) procedures.

### DIFF
--- a/src/devices/maverick_et73.c
+++ b/src/devices/maverick_et73.c
@@ -45,7 +45,8 @@ Layout appears to be:
 
 static int maverick_et73_sensor_callback(r_device *decoder, bitbuffer_t *bitbuffer)
 {
-    int temp1_raw, temp2_raw, row;
+  /*[hdt] add -1 mask for integers to make 12-bit negative into a negative full int */
+  int temp1_raw, temp2_raw, row, m1mask = -1 ^ 0xfff;
     float temp1_c, temp2_c;
     uint8_t *bytes;
     unsigned int device;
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks!
If I've understood the flipping then it's a sign-extend (filling the left-most/sign bit all the way to the left), right?
We then usually shift the raw value into 16 bit instead of 12, and use a signed type. See e.g. https://github.com/merbanan/rtl_433/blob/master/src/devices/alecto.c#L189-L190 (ignore the reverse8, that's something else)

hdtodd:
I didn't want to assume a default size for the integer on whatever system the code gets compiled on.   So I used the masking approach.  ORing the int -1 (at whatever the default int size is for individual compilers/systems) with the right-hand 12 bits masked out,  with the 12-bit 2's complement integer coming from the ET73, gives the 2's complement representation of the ET73 reading in the size of the int native to the system it's being compiled on.  

I first thought about using the shift-extension approach but thought it would require either determining and using the size of int's on the native system or picking a size (as your link above does).  It seemed less confusing to compile to the native size int.

But I may have missed something.  Is there a use case in which that fails, or does it complicate something I missed?

zuckschwerdt:
Technically the minimum size for an int is 16-bit and that still works with sign extension: shift to 16-bit left aligned (using implicit integer promotion), cast to `int16_t` and store anywhere, it's now properly signed, no matter the size of int.

In the end it's just useful (for maintainance) to have the same uniform logic across decoders (not much emphasis on what's most elegant). As a bonus we don't need the conditional and sign-extend likely compiles to a native instruction ;)

hdtodd:
Fixed the trailing spaces in maverick-et73 and pushed to my fork from the clone on my Pi.

hdtodd:
You're right: shifting requires fewer instructions and fewer memory accesses than masking.  

To make sure my revised code worked correctly with shifting, I wrote a small program to compute the two temps both ways.  Compiled on a Raspberry Pi 4, armv8 architecture with gcc version 10.2.1 20210110 (Debian 10.2.1-6), to convert both temperatures to float, the shifting algorithm took 35 instructions + 10 memory accesses and the masking algorithm took 36 to 39 instructions (depending on signs) + 12-14 memory accesses.

Not a critical difference, and results would depend on the architecture and compiler, but it was a surprise to me.  I should have realized that the test and branch would cost something in extra instructions, besides breaking the pipeline.  Useful exercise.

Anyway, revised code with shifting has been uploaded, with an additional test result file in the testing area.

zuckschwerdt:
Thanks for looking into this so in-depth, interesting results (ops and mem)!

Can you check the indentation on the declaration; usually we use EOL comments (`//`) in the code; wouldn't the shift-right look cleaner as `(temp2_raw >> 4) * 0.1f;`, there is too much going on in the raw line now, I think.

hdtodd:
I'd have preferred the int temp calculation to be all on one line, but yes, there's a lot going on in that raw line.  Edited comment format, removed trailing white space, and moved the raw >> to the line that converts to float.

PR #1967, opened 2022-01-27
Add missing angle bracket to man page

djcrock:
Close the angle brackets around the -p option's required ppm_error parameter.


zuckschwerdt:
Thanks! That's generated output, change at https://github.com/merbanan/rtl_433/blob/master/src/rtl_433.c#L150 please.

djcrock:
Thanks for the heads up! Added :+1:

PR #1970, opened 2022-01-29
Add support for GEO minim+ energy monitor

UnscrewLater:
Recently purchased this electricity monitor hoping to keep a lid on spiraling energy costs. Datasheet here:
https://assets.geotogether.com/sites/4/20170719152420/Minim-Data-sheet.pdf

Creating a Flex decoder was straightforward so I wrote a device to pretty print the data. Hope this is of use to others .

Thanks to all the contributors to rtl_433 - a great program!

Attached signals:
[geo_minim_signals.zip](https://github.com/merbanan/rtl_433/files/7963545/geo_minim_signals.zip)

PS would be interested to see (maybe fix my code) how the header bytes change for different sensors & display units.


zuckschwerdt:
Thanks! Some initial notes:
- "model" needs to be a fixed string of e.g. "Minim-Sensor". remove the define and put the long description directly in ".name".
- Continued lines need to be indented 8 spaces.
- Remove the `UINT8_BITS` define and assume a fixed `8` bits per byte ;)
- put parens around the sizeof arg
- Don't space inside parens, e.g. `(x)` not `( x )`

UnscrewLater:
Hi,

Many thanks for reviewing this and for your comments.

I have addressed the points you mentioned and pushed a cosmetic fix.
Hope that is OK?

zuckschwerdt:
- data_make, output_fields, and r_device lines are continuued, so need 8 spaces.
- we don't provide derived values form sensors, so s_watt_hours s_last statics and tm shouldn't be there
- the `static const uint8_t` should be `uint8_t const`
- kType is good practice, we usually would just use `type = 1`
- the switch should likely be an if/else
- no need for strftime, we'd rather just glue the raw values together, see e.g. fineoffset_wh1080
- mic needs to be last in the list
- you likely shouldn't pass a data_t out around, but rather output directly in the functions.
- you shouldn't pass a id string in, keep it plain and dumb and compute the id where it's used (along with the other fields)

UnscrewLater:
Thanks for the advice.

I've done all that you suggested except for removing strftime. My
reasoning is that the message contains real hours and minutes but the
date is expressed as days since 1/1/2007. Printing the raw value
doesn't provide much help so I used mktime to convert the date and
strftime to print it.

I have kept the enum & switch statement for the message type as I was
hoping to obtain an LED sensor for use with electricity meters that
output a light pulse. I'm assuming the data packet will be different so
the switch will be necessary to handle a different decode function.

zuckschwerdt:
My intention was to snprintf() the date/time values, not pass them through mktime(), see e.g. https://github.com/merbanan/rtl_433/blob/master/src/devices/fineoffset_wh1080.c#L255-L283
In any case you can't use setlocale() as that changes global environment. The libc APIs for these things are very inconvenient and troublesome.

I do get that the GEO_ZERO_WH state is useful, but as a general rule we don't keep state and offsets in decoders. It's a task for consumers to track and derive these things. A common case is accumulative rainfall, consumers need to offset and derive themselves. E.g. grafana can do this nicely on display :)

UnscrewLater:
Agreed, setlocale() is a no-go so I have removed that together with
strftime() and replaced it with snprintf. I have retained the call to
mktime() as this converts the days since 1/1/2007 into a proper
calendar date. Sure, it would be possible to write my own conversion
but at the risk of undetected bugs.

I have also pushed a bug fix to the handling of oversized packets to
prevent stack smashing.

On the subject of bugs, is there a simple way of enabling gcc's
sanitizers from cmake?

Thanks.

zuckschwerdt:
Yes, using `-DCMAKE_BUILD_TYPE=Debug` will enable ASAN.

zuckschwerdt:
Needs to be update to current code style and a few issues need to be addressed (e.g. model name, state).

PR #1971, opened 2022-01-29
Improve digitech_xc0324.c for Humidity readings. (FT-005TH Sensor)

adellafave:
Model Number of unit is FT-005TH 

Updated for Humidity Field in Byte 4. Previously unknown Byte that author implied wasn't in use on their sensor.


zuckschwerdt:
This is great, thanks!

should be:
- `int humidity;`
- `humidity   = reverse8(b[4]);`
- `"humidity",         "Humidity",     DATA_FORMAT, "%u %%", DATA_INT,    humidity,`

PR #1972, opened 2022-02-02
Added DATA_BYTE 

FuzzIO:
Added support for DATA_BYTE, which is a uint8_t data (one byte). This is useful to print out raw byte values and compatible with the DATA_ARRAY structure. The output is printed in hexadecimal values.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -22,6 +22,7 @@
 #define INCLUDE_DATA_H_
 
 #include <stddef.h>
+#include <stdint.h>
 
 typedef enum {
     DATA_DATA,   /**< pointer to data is stored */
@@ -29,6 +30,7 @@ typedef enum {
     DATA_DOUBLE, /**< pointer to a double is stored */
     DATA_STRING, /**< pointer to a string is stored */
     DATA_ARRAY,  /**< pointer to an array of values is stored */
+    DATA_BYTE,   /**< pointer to a uint8_t is stored */
     DATA_COUNT,  /**< invalid */
     DATA_FORMAT, /**< indicates the following value is formatted */
     DATA_COND,   /**< add data only if condition is true, skip otherwise */
@@ -43,6 +45,7 @@ typedef struct data_array {
 typedef union data_value {
     int         v_int;
     double      v_dbl;
+    uint8_t     v_byte;
     void        *v_ptr;
 } data_value_t;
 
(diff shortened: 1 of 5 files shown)


zuckschwerdt:
This is covered with DATA_INT + DATA_FORMAT, isn't it?

We don't like to add more types to data_t if there isn't a very solid reason like memory management or mapping requirements in the outputs.

I guess an array type might be more compact with shorter values then int, but it shouldn't be that long anyway.

There were ideas about a DATA_BITROW type, but that is more intrusive (fat pointer) and of limited value, rtl_433 really should decode bytes.

FuzzIO:
The use case I have encountered was that I had to work on decoding a device with fields that were unknown. The unknown field could be archived as an array of bytes in the file output for ulterior analysis or if we needed it. 
Using DATA_FORMAT and DATA_INT consumed 32 bits (the size of int). As the value of interest was a Byte (8 bit), 32-bit was too much. 
Also, when a field covers more than one byte, but is not a multiple of 4 bytes (again, size of an int), declaring a DATA_ARRAY with DATA_INT would not cover the value properly. 

Let me know if more explanations are needed or if you think my use case could be handled with existing data structures. I did try before implementing DATA_BYTE, but I may have missed something.

Cheers!

FuzzIO:
An example of  this is if a source and a destination address are encoded on a number of bytes unaligned with an int (e.g. 24 bits or 48 bits). Displaying the output in a an array of byte would then make sense. Again, let me know if something is already existent that covers this use case. More than happy to avoid "reinventing the wheel" !

FuzzIO:
Here is an example that could be representative for a device.
```
data_make(
...
"raw",              "Raw data",      DATA_ARRAY, data_array(9, DATA_BYTE, byte_buffer)
...
)
```
Produces something like : 
image

zuckschwerdt:
I see. Using a DATA_ARRAY is not recommended for general use. Some outputs don't handle that well.
In this case you likely want `bitrow_snprint()` to a buffer of sufficient size, then DATA_STRING as output.

zuckschwerdt:
This should be picked up again – the main idea is a `DATA_STRING` variant with unsafe bytes vs. a hex formatted actual `DATA_STRING`.

gdt:
Status?   still on the table, should be closed, something else?

zuckschwerdt:
The verdict is: no to the change as-is but generally yes to the idea of some raw `DATA_STRING`.

gdt:
So therefore close this PR, and if there is DATA_STRING later, we are happy to look at it?

zuckschwerdt:
Yes, close. The implementation is good, and sound in principle, but we don't want more types in data_t.

The `DATA_STRING` currently prints as-is and there could be a variant that prints as hex. This needs a good case (e.g. some output that makes use of a better representation) to be really useful as we can just print to hex and use plain `DATA_STRING` right now.

PR #1980, opened 2022-02-14
Add proper MQTT QoS

zuckschwerdt:
This adds proper MQTT QoS support by added a list of in-flight messages and retry timer.

Early draft and needs to wait on #1978 as timers don't work well with the SDR loop blocking events.

### DIFF
--- a/src/output_mqtt.c
+++ b/src/output_mqtt.c
@@ -23,6 +23,122 @@
 
 #include "mongoose.h"
 
+/* MQTT transmission list */
+
+typedef struct mqtt_msg {
+    char *topic; //!< NULL after QoS 2 release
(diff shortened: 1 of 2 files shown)


merbanan:
#1978 was merged, is this code ready for use now ?

zuckschwerdt:
I've rebased and updated the code -- the timer needs to be multiplexed now as the reconnect is also using it.
It needs thorough read and test. I think QoS 2 also needs work.

zuckschwerdt:
QoS 2 is done. Good for stress testing if someone is interested (I don't use MQTT QoS 1 or 2).
Otherwise let's merge soon.

PR #1983, opened 2022-02-16
Add support for EMOS 6016

zuckschwerdt:
Proxy PR for source from #1710


zuckschwerdt:
The code is nearly good to merge. There is an API change to `bitbuffer_find_repeated_row()` pending (compare just up to the limit, not whole rows). I hope to get that done in the next days.

PR #1988, opened 2022-02-20
Add device id include list for mqtt hass discovery

ScottRoach:
New argument (`-I` / `--ids`) for a list of numeric ID's that will be discovered. If this argument is supplied then any devices outside of these ID's will be ignored.

Example:
```
python rtl_433_mqtt_hass.py -u user -P password -H host.mqtt.com --ids 1234 5678 9012
```

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -77,12 +77,6 @@ MQTT Explorer also makes it easy to publish an empty config topic to delete an
 entity from Home Assistant.
 
 
-Known Issues:
-
-Currently there is no white or black lists, so any device that rtl_433 receives
-including transients, false positives, will create a bunch of entities in
-Home Assistant.
-
 As of 2020-10, Home Assistant MQTT auto discovery doesn't currently support
 supplying "friendly name", and "area" key, so some configuration must be
 done in Home Assistant.
@@ -648,6 +642,11 @@ def bridge_event_to_hass(mqttc, topicprefix, data):
         logging.warning("No suitable identifier found for model: ", model)
         return
 
+    if args.ids and data.get("id") not in args.ids:
+        # not in the safe list
+        logging.debug("Device (%s) is not in the desired list of device ids: [%s]" % (data["id"], ids))
+        return
+
     # detect known attributes
     for key in data.keys():
         if key in mappings:
@@ -727,6 +726,8 @@ if __name__ == "__main__":
                         dest="discovery_interval",
                         default=600,
                         help="Interval to republish config topics in seconds (default: %(default)d)")
+    parser.add_argument("-I", "--ids", type=int, nargs="+",
+                        help="ID's of devices that will be discovered (omit for all)")
     args = parser.parse_args()
 
     if args.debug and args.quiet:
@@ -748,5 +749,11 @@ if __name__ == "__main__":
 
     if not args.user or not args.password:
         logging.warning("User or password is not set. Check credentials if subscriptions do not return messages.")
+
+    if args.ids:
+        ids = ', '.join(str(id) for id in args.ids)
+        logging.info("Only discovering devices with ids: [%s]" % ids)
+    else:
+        logging.info("Discovering all devices")
 
     run()


zuckschwerdt:
This looks good. Any comments or objections from other current users of the script?

Squixx:
This would be prefect, and allow low-config filtering. :)

robloh:
This would be great for sure! I'd also (or instead of) love to have this in the main rtl_433 executable so the stuff I don't care about doesn't even make it to my mqtt server. Now I periodically manually remove things from there with MQTT Explorer when I'm debugging some stuff

rct:
So this very simplistically only matches on `ID` and not `Model` + `ID`, which is fairly easy to use.

This should certainly cut down on false positives creating unwanted devices in Hass, but there will be edge cases where matching `ID` only won't do what you want, particular for larger setups with a variety of models (device types).

It's low probability but you'll still get a device created if any of the enabled protocols (other models) have a false positive with one of the white listed IDs.

Doing whitelisting/blacklisting based on the composite identifier `/model[/subdevice][/channel][/ID]` would be arguably more correct but is a bit cumbersome to implement and use.

sotatech:
I can't seem to get this to work:

`pi@rpi2-rtl-433:~ $ python ./rtl_433_mqtt_hass.py -H 192.168.16.5 -u MQTT -P password --ids 77
INFO:root:Only discovering devices with ids: [77]
INFO:root:MQTT connected: Connection Accepted.
INFO:root:Subscribing to: rtl_433/+/events
Exception in thread Thread-1:
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 3452, in _thread_main
    self.loop_forever(retry_first_connection=True)
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 1779, in loop_forever
    rc = self.loop(timeout, max_packets)
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 1181, in loop
    rc = self.loop_read(max_packets)
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 1572, in loop_read
    rc = self._packet_read()
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 2310, in _packet_read
    rc = self._packet_handle()
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 2936, in _packet_handle
    return self._handle_publish()
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 3216, in _handle_publish
    self._handle_on_message(message)
  File "/usr/lib/python3/dist-packages/paho/mqtt/client.py", line 3444, in _handle_on_message
    self.on_message(self, self._userdata, message)
  File "/home/pi/./rtl_433_mqtt_hass.py", line 576, in mqtt_message
    bridge_event_to_hass(client, topicprefix, data)
  File "/home/pi/./rtl_433_mqtt_hass.py", line 660, in bridge_event_to_hass
    logging.debug("Device (%s) is not in the desired list of device ids: [%s]" % (data["id"], ids))
KeyError: 'id'

^CTraceback (most recent call last):
  File "/home/pi/./rtl_433_mqtt_hass.py", line 780, in <module>
    run()
  File "/home/pi/./rtl_433_mqtt_hass.py", line 713, in run
    rtl_433_bridge()
  File "/home/pi/./rtl_433_mqtt_hass.py", line 703, in rtl_433_bridge
    time.sleep(1)
KeyboardInterrupt

`

zuckschwerdt:
Thanks. I guess [line 645](https://github.com/merbanan/rtl_433/pull/1988/files#diff-571535de411bd6413429fcb2a96b90f57d27c1bbe19194ad1bfb3ddbdfbce1beR645) should have been:
`if args.ids and id in data and data.get("id") not in args.ids:`

sotatech:
Thanks, that change worked. I had worked around the problem by using the -R switch on the rtl_433 command line.

zuckschwerdt:
Thanks for the feedback, fixed with 315b98b now.

PR #1991, opened 2022-02-21
Initial pass at adding simplistic statsd output support

ross:
This adds statsd output support along side the existing udp syslog output (since they share required functionality.) 

It's a rough first pass, but appears to work as expected/desired for the Acurite temp & humidity towers I have to test with. 

Opening this as a draft since it's by no means ready to review & ship, but I'd welcome feedback/throughts and would like to get a general idea whether this is of interest before I put more time & effort into it.

### DIFF
--- a/include/output_udp.h
+++ b/include/output_udp.h
@@ -15,5 +15,6 @@
 #include "data.h"
 
 struct data_output *data_output_syslog_create(const char *host, const char *port);
+struct data_output *data_output_statsd_create(const char *host, const char *port);
 
 #endif /* INCLUDE_OUTPUT_UDP_H_ */
--- a/include/r_api.h
+++ b/include/r_api.h
@@ -77,6 +77,8 @@ void add_mqtt_output(struct r_cfg *cfg, char *param);
 
 void add_influx_output(struct r_cfg *cfg, char *param);
 
+void add_statsd_output(struct r_cfg *cfg, char *param);
+
 void add_syslog_output(struct r_cfg *cfg, char *param);
 
 void add_http_output(struct r_cfg *cfg, char *param);
--- a/src/output_udp.c
+++ b/src/output_udp.c
@@ -231,3 +231,127 @@ struct data_output *data_output_syslog_create(const char *host, const char *port
 
     return &syslog->output;
 }
+
+/* statsd UDP printer */
+
+typedef struct {
+    struct data_output output;
+    datagram_client_t client;
+    char hostname[_POSIX_HOST_NAME_MAX + 1];
+} data_output_statsd_t;
+
+static void emit_statsd(data_output_statsd_t *statsd, char const *metric_key_base, char const *metric_key, double metric_value) {
+    // TODO: buffering/batching?
+
+    char message[1024];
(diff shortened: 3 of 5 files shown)


zuckschwerdt:
I would like to see more outputs, e.g. Graphite and Collectd come to mind. But people will expect high quality (as other tools would bring) not just bare minimum (see the QoS mishap with MQTT). Also there will be requests for e.g. filtering and mapping, as well as discussions which number formatting to use and so on.
It might be a safer approach to use user-adaptable "bridges" (like the python examples we have).

What consumer are you targeting with statsd? I get the feeling that protocols have consolidated to Influx, Prometheus, and maybe Graphite (all with tags support) and Collectd, Statsd, and RRD are phasing out.

ross:
> I would like to see more outputs, e.g. Graphite and Collectd come to mind. But people will expect high quality (as other tools would bring) not just bare minimum (see the QoS mishap with MQTT).

Graphite would be trivial to do and is fairly similar to statsd, main difference being the timestamp. Collectd isn't really the same sort of thing, iirc it just wants to read k/v per line from running the command. That'd probably use the same metric name and value logic, but otherwise share none of the network sending bits.

> Also there will be requests for e.g. filtering and mapping, as well as discussions which number formatting to use and so on.

Don't see support for that with any of the other outputs so not sure it'd make sense to have it just for this one. If done those seem more like things that should be implemented at an earlier point before the outputs get sent the data.

> It might be a safer approach to use user-adaptable "bridges" (like the python examples we have).

There's an existing one for statsd using the python library, so if that's the preferred route there's not much sense in pursuing this. Just seemed cleaner and more in line with how MQTT and influx worked to have direct support w/o having to pipe things into python. In envronments that are running on low power microcontrollers I could imagine python not being an option, but not super familiar with where all/how rtt_433 is used.

> What consumer are you targeting with statsd?

I have statsd hooked up to a self hosted graphite and dogstatsd (Datadog.)

> I get the feeling that protocols have consolidated to Influx, Prometheus, and maybe Graphite (all with tags support) and Collectd, Statsd, and RRD are phasing out.

I planned to implement a mode that fully supported tagging rather than stuffing things into the metric names. That could easily support dogstatsd and the other flavors of statsd+tags. 

I don't see any existing graphite support/functionality in rtl_433, but as mentioned above it would be easy enough to do and I would have implemented that if I didn't also want things to go to dogstatsd.

zuckschwerdt:
re mapping: e.g. you are using "rtl_433.%s.%d" and users will request "<hostname>.%s.ID%d".
Not an entirely unreasonable request ;) might need something like our MQTT mapping string.

re filtering: e.g. users might not want any random data to be sent, but just temperature and humidity.
Some backends will auto-create new metrics, random metrics might make a mess there, hard to then not implement a fix ;)

I'm using Graphite myself, so a UDP "line format" supporting some flavors like StatsD, DogStatsD (tags), Graphite would be nice. Not a fan of pipes either, but Syslog UDP to bridges (running on the metrics server, not the sensor edge) works well.

There should be some other users of Graphite or StatsD, would be interesting to hear what key format (and maybe filtering) they use.

merbanan:
No work for several years, and still marked as a draft, closing.

PR #2004, opened 2022-03-02
Add new device decoder for ANT and ANT+ devices

robcazzaro:
This is a proposed ANT and ANT+ decoder. I verified it works with multiple devices.

Please note that, unlike most protocols here, ANT and ANT+ devices operate in the ISM band and require 4Msps sample rates. I also captured and saved samples in the rtl_433_tests projects and have created a pull request for that https://github.com/merbanan/rtl_433_tests/pull/421

### DIFF
--- a/src/devices/ant_antplus.c
+++ b/src/devices/ant_antplus.c
@@ -11,7 +11,7 @@
 /**
 ANT and ANT+ decoder.
 
-ANT and ANT+ comunication standards are defined by a division of Garmin
+ANT and ANT+ communication standards are defined by a division of Garmin
 https://www.thisisant.com/ and used widely for low power devices.
 The ANT radio transmits for less than 150 µs per message, allowing a single
 channel to be divided into hundreds of time slots and avoiding collisions.
@@ -32,7 +32,10 @@ use of a downconverter for rtl_sdr. The ISM band is very noisy, so it's
 recommended to get the device very close, use only a ~30mm wire antenna and
 use mid-low gains. Finally, since the protocol encodes at 1Mbps, sampling
 rate should be -s 4M or higher. It can work with a sampling rate as low as
-2Msps, but unreliably. The following works well with PlutoSDR:
+2Msps, but unreliably. To avoid excessive warnings when running with default
+250k sampling rate, the decoder is disabled by default.
+
+The following works well with PlutoSDR:
 
 rtl_433 -d driver=plutosdr,uri=ip:192.168.2.1 -g 20 -f 2457.025M -s 4M
 
@@ -40,16 +43,16 @@ Without knowing the 8 byte key, existing ANT chips do not allow sniffing
 of the data packets. Hence the need for this decoder. At the moment it's
 not possible to recover the ANT key from the 16 bit on-air network key.
 
-This decoder only captures and displays the low level packets, identifying the
+This decoder only captures and displays the low-level packets, identifying the
 2 byte network key plus all other device characteristics and 8 byte ANT payload.
 It identifies ANT+ packets using the unique network key used (0xa6c5)
 Refer to ANT+ documentation for ech specific device to parse the content of
 the pages sent as 8 byte ANT+ payload.
 
-The ANT protocol is using an uncommon strategy for the premable: either 0x55 or
+The ANT protocol is using an uncommon strategy for the preamble: either 0x55 or
 0xaa, depending on the value of the first bit of the following byte (0 and 1
 respectively). The nRF24L01, on which the ANT protocol is based, uses the same
-premable strategy. In order to determine if the packet is using 0x55 or 0xaa,
+preamble strategy. In order to determine if the packet is using 0x55 or 0xaa,
 both packets are extracted and the only valid one is determined by matching
 the packet CRC with a calculated CRC value for both alternative packets.
 
@@ -83,7 +86,7 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int antplus_flag = 0;
 
     // validate buffer: ANT messages are shorter than 150us, i.e. ~140 bits at 1Mbps
-    if (bitbuffer->bits_per_row[0] < 137 || bitbuffer->bits_per_row[0] > 160) {
+    if (bitbuffer->bits_per_row[0] < 120 || bitbuffer->bits_per_row[0] > 200) {
         return DECODE_ABORT_LENGTH;
     }
 
@@ -101,9 +104,9 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
     // calculate CRC for both alternatives, starting with aa (used by all ANT+ devices)
     // if the two crc bytes b[15] and b[16] are included in the crc calculation, a valid packet returns 0
-    if (crc16(b, 17, 0x1021, 0xffff) != 0) {  // no, preamble is not aa
+    if (crc16(b, 17, 0x1021, 0xffff) != 0) { // no, preamble is not aa
         bitbuffer_extract_bytes(bitbuffer, 0, bit_offset + 1, b, sizeof(b) * 8); // shift one bit right for 55 preamble
-        if (crc16(b, 17, 0x1021, 0xffff) != 0)  // nope, not preamble = 55 either, invalid packet, abort
+        if (crc16(b, 17, 0x1021, 0xffff) != 0) // nope, not preamble = 55 either, invalid packet, abort
             return DECODE_FAIL_MIC;
     }
 
@@ -114,20 +117,21 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     // display ANT and ANT+ payload in the same format used by ANT tools
     sprintf(payload, "%02x %02x %02x %02x %02x %02x %02x %02x", b[7], b[8], b[9], b[10], b[11], b[12], b[13], b[14]);
 
-    // display ANT or ANT+ depending on the network key used
+    // display ANT or ANT+ depending on the network key used.
     if (0xc5a6 == net_key)
         antplus_flag = 1;
 
     /* clang-format off */
     data_t *data = data_make(
-        "net_type",    "Network",        DATA_COND,   antplus_flag == 1,  DATA_STRING, "ANT+",
-        "net_type",    "Network",        DATA_COND,   antplus_flag == 0,  DATA_STRING, "ANT",
-        "net_key",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
+        "model",       "",               DATA_STRING, "Garmin-ANT",
+        "network",     "Network",        DATA_COND,   antplus_flag == 1,  DATA_STRING, "ANT+",
+        "network",     "Network",        DATA_COND,   antplus_flag == 0,  DATA_STRING, "ANT",
+        "channel",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
         "id",          "Device #",       DATA_FORMAT, "0x%04x", DATA_INT, id,
         "device_type", "Device type",    DATA_INT,    device_type,
         "tx_type",     "TX type",        DATA_INT,    tx_type,
         "payload",     "Payload",        DATA_STRING, payload,
-        "mic",         "Integrity",      DATA_STRING, "CHECKSUM",
+        "mic",         "Integrity",      DATA_STRING, "CRC",
         NULL);
     /* clang-format on */
 
@@ -136,8 +140,9 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 }
 
 static char *output_fields[] = {
-        "net_type",
-        "net_key",
+        "model",
+        "network",
+        "channel",
         "id",
         "device_type",
         "tx_type",
@@ -151,9 +156,10 @@ r_device ant_antplus = {
         .modulation  = FSK_PULSE_PCM,
         .short_width = 1,
         .long_width  = 1,
-        .sync_width  = 10,
+        .sync_width  = 8,
         .gap_limit   = 500,
         .reset_limit = 500,
         .decode_fn   = &ant_antplus_decode,
         .fields      = output_fields,
+        .disabled    = 1, // disabled by default, because of higher than default sampling requirements (s = 4M)
 };
--- a/src/devices/ant_antplus.c
+++ b/src/devices/ant_antplus.c
@@ -73,8 +73,6 @@ https://reveng.sourceforge.io/ to reverse engineer the CRC algorithm used
 */
 
 #include "decoder.h"
-#include "util.h"
-#include "stdbool.h"
 
 static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 {
@@ -82,9 +80,7 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     uint8_t b[17]; // aligned packet data for both preambles/offsets
     unsigned bit_offset;
     char payload[8 * 3 + 1]; // payload is 8 hex pairs for ANT and ANT+
-    bool antplus_flag = false;
-
-    data_t *data;
+    int antplus_flag = 0;
 
     // validate buffer: ANT messages are shorter than 150us, i.e. ~140 bits at 1Mbps
     if (bitbuffer->bits_per_row[0] < 137 || bitbuffer->bits_per_row[0] > 160) {
@@ -104,11 +100,10 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     bitbuffer_extract_bytes(bitbuffer, 0, bit_offset, b, sizeof(b) * 8); // assume preamble is aa
 
     // calculate CRC for both alternatives, starting with aa (used by all ANT+ devices)
-    uint16_t crc = crc16(b, 15, 0x1021, 0xffff);
-    if (((b[15] << 8) | b[16]) != crc) {                                         // no, preamble is not aa
+    // if the two crc bytes b[15] and b[16] are included in the crc calculation, a valid packet returns 0
+    if (crc16(b, 17, 0x1021, 0xffff) != 0) {  // no, preamble is not aa
         bitbuffer_extract_bytes(bitbuffer, 0, bit_offset + 1, b, sizeof(b) * 8); // shift one bit right for 55 preamble
-        crc = crc16(b, 15, 0x1021, 0xffff);
-        if (((b[15] << 8) | b[16]) != crc) // nope, not premable = 55 either, invalid packet, abort
+        if (crc16(b, 17, 0x1021, 0xffff) != 0)  // nope, not preamble = 55 either, invalid packet, abort
             return DECODE_FAIL_MIC;
     }
 
@@ -121,19 +116,19 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
     // display ANT or ANT+ depending on the network key used
     if (0xc5a6 == net_key)
-        antplus_flag = true;
+        antplus_flag = 1;
 
     /* clang-format off */
-	data = data_make(
-			"model",       "",               DATA_COND,   antplus_flag == true,  DATA_STRING, "ANT+",
-			"model",       "",               DATA_COND,   antplus_flag == false, DATA_STRING, "ANT",
-			"net_key",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
-			"device_num",  "Device #",       DATA_FORMAT, "0x%04x", DATA_INT, id,
-			"device_type", "Device type",    DATA_INT,    device_type,
-			"tx_type",     "TX type",        DATA_INT,    tx_type,
-			"payload",     "Payload",        DATA_STRING, payload,
-			"CRC",         "CRC",            DATA_FORMAT, "0x%04x", DATA_INT, crc,
-			NULL);
+    data_t *data = data_make(
+        "net_type",    "Network",        DATA_COND,   antplus_flag == 1,  DATA_STRING, "ANT+",
+        "net_type",    "Network",        DATA_COND,   antplus_flag == 0,  DATA_STRING, "ANT",
+        "net_key",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
+        "id",          "Device #",       DATA_FORMAT, "0x%04x", DATA_INT, id,
+        "device_type", "Device type",    DATA_INT,    device_type,
+        "tx_type",     "TX type",        DATA_INT,    tx_type,
+        "payload",     "Payload",        DATA_STRING, payload,
+        "mic",         "Integrity",      DATA_STRING, "CHECKSUM",
+        NULL);
     /* clang-format on */
 
     decoder_output_data(decoder, data);
@@ -141,13 +136,13 @@ static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 }
 
 static char *output_fields[] = {
-        "model",
+        "net_type",
         "net_key",
-        "device_num",
+        "id",
         "device_type",
         "tx_type",
         "payload",
-        "CRC",
+        "mic",
         NULL,
 };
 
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
- `crc16ccitt(b, len)` is our `crc16(b, len, 0x1021, 0xffff)` right?
- if crc bits are in the right order it should be `crc16ccitt(b1, 17) == 0`
- `model[5] = "ANT\0";` what? ;) Literal strings are terminated already.
- `strncpy(model, "ANT+\0", 5);` overflow, kinda, `"ANT+\0"` is 6 bytes.
- `DATA_STRING, model,` make model just a number and use
```
"model",            "",             DATA_COND, type == 1, DATA_STRING, "ANT-data",
"model",            "",             DATA_COND, type == 2, DATA_STRING, "ANT-plus",
```
but really isn't ANT+ just a mode of ANT, i.e. we can know structure?
- If `device_num` is an ID then output as key `id`
- Do not output the CRC, but add `mic`
- you don't need two copys of the output code detect the aa/55 then `b=b1` or `b=b2`

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,198 @@
...
+network ID using an unspecified algorithm. Valid keys are only assigned by Garmin and
+require specific licensing terms (and in some cases a payment)
+ANT+ uses the basic ANT message structure but is a managed network with a specific
+network key and well defined device types, each sending "data pages" of 8 bytes
+with specific data for the device type. Most ANT+ devices are sports focused like
+heart rate monitors, bicycle sensors, or environmental sensors.
+
+Please note that unlike most devices in the rtl_433 repository, ANT+ devices
+operate in the ISM band between 2.4GHz and 2.5GHz. Decoding these signals
+requires an SDR capable of operating above 2.4GHz (e.g. PlutoSDR) or the
```
merbanan, line 30: I would like a working command line example here.

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,198 @@
...
+    int i;
+    uint16_t crc = 0xffff;
+
+    while (length--) {
+        crc ^= *(unsigned char *)data++ << 8;
+        for (i = 0; i < 8; ++i)
+            crc = crc & 0x8000 ? (crc << 1) ^ 0x1021 : crc << 1;
+    }
+    return crc & 0xffff;
+}
```
merbanan, line 83: We should have proper native crc functions matching this.

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,198 @@
...
+    // find a data package and extract data buffer
+    bit_offset = bitbuffer_search(bitbuffer, 0, 0, preamble, sizeof(preamble) * 8) + sizeof(preamble) * 8;
+    if (bit_offset + sizeof(b1) * 8 > bitbuffer->bits_per_row[0]) { // did not find a big enough package
+        return DECODE_ABORT_LENGTH;
+    }
+
+    // ANT and ANT+ packes have either aa or 55 preamble, depending on the first bit of
+    // the following byte. The best way to know which is being used, is to rely on the CRC
+    bitbuffer_extract_bytes(bitbuffer, 0, bit_offset, b1, sizeof(b1) * 8);     // if preamble is AA
+    bitbuffer_extract_bytes(bitbuffer, 0, bit_offset + 1, b2, sizeof(b2) * 8); // if preamble is 55, read starting 1 bit further right
```
merbanan, line 107: Check this first. If the crc is valid then continue with the code. If not then extract one more bit and check this crc. If it passes then continue with the code and if not return error. Look in other decoder how to do that.

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,198 @@
...
+                "tx_type",     "TX type",        DATA_INT,    tx_type,
+                "payload",     "Payload",        DATA_STRING, payload,
+                "CRC",         "CRC",            DATA_FORMAT, "0x%04x", DATA_INT, crc1,
+                NULL);
+        /* clang-format on */
+
+        decoder_output_data(decoder, data);
+        return 1;
+    }
+    else if (((b2[15] << 8) | b2[16]) == crc2) { // valid preamble is 55
```
merbanan, line 150: With restructured code logic this code section will not be needed.

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,198 @@
...
+require specific licensing terms (and in some cases a payment)
+ANT+ uses the basic ANT message structure but is a managed network with a specific
+network key and well defined device types, each sending "data pages" of 8 bytes
+with specific data for the device type. Most ANT+ devices are sports focused like
+heart rate monitors, bicycle sensors, or environmental sensors.
+
+Please note that unlike most devices in the rtl_433 repository, ANT+ devices
+operate in the ISM band between 2.4GHz and 2.5GHz. Decoding these signals
+requires an SDR capable of operating above 2.4GHz (e.g. PlutoSDR) or the
+use of a downconverter for rtl_sdr. Finally, since the protocol encodes at
```
merbanan, line 31: It would be interesting if this could be used for this case. 

https://www.rtl-sdr.com/tuning-an-r820t2-rtl-sdr-up-to-6-ghz-via-a-harmonic-mixing-driver-hack/

A proper filter seems expensive though.

src/devices/ant_antplus.c
robcazzaro: Good idea, will do in the next check in (once I address everything else)

src/devices/ant_antplus.c
robcazzaro: I must have missed it, sorry, not familiar with the code yet.

src/devices/ant_antplus.c
robcazzaro: I was struggling on how to structure this. In the beginning I was printing both alternatives to help me understand how it worked, and I kinda got stuck in that mindset. But you are absolutely right, your solution is much more elegant. Also, now that I think of it, most devices that users are likely to find are ANT+, and that uses aa as preamble. So testing the aa alternative first is likely to be successful, avoiding extra code execution.

src/devices/ant_antplus.c
robcazzaro: Yep

src/devices/ant_antplus.c
merbanan: Sometimes I get sad when I see the amount of code that represent a decoder. You spent so much time and its only a few lines of code. But then again it is very easy to maintain.

src/devices/ant_antplus.c
robcazzaro: Before breaking down and buying the PlutoSDR, https://www.kd0cq.com/2016/04/part-ii-the-mod-receive-up-to-4-5ghz-on-your-rtlsdr-for-5-00/ and(http://blog.cyberexplorer.me/2014/01/sniffing-and-decoding-nrf24l01-and.html. I wrote a short note about it here https://github.com/robcazzaro/nRF24L01-SDR-testing

Both solutions worked, but with a lot of noise. And in any case, the rtl-sdr had too low a sampling rate to be useful for ANT and ANT+. I even looked into a mixer, but then more knowledgeable people said what you just said: need an expensive filter. A PlutoSDR looked like a much more palatable alternative :)

robcazzaro:
Once again, thanks for the feedback. I reworked the code as suggested (using native crc16(), better logic flow and DATA_COND conditional formatting. As I mentioned, I got stuck in the "dual data streams" and the old code was clunky. By forcing me to rethink the logic I realized it could be streamlined and faster. Hopefully still readable, which was one of the goals all along. I re-run all my tests and the code works as expected

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+
+Weirdly, L is always 10, because at the moment the ANT payload is always 8 bytes
+
+CREDITS:
+https://github.com/sghctoma/antfs-poc-defcon24
+https://reveng.sourceforge.io/ to reverse engineer the CRC algorithm used
+*/
+
+#include "decoder.h"
+#include "util.h"
```
zuckschwerdt, line 76: remove this, all needed includes are in decoder.h

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+#include "stdbool.h"
```
zuckschwerdt, line 77: use int instead of bool

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+
+static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    uint8_t const preamble[] = {0xAA};
+    uint8_t b[17]; // aligned packet data for both preambles/offsets
+    unsigned bit_offset;
+    char payload[8 * 3 + 1]; // payload is 8 hex pairs for ANT and ANT+
```
zuckschwerdt, line 84: use uint8_t, never char (could be signed)

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    // ANT and ANT+ packes have either aa or 55 preamble, depending on the first bit of
+    // the following byte. i.e. 10101010 1xxxxxxx or 01010101 0xxxxxxx
+    // the best way to know which is being used, is to verify which one has a valid CRC
+    // the following code relies on the fact that 55 is aa shifted right
+    bitbuffer_extract_bytes(bitbuffer, 0, bit_offset, b, sizeof(b) * 8); // assume preamble is aa
+
+    // calculate CRC for both alternatives, starting with aa (used by all ANT+ devices)
+    uint16_t crc = crc16(b, 15, 0x1021, 0xffff);
```
zuckschwerdt, line 107: if crc bits are in the right order it should be crc16ccitt(b, 17) == 0

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+    // display ANT and ANT+ payload in the same format used by ANT tools
+    sprintf(payload, "%02x %02x %02x %02x %02x %02x %02x %02x", b[7], b[8], b[9], b[10], b[11], b[12], b[13], b[14]);
+
+    // display ANT or ANT+ depending on the network key used
+    if (0xc5a6 == net_key)
+        antplus_flag = true;
+
+    /* clang-format off */
+	data = data_make(
+			"model",       "",               DATA_COND,   antplus_flag == true,  DATA_STRING, "ANT+",
```
zuckschwerdt, line 128: model needs to be e.g. "ANT-data" or "ANT-plus"

zuckschwerdt: but really isn't ANT+ just a mode of ANT, i.e. we can know structure?

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+			"model",       "",               DATA_COND,   antplus_flag == false, DATA_STRING, "ANT",
+			"net_key",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
+			"device_num",  "Device #",       DATA_FORMAT, "0x%04x", DATA_INT, id,
```
zuckschwerdt, line 131: If device_num is an ID then output as key id

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+			"device_type", "Device type",    DATA_INT,    device_type,
+			"tx_type",     "TX type",        DATA_INT,    tx_type,
+			"payload",     "Payload",        DATA_STRING, payload,
+			"CRC",         "CRC",            DATA_FORMAT, "0x%04x", DATA_INT, crc,
```
zuckschwerdt, line 135: Do not output the CRC, but add mic

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,164 @@
...
+
+static int ant_antplus_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    uint8_t const preamble[] = {0xAA};
+    uint8_t b[17]; // aligned packet data for both preambles/offsets
+    unsigned bit_offset;
+    char payload[8 * 3 + 1]; // payload is 8 hex pairs for ANT and ANT+
+    bool antplus_flag = false;
+
+    data_t *data;
```
zuckschwerdt, line 87: put this on the data_make() line: `data_t *data = data_make(`

src/devices/ant_antplus.c
robcazzaro: All ANT+ packets are ANT packets. The specs first define ANT as both a physical and logical point of view. ANT devices can use any valid ISM frequency (aside from 2457GHz)use any valid network key, and can format the payload any which way they want. There is a public ANT key defined, and any organization can buy a private key from Garmin to ensure that their data is "invisible" to the chips supporting the ANT protocol: if you don't know the key, cannot use an ANT chip to capture data. This was the driving reason behind my investigation. ANT devices are not design to interoperate between manufacturers

ANT+ devices can operate only in the 2457GHz channel, must use the ANT+ key, and must use one of the predefined payload formats defined by Garmin for each type of device (heart rate monitor, bicycle cadence and speed, etc). All ANT+ devices can interoperate because they use well defined packets

And the proper format to display which packet is is, it's ANT and ANT+.

src/devices/ant_antplus.c
robcazzaro: will do

src/devices/ant_antplus.c
zuckschwerdt: then perhaps rename that key to "device_format", the "model" key really needs to be `alphanum+ dash alphanum+`, no special chars exactly one dash in the middle, with usually manufacturer, brand, or class on the left and a short model-like thing on the right.

src/devices/ant_antplus.c
robcazzaro: I'm not sure I understand this comment, sorry. Are you saying that it's a property of that crc algorithm that including the crc itself results in a value of 0?

src/devices/ant_antplus.c
robcazzaro: Got it, thanks

src/devices/ant_antplus.c
zuckschwerdt: Yes, the CRC is the remainder of division, adding that in the remainder is zero.
So generally, if the byte/bit order fits, just compute the CRC over data+remainder and check for `0`.

src/devices/ant_antplus.c
zuckschwerdt: the model name is used in e.g. the MQTT topic, and `+` is problematic there.

src/devices/ant_antplus.c
robcazzaro: Since I rely on sprintf(payload), changing to uint8_t would require a lot more changes. I see that a lot of other decoders use the exact same strategy (char and sprintf), and as a matter of fact my code started from one of those. Is this change really needed?

zuckschwerdt: Ah, I didn't realize that. Looking at a buffer named "payload" I assumed it was a bit row.

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,159 @@
...
+    /* clang-format off */
+    data_t *data = data_make(
+        "net_type",    "Network",        DATA_COND,   antplus_flag == 1,  DATA_STRING, "ANT+",
+        "net_type",    "Network",        DATA_COND,   antplus_flag == 0,  DATA_STRING, "ANT",
+        "net_key",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
+        "id",          "Device #",       DATA_FORMAT, "0x%04x", DATA_INT, id,
+        "device_type", "Device type",    DATA_INT,    device_type,
+        "tx_type",     "TX type",        DATA_INT,    tx_type,
+        "payload",     "Payload",        DATA_STRING, payload,
+        "mic",         "Integrity",      DATA_STRING, "CHECKSUM",
```
zuckschwerdt, line 130: you can use "CRC" instead of "CHECKSUM"

src/devices/ant_antplus.c
```diff
@@ -0,0 +1,159 @@
...
+    // display ANT and ANT+ payload in the same format used by ANT tools
+    sprintf(payload, "%02x %02x %02x %02x %02x %02x %02x %02x", b[7], b[8], b[9], b[10], b[11], b[12], b[13], b[14]);
+
+    // display ANT or ANT+ depending on the network key used
+    if (0xc5a6 == net_key)
+        antplus_flag = 1;
+
+    /* clang-format off */
+    data_t *data = data_make(
+        "net_type",    "Network",        DATA_COND,   antplus_flag == 1,  DATA_STRING, "ANT+",
```
zuckschwerdt, line 123: I was confusing with my notes on "model", sorry. We do need "model", but it must be short and simple.
Then "id", then other fields, "mic" last.

src/devices/ant_antplus.c
robcazzaro: I'll leave as it is, then. Thanks for all the feedback

src/devices/ant_antplus.c
robcazzaro: Will do

src/devices/ant_antplus.c
robcazzaro: Just to make sure I understand correctly:

Can I use model = "ANT" or model = "ANT+" depending on the type of network?
Does id need to immediately follow model? For the purposes of this decoder, knowing the 16bit net_key is really important, and usually goes together with the ANT or ANT+ determination, so it's preferable to have it just following the ANT/ANT+ part. But not critical, can be moved if needed

zuckschwerdt: No, "model" needs to be the decoder name, some details are here: https://triq.org/rtl_433/DATA_FORMAT.html
All well-known fields (except mic) should be first (model, id), then uncommon fields, like "net_type", last "mic".
"net_type" can be "ANT+", no problem. But "model" must be a valid key for all outputs.

This is a special case I guess, it would be nice if "id" could be net_key+id to identify a device always. But the number would be full 32bit and thus negative :(
If net_key is logically part of the ID, then well maybe put it before "id", but make comment that it should not be there but it's a special case. (other people will copy things and chaos creeps in otherwise).

robcazzaro: It all makes sense, thanks for the explanations (I will add it to the wiki once this is closed, to ensure others can benefit). ANT and ANT+ devices do not fit neatly into the usual rtl_433 paradigm, but I still think that there is real value in adding this decoder. At the moment, this is the only way to really decode these devices without having to write custom modules in GNU radio or similar.

And unfortunately in these devices, the device # is not unique, and the device itself can change the device # on the fly, and many change it at boot time. So not even net_key+id would be unique in any real way. 

Would it be acceptable to have something like this:

```
    /* clang-format off */
    data_t *data = data_make(
        "model",       "Network",        DATA_COND,   antplus_flag == 1,  DATA_STRING, "ANTplus",
        "model",       "Network",        DATA_COND,   antplus_flag == 0,  DATA_STRING, "ANT",
        "net_key",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
        "id",          "Device #",       DATA_FORMAT, "0x%04x", DATA_INT, id,
        "device_type", "Device type",    DATA_INT,    device_type,
        "tx_type",     "TX type",        DATA_INT,    tx_type,
        "payload",     "Payload",        DATA_STRING, payload,
        "mic",         "Integrity",      DATA_STRING, "CRC",
        NULL);
    /* clang-format on */

static char *output_fields[] = {
        "model",
        "net_key",
        "id",
        "device_type",
        "tx_type",
        "payload",
        "mic",
        NULL,
};
```

Using ANT and ANTplus for the model, but displaying "Network" in the terminal UI, given that it's a network typology. As much as I'd like to use ANT+ instead of ANTplus, if the only option is to have both a ANT id and another ANT or ANT+, I'd rather go with a single ANT or ANTplus.

If `"model", "Network"` doesn't work, I'll go with `"model", ""` as other decoders do (but I verified that my version works)

zuckschwerdt: We have "id" and "channel" to identify senders, this maps to e.g. MQTT topics nicely. How about making net_key and id those two? Would that make sense?

About the "model": I'd rather see one model string in the usual format (naming the decoder, not the message contents) and then another key listing the coding type of "ANT"/"ANT+", as it's really not a different transmission but just with different rules for parsing the payload, right?

It's great to have these special-case decoders. It helps to push the conventions further. And there are others, e.g. M-Bus and EnOcean, that need a special setup. It's nice to see rtl_433 can be a simple framework for other things then weather sensors ;)

robcazzaro: Instead of checking in a different version, I'll keep iterating here until we have something good, if that works for you

And, yes, ANT/ANT+ is the same physical transmission, the only things that identifies ANT+ are fixed frequency (2457GHz), specific net key, well defined payloads for each device type and specific time slot in the time-division protocol. ANT+ defines device types, and in theory would be possible to write a specific ANT+ decoder that can interpret each device data in details. But it would get massively complicated, given that each device can send multiple pages with data (each "page" is 1 byte page number followed by 7 bytes data). For example, an heart rate monitor sends a few pages with the current heartrate, then one page with battery status, then more pages with heart rate, then manufacturer data, then more data rates, then extended info specific to the device, etc. But there are many tools capable of decoding ANT+, and those are much more functional than anything I could do in rtl_433. The critical part of this decoder is the ability to decode ANT devices using non standard net keys and transmission frequencies

How about something like this. Model is Garmin-ANT, given it's their proprietary format, network is ANT or ANT+, net_ney is shown as Net Key but it's channel, and id is the device #

```
    /* clang-format off */
    data_t *data = data_make(
        "model",       "",               DATA_STRING, "Garmin-ANT",
        "network",     "Network",        DATA_COND,   antplus_flag == 1,  DATA_STRING, "ANT+",
        "network",     "Network",        DATA_COND,   antplus_flag == 0,  DATA_STRING, "ANT",
        "channel",     "Net key",        DATA_FORMAT, "0x%04x", DATA_INT, net_key,
        "id",          "Device #",       DATA_FORMAT, "0x%04x", DATA_INT, id,
        "device_type", "Device type",    DATA_INT,    device_type,
        "tx_type",     "TX type",        DATA_INT,    tx_type,
        "payload",     "Payload",        DATA_STRING, payload,
        "mic",         "Integrity",      DATA_STRING, "CRC",
        NULL);
    /* clang-format on */

static char *output_fields[] = {
        "model",
        "network",
        "channel",
        "id",
        "device_type",
        "tx_type",
        "payload",
        "mic",
        NULL,
};
```

robcazzaro: One more question sorry: even if this protocol won't have false positives, given that it's for devices operating on a frequency that most consumer SDR devices can use, it might be appropriate to disable it by default.

If you agree with that, is there anything I need to do as part of the check-in to disable the protocol by default?

zuckschwerdt: Yes, this looks good. Note that normally we'd have model, id, channel, ..., mic. But I know you want a specific display for the kv output here, it's ok for this decoder.

Yes, the width of 1µs will not generally work, default-disable the decoder by adding `.disabled    = 1` in the r_device.

merbanan: We do support "payload" protocols also. For example wmbus is known to carry semi-generic data. The purpose of rtl_433 is to translate the rf signal into easily consumable data while still representing the rf signal as close as possible.

So the first step to only present the payload is good. And if you want you can as a next step decode the payload packages. But I guess this can be a project as big as rtl_433 it self.

And I don't see the reason for this to be disabled by default. The payload is protected by crc thus should be enabled by default.

zuckschwerdt: disabled just because at 250k rate it would spam "Sample rate too low for protocol", like EnOcean ;)

merbanan: well ok, but then maybe add a note about that

robcazzaro: I think I implemented all the feedback, hopefully this is now ready to be integrated

robcazzaro:
I pushed a new commit to my repository, but unlike in the past I do not see it reflected here. It looks as if there have been changes in master I'm really new to the "pull request" thing and I'm not sure if I should do something to address the situation, or just wait for you to do your magic. Please advise if there's anything I should be doing

I incorporated all the feedback and disabled the decoder by default. I verified it works as expected (disabled and not)

robcazzaro:
I have not received feedback on the last changes from a couple of weeks ago. Is there something else I need to address, or is this pull request simply waiting in line to be committed once there is free maintainer time?

zuckschwerdt:
Sorry, it will need a few steps to get integrated, so waiting on maintainer time, yes ;)

robcazzaro:
Thanks for the update. I was just worried that I did something wrong. No hurry on my side :) For what is worth, I kept using and testing this with a variety of ANT and ANT+ devices, and everything works as expected. I was finally able to figure out how the device that started my quest works.

So, thanks for all the help along the way, and I hope my contribution will be one day useful to others

PR #2021, opened 2022-03-23
Initial work on expression evaluation

triptec:
Running when v1=229,v2=230,v3=231
`rtl_433 -Xn=LOLZ,m=OOK_PPM,s=256,l=508,g=520,r=1520,preamble={119}000000000000000000000000000000,get=@0:{8}:length,get=@8:{8}:crc,get=@16:{8}:seq,get=@24:{8}:v1,get=@32:{8}:v2,get=@40:{8}:(value/10+230):v3 -r g001_433.92M_250k.cu8`
```
time      : @0.312192s
model     : LOLZ         count     : 1             num_rows  : 2             rows      : 
len       : 0            data      :               length    : 0             crc       : 0             seq       : 0             v1        : 0             v2        : 0             v3        : 230.000, 
len       : 48           data      : 03565ae5e6e7  length    : 3             crc       : 86            seq       : 90            v1        : 229           v2        : 230           v3        : 253.100
codes     : {0}, {48}03565ae5e6e7
```

Would it be okay to use tinyexpr?
It seems it also have support for passing functions into the evaluator so making things like rh_6to4 or other functions that would be useful in the evaluation of things.

I just got it to compile so stuff that just been copy pasted but I don't want to clean up unless you think I should continue.

### DIFF
--- a/include/tinyexpr.h
+++ b/include/tinyexpr.h
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: Zlib
+/*
+ * TINYEXPR - Tiny recursive descent parser and evaluation engine in C
+ *
+ * Copyright (c) 2015-2020 Lewis Van Winkle
+ *
+ * http://CodePlea.com
+ *
+ * This software is provided 'as-is', without any express or implied
+ * warranty. In no event will the authors be held liable for any damages
+ * arising from the use of this software.
+ *
+ * Permission is granted to anyone to use this software for any purpose,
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Yes, embedded ("header-only") libs will be okay.

triptec:
So looking at the radiohead code I don't think it's going to be nice solution just using expressions. You guys might already come to that conclusion.

merbanan:
No work for several years, closing.

PR #2027, opened 2022-03-27
Format for data_len instead of msg_len

triptec:
I think the loop should only look at data_len and not msg_len. msg_len is the length of the full message with headers and all as opposed to data length that's just the payload length and that what is being formated.

If I have understood this correctly


zuckschwerdt:
Thanks. Looks wrong to me too.

PR #2043, opened 2022-04-11
Add an optional TLS configuration param to the mqtt-hass discovery script

triplepoint:
This allows the script to subscribe to an MQTT broker over a secure TLS channel, and which has a custom, private TLS certificate.

Not passing the `-c` or `--ca_cert` CLI parameters results in the same behavior as before.


zuckschwerdt:
Thanks!

PR #2044, opened 2022-04-11
Add Aaronia RTSA HTTP interface support

g3gg0:
Hi,

This PR adds support to connect to a commercially available spectrum analyzer called Sepctran V6, made by the german company Aaronia.
It connects to the HTTP interface of their RTSA Suite (https://aaronia.com/software/rtsa-suite/) and passes the int16 complex data to the processing engine.
This device would also support CF32, but I was not sure how well it is supported in the sdr engine code.
To me it seemed the maximum is CS16.

Regards,
g3gg0


zuckschwerdt:
Interessting. But I think this would be better done as SoapySDR module?

g3gg0:
I see. Checking that lib, I could not make out if it is active, wip or stale.
Are there already implementations for devices available? 
I Can only find the ExampleDriver.

zuckschwerdt:
SoapySDR has lots of driver modules at https://github.com/pothosware/
A somewhat minimal networked driver is SoapyRTLTCP: https://github.com/pothosware/SoapyRTLTCP
Feel free to take that as a template if you like.

g3gg0:
Thanks for the hints.

While I find the idea behind it great and might even add support for SoapySDR to GSMAnalyzer or RX-FFT 
https://www.g3gg0.de/wordpress/programming/gsm-source-code-of-rx-fft-and-gsm-analyzer/
https://www.g3gg0.de/wordpress/projects/gsm-analyzer/

I am not sure if it will be within a short timeframe. This mainly depends on how quick I get into that code or if it requires a zillion other dependencies.

That's what I liked so much in rtl_433 - to nothing externally required, kind of selfcontained :)

zuckschwerdt:
> That's what I liked so much in rtl_433 - to nothing externally required, kind of selfcontained :)

Yes, that's a goal, required for e.g. small and constraint hardware.

SoapySDR is an external lib, and the modules are dynamically loaded but in that usecase (drivers) it's reasonable.
And SoapySDR is pretty self-contained too.

hb9fxq:
SoapySDR adapter is actually work in progress and currently in testing in will be released open source in just a few days ;-)

g3gg0:
haha okay then i'll skip that :)

did you implement both HTTP protocol and native lib?
because the HTTP channel has a great advantage - i can preprocess (LPF, gain) the signal which I found quite helpful.

regards,
g3gg0

PR #2050, opened 2022-04-20
Push the first row into the parent dataframe

klattimer:
When creating a custom flex decoder such as 
```
decoder n=Doorbell,m=OOK_PWM,s=300,l=600,r=5256,g=548,t=134,y=0,bits=25,repeats>=3,match=cba,get={12}:protocol,get=:12:{8}:id
```

I'd like to collect the first row data into the parent object. This small patch (1 added line) allows this to happen. 

This means that my MQTT event is received as this structure

```
{
	"codes": [
		"{25}cba3fa8",
		"{25}cba3fa8",
		"{25}cba3fa8",
		"{25}cba3fa8",
		"{25}cba3fa8",
		"{25}cba3fa8",
		"{24}cb47f5"
	],
	"count": 6,
	"freq": 433.85766,
	"id": 203,
	"mod": "ASK",
	"model": "Doorbell",
	"noise": -16.4857,
	"num_rows": 7,
	"protocol": 3258,
	"rows": [
		{
			"data": "cba3fa8",
			"id": 203,
			"len": 25,
			"protocol": 3258
		},
		# more rows
	],
	"rssi": -0.09784320000000001,
	"snr": 16.38788,
	"time": "2022-04-19T15:30:45.095780"
}
```

This makes the device more usable in Home assistant.

### DIFF
--- a/src/devices/flex.c
+++ b/src/devices/flex.c
@@ -298,6 +298,7 @@ static int flex_callback(r_device *decoder, bitbuffer_t *bitbuffer)
             "codes", "", DATA_ARRAY, data_array(bitbuffer->num_rows, DATA_STRING, row_codes),
             NULL);
     /* clang-format on */
+    render_getters(data, bitbuffer->bb[0], params);
 
     decoder_output_data(decoder, data);
     for (i = 0; i < bitbuffer->num_rows; i++) {


zuckschwerdt:
I like the idea and we do need something like this. But the feature needs more logic to it. Perhaps we should only "promote" to a single top-level output if exactly one row (or multiple identical ones) match.

Usually doorbell-style transmissions will have the first (few) and last (few) rows chopped off or wrong. Ideally we want a good line from the "middle".

klattimer:
Yes, I've been testing it and you're absolutely right - I may try to collect the most frequent data and build from that. I'll take a look when I get a minute. It shouldn't be overly difficult. 

I've also made a change to output_mqtt.c which includes the model in the MQTT devices structure, it seems type and subtype were also being stripped. I can't quite understand the logic of stripping keys. Do you know why this was done?

PR #2051, opened 2022-04-20
DSC decoder: Update reset_limit to fix ws4945

### LINKED ISSUE #1900: dsc: WS4975 door contact not decoding open events
My WS4975W gets closed events decodes correctly (x4):

image

But open events don't get decoded at all. Here are two recordings of the unknown signal (might be other signals mixed in):

[g002_433.92M_250k.zip](https://github.com/merbanan/rtl_433/files/7670283/g002_433.92M_250k.zip)
[g001_433.92M_250k.zip](https://github.com/merbanan/rtl_433/files/7670284/g001_433.92M_250k.zip)

Tried shifting the frequency and raising the sample rate, but still not decoding.

[This](https://triq.org/pdv/#AAB01607011358020C1AE80A60063C0E7CC17481919191919255+AAB01E07011358020C1AE80A60063C0E7CC174919394939191919494919191919555+AAB01907011358020C1AE80A60063C0E7CC17491949191919493919655)  should be the signal on triq BTW.

Appreciate any help!

GhettoDubs:
updated reset_limit to 7000us for ws4945 to fix #1900

### DIFF
--- a/src/devices/dsc.c
+++ b/src/devices/dsc.c
@@ -271,7 +271,7 @@ r_device dsc_security_ws4945 = {
         .modulation  = OOK_PULSE_RZ,
         .short_width = 536,  // Pulse length, 536 µs
         .long_width  = 1072, // Bit period, 1072 µs
-        .reset_limit = 6000, // Max gap,
+        .reset_limit = 7000, // Max gap,
         .decode_fn   = &dsc_callback,
         .fields      = output_fields,
 };


zuckschwerdt:
Looks good. But needs some explanation on how this fits the protocol.

zuckschwerdt:
Please add below line 33: "- Note the WS4945 doubles length of those timings."
And make the reset limit 9000.

GhettoDubs:
Updated as requested. Ive also tested this and it works with EV-DW4927 door/glass break sensor and WS4975 and works fine. 

For the EV-DW4927 it sends out two separate signals. Glass break uses the original ESN as written on the case and door sensor uses ESN with last digit +1.

Do you want any additional comments added?

GhettoDubs:
Also seems to work with DSC WS4904P motion sensors.

zuckschwerdt:
> Do you want any additional comments added?

Sure! Add something like both above comments -- always valueable to have!

GhettoDubs:
Ok done, additional comments added.

zuckschwerdt:
Oops, trailing whitespace in the comments. Otherwise perfect!

rct:
@zuckschwerdt wrote
> Please add below line 33: "- Note the WS4945 doubles length of those timings."
> And make the reset limit 9000.

This isn't quite a correct description. There might be some (newer?) versions of the WS4945 that double the timings, but the original WS4945 did not.

I have over 25 DSC contacts that were installed 2014. At least half of them are WS4945s, some are using the normally closed hardwired terminals for things like heat/rate of rise sensors. I also have smoke detectors, CO, glass break, the 4904P motion, a bunch of the EV-* door/window sensors.  *All of those* have worked with the original code for the years since this decoder was originally added. 

Off the to of my head, I *don't think* increasing the reset_limit would break anything with the existing/older sensors. I'll test and let you know.

zuckschwerdt:
There are two possible timings: https://github.com/merbanan/rtl_433/blob/master/src/devices/dsc.c#L265-L284
The second is labeled "WS4945", but we can update or annotate that with a better description anytime.

PR #2052, opened 2022-04-20
Add support for HCS200/HCS300 based remotes with FSK transmitters

Pietro210:
The existing decoder works fine but some remotes based on the same chips transmit using FSK instead of OOK.

I own one of such remotes so I initially tried running rtl_433 with this custom decoder:

`rtl_433 -R 0 -X 'n=hcs200,m=FSK_PWM,s=370,l=772,r=9000,g=1500,t=152'`

I confirmed that it works but it's a bit inconvenient, I'm not sure if adding a separate device protocol is the way to go but it might be useful to somebody else.


zuckschwerdt:
Thanks! Yes, adding a device is preferred.
Can you run `./maintainer_update.py` and add those changes?

Pietro210:
Done!

PR #2061, opened 2022-04-28
feat: add light_klx from bresser 7 in 1

acesyde:
Add missing `light_klx`

Logs :
```
INFO:root:Published Bresser-7in1/47539: time, temperature_C, humidity, wind_max_m_s, wind_avg_m_s, wind_dir_deg, rain_mm, uv, battery_ok
INFO:root:Skipped Bresser-7in1/47539: light_klx
```

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -477,7 +477,16 @@ mappings = {
             "state_class": "measurement"
         }
     },
-
+    "light_klx": {
+        "device_type": "sensor",
+        "object_suffix": "lux",
+        "config": {
+            "name": "Outside Luminance",
+            "unit_of_measurement": "lux",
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Oh, that slipped through on review and should always have been `light_lux`, we need to change that in Bresser-7in1. (and having "lux" as unit name everywhere is also unfortunate, the SI unit is "lx".)

Please change the PR to fix the key and scaling, then we'd need to let it sit a week or so with "Breaking change" warning to allow users a chance the comment.

acesyde:
Hi @zuckschwerdt 

Like that ?

zuckschwerdt:
Looks good. Also comment all klx lines with `// TODO: remove this`

acesyde:
@zuckschwerdt done ;)

PR #2062, opened 2022-05-04
rtl_433_mqtt_hass.py: Add expire_after option to show sensors as `unavailable` after x seconds

peterdey:
I had an issue where a sensor started reporting only intermittently (once every few days - instead of every few minutes).

Home Assistant would continue to show the last-received value, even when no value had been received for days.

MQTT sensors have a `expire_after` option, where HASS marks the sensor as `unavailable` when no value has been received after _x_ seconds.

This PR adds a command line option to `examples/rtl_433_mqtt_hass.py` to specify this behaviour.


zuckschwerdt:
Looks good to me.

PR #2066, opened 2022-05-09
Add support for ELRO home control remote

jenskdsgn:
ELRO AB440R is a simple power socket on and off switch


jenskdsgn:
@zuckschwerdt can you point me to some documentation of flex decoder so I can give it a try?

zuckschwerdt:
We only have the examples in the conf files and the short help from `rtl_433 -X` for now.

jenskdsgn:
I don't see how I can only take even-bits into consideration and use the odd-bits which are always 1 as some kind of sanity check.

zuckschwerdt:
They are only always on by chance, the third ternary symbol just isn't used here.

For the
```
BBBB:  1000 = button A
       0100 = button B
       0010 = button C
       0001 = button D
SS:    10 = ON
       01 = OFF
```
You can view the full pattern as
```
1B1B1B1B:  11101010 = 0xea = 234 = button A
           10111010 = 0xba = 186 = button B
           10101110 = 0xae = 174 = button C
           10101011 = 0xab = 171 = button D
1S1S:      1110 = 0xe = 14 = ON
           1011 = 0xb = 11 = OFF
```
And e.g. then write getters:
```
get=@10:{8}:button:[234:A 186:B 174:C 171:D],
get=@20:{4}:toggle:[11:OFF 14:ON],
```
Since channel is fixed you best filter that with a `match=`

jenskdsgn:
> They are only always on by chance, the third ternary symbol just isn't used here.
> 
> For the
> 
> ```
> BBBB:  1000 = button A
>        0100 = button B
>        0010 = button C
>        0001 = button D
> SS:    10 = ON
>        01 = OFF
> ```
> 
> You can view the full pattern as
> 
> ```
> 1B1B1B1B:  11101010 = 0xea = 234 = button A
>            10111010 = 0xba = 186 = button B
>            10101110 = 0xae = 174 = button C
>            10101011 = 0xab = 171 = button D
> 1S1S:      1110 = 0xe = 14 = ON
>            1011 = 0xb = 11 = OFF
> ```
> 
> And e.g. then write getters:
> 
> ```
> get=@10:{8}:button:[234:A 186:B 174:C 171:D],
> get=@20:{4}:toggle:[11:OFF 14:ON],
> ```
> 
> Since channel is fixed you best filter that with a `match=`

This is clear but the bits that are reversed for the channel information is quite tricky and I think can only be programmatically be solved unless we want to map each channel to the pattern which would be 32 pattern.
Let me know how to proceed here.

zuckschwerdt:
Ideally we need more decode helpers like the `decode_uart` in the flex spec. A `decode_ternary` or `decode_x1527` would be nice (and could also sanity check).

zuckschwerdt:
A generic symbol decode for flex was now added with #2161

To decode a `1B` two-bit to one-bit format use e.g., note that the symbols bits need to be MSB aligned:
`symbol_zero={2}8,symbol_one={2}c`

zuckschwerdt:
For a payload format of `1C1C1C1C1C 1B1B1B1B 10 1S1S 1`
e.g. `1011101110 10101110 10 1110 1` which is `{25}bbabae8`
you get `{12}512` which is `010100010010`:

`rtl_433 -R 0 -X 'n=name,m=OOK_PWM,s=330,l=970,r=1200,symbol_zero={2}8,symbol_one={2}c' -y '{25}bbabae8'`

The getters now would be:
`get=@0:{5}:channel,`
`get=@5:{4}:button:[8:A 4:B 2:C 1:D],`
`get=@10:{2}:toggle:[2:ON 1:OFF],`

petterreinholdtsen:
There are two automated checks failing for this pull request.  Perhaps it need an update?

zuckschwerdt:
The goal is to make decoders like this easy as flex conf files. That way most users can tweak timings or adapt them for new devices.

PR #871, opened 2018-11-18
Add HTTP WS user interface, REST api

zuckschwerdt:
Preview of an embedded HTTP WebSocket user interface, ReSTful API to to follow.
Very early proof-of-concept but basically working. WIP.

Start with `rtl_433 -M 1 -F http` and go to http://127.0.0.1:8433/ (initialization takes abound 30 seconds currently).

Comments and requests welcome, features planned so far:

- [x] Improved UI
- [ ] Mobile/Tablet suport
- [ ] Color coding (known fields, value ranges)
- [ ] Unit conversion
- [ ] Precision rounding
- [ ] History control
- [ ] Blacklist/Whitelist
- [ ] View / change active protocols
- [x] Changing frequency, hopping, rate
- [x] JSON-RPC API
- [x] HTTP-POST API
- [ ] ReSTful API
- [ ] Maybe analyzer support

PR #2068, opened 2022-05-10
fix crash when reading a conf file

anthyz:
I have rtl_433 running on an RPi4 in a docker container based on debian buster-slim, and when I use a .conf file to specify options, rtl_433 crashes if I try to set up an rtl_tcp output.  I dug down a bit and found that a strncmp() in parse_conf_option() blows up because optarg is NULL when reading a conf file.


zuckschwerdt:
Uhh, good catch, thanks! That's been there unnoticed since #871

PR #2071, opened 2022-05-11
Improve/change Ford TPMS decoder

gdt:
Change pressure encoding to psi*4.
Stop using a bit to flip the top bit of temperature.
Add bits for moving and learn (packet commanded by learn tool).
Add output variables for unexpected bits.

### DIFF
--- a/src/devices/tpms_ford.c
+++ b/src/devices/tpms_ford.c
@@ -10,20 +10,44 @@
 */
 /**
 FSK 8 byte Manchester encoded TPMS with simple checksum.
-Seen on Ford Fiesta, Focus, Kuga, Escape ...
+Seen on Ford Fiesta, Focus, Kuga, Escape, Transit...
+
+Seen on 315.00 MHz (United States).
 
 Seen on 433.92 MHz.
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Looks great! Esp. that you figured out the flags in that much detail.
I was never convinced that the manually calibrated PSI values were genuine. I guess the measurements were temperature compensated or such.

For temp you can use this inline without an `if`:
`"temperature_C",    "Temperature",  DATA_COND, temperature_valid, DATA_FORMAT, "%.1f C",   DATA_DOUBLE, (doubletemperature_c,`

gdt:
I will change to the conditional form (news to me but makes a lot of sense), retest and force push.   I definitely want to capture some live data after that change.

As for curve fitting, it's easy to fit noise, and get a slope/offset that is a bit wonky.  When using it for what regression is intended for, to predict bits from pressure, all of them will work just fine over the region it was developed from.   But I think it needs essentially snapping to sensible points, from a (correct, hard) guess at what is a priori likely.

gdt:
I think it's ready to merge now.

PR #2082, opened 2022-06-03
Fix spelling errors

dforsi:
The only change in code is "Accelaration" in src/devices/tpms_renault_0435r.c the other changes are in documentation.

Fixed with:
codespell --interactive=1 --write-changes --ignore-words-list=als,ans,hass,produkt,proove,temperatur


zuckschwerdt:
Thanks. Changing "Acceleration" is ok, but we mustn't change mongoose (3rd party code) and we likely shouldn't change old Changelog entries.

dforsi:
I reverted the changes to mongoose.* and to the changelog.
Editing old changelog entries is not like rewriting git history, but I can see that having immutable entries has its advantages.

PR #2083, opened 2022-06-04
Add support for Kia Rio III (UB) and Hyundai TPMS sensors

rrrlasse:


src/devices/tpms_kia.c
```diff
@@ -0,0 +1,149 @@
...
+
+    Copyright (C) 2022 Lasse Mikkel Reinhold, Todor Uzunov aka teou, TTiges, 2019 Andreas
+    Spiess, 2017 Christian W. Zuckschwerdt <zany@triq.net>
+
+    This program is free software; you can redistribute it and/or modify it under the terms of
+    the GNU General Public License as published by the Free Software Foundation; either version
+    2 of the License, or (at your option) any later version.
+*/
+
+/*
```
zuckschwerdt, line 12: should be `/**`
and remove 4 spaces for each line.

rrrlasse:
Sorry, I had one more commit because I just saw that other devices tell users to use the "-s 1000000" option through the device name. Let me know if I need to do anything else.

zuckschwerdt:
Thanks. I both instances "-s 1M"  would be preferred.

rrrlasse:
Hmm please wait merging this. I think the pulse widths need to be 50 instead of 200. I missed that when you read a file (-r flag), it assumes 250k by default. And I recorded at 1M. Need a few days to sort it out :)

zuckschwerdt:
If samples are tagged with e.g. ...`_1000k`... then reading will honor that sample rate.
https://triq.org/rtl_433/IQ_FORMATS.html#file-name-meta-data

PR #2086, opened 2022-06-08
Update oregon_scientific.c

elljay23:
Hi, I've added some sanity checks to confirm BCD digits are only 0-9 and check sensor values aren't outside their maximum range.  I've also corrected the flag to set the negative sign on temperature values.  It was incorrectly including the bit that denotes "maximum range limit reached".  I've been running the updates against my sensors for some weeks and they've stopped the screwy values I was seeing from rtl_433 every day or so.

This is my first pull request so I hope I've got the process right! :)

Thanks, Laurence

### DIFF
--- a/src/devices/oregon_scientific.c
+++ b/src/devices/oregon_scientific.c
@@ -50,23 +50,25 @@ static float get_os_temperature(unsigned char *message)
 {
     float temp_c = 0;
     temp_c = (((message[5] >> 4) * 100) + ((message[4] & 0x0f) * 10) + ((message[4] >> 4) & 0x0f)) / 10.0F;
-    if (message[5] & 0x0f)
+    // Correct 0x0f to 0x08:
+    if (message[5] & 0x08) {
         temp_c = -temp_c;
+    }
     return temp_c;
 }
 
 static float get_os_rain_rate(unsigned char *message)
 {
-    float rain_rate = 0; // Nibbles 11..8 rain rate, LSD = 0.01 inches per hour
-    rain_rate = (((message[5] & 0x0f) * 1000) + ((message[5] >> 4) * 100) + ((message[4] & 0x0f) * 10) + ((message[4] >> 4) & 0x0f)) / 100.0F;
+    float rain_rate = 0.0F; // Nibbles 11..8 rain rate, LSD = 0.1 units per hour, 4321 = 123.4 units per hour
+    rain_rate = (((message[5] & 0x0f) * 100) + ((message[5] >> 4) * 10) + (message[4] & 0x0f) + ((message[4] >> 4) & 0x0f)) / 10.0F;
     return rain_rate;
 }
 
 static float get_os_total_rain(unsigned char *message)
 {
-    float total_rain = 0.0F; // Nibbles 17..12 Total rain, LSD = 0.001, 543210 = 012.345 inches
+    float total_rain = 0.0F; // Nibbles 17..12 Total rain, LSD = 0.001, 654321 = 123.456
     total_rain = (message[8] & 0x0f) * 100.0F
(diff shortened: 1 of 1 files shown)


src/devices/oregon_scientific.c
```diff
@@ -50,23 +50,25 @@ static float get_os_temperature(unsigned char *message)
...
         temp_c = -temp_c;
+    }
     return temp_c;
 }
 
 static float get_os_rain_rate(unsigned char *message)
 {
-    float rain_rate = 0; // Nibbles 11..8 rain rate, LSD = 0.01 inches per hour
-    rain_rate = (((message[5] & 0x0f) * 1000) + ((message[5] >> 4) * 100) + ((message[4] & 0x0f) * 10) + ((message[4] >> 4) & 0x0f)) / 100.0F;
+    float rain_rate = 0.0F; // Nibbles 11..8 rain rate, LSD = 0.1 units per hour, 4321 = 123.4 units per hour
```
zuckschwerdt, line 62: This is later used as "rain_rate_in_h", can you confirm that units is in/h here? Or Is a unit 0.1 in, thus the change of scale wrong here?

src/devices/oregon_scientific.c
```diff
@@ -50,23 +50,25 @@ static float get_os_temperature(unsigned char *message)
...
     return rain_rate;
 }
 
 static float get_os_total_rain(unsigned char *message)
 {
-    float total_rain = 0.0F; // Nibbles 17..12 Total rain, LSD = 0.001, 543210 = 012.345 inches
+    float total_rain = 0.0F; // Nibbles 17..12 Total rain, LSD = 0.001, 654321 = 123.456
     total_rain = (message[8] & 0x0f) * 100.0F
-            + ((message[8] >> 4) & 0x0f) * 10.0F + (message[7] & 0x0f)
+            + ((message[8] >> 4) & 0x0f) * 10.0F + (message[7] & 0x0f)
```
zuckschwerdt, line 71: Trailing whitespace

src/devices/oregon_scientific.c
```diff
@@ -868,7 +922,7 @@ static char *output_fields[] = {
 r_device oregon_scientific = {
         .name        = "Oregon Scientific Weather Sensor",
         .modulation  = OOK_PULSE_MANCHESTER_ZEROBIT,
-        .short_width = 440, // Nominal 1024Hz (488µs), but pulses are shorter than pauses
+        .short_width = 440, // Nominal 1024Hz (488Âµs), but pulses are shorter than pauses
```
zuckschwerdt, line 925: UTF8/Latin1 conversion problems introduced an invalid code here. Best to replace `µs` with `us`

zuckschwerdt:
How did this slip through? A genuine fix for negative temps and needed sanity checks, thanks!

src/devices/oregon_scientific.c
elljay23: Hi, good question.  Not sure why I edited that section...  The sensor generated rain_rate lags so much I don't use it myself (I calculate it based on when the rain_total field changes.)  I used the page here: https://www.bashewa.com/wmr200-protocol.php as a starting point and compared it with the data I get from a PCR800 sensor.  I believe the PCR800 reports it in units of 0.1 inches per hour.

elljay23:
> How did this slip through? A genuine fix for negative temps and needed sanity checks, thanks!

Welcome :)  I've been running this code 24 hours a day since June and I no longer see any issues with "BCD" values outside A to F being converted to decimal and producing screwy jumps in the data.

zuckschwerdt:
Can you fix those 3 places (revert to "units" to inches per hour, remove the trailing space, change the mangeld "µ" char to "u") and just add another commit on to. Good to merge then!

zuckschwerdt:
Merging with a force push to correct the above now.

PR #2089, opened 2022-06-09
Add support for Badger ORION water meter

nickovs:
This PR provides support for the [Badger ORION](https://www.badgermeter.com/en-gb/products/endpoints/orion/) series of remotely readable water meters. According to the [FCC certification](https://fccid.io/GIF2006B) there are two variants, one with a fixed 916.45MHz frequency and one frequency hopping; this code has only been tested with the single frequency version (since that's all they have around me) but it should work with the other version if given the correct list of hop frequencies.

The encoding is similar to M-Bus mode T, using the same 4:6 DC-balanced encoding and the same CRC16 polynomial, but with a totally different payload format. This is documented in a comment in the code.

This is my first PR to this project; I've tried to follow all of the contributing guidelines but if there are any changes required then I'm more than happy to squash them in.

### DIFF
--- a/src/devices/badger_water.c
+++ b/src/devices/badger_water.c
@@ -33,7 +33,7 @@ preamble of 0000 1111 01. This is followed by 10 bytes encoded using a 4:6 NRZ
 encoding. This code treats 6 bits of the sync sequence as part of a 16 bit preamble.
 
 Once the data has been decoded with the NRZ 6:4 decoding, it has the following format:
-- Device ID: 3 bytes, little-endian. Value as stamped on cover, mod 2^24
+- Device ID: 3 bytes, little-endian. Typically utility provider's number, mod 2^24 or mod 10^7.
 - Device flags: 1 byte. Fields not known
 - Meter reading: 3 bytes, little-endian. Value in gallons for meters with 1-gallon resolution.
 - Status flags: 1 byte. Fields not known
--- a/src/devices/badger_water.c
+++ b/src/devices/badger_water.c
@@ -1,0 +1,162 @@
+/** @file
+    Badger ORION water meter support.
+
+    Copyright (C) 2022 Nicko van Someren
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "decoder.h"
+
+/** @fn static int badger_orion_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+Badger ORION water meter.
+
+S.a. https://fccid.io/GIF2006B
+
+For the single-frequency models the center frequency is 916.45MHz. The bit rate is
+100KHz, so the sample rate should be at least 1.2MHz; using 1.6MHz may work better
+when the signal is weak or noisy.
+
+The low-level encoding is much the same as M-Bus mode T, but the payload differs.
+
+The specification sheet states that "The endpoint broadcasts its unique endpoint
+serial number, current meter reading and applicable status indicators" and also that
+status reports include "Premise Leak Detection", "Cut-Wire Indication", "Reverse Flow
+Indication", "No Usage Indication" and "Encoder Error", but the specific flag values
+are not known.
+
+The data is preceded by several sync bytes of 01010101, followed by the ten bit
+preamble of 0000 1111 01. This is followed by 10 bytes encoded using a 4:6 NRZ
+encoding. This code treats 6 bits of the sync sequence as part of a 16 bit preamble.
+
+Once the data has been decoded with the NRZ 6:4 decoding, it has the following format:
+- Device ID: 3 bytes, little-endian. Value as stamped on cover, mod 2^24
+- Device flags: 1 byte. Fields not known
+- Meter reading: 3 bytes, little-endian. Value in gallons for meters with 1-gallon resolution.
+- Status flags: 1 byte. Fields not known
+- CRC: 2 bytes, crc16, polynomial 0x3D65
+
+*/
+
+// Mapping from 6 bits to 4 bits. "3of6" coding used for Mode T
+static uint8_t badger_decode_3of6(uint8_t byte)
+{
+    uint8_t out = 0xFF; // Error
+    //fprintf(stderr,"Decode %0d\n", byte);
+    switch(byte) {
+    case 22:    out = 0x0;  break;  // 0x16
+    case 13:    out = 0x1;  break;  // 0x0D
+    case 14:    out = 0x2;  break;  // 0x0E
+    case 11:    out = 0x3;  break;  // 0x0B
+    case 28:    out = 0x4;  break;  // 0x1C
+    case 25:    out = 0x5;  break;  // 0x19
+    case 26:    out = 0x6;  break;  // 0x1A
+    case 19:    out = 0x7;  break;  // 0x13
+    case 44:    out = 0x8;  break;  // 0x2C
+    case 37:    out = 0x9;  break;  // 0x25
+    case 38:    out = 0xA;  break;  // 0x26
+    case 35:    out = 0xB;  break;  // 0x23
+    case 52:    out = 0xC;  break;  // 0x34
+    case 49:    out = 0xD;  break;  // 0x31
+    case 50:    out = 0xE;  break;  // 0x32
+    case 41:    out = 0xF;  break;  // 0x29
+    default:    break;  // Error
+    }
+    return out;
+}
+
+// Decode the DC-free 4:6 encoding
+static int badger_decode_3of6_buffer(uint8_t const *bits, unsigned bit_offset, uint8_t *output)
+{
+    for (unsigned n=0; n<10; ++n) {
+        uint8_t nibble_h = badger_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset) >> 2);
+        uint8_t nibble_l = badger_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset+6) >> 2);
+        if ((nibble_h | nibble_l) > 15) {
+            return 1;
+        }
+        output[n] = (nibble_h << 4) | nibble_l;
+    }
+    return 0;
+}
+
+static int badger_orion_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    static uint8_t const preamble_pattern[] = {0x54, 0x3D};
+
+    uint8_t data_in[10] = {0}; // Data from Physical layer decoded to bytes
+    data_t *data;
+
+    // Validate package length
+    // The minimum preamble is 16 bits and the payload is 10 4:6 encoded bytes.
+    // There is often a long preamble and a 64 or more bits of tail, so the maximum likely length is longer
+    if (bitbuffer->bits_per_row[0] < (16 + 12 * 10) || bitbuffer->bits_per_row[0] > (128 + 16 + 12 * 10 + 96)) {
+        return DECODE_ABORT_LENGTH;
+    }
+
+    // Find the preamble
+    unsigned bit_offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, sizeof(preamble_pattern) * 8);
+    if (bit_offset + 12 * 10 >= bitbuffer->bits_per_row[0]) {
+        return DECODE_ABORT_EARLY;
+    }
+
+    decoder_logf_bitbuffer(decoder, 2, __func__, bitbuffer, "Preamble found at: %u", bit_offset);
+    bit_offset += sizeof(preamble_pattern) * 8; // skip preamble
+
+    if (badger_decode_3of6_buffer(bitbuffer->bb[0], bit_offset, data_in) < 0) {
+        return DECODE_FAIL_MIC;
+    }
+
+    uint16_t crc_read = (((uint16_t)data_in[8] << 8) | data_in[9]);
+    uint16_t crc_calc = ~crc16(data_in, 8, 0x3D65, 0);
+    if (crc_calc != crc_read) {
+        decoder_logf(decoder, 1, __func__,
+                "Badger ORION: CRC error: Calculated 0x%0X, Read 0x%0X",
+                (unsigned)crc_calc, (unsigned) crc_read);
+        return DECODE_FAIL_MIC;
+    }
+
+    uint32_t device_id = data_in[0] | (data_in[1] << 8) | (data_in[2] << 16);
+    uint8_t flags_1 = data_in[3];
+    uint32_t volume = data_in[4] | (data_in[5] << 8) | (data_in[6] << 16);
+    uint8_t flags_2 = data_in[7];
+
+    /* clang-format off */
+    data = data_make(
+            "model",      "",          DATA_STRING, "Badger-ORION",
+            "id",         "ID",        DATA_INT,    device_id,
+            "flags_1",    "Flags-1",   DATA_INT,    flags_1,
+            "volume_gal", "Volume",    DATA_INT,    volume,
+            "flags_2",    "Flags-2",   DATA_INT,    flags_2,
+            "mic",        "Integrity", DATA_STRING, "CRC",
+            NULL);
+    /* clang-format on */
+
+    decoder_output_data(decoder, data);
+    return 0;
+}
+
+// Note: At this time the exact meaning of the flags is not known.
+static char *badger_output_fields[] = {
+        "model",
+        "id",
+        "flags_1",
+        "volume_gal",
+        "flags_2",
+        "mic",
+        NULL,
+};
+
+// Badger ORION water meter,
+// Frequency 916.45 MHz, Bitrate 100 kbps, Modulation NRZ FSK
+r_device badger_orion = {
+        .name        = "Badger ORION water meter, 100kbps (-f 916450000 -s 1200000)", // Minimum samplerate = 1.2 MHz (12 samples of 100kb/s)
+        .modulation  = FSK_PULSE_PCM,
+        .short_width = 10,   // Bit rate: 100 kb/s
+        .long_width  = 10,   // NRZ encoding (bit width = pulse width)
+        .reset_limit = 1000, //
+        .decode_fn   = &badger_orion_decode,
+        .fields      = badger_output_fields,
+};
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -226,6 +226,7 @@
     DECL(ant_antplus) \
     DECL(emos_e6016_rain) \
     DECL(hcs200_fsk) \
+    DECL(badger_orion) \
 
     /* Add new decoders here. */
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -231,6 +231,7 @@ add_library(r_433 STATIC
     devices/x10_rf.c
     devices/x10_sec.c
     devices/yale_hsa.c
+    devices/badger_water.c
 )
 
 if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_C_COMPILER_ID}" MATCHES "Clang")


src/devices/badger_water.c
```diff
@@ -0,0 +1,157 @@
+/** @file
+    Badger ORION water meter support
```
zuckschwerdt, line 2: Add a "." to end the line.

src/devices/badger_water.c
```diff
@@ -0,0 +1,157 @@
...
+    case 35:    out = 0xB;  break;  // 0x23
+    case 52:    out = 0xC;  break;  // 0x34
+    case 49:    out = 0xD;  break;  // 0x31
+    case 50:    out = 0xE;  break;  // 0x32
+    case 41:    out = 0xF;  break;  // 0x29
+    default:    break;  // Error
+    }
+    return out;
+}
+
```
zuckschwerdt, line 68: no double newlines please

src/devices/badger_water.c
```diff
@@ -0,0 +1,157 @@
...
+            return 1;
+        }
+        output[n] = (nibble_h << 4) | nibble_l;
+    }
+    return 0;
+}
+
+static int badger_orion_callback(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    static const uint8_t PREAMBLE_ORION[] = {0x54, 0x3D};
```
zuckschwerdt, line 86: should be `uint8_t const preamble_pattern...`

src/devices/badger_water.c
```diff
@@ -0,0 +1,157 @@
...
+        uint8_t nibble_l = badger_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset+6) >> 2);
+        if ((nibble_h | nibble_l) > 15) {
+            return 1;
+        }
+        output[n] = (nibble_h << 4) | nibble_l;
+    }
+    return 0;
+}
+
+static int badger_orion_callback(r_device *decoder, bitbuffer_t *bitbuffer)
```
zuckschwerdt, line 84: should be named `..._decode`

src/devices/badger_water.c
```diff
@@ -0,0 +1,157 @@
...
+        return DECODE_ABORT_LENGTH;
+    }
+
+    // Find the preamble
+    unsigned bit_offset = bitbuffer_search(bitbuffer, 0, 0, PREAMBLE_ORION, sizeof(PREAMBLE_ORION) * 8);
+    if (bit_offset + 12 * 10 >= bitbuffer->bits_per_row[0]) {
+        return DECODE_ABORT_EARLY;
+    }
+
+    decoder_logf_bitbuffer(decoder, 2, __func__, bitbuffer, "PREAMBLE_ORION: found at: %u", bit_offset);
```
zuckschwerdt, line 104: `Preamble found at`  should be used, the context for the log is decoder and func automatically.

src/devices/badger_water.c
```diff
@@ -0,0 +1,157 @@
...
+        decoder_logf(decoder, 1, __func__,
+                "Badger ORION: CRC error: Calculated 0x%0X, Read 0x%0X",
+                (unsigned)crc_calc, (unsigned) crc_read);
+        return DECODE_FAIL_MIC;
+    }
+
+    /* clang-format off */
+    data = data_make(
+            "model",      "",          DATA_STRING, "Badger-ORION",
+            "id",         "ID",        DATA_INT,    data_in[0] | (data_in[1] << 8) | (data_in[1] << 16),
```
zuckschwerdt, line 123: the data values should be assembled in a preceeding block, easier to read :)

src/devices/badger_water.c
```diff
@@ -0,0 +1,157 @@
...
+            "flags_1",    "Flags-1",   DATA_INT,    data_in[3],
+            "volume_gal", "Volume",    DATA_INT,    data_in[4] | (data_in[5] << 8) | (data_in[6] << 16),
+            "flags_2",    "Flags-2",   DATA_INT,    data_in[7],
+            "mic",        "Integrity", DATA_STRING, "CRC",
+            NULL
+            );
```
zuckschwerdt, line 129: use `NULL);`

zuckschwerdt:
Looks very good, thanks!

nickovs:
@zuckschwerdt Thanks for the feedback. I have made the changes that you suggested.

nickovs:
@CFSworks Thanks for testing this and for your pervious analysis which set me on the right path. Thanks also for your pointing me to the rtl_433 project in the first place, since otherwise I wouldn't have done this PR! 

Regarding the ID field, can you check if the extracted ID field matches the number stamped on your meter mod 2^24? This is the case for my meter and the meters on the adjacent houses to mine. I've documented this assumption in the comment on [line 36](https://github.com/nickovs/rtl_433/blob/5114a7c87df97a970c98872d11e302de7a0b585e/src/devices/badger_water.c#L36) but it would be good to get further validation.

Regarding the flags, my guess (based on far too few observations) is that the byte at offset 3 has flags about the device hardware itself and the byte at offset 7 has status flags for the "Premise Leak Detection", "Cut-Wire Indication", "Reverse Flow Indication", "No Usage Indication" and "Encoder Error" conditions described in the documentation. In my neighbourhood most devices have data[3] == 0 but one device (which looks noticeably different and seems more modern) has one of these bits set. There is more variation in the bits in byte 7. Unfortunately without deliberately triggering a variety of fault conditions it is a little hard to accurately identify which flag is which.

CFSworks:
> @CFSworks You might want to try your tests on the device ID field again now that I've fixed a very dumb bug in my code!

Much better now! The ID field matches the low 7 digits of the serial on the label (so, not mod 2^24, but mod 10^7)

nickovs:
> Much better now! The ID field matches the low 7 digits of the serial on the label (so, not mod 2^24, but mod 10^7)

@CFSworks That is interesting. My guess is that the allocation of meter IDs is performed by the utility company that deploys them rather than by Badger, and the utility choose how these get programmed into the meter. I guess that some companies truncate the decimal value to fit it in and some truncate the binary value. I will update the comment in the code to reflect this variation.

Adminiuga:
> My guess is that the allocation of meter IDs is performed by the utility company that deploys them rather than by Badger

Seems so, my meter ID, the one which is indicated on the wire is `ReportedMeterID | 0x4000000`

Thanks for the PR! Finally can monitor the water usage

nickovs:
@merbanan @zuckschwerdt Any update on if this could be merged?

caliKev:
I think I have the frequency hopping version I can’t seem to receive any readings from my badger meter on the 916.45 frequency. Do you have any suggested next steps to pin down the frequency? I’m in an urban area so there is a lot of background.

mumixam:
@caliKev what's the fccid of yours?

timelf123:
I have a Badger 100WRD - FCC ID ewq100wd. I'm new to radio - does this spectrum grant mean that the 100WRD could possibly be listening on the same 916.45 freq? I'm not at the location with this meter right now to test

image

merbanan:
You can add frequency hopping parameters to rtl_433 command line. You can then add the option to output the frequency in the decoded output and then just let rtl_433 scan the frequency band for a while.

timelf123:
For later readers, The Badger 100WRD is actually using an Itron radio and is supported already with SCM+. Example:
> rtlamr -msgtype=scm+

```
{"FrameSync": 5795, "ProtocolID": 30, "EndpointType": 171, "EndpointID": FOOBAR, "Consumption": 121964, "Tamper": 18432, "PacketCRC": 53135}
```

segdy:
My water company recently installed a new meter. It's a "BadgerMeter Recordall Model 25". I have no idea if this is a smart meter or not (I know _previously_ water company would read the meters manually). Unfortunately water meter is out on the street so I can't easily use a pulse counter or similar.

Any chance this is supported as well?

I have tried

`rtl_433 -f 916.45M -s 1200k -R 223`

but I don't get any output. Any advice for debug?

nickovs:
Does the metal cover for the meter have a plastic insert, and if so, does it have an FCC ID stamped into it (or can you find such an ID anywhere else in the meter housing)? If so it's possible to look up the FCC certification and get quite a bit of detail about the transmissions.

segdy:
Hi Nick, thanks for the reply. No FCC indicator or anything else. Looks like this:





Maybe still not a smart meter?
(I’d be really surprised if they replace the meter but still don’t put a state of the art solution there …)

mumixam:
@segdy the meter itself does not need to be "smart". they can attached a device to the pipe that is able to calculate flow using ultrasonic 'waves'. that being said it doesn't look like you have one installed unless its outside of the photo. most of the time you'll see a wire that corrects to a antenna on the "lid" of the meter enclosure lid

segdy:
Thanks, so probably no smart meter the …
There is definitely nothing else outside and no cables / antenna. Lid is simply plastic.

One question, slightly OT: I know many meters have a pulse counter. Do you know from how far this can be read out? Does it have to be immediately on top of the meter? Or could it be “hidden” a few centimeters on the side of the meter?

nickovs:
@segdy That doesn't look like a meter that is remotely readable. For most models of Badger water meter the radio interface that allows remote reading is a separate unit that is attached by a wire and is fitted into the lid that covers the hole where the mechanical water meter is mounted to the pipe.

To answer your later question, at least for Badger, the pulse recorder is connected to the flow meter by a cable, but the radio signal from the recorder can be picked tens of meters away; with a tuned directional antenna pointing straight at it you could probably do even better. The receiver that I have picks up some signal from each of the meters on my block, although with a small, cheap, omnidirectional antenna the more distant ones tend to have rather weak signals. In my town the water company reads the all the meters just by driving slowly down the street in a car with an antenna on the roof.

caliKev:
5A56CA89-AE52-42E3-9A38-F66D293D009C

Sorry for the long delay. I don’t see an fcc ID. I’m guessing this is lte modem. Is there any way to read it?

jschwalbe:
Greetings! It appears my device (FCC ID [GIF2006B](https://fcc.report/FCC-ID/GIF2006B)) also uses frequency hopping. Hoping to find a way to utilize it. Any ideas?

nickovs:
@jschwalbe First off, it should be noted that meters with FCC ID GIF2006B can operate in either fixed frequency or hopping mode. It can run continuously on 911.65MHz, 916.45MHz and 921.25MHz, or it can hop on the 25 channels at 400kHz spacing starting with the lowest fixed frequency and running up to the highest fixed frequency (i.e. hops are at `911.65 + n * 0.4` MHz for `n` in the range `0` to `24`). So, before you go down the rabbit hole of trying to deal with frequency hopping you should try listing on each of those frequencies and see if you can pick anything up.

If that fails, you have two option: try to work out the order in which the hops are made, or ignore the issue and only collect 1 in 25 packets. The later option is much simpler, but the unit transmits ever 4 seconds so you'll only have the chance to pick up a reading every 100 seconds, and if you have interference then you'd need to wait another 100 seconds for the next one.

Working out the hopping sequence is not horribly hard, but it's sort of time consuming. The way I'd do it is to write a script that runs the `rtl_433` for a while (`-T 1000`) on each of the 25 channels, only listening for the Badger Orion (`-R 223`) with the output set to some machine readable format like JSON (`-F json`) and with the metadata set to include a high resolution time stamp (`-M time:unix:usec`). You could then analyse the timestamps for the transmissions on each frequency; records on a single frequency will show up with an interval that is a multiple of the total cycle time but the records on the different frequencies will have a different phase. A short Python script of similar should be able to work out the correct order for you.

I hope this helps!

nijhawank:
Hi @nickovs any help on this meter https://fccid.io/GIF2014W-OSE

I couldn’t find what frequencies I should try listen on

jschwalbe:
@nickovs Thank you so much for the detailed reply!

I have two units in the basement. One for inside water, and one for outside water.

`rtl_433 -f 911.65M -R 223` yielded nothing
`rtl_433 -f 916.45M -R 223` yielded two Badgers; updating at 4 sec intervals! (one is certainly mine as the Volumes match, but the other is off: 105323 vs 105238 and not changing despite changing on the physical readout)
`rtl_433 -f 921.25M -R 223` yielded nothing

For each time, I gave it a good 3 minutes (and much longer on the middle one). Didn't pick up any others, so that makes me wonder if the update interval is slower than the transmit interval. I'll report back on my findings as I take more readings over time! Thanks again!!


@nijhawank try this command `rtl_433 -f 904.9M -R 223`.  I honestly have no idea if that's going to work, but the FCC page says it transmits 904.9  -  924.5 MHz, so it's a start. Also try replacing 904.9M with 924.5M.

mgorbach:
@nickovs I have a badger meter labeled "Orion Cellular LTE" but FCC ID GIF20170CEV5D0M. I am able to see transmissions when running `rtl_433 -v -f 916.45M -s 1000000 -X 'n=badger,m=FSK_PWM,short=10,long=10,reset=1000'  every couple seconds. However, when I set the `preamble` as in your script, I no longer get any information decoded. Any idea how I can figure out how to decode this?

nickovs:
@mgorbach Firstly, given that the device as a cellular interface, do you know if it is going to also transmit in the 915MHz band. Finding a certification number on the device and looking up FCC certification documents would tell you this, even if the rest of the documentation is somewhat opaque.

The way I decoded the data in the first place was to read this [wiki page (https://github.com/merbanan/rtl_433/wiki/Decipher-a-new-device-and-write-a-decoder) and the FCC certification documents and then write some Python code to first visualise the data, followed by some trial and error. I can't really recommend a better option. Given that you seem to already be seeing some data, just printing it as binary and eyeballing it is not a bad place to start.

Looking at your test command line, I note that you didn't disable the other decoders. Just to check, are you seeing output packets that are tagged as `badger`? Are you seeing any other packets coming in on a similar cadence? If you are getting `badger` packets and nothing else with a similar timing then you are probably on the right track and it's just going to be a matter of figuring out the preamble.

Michael55123:
I'm in the same boat @mgorbach did you ever fin da solution?  I get this kind of data when I run you above command.
root@notlinux:~# rtl_433 -v -f 916.45M -s 1000000 -X 'n=badger,m=FSK_PWM,short=10,long=10,reset=1000'
bash: /usr/bin/rtl_433: No such file or directory
root@notlinux:~# /usr/local/bin/rtl_433 -v -f 916.45M -s 1000000 -X 'n=badger,m=FSK_PWM,short=10,long=10,reset=1000'
rtl_433 version 23.11-161-g71bcb5b7 branch master at 202409100934 inputs file rtl_tcp RTL-SDR

New defaults active, use "-Y classic -s 250k" if you need the old defaults

[Protocols] Registered 228 out of 263 device decoding protocols [ 1-4 8 10-12 15-17 19-23 25-26 29-36 38-47 49-60 63 67-71 73-85 87-100 102-105 108-116 119-122 124-128 130-149 151-161 163-168 170-175 177-197 199 201-215 217-232 234-241 243-244 246-247 249-259 261-263
[Input] The internals of input handling changed, read about and report problems on PR #1978
[SDR] Found 1 device(s)
[SDR] trying device 0: Realtek, RTL2838UHIDIR, SN: 00000001
[SDR] Using device 0: Realtek, RTL2838UHIDIR, SN: 00000001, "Generic RTL2832U OEM"
Exact sample rate is: 1000000.026491 Hz
[R82XX] PLL not locked!
[SDR] Sample rate set to 1000000 S/s.
[Input] Bit detection level set to 0.0 (Auto).
[SDR] Tuner gain set to Auto.
[Input] Reading samples in async mode...
[SDR] Tuned to 916.450MHz.
[Baseband] low pass filter for 1000000 Hz at cutoff 200000 Hz, 5.0 us
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:01:44
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 68           data      : fffffffffffffffff
codes     : {68}fffffffffffffffff
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:01:51
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 188          data      : c0000000000000000000000000000000000000000000000
codes     : {188}c0000000000000000000000000000000000000000000000
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:01
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 78           data      : fffffffffffffffffffc
codes     : {78}fffffffffffffffffffc
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:17
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 64           data      : fffffffffffffffe
codes     : {64}fffffffffffffffe
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:32
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 805          data      : 802080000a0408000000005250242080020012091108802080804204a04044060050a4808122204008a42a1540cd0010002800c402103440400902242020456418402a11488a208820a09a12092800414481580a0a805248100a8841006010801011084088
codes     : {805}802080000a0408000000005250242080020012091108802080804204a04044060050a4808122204008a42a1540cd0010002800c402103440400902242020456418402a11488a208820a09a12092800414481580a0a805248100a8841006010801011084088
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:34
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 67           data      : ffffffffffffffffe
codes     : {67}ffffffffffffffffe
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:46
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 37           data      : e000000008
codes     : {37}e000000008
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:46
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 401          data      : fffffffed5511169cd6bd79f101216858058b5ec31bed080969c0d73aa1f847c59a9942cf2d739c7625aeaaddeb4878e4ed00
codes     : {401}fffffffed5511169cd6bd79f101216858058b5ec31bed080969c0d73aa1f847c59a9942cf2d739c7625aeaaddeb4878e4ed00
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:46
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 93           data      : fffffff696a3e966c5f7a9c0
codes     : {93}fffffff696a3e966c5f7a9c0
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:02:50
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 79           data      : fffffffffffffffffffe
codes     : {79}fffffffffffffffffffe
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:03:06
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 38           data      : fffffffffc
codes     : {38}fffffffffc
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:03:06
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 24           data      : 7fffff
codes     : {24}7fffff
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2024-09-11 17:03:22
model     : badger       count     : 1             num_rows  : 1             rows      : 
len       : 74           data      : ffffffffffffffffffc
codes     : {74}ffffffffffffffffffc

baf:
@nijhawank @Michael55123 - just had my meter replaced with a similar model to what you seem to have. Both the old one and new one are Badger branded, but the old one worked fine with the `-R 223` decoder whereas the new one (FCC ID `GIF2018OCELMDOM`) does not. I also see plenty of transmissions on 916.45MHz, but I don’t have the experience to create my own decoder. Did either of you have any luck? @nickovs, would be happy to provide some sample output if you could help update your decoder.

PR #2102, opened 2022-07-04
Update mqtt hass to include triggers

deviantintegral:
This PR replaces #1665 . Currently, the significant difference is that it removes some of the protocol / model mapping the prior PR was doing, and also resolves conflicts against master.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -105,7 +105,7 @@ NAMING_KEYS = [ "type", "model", "subtype", "channel", "id" ]
 
 # Fields that get ignored when publishing to Home Assistant
 # (reduces noise to help spot missing field mappings)
-SKIP_KEYS = NAMING_KEYS + [ "time", "mic", "mod", "freq", "sequence_num",
+SKIP_KEYS = NAMING_KEYS + [ "mic", "mod", "freq", "sequence_num",
                             "message_type", "exception", "raw_msg" ]
 
 
@@ -156,6 +156,22 @@ mappings = {
             "unit_of_measurement": "°F",
             "value_template": "{{ value|float|round(1) }}",
             "state_class": "measurement"
+        }
+    },
+
+    # This diagnostic sensor is useful to see when a device last sent a value,
+    # even if the value didn't change.
+    # https://community.home-assistant.io/t/send-metrics-to-influxdb-at-regular-intervals/9096
+    # https://github.com/home-assistant/frontend/discussions/13687
+    "time": {
+        "device_type": "sensor",
+        "object_suffix": "UTC",
+        "config": {
+            "device_class": "timestamp",
+            "name": "Timestamp",
+            "entity_category": "diagnostic",
+            "enabled_by_default": False,
+            "icon": "mdi:clock-in"
         }
     },
 
@@ -541,6 +557,26 @@ mappings = {
         }
     },
 
+    "channel": {
+        "device_type": "device_automation",
+        "object_suffix": "CH",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+
+    "button": {
+        "device_type": "device_automation",
+        "object_suffix": "BTN",
+        "config": {
+           "automation_type": "trigger",
+           "type": "button_short_release",
+           "subtype": "button_1",
+        }
+    },
+
 }
 
 
@@ -597,7 +633,7 @@ def rtl_433_device_topic(data):
     return '/'.join(path_elements)
 
 
-def publish_config(mqttc, topic, model, instance, mapping):
+def publish_config(mqttc, topic, model, instance, mapping, value=None):
     """Publish Home Assistant auto discovery data."""
     global discovery_timeouts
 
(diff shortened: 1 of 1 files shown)


examples/rtl_433_mqtt_hass.py
```diff
@@ -619,10 +686,16 @@ def publish_config(mqttc, topic, model, instance, mapping):
     discovery_timeouts[path] = now + args.discovery_interval
 
     config = mapping["config"].copy()
-    config["name"] = object_name
-    config["state_topic"] = topic
-    config["unique_id"] = object_name
-    config["device"] = { "identifiers": object_id, "name": object_id, "model": model, "manufacturer": "rtl_433" }
+
+    if device_type == 'device_automation':
```
gdt, line 690: I would like to see a comment explaining the split.   I don't yet understand the details of MQTT discovery, but I find it hard to convince myself that this is right.  (That's not an assertion it isn't.)

examples/rtl_433_mqtt_hass.py
```diff
@@ -158,7 +158,15 @@
             "state_class": "measurement"
         }
     },
-
+    "time": {
```
gdt, line 161: It seems like there was a lot of thought into what to do and why but it is not explained in comments.

gdt:
If you are comfortable with a rebase into a single change, please do, but rtl_433 does not require that and we can squash merge.

examples/rtl_433_mqtt_hass.py
deviantintegral: I had left notes about this at https://github.com/merbanan/rtl_433/pull/1665#discussion_r799932852

However, I did a deeper look, and I am back on the side of "this isn't needed". Getting the "Last updated" value in home assistant requires setting `force_update` in the MQTT configuration: https://www.home-assistant.io/integrations/sensor.mqtt/#force_update

I've filed a docs PR as currently "updated" doesn't match anything to this: https://github.com/home-assistant/home-assistant.io/pull/23900

The only reason I can think to keep this is if somehow there are sensors out there that don't provide _any_ values  when transmitting to associate a "Last changed" value with?

I'm going to post in the discourse thread to ask for feedback on removing this.

examples/rtl_433_mqtt_hass.py
deviantintegral: device_automation has a different configuration structure than many other types. Would a link to https://www.home-assistant.io/integrations/device_trigger.mqtt/ suffice?

examples/rtl_433_mqtt_hass.py
deviantintegral: I ended up posting a home assistant docs PR, and in summary I misunderstood what "Last updated" referred to: https://github.com/home-assistant/home-assistant.io/pull/23900

Instead, we should set a timestamp as a diagnostic entity, and possibly have them disabled by default like zigbee2mqtt does: https://developers.home-assistant.io/blog/2021/10/26/config-entity/

deviantintegral: This sensor is now marked as a diagnostic sensor and disabled by default. This matches what zigbee2mqtt does.

examples/rtl_433_mqtt_hass.py
```diff
@@ -159,12 +159,29 @@
...
+        }
+    },
+
     "battery_ok": {
         "device_type": "sensor",
         "object_suffix": "B",
         "config": {
             "device_class": "battery",
             "name": "Battery",
+            "entity_category": "diagnostic",
```
deviantintegral, line 184: Both zwavejs and zigbee2mqtt mark batteries as diagnostic by default, so I've done the same here.

examples/rtl_433_mqtt_hass.py
```diff
@@ -596,8 +664,7 @@ def rtl_433_device_topic(data):
 
     return '/'.join(path_elements)
 
-
```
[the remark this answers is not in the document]
deviantintegral: Fixed!

examples/rtl_433_mqtt_hass.py
deviantintegral: I've added a comment for this.

deviantintegral (commented):
I think this is ready for another review.

I've removed the fields without standard interpretations as per https://github.com/merbanan/rtl_433/pull/2102#issuecomment-1195581778.

gdt:
@deviantintegral Are you able to rebase?   I will dig into this, but it would really help if there were clean commits each with defined scope.   This has a bunch of backwards merges and thus needs squashing, but then all the changes have to be understood at once.

It will take me a bit  as I want to actually run it -- I have a test HA instance for update pre-qualification.

deviantintegral:
Done! The only difference compared to the pre-rebase branch is removing a duplicate key for marking the battery as a diagnostic entity that's been fixed in master.

gdt:
@deviantintegral Thank you for your patience.  I have now recocvered from a hardware failure and some HA issues and am paging this in.   I'm glad to see it's under 200 lines of `log -p` and hope to digest soonish.   I don't really understand mqtt autodiscovery details but should -- right now I use manually-configured sensors, and I'm going to learn that so this gets some thought.

PR #2104, opened 2022-07-05
Requirements

JEENB:
added requirements for py and c


zuckschwerdt:
If this was an attempt do add D-Bus and a Python example, then please discuss this in an issue first.

rtl_433 is not likely to add any external lib dependencies. You might be better served to add D-Bus as a bridge script from UDP input. See the e.g. https://github.com/merbanan/rtl_433/blob/master/examples/rtl_433_custom.py

PR #2114, opened 2022-07-19
Add ESIC/SCMplus fields to rtl_433_mqtt_hass

stevieg2123:
Adding power meter energy/current/voltage values, as well as consumption from SCMplus meters (water/gas)

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -456,6 +456,42 @@ mappings = {
             "state_class": "measurement"
         }
     },
+  
+    "energy_kWh": {
+        "device_type": "sensor",
+        "object_suffix": "kwh",
+        "config": {
+            "device_class": "power",
+            "name": "Energy",
+            "unit_of_measurement": "kWh",
+            "value_template": "{{ value|float }}",
+            "state_class": "measurement"
+        }
+    },
+  
+    "current_A": {
+        "device_type": "sensor",
+        "object_suffix": "A",
+        "config": {
+            "device_class": "power",
+            "name": "Current",
+            "unit_of_measurement": "A",
+            "value_template": "{{ value|float }}",
+            "state_class": "measurement"
+        }
+    },
+  
+    "voltage_V": {
+        "device_type": "sensor",
+        "object_suffix": "V",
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
"consumption" is not a common field (and should not be named "SCMplus Consumption Value").
We don't want to publish one-off custom fields. IIRC there is no indication what scale (x kWh?) that field is?

stevieg2123:
> "consumption" is not a common field (and should not be named "SCMplus Consumption Value"). We don't want to publish one-off custom fields.

Good point re:consumption. Potentially something to look at upstream comparing ERT-SCM to SCMplus - might be worth standardizing on output field? 
https://github.com/merbanan/rtl_433/blob/10891fb732e65f3c6962a72ac05dd9656f07638e/src/devices/scmplus.c#L153
https://github.com/merbanan/rtl_433/blob/10891fb732e65f3c6962a72ac05dd9656f07638e/src/devices/ert_scm.c#L83

Was basing off previous commit for consumption_data, thus the misunderstanding for the naming. Let me know what your thoughts are and I can overhaul this pull, or go after the src/devices route.

> IIRC there is no indication what scale (x kWh?) that field is?

What's interesting with this one is that I am not receiving consistent readings - my assumption is that it is similar to consumption with the SCM meters; however, I cannot confirm/deny as to whether this resets at 24 hour, etc. Either way I'm inclined to remove it from the PR and if people want to pull it in they can manual MQTT entry in the config.

zuckschwerdt:
> Good point re:consumption. Potentially something to look at upstream comparing ERT-SCM to SCMplus - might be worth standardizing on output field?

Ideally it would be "energy_kWh" -- but we don't have the scaling.

> Was basing off previous commit for consumption_data,

Oops, so that one-off ship already sailed ;) Can you maybe add a warning comment to both so we hopefully clean that up in the future? `# NOTE: Only used in ... and should be normalized and removed someday.`

> What's interesting with this one is that I am not receiving consistent readings

The source noted a 1 or 10 kWh scale. You are seeing something even different?

zuckschwerdt:
Merging as it's currently useful. The "SCMplus Consumption Value" needs to be changed someday.

Issue #2105, opened 2022-07-07
Rainfall reading spikes with Acurite-Rain899 rain gauge


deviantintegral:
I recently set up an Acurite 0899 rain gauge, and I am getting spurious readings every few days from it:

> [weather_station] time      : 2022-07-07T01:19:57-0400
> [weather_station] model     : Acurite-Rain899
> [weather_station] id        : 3248
> [weather_station] channel   : 0
> [weather_station] Battery   : 1
> [weather_station] Rainfall Accumulation: 4.83 mm
> [weather_station] Integrity : CHECKSUM
> [weather_station] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> [weather_station]
> [weather_station] time      : 2022-07-07T01:19:57-0400
> [weather_station] model     : Acurite-Rain899
> [weather_station] id        : 3248
> [weather_station] channel   : 0
> [weather_station] Battery   : 1
> [weather_station] Rainfall Accumulation: 645.92 mm
> [weather_station] Integrity : CHECKSUM
> 
> <snip>
> 
> [weather_station] time      : 2022-07-07T02:45:58-0400
> [weather_station] model     : Acurite-Rain899
> [weather_station] id        : 3248
> [weather_station] channel   : 0
> [weather_station] Battery   : 1
> [weather_station] Rainfall Accumulation: 645.92 mm
> [weather_station] Integrity : CHECKSUM
> [weather_station] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> [weather_station]
> [weather_station] time      : 2022-07-07T02:46:57-0400
> [weather_station] model     : Acurite-Rain899
> [weather_station] id        : 3248
> [weather_station] channel   : 0
> [weather_station] Battery   : 1
> [weather_station] Rainfall Accumulation: 4.83 mm
> [weather_station] Integrity : CHECKSUM

What's interesting is that at least in the last few cases, the amount read is `645.92 mm`. The display unit does not show these values at all, so either rtl_433 is misinterpreting the signal or it has some sort of built-in filter.

Is this a common issue? I'm wondering if this should be filtered out in rtl_433 or if the reading is actually "correct" and I should filter it out downstream.

zuckschwerdt:
Rainfall here is just two bytes (7 bits each only): https://github.com/merbanan/rtl_433/blob/master/src/devices/acurite.c#L998-L1002
[...]

PR #2116, opened 2022-07-20
Reject short messages for Acurite-Rain899 #2105

deviantintegral:
Fixes the issue discussed at https://github.com/merbanan/rtl_433/issues/2105.

To do:

- [ ] Tests

### DIFF
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -996,6 +996,13 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
                 valid++;
             }
             else if (message_type == ACURITE_MSGTYPE_RAINFALL) {
+                // If the row is missing zeros, it can still pass the checksum.
+                // The earlier length check passes as ACURITE_TXR_BITLEN is 56.
+                if (browlen < 8) {
+                    decoder_log(decoder, 2, __func__, "skipping wrong len");
+                    continue;
+                } // DECODE_ABORT_LENGTH
+
                 // Rain Fall Gauge 899
                 // The high 2 bits of byte zero are the channel (bits 7,6), 00 = A, 01 = B, 10 = C
                 int channel = bb[0] >> 6;
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -996,9 +996,9 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
                 valid++;
             }
             else if (message_type == ACURITE_MSGTYPE_RAINFALL) {
-                // If the row is missing zeros, it can still pass the checksum.
-                // The earlier length check passes as ACURITE_TXR_BITLEN is 56.
-                if (browlen < 8) {
+                // The earlier length check passes as ACURITE_TXR_BITLEN is 56
+                // bits / 7 bytes. This message type is always 8 bytes.
+                if (browlen != 8) {
                     decoder_log(decoder, 2, __func__, "skipping wrong len");
                     continue;
                 } // DECODE_ABORT_LENGTH


deviantintegral:
I'd like to add a test for this. Looking at https://github.com/merbanan/rtl_433_tests, it appears all tests are recorded samples. Since `rtl_433 -R40:vv -y 'c34f0fffffeb0f/'` will replicate the issue, is there a way I can create a test that uses that instead of a recording?

src/devices/acurite.c
```diff
@@ -996,6 +996,13 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
                 valid++;
             }
             else if (message_type == ACURITE_MSGTYPE_RAINFALL) {
+                // If the row is missing zeros, it can still pass the checksum.
+                // The earlier length check passes as ACURITE_TXR_BITLEN is 56.
+                if (browlen < 8) {
+                    decoder_log(decoder, 2, __func__, "skipping wrong len");
+                    continue;
+                } // DECODE_ABORT_LENGTH
```
deviantintegral, line 1004: I noticed that this causes a zero return, and that is marked as legacy. I could refactor this function to the new return codes, but I think it makes more sense to just fix the bug first. Let me know if that should instead be done in this PR.

zuckschwerdt: Looks good. Clean up of return codes in a new PR would be great, if you have the time!

zuckschwerdt:
Then `-y` option was though as a first step to tests with just bitbuffers and not the excessive sample files. There was no good idea where/how to put those tests yet. Perhaps in the documentation comment, but format and returned output checking wouldn't be a good fit.

src/devices/acurite.c
```diff
@@ -996,6 +996,13 @@ static int acurite_txr_decode(r_device *decoder, bitbuffer_t *bitbuffer)
                 valid++;
             }
             else if (message_type == ACURITE_MSGTYPE_RAINFALL) {
+                // If the row is missing zeros, it can still pass the checksum.
+                // The earlier length check passes as ACURITE_TXR_BITLEN is 56.
```
gdt, line 1000: I don't like this comment, but I admit up front I am nitpicking.

First, it's a truism that corrupted messages could still pass a checksum, and we don't check other things only because of that -- we check everything we can, meaning every property that is believed to be true of all messages.

Second, and this is the bigger thing, the "earlier length check" is confusing.  I see 56, and that feels like 7.   If the earlier check is "every valid message is >= 7 bytes", then it's great to reject early.   But if different types are different lengths, then every type should have a length check.   And if the longest messages is 12 bytes (totally making that up) then the earlier check should also require <= 12.

Finally, the next line says <8, but is 9 valid?  20?  Should this be ==8?   If not ==, is there some upper bound?  If it is variable, is there an embedded length that can be checked after the min check?

I realize my specific comments may be off here, but I hope the approach of trying to check as aggressively as possible is useful.

deviantintegral: Thanks for the feedback, it's appreciated!

> But if different types are different lengths, then every type should have a length check.

I agree. I think this function is somewhat confusing because it combines several different message formats all at once. And given that we're working off of reverse engineered samples (and not committed specifications), it's entirely possible that future devices emit valid signals that break these assumptions.

I'm tempted to consider refactoring this along with the return codes mentioned above, but would probably do that as a followup PR as it's hard for me to know exactly how much time I'll have to commit in the next few months.

> Finally, the next line says <8, but is 9 valid? 20? Should this be ==8?

I'm not sure how to confirm that. I did `<` because that seemed to be the most targeted fix for corrupted messages. I'm OK changing it to 8, and if a future user somehow finds a valid message that is longer, at least we can ask for a sample to be included in the tests.

deviantintegral:
A PR in the testing repository is at https://github.com/merbanan/rtl_433_tests/pull/428.

rct:
FWIW, I tracked down some occasional bad data from my Acurite Atlas to short messages.   An 8 byte message was passing the MIC check, but it was the message type for the lightning sensor being installed so 10 bytes in the minimum.  This would cause me to get random lightning values for one message when the values were read past the end of the bit buffer row.

This PR looks fine to merge.   I'll be submitting something to more sanely check message type and length for at least the Atlas and possible other TXR format devices.

PR #2122, opened 2022-07-21
Add support for Fine Offset WN34 temperature sensor

anthyz:
I've tested this with the WN34S soil temperature sensor. It should also work with the WN34L liquid temperature variation. This decoder follows established fine offset protocol. The only thing unusual I found when comparing data with the Ecowitt WS View app is that the app displays battery bars calculated from the battery voltage in a non-linear manner. As a result I calculate battery_ok from the calculated bars rather than directly from reported voltage.


err53:
Did a build of the branch and can validate that it works as expected for the Ecowitt WN34L

zuckschwerdt:
Looks very nice. No comments so far ;)

Issue #2119, opened 2022-07-21
New device InFactory PT310 (ZX-7074/7073) pool thermometer


unverbraucht:
I bought this pool thermometer that's not currently decoded. I added samples and what I think might be the decoder in the [tests repo](https://github.com/merbanan/rtl_433_tests/compare/master...unverbraucht:rtl_433_tests:master), along with some samples:

Test output:
Data can be decoded by `-X 'n=pool3,m=OOK_PPM,s=1000,l=2008,g=2024,r=4064'`. Data packets get repeated 6 times with an empty packet between - maybe I'm missreading the encoding?

Decoding: 
```
Test mode active. Reading samples from file: ch1-27.7.cu8
baseband_demod_FM: low pass filter for 250000 Hz at cutoff 25000 Hz, 40.0 us
Detected OOK package    @0.247600s
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
time      : @0.247600s
model     : pool3        count     : 11            num_rows  : 11            rows      : 
len       : 37           data      : 8e8115f6e0, 
len       : 0            data      : , 
len       : 37           data      : 8e8115f6e0, 
len       : 0            data      : , 
len       : 37           data      : 8e8115f6e0, 
len       : 0            data      : , 
len       : 37           data      : 8e8115f6e0, 
len       : 0            data      : , 
len       : 37           data      : 8e8115f6e0, 
len       : 0            data      : , 
len       : 37           data      : 8e8115f6e0
codes     : {37}8e8115f6e0, {0}, {37}8e8115f6e0, {0}, {37}8e8115f6e0, {0}, {37}8e8115f6e0, {0}, {37}8e8115f6e0, {0}, {37}8e8115f6e0
Analyzing pulses...
Total count:  233,  width: 463.29 ms            (115823 S)
Pulse width distribution:
 [ 0] count:  227,  width:  472 us [424;500]    ( 118 S)
[...]

PR #2123, opened 2022-07-22
Add InFactory PT-310 parsing to Rubicson (longer packet length)

unverbraucht:
Adds the parsing for InFactory PT-310 (aka ZX...) to the Rubicson device.

As discussed in issue #2119.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
 
 		= Supported device protocols =
     [01]  Silvercrest Remote Control
-    [02]  Rubicson Temperature Sensor
+    [02]  Rubicson or InFactory PT-310 Temperature Sensor
     [03]  Prologue, FreeTec NC-7104, NC-7159-675 temperature sensor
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
Looks good, thanks! We can't change "model" -- can you undo that change?
And if possible: run `./maintainer-update.py` and commit afterwards.
Just add more commits, we squash on merge.

unverbraucht:
Done. I don't think that maintainer-update changed anything this time.

zuckschwerdt:
Thanks. It should have though. README and example conf are built using the .name value, and that looks different? Maybe you didn't compile recently?

unverbraucht:
Ah yes, that was it, I forgot to build - thanks. I also pulled master just in case.

zuckschwerdt:
Thanks for sticking with this. It can be a nuisance, but we tend to forget to fix small things once PRs are merged ;)

unverbraucht:
Not at all, it's great that the small things matter :)
I've observed around 1-2 wrong measurements per day, I'll investigate and open a separate PR for this.

PR #2127, opened 2022-07-26
examples: Revise comments near the MQTT relay example, and support MQTT authentication

gdt:
These commits aren't super related, but are all under the general umbrella of improving the MQTT relay and specfically documentation.  I'm submitting this PR now so that more invasive and perhaps controversial changes will be separate.
There are three unrelated commits (please don't squash):
  - Rototill README
  - Remove unnecessary dependency on bash
  - Add comments to the mqtt relay script, and add the ability to set the username/password and TLS variables

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,16 +1,94 @@
-# rtl_433 examples
+# rtl_433 scripts (example and useful)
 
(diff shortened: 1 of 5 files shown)


zuckschwerdt:
Looks good. Thanks!

PR #2147, opened 2022-08-12
add support for SRSmith SRS-2C-TX Pool Remote Control

gcohen55:


zuckschwerdt:
Looks good. `id` needs to be right after model. Can you explain about "baby_crc"? Preferably we only check and don't output those fields.

gcohen55:
they calculated a single byte crc for a subset of the packet even though the underlying modem supports performing a crc-16 on the whole packet (and the trailing two bytes at the end of the packet).

gcohen55:
removed a few fields from output. PR updated.

gcohen55:
I also provided samples on the rtl_433_tests repository: https://github.com/merbanan/rtl_433_tests/pull/432

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+    Copyright (C) 2022 gcohen55
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "decoder.h"
+#include "util.h"
```
zuckschwerdt, line 13: why "util.h"?
only include "decoder.h"

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+- P: 16 bit CRC-16, poly 0x8005, init 0xFFFF, of the packet from the size (S) until the CRC-8 (T)
+
+Format String:
+    PRE:32h SYNC: 32h SIZE: hh UNSURE:32h | UNSURE: 4b | PIN ~^4b |  BTN: hh | CRC-8: hh | CRC-16: hhhh
+
+Capture raw but often misshifted :| data. if you want to work with this data, throw it into bitbench and shift it over until you see A's instead of 5's
+    -f 915000000 -X n=SRSmith,m=FSK_PCM,s=100,l=100,r=4096,match=d391d391
+*/
+
+static int srsmith_pool_srs_2c_tx_callback(r_device *decoder, bitbuffer_t *bitbuffer)
```
zuckschwerdt, line 47: Use "srsmith_pool_srs_2c_tx_decode"

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+    if (bitbuffer->num_rows != 1)
+        return DECODE_ABORT_EARLY;
+
+    if (bitbuffer->bits_per_row[0] != 144)
+        return DECODE_ABORT_LENGTH;
+
+    // this is heavily derived from Maverick XR30 BBQ Sensor
+    uint8_t b[18];
+
+    // check for correct preamble/sync (0xaaaaaad391d391)
```
zuckschwerdt, line 60: It would improve robust reception to use `bitbuffer_search()` for just `aad391d391`

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+*/
+
+static int srsmith_pool_srs_2c_tx_callback(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    data_t *data;
+
+    if (bitbuffer->num_rows != 1)
+        return DECODE_ABORT_EARLY;
+
+    if (bitbuffer->bits_per_row[0] != 144)
```
zuckschwerdt, line 54: Better to use minimum (just `aad391d391`...) and maximum length (full `aaaaaad391d391`...)

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+    else if (bitbuffer->bb[0][0] == 0xaa) {
+        bitbuffer_extract_bytes(bitbuffer, 0, 8, b, 18 * 8); // shift in case first bit was received properly
+    }
+    else {
+        return DECODE_ABORT_EARLY; // preamble/sync missing
+    }
+    if (b[0] != 0xaa || b[1] != 0xaa || b[2] != 0xaa || b[3] != 0xd3 || b[4] != 0x91 || b[5] != 0xd3 || b[6] != 0x91)
+        return DECODE_ABORT_EARLY; // preamble/sync missing
+
+    int totallength = bitbuffer->bits_per_row[0];
```
zuckschwerdt, line 73: We don't need `totallength`

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+    int packetsize  = (int)b[7];
```
zuckschwerdt, line 74: no cast needed, remove `(int)`

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+            (reversed_pin & 0x80 ? 1 : 0),
+            (reversed_pin & 0x40 ? 1 : 0),
+            (reversed_pin & 0x20 ? 1 : 0),
+            (reversed_pin & 0x10 ? 1 : 0));
+
+    // button that was pushed
+    uint8_t button_id = b[13];
+
+    // get label
+    char button_string[50] = {'\0'};
```
zuckschwerdt, line 96: should be `char *button_string;`
and replace all `snprintf(button_string, sizeof(button_string), "On/Off Channel 1");` with `button_string = "On/Off Channel 1";`

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+        break;
+    }
+
+    // grab CRCs
+    // commenting out the below two lines because we're not using this crc
+    // see above for description.
+    //
+    // however, i'd still like the code to be here in case we need it in the future.
+    // uint8_t baby_crc            = b[14];
+    // uint8_t baby_calculated_crc = crc8(&b[8], 6, 1, 1);
```
zuckschwerdt, line 122: Is the CRC poly really `1`? That's parity not CRC.

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+
+    if (total_crc != calculated_total_crt) {
+        // parity fail.
+        return DECODE_FAIL_MIC;
+    }
+
+    /* clang-format off */
+    data = data_make(
+            "model",                  "",                         DATA_STRING, "SRSmith-SRS2CTX",
+            "id",                     "Id",                       DATA_STRING, pin_string, // technically peoples pins should be different on each remote
```
zuckschwerdt, line 137: this should be `reversed_pin` not `pin_string`, right?

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+            "totallength",            "Total Length",             DATA_INT, totallength,
```
zuckschwerdt, line 138: remove debug fields: "totallength", "packetsize", "extracted_pin", "total_crc"

src/devices/srsmith_pool_srs_2c_tx.c
```diff
@@ -0,0 +1,175 @@
...
+            "packetsize",             "Packet Size",              DATA_INT, packetsize,
+            "extracted_pin",          "Extracted PIN",            DATA_STRING, pin_string,
+            "button_pressed_id",      "Pushed Button ID",         DATA_FORMAT, "%02x", DATA_INT, button_id,
```
zuckschwerdt, line 141: Maybe better names are "button_press", "button_press_name"

zuckschwerdt (changes requested):

src/devices/srsmith_pool_srs_2c_tx.c
gcohen55: `pin_string` is correct in this case, imho, i'd rather output something that makes sense to whoever is working with this (e.g., have it match the dip switches on/off on the battery compartment of the remote

zuckschwerdt: Sorry, we really want simple numbers. Presentation (pretty printing) is not a task for rtl_433.
There are devices that have ID or channel code settings like A,B,C, or I/O, or X,- -- we want to always show a simple number to make database logging and such easier.
But you can add something like `decoder_logf(decoder, 2, __func__, "Pin setting is %c%c%c%c", pin0, pin1, pin2, pin3);` for debugging.

src/devices/srsmith_pool_srs_2c_tx.c
gcohen55: i'll grab and log for debugging. we just won't output.

src/devices/srsmith_pool_srs_2c_tx.c
gcohen55: renamed in code.

src/devices/srsmith_pool_srs_2c_tx.c
gcohen55: understood.

gcohen55:
hi team, i've made the requested changes.

gcohen55:
hi just following up on this PR - I believe I've taken care of all the requests here

PR #2155, opened 2022-08-19
Sanity check Ambient Weather F007th temperature and humidity

rct:
Sanity checks to reduce false positives and other bad data from Ambient Weather F007TH and related thermo-hygrometers.

Packets with Bad data often pass the MIC check.

- humidity > 100 (such as 255) and
- temperatures > 140 F (such as 369.5 F (raw 0xFFF) and 348.8 F)

Specs in the F007TH and F012TH manuals state the range is:

- Temperature: -40 to 140 F
- Humidity: 10 to 99%

Note: Allowing humidity 0 and above because broken humidity sensors may report 1% or 0% but still have functional temperature sensors.

### DIFF
--- a/src/devices/ambient_weather.c
+++ b/src/devices/ambient_weather.c
@@ -11,12 +11,29 @@
 */
 
 /**
-Ambient Weather F007TH Thermo-Hygrometer.
+Devices supported:
 
-Also: TFA senders 30.3208.02 from the TFA "Klima-Monitor" 30.3054,
-also: SwitchDoc Labs F016TH.
+- Ambient Weather F007TH Thermo-Hygrometer.
+- Ambient Weather F012TH Indoor/Display Thermo-Hygrometer.
+- TFA senders 30.3208.02 from the TFA "Klima-Monitor" 30.3054,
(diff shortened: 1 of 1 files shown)


rct:
@zuckschwerdt - I may not have had enough coffee yet, but I'm not sure what to do about these symbol errors from these the symbol error check.  It is building fine on my system.

```
Error: No func
Error: models not found (ambient_weather.c)
Error: but @sa (ambient_weather_decode)
```

zuckschwerdt:
The checker is not really context sensitive and get's confused with the doc-comment at line 69. That should be `/*` not `/**`.
Also ideally the first line in the decoder doc-comment should be just a headline style sentence, mostly the same as line 2.
And lists (29 to 36) should not be indented but each line start with a dash.

rct:
Thanks, changes to the comments made the symbol checker happy.

Also, while testing and looking for bad data I noticed I get a lot of false positives from Oregon SL109H that all have humidity > 100.

rct:
@zuckschwerdt - do you need anything else from me on this?

PR #2156, opened 2022-08-21
Adjust reset limit for Emos-E6016 to improve detection

vidarino:
Slightly increased reset limit for improved Emos-E6016 detection.

Sample capture that fails detection unless patched:
[emos-e6016.zip](https://github.com/vidarino/rtl_433/files/9389138/emos-e6016.zip)


zuckschwerdt:
Thanks, yes that reset limit is way to tight to the expected long gap of around 800 µs.
Not sure why there is a gap limit (of 3000 µs) given. The gaps on sync are (nominal) 260 µs, right?

PR #2158, opened 2022-08-23
Add open in browser script

zuckschwerdt:
A quick proof-of-concept script to parse a sample file with rtl_433 and open the pulse data analysis in the browser.
(just download the script and run `./open_rtl433.py sample.cu8`)

Quite useful shortcut for a common workflow: *show the pulse data (OOK or FSK) that rtl_433 found in some sample file.*

Something like this should also work for the I/Q data spectrogram. And we could perhaps add a "publish" feature to stash the data somewhere and return a url the user can share.

A draft just to get some initial feedback, ideally this should be integrated in rtl_433 itself (i.e. a mode to serve one http request and exit).

Comments and further ideas very much welcome!

### DIFF
--- a/examples/open_rtl433.py
+++ b/examples/open_rtl433.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+"""Helper command for rtl_433 to visualize a sample file in a web browser."""
+
+from http.server import BaseHTTPRequestHandler, HTTPServer
+import sys
+import subprocess
+import webbrowser
+
+hostName = "localhost"
+serverPort = 8080
+
+
+def parseToPulseData(filename):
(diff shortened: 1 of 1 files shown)


anthyz:
First I just have to say that the triq.org documentation and tools are well done and have been very informative and useful. Excellent work! That said this script is a welcome shortcut. One thing I noticed, though, is that pulse data is not properly generated for .cu8 files for FSK signals since all decoders are disabled and therefore enable_FM_demod=0.
Along the same lines I wonder if there are ways to streamline getting the bits into bitbench. For example, a button to copy the bits to the clipboard from Pulseview.

zuckschwerdt:
Thanks! I admit I just threw the script out there and didn't even test. We need to remove `"-R", "0",` from line 15 and keep the decoders active.

PR #2163, opened 2022-08-31
Add Dockerfile for creating a Docker image around rtl_433_mqtt_hass.py

unverbraucht:
I made a Dockerfile that wraps the Home Assistant MQTT wrapper because I run RTL 433 dockerized and missed the possibility to also run the wrapper in Docker.

I'll try to keep the Image updated (the MQTT wrapper doesn't change that often I think). Does it make sense to add the Dockerfile to the repo so that others can also build the image or improve it?

### DIFF
--- a/examples/Dockerfile-rtl_433_mqtt_hass
+++ b/examples/Dockerfile-rtl_433_mqtt_hass
@@ -1,0 +1,16 @@
+# Install python/pip
+FROM alpine
+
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
I'd rather see this as a project on it's own. We can split this off in a repo with access for interested contributors like you.

unverbraucht:
Great, then this can be closed. I'll link to the maintained repo.

PR #2173, opened 2022-09-15
Update cotech_36_7959.c

nepozs:
Added more compatible models


zuckschwerdt:
It would be better to list the actual sensors not just the stations the sensors work with. A useful hint anyway.

PR #2175, opened 2022-09-18
Add support for Neptune R900 flow meters

jruby411:
Using the work of many others, I added a decoder for the r900 neptune flow meters. Might as well share it with everyone. Thanks.

PR #2180, opened 2022-09-20
Add support for Neptune R900 flow meters

jruby411:
I tried to follow the guidelines and suggestion given on the last pull request for adding the Neptune flow meters that use the R900 protocol. If there is anything else that needs to be fixed, let me know.

Thanks.


zuckschwerdt:
Note that you could have added commits to #2175. Which would be preferred to track conversation and review.

zuckschwerdt:
Looks like all my review notes were addressed, thanks. I'll need to look at this again, leave it as-is for now if you don't feel comfortable fixing the whitespace warnings.

jruby411:
Sorry about not adding to the original Pull Request. This was nearly a re-write. I am still an amateur at this. Please forgive.

src/devices/neptune_r900.c
```diff
@@ -0,0 +1,258 @@
...
+#include "decoder.h"
+
+int const map16to6[16] = { -1, -1, -1, 0, -1, 1, 2, -1, -1, 5, 4, -1, 3, -1, -1, -1 };
+
+static void decode_5to8(bitbuffer_t *bytes, uint8_t *base6_dec) {
+    // is there a better way to convert groups of 5 bits to groups of 8 bits?
+    for (int i=0; i < 21; i++) {
+        uint8_t data = base6_dec[i];
+        if (data < 0x0F) {
+            bitbuffer_add_bit(bytes, 0);
```
merbanan, line 70: Is the else case really needed?

jruby411: Nope. Sorry, I was still thinking in "string mode" when I was building 5-bit strings out of 4-bit numbers.

I fixed this in my latest commit.

Thank you for reviewing my code.

-Jeff.

roger-:
FWIW I tested this and it seems to be working. 

Only issue is it's less sensitive than rtlamr and only picks up one R900 device, but that could partially due to the receiver settings (I tried to match rtlamr's settings with `rtl_433 -f 912380000 -R 224 -s 2359296`).

Would be great to have this merged.

jruby411:
I have been running this code 24/7 on my server for the last couple of months without issues. I pick up a lot of meters with these settings: ` rtl_433 -f 912600155 -Y classic -s 2000k -R 224` Of course I am only interested in my meter, and it works great for that.

roger-:
@jruby411 interesting because with those settings I only get my meter, whereas rtlamr gets several. Have you compared this to rtlamr?

Only difference I can think of besides the demodulator is the gain settings.

roger-:
Any chance of merging this?

zuckschwerdt:
Yes. I justed started catching up on PRs and this is coming up next now.

Issue #2186, opened 2022-09-24
How to publish unixtime in utc?


HolgerAusB:
Hi,
I tried for `output mqtt`:

```
report_meta time:unix:utc:usec
report_meta time:unix:utc
report_meta time:utc:unix
report_meta utc:unix
```
only one row at the time. But I never get the unixtime for UTC, always in localtime. Is that a bug or did I miss something else?

Tried on two Bullseye-Raspies
- one with rtl-433 from bullseye-repo: version 20.11-1
- one installed from scratch: version 21.05-79-ga43c2a05

zuckschwerdt:
Those are correct. Strange, if I don't use `unix` it works fine. Same for you?
The `utc` option just sets the env var "TZ" to "UTC" and `unix` or `iso` just set a format string.

HolgerAusB:
Yes, @zuckschwerdt!
`report_meta time:utc` and even undocumented
`report_meta utc`

brings the UTC-time in human readble format

gdt:
It seems very strange for "unix time" (as returned by time(2)) to be in anything but UTC, no matter what someone tries to ask for.    What does it mean "unix time in local time"?  Is that seconds since 1970-01-01 at 0000 local?  If so, I would say it's a bug that it can be produced.

zuckschwerdt:
@gdt good point. Somewhat unexpected how `localtime_r()` and `strftime()` interact still.
https://github.com/merbanan/rtl_433/blob/master/src/r_api.c#L359-L363
https://github.com/merbanan/rtl_433/blob/master/src/r_util.c#L66-L75

I don't think I'll investigate that further, PR to clear up the docs or fix something welcome though.

HolgerAusB:
OK, seems to be my fault. As I now read at wikipedia, unixtime **is** always UTC. I fell for a Python example that obviously gave a wrong value here. And now knowing that unix=utc it is oviously wrong:

```
from datetime import timezone, datetime
currentTime = int(datetime.strftime(datetime.utcnow(), "%s"))
```
In comparison, I naturally thought that the error was with rtl_433, since example `time:unix:utc:usec` is explicitly listed in example.conf.

gdt:
> @gdt good point. Somewhat unexpected how `localtime_r()` and `strftime()` interact still. https://github.com/merbanan/rtl_433/blob/master/src/r_api.c#L359-L363 https://github.com/merbanan/rtl_433/blob/master/src/r_util.c#L66-L75
> 
[...]

PR #2187, opened 2022-09-24
Update rtl_433.example.conf

HolgerAusB:
As unix-time allways is in utc, the 'utc' the example 'time:unix:utc:usec' didn't change anything.
see #2186


zuckschwerdt:
Thanks! Can you also apply the same to `help_meta()` in `src/rtl_433.c` so they match? Just add another commit to this PR.

HolgerAusB:
mhh, will try

PR #2205, opened 2022-10-17
Fix Chuango devices with some zeroes in their ID

LuccoJ:
My PIR sensor wasn't being detected by rtl_433 anymore since after the last refactor of this code, but it works again with this change


zuckschwerdt:
Good catch, thanks. I broke that check 4 years ago in 90fc4e1 oO.

PR #2214, opened 2022-10-24
Improve false positives for acurite_01185m

knarrff:
This decoder treated a row of 56 '0's as valid. However, this would be unreasonable temperatures (already handled in existing code by excluding those from being reported), both channel and id of '0', an 'ok' battery. The checksum does not help here, as the valid value is '0' as well.

### DIFF
--- a/src/devices/acurite_01185m.c
+++ b/src/devices/acurite_01185m.c
@@ -74,6 +74,14 @@ static int acurite_01185m_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             result = DECODE_FAIL_MIC;
             continue; // return DECODE_FAIL_MIC;
         }
+        /* A sanity check to detect some false positives. The following in
+           particular checks for a row of 56 "0"s, which would be unreasonable
+           temperatures, channel and id of 0, an 'ok' battery, which all
+           happens to result in a '0' checksum as well.
+        */
+        if (chk == 0) {
+            return DECODE_FAIL_SANITY;
+        }
 
         // Decode fields
         int id        = (b[0]);


src/devices/acurite_01185m.c
```diff
@@ -74,6 +74,14 @@ static int acurite_01185m_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             result = DECODE_FAIL_MIC;
             continue; // return DECODE_FAIL_MIC;
         }
+        /* A sanity check to detect some false positives. The following in
+           particular checks for a row of 56 "0"s, which would be unreasonable
+           temperatures, channel and id of 0, an 'ok' battery, which all
+           happens to result in a '0' checksum as well.
+        */
+        if (chk == 0) {
```
rct, line 82: Thanks, more sanity checking is a good thing.  

However, are you sure a checksum of 0 can't happen? (On other acurite devices 0 can be a valid checksum even on these relatively short messages.)

This isn't a test for a row a zeros . At a minimum, I'd check that the ID/channel are also zero.

zuckschwerdt: @rct I thought so too, but there is a nuance: `chk = add_bytes(b, 6);` -- so `chk == 0` is indeed a test for 6 bytes of zero ;)

rct: That's a bit subtle for my tastes, but you are right.  I looked too quickly, I saw the `chk & 0xff` on the next line and though it was `chk = chk & 0xff`. because I thought `chk` would have the actual 8 bit checksum.

knarrff: I agree with the sentiment that this made the code hard to read, and code should be easy to read and to understand. I have now renamed the variable 'chk' to 'sum'. I believe this improves readability. Thanks for your feedback.

PR #2236, opened 2022-11-13
Add note regarding WH51 915MHz decoding

loopy321:
Add comment note to use  "-Y classic".

Might be useful to add in case someone else runs into the same problem.


zuckschwerdt:
Can you make that:
```
(empty line)
Note: for WH51 at 915MHz: try also "-Y classic" i.e. : rtl_433 -f 915M -Y classic -- see https://github.com/merbanan/rtl_433/issues/2235
```

PR #2248, opened 2022-11-27
GE smartremote plus

mores:


zuckschwerdt:
Thanks. But "dumb" protocols should really be using a simple flex conf. Much easier for other users to adapt and tune.
Your check translates to `,bits>=25` (actually 24, as the last bit in EV1527 is a sync bit)
Channel could be `,get=channel:@1:{1}`
And the state would be `,get=state:@22:{1}`

I'd recommend:
`rtl_433 -f 319.56M -X 'n=GE-Smartremote-RF108,m=OOK_PWM,s=330,l=1000,r=1500,invert,bits>=24,bits<=25,get=channel:@1:{1},get=action:@22:{1}'`

zuckschwerdt:
Oh, and the ID: `,get=id:@2:{20}`

mores:
Closing in favor of: https://github.com/merbanan/rtl_433/pull/2249

PR #2254, opened 2022-11-29
Add log option to all outputs

zuckschwerdt:
Logging used to be simple messages on stderr. Limitations:
- Always visible on the console (KV output), not available at other outputs (JSON, Syslog, HTTP)
- Not really inline with events (events: stdout, log: stderr)
- No severity visible (by color or common markers)

Logging now changed to be messages along the same path as events:
- Selectable for any output
- Inline with events
- Severity by color (or level markers)

Usage changes:
- If you **don't select any outputs** then **logging will go to KV** output (equivalent to `-F kv`, i.e. `-F kv,v=8`)
- If you **select the KV** output then **logging will be shown**, suppress with `-F kv,v=0`)
- If you **don't select the KV** output then **no logging** will be shown, add it with `-F log`)
- If you want **logging back to stderr and suppressed in KV use** `-F log -F kv,v=0`

### Implementation
Now that refactoring from `fprintf()`s in decoders to `decoder_log*()` calls is done this will finally allow all logging to selectively (verbosity levels) go to outputs (KV, JSON, CSV, MQTT, HTTP, …).

One would raise the verbosity globally (`-v`) or per decoder (`-R 99,v`) and configure outputs to display those messages (e.g. `-F json,v`).

### Tasks
- [x] Outputs need to parse the level option
- [x] Global prints ("noise level") need to be converted to log
- [x] KV ouput should prettify the logs
- [x] Looks like there is an overflow somewhere? (6e1f120)

### DIFF
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -176,8 +176,8 @@ signal_grabber none
 ## Data output options
 
 # as command line option:
-#   [-F kv|json|csv|mqtt|influx|syslog|trigger|null] Produce decoded output in given format.
-#     Without this option the default is KV output. Use "-F null" to remove the default.
+#   [-F log|kv|json|csv|mqtt|influx|syslog|trigger|null] Produce decoded output in given format.
+#     Without this option the default is LOG and KV output. Use "-F null" to remove the default.
 #     Append output to file with :<filename> (e.g. -F csv:log.csv), defaults to stdout.
 #     Specify MQTT server with e.g. -F mqtt://localhost:1883
 #     Add MQTT options with e.g. -F "mqtt://host:1883,opt=arg"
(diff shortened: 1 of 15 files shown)


zuckschwerdt:
I'm pondering what label coloring would appeal to most? (dim or intense bg? regular or bold fg?)

intense-norm
intense-bold

dim-norm
dim-bold

The "intense-bold" variant is what we would use by default right now, the "dim" variants would be slightly more compatible with Windows 8.1 (only Windows 10+ offers bright bg).

The labels would show the logging "source" as seen here:

log-src

zuckschwerdt:
Personally, I prefer the crisper looking non-bold text (also it's less of a distraction).
And the dim backgrounds might be a problem as many custom color schemes "wash out" those colors.
For me it then would be the first example above.

zuckschwerdt:
Ok, let's use variant 1 for now (intense background, crisp non-bold text).

I'll merge this now. There is still need to fine-tune some log levels and messages, feel free to leave some feedback!

PR #2258, opened 2022-12-01
Change somfy_iohc to recognize more messages

mtdcr:
- Messages can be as short as 11 bytes.
- Message length has 5 bits instead of 4.
- Flags were assumed to be constant, causing two-way remotes not to get recognized.
- The message type (command id) is not indicated by the 1st but the 9th byte.
- Trailing bytes were not removed, leading to invalid length and failing CRC.
- Sequence numbers and MAC are limited to authenticated one-way frames.

This was briefly tested using both one- and two-way remotes.

### DIFF
--- a/src/devices/somfy_iohc.c
+++ b/src/devices/somfy_iohc.c
@@ -136,8 +136,7 @@ static int somfy_iohc_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     if (num_bits <= 0)
         return DECODE_ABORT_EARLY;
 
-    if ((size_t)num_bits > sizeof b * 8)
-        num_bits = sizeof b * 8;
+    num_bits = MIN((size_t)num_bits, sizeof (b) * 8);
 
     int len = extract_bytes_uart(bitbuffer->bb[0], offset, num_bits, b);
     if (len < 11)
--- a/src/devices/somfy_iohc.c
+++ b/src/devices/somfy_iohc.c
@@ -91,59 +91,122 @@ Example packets:
 
 #include "decoder.h"
 
+struct iohc_msg {
+    /* Mandatory fields */
+    /* Control byte 1 */
+    unsigned end_flag : 1;
+    unsigned start_flag : 1;
+    unsigned protocol_mode : 1;
+    unsigned frame_length : 5;
+    /* Control byte 2 */
+    unsigned use_beacon : 1;
+    unsigned is_routed : 1;
+    unsigned low_power_mode : 1;
+    unsigned protocol_version : 3;
+    /* Addresses */
+    unsigned dst_addr : 24;
+    unsigned src_addr : 24;
+    /* Command ID */
+    unsigned cmd_id : 8;
+    char data[31 * 2 + 1]; /* variable length, converted to hex string */
+    unsigned crc : 16;
+
+    /* optional fields */
+    int seq_nr;
+    char mac[13];
+};
+
 static int somfy_iohc_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 {
     uint8_t const preamble_pattern[] = {0x57, 0xfd, 0x99};
+    struct iohc_msg msg = {0};
 
-    uint8_t b[19 + 15]; // 19 byte + up 15 byte payload
+    uint8_t b[1 + 31 + 2]; // Length, payload, CRC
 
     if (bitbuffer->num_rows != 1)
         return DECODE_ABORT_EARLY;
 
-    unsigned offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, 24) + 24;
-    if (offset + 19 * 10 >= bitbuffer->bits_per_row[0])
+    unsigned offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, 24);
+    if (offset == bitbuffer->bits_per_row[0])
         return DECODE_ABORT_EARLY;
 
-    unsigned num_bits = bitbuffer->bits_per_row[0] - offset;
-    num_bits = MIN(num_bits, sizeof (b) * 8);
+    offset += 24;
+
+    int num_bits = bitbuffer->bits_per_row[0] - offset;
+    if (num_bits <= 0)
+        return DECODE_ABORT_EARLY;
+
+    if ((size_t)num_bits > sizeof b * 8)
+        num_bits = sizeof b * 8;
 
     int len = extract_bytes_uart(bitbuffer->bb[0], offset, num_bits, b);
-    if (len < 19)
+    if (len < 11)
         return DECODE_ABORT_LENGTH;
 
-    if ((b[0] & 0xf0) != 0xf0)
-        return DECODE_ABORT_EARLY;
-
-    int msg_len = b[0] & 0xf; // should be 6 or 8
-    if (len < 19 + msg_len)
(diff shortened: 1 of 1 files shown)


src/devices/somfy_iohc.c
```diff
@@ -91,60 +91,123 @@ Example packets:
...
-    uint8_t b[19 + 15]; // 19 byte + up 15 byte payload
+    uint8_t b[1 + 31 + 2]; // Length, payload, CRC
 
     if (bitbuffer->num_rows != 1)
         return DECODE_ABORT_EARLY;
 
-    unsigned offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, 24) + 24;
-    if (offset + 19 * 10 >= bitbuffer->bits_per_row[0])
+    unsigned offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, 24);
+    if (offset == bitbuffer->bits_per_row[0])
```
zuckschwerdt, line 130: Use `>=` to be safe

src/devices/somfy_iohc.c
```diff
@@ -91,60 +91,123 @@ Example packets:
...
+        return DECODE_ABORT_EARLY;
+
+    offset += 24;
+
+    int num_bits = bitbuffer->bits_per_row[0] - offset;
+    if (num_bits <= 0)
```
zuckschwerdt, line 136: This check can never be true, remove.

src/devices/somfy_iohc.c
```diff
@@ -91,60 +91,123 @@ Example packets:
...
         return DECODE_ABORT_EARLY;
 
-    unsigned num_bits = bitbuffer->bits_per_row[0] - offset;
-    num_bits = MIN(num_bits, sizeof (b) * 8);
+    if ((size_t)num_bits > sizeof b * 8)
```
zuckschwerdt, line 139: keep the parens on sizeof and add braces for the block, otherwise maybe keep `MIN`

src/devices/somfy_iohc.c
```diff
@@ -91,60 +91,123 @@ Example packets:
...
+            "msg_type",         "Command",          DATA_FORMAT, "%02x", DATA_INT, msg.cmd_id,
+            "msg",              "Message",          DATA_STRING, msg.data,
             "mic",              "Integrity",        DATA_STRING, "CRC",
+            "mode",             "Mode",             DATA_STRING, msg.protocol_mode ? "One-way" : "Two-way",
+            "version",          "Version",          DATA_INT, msg.protocol_version,
+            NULL);
+
+    if (msg.protocol_mode == 1)
+        data = data_append(data,
+            "counter",          "Counter",          DATA_INT,    msg.seq_nr,
```
zuckschwerdt, line 194: You can also use `DATA_COND` here an put in the data_make above.
E.g. 
`"counter",          "Counter",          DATA_COND, msg.protocol_mode == 1, DATA_INT,    msg.seq_nr,`

src/devices/somfy_iohc.c
```diff
@@ -91,60 +91,123 @@ Example packets:
...
+            "mac",              "MAC",              DATA_STRING, msg.mac,
             NULL);
     /* clang-format on */
 
+    if (decoder->verbose) {
```
zuckschwerdt, line 199: Not a good idea to have decoder data depended on verbosity level.
If this is just debug output please use `decoder_logf()`

src/devices/somfy_iohc.c
```diff
@@ -91,60 +91,123 @@ Example packets:
 
 #include "decoder.h"
 
+struct iohc_msg {
```
zuckschwerdt, line 94: Do you really need a struct? I.e. using just e.g. `iohc_end_flag` or `msg_end_flag` would be simpler to maintain.
Otherwise you will need to put a `@fn` on the doc-comment above.

src/devices/somfy_iohc.c
mtdcr: Safe from what?

src/devices/somfy_iohc.c
mtdcr: I did this because of the cast to size_t below. This way a reviewer will see it's safe to cast, even without knowing the data type or value range of bits_per_row[0].

Using <= instead of < was just an optimization, because it's possible to find the preamble_pattern without following data, therefore getting num_bits==0, right?

src/devices/somfy_iohc.c
mtdcr: Why braces? Other single-line blocks don't have any. `sizeof` is an operator, but well... I'll change it back to MIN with a cast.

src/devices/somfy_iohc.c
mtdcr: It clutters the output, but some people may want to see it. According to the help screen, -vv is supposed to make decoders verbose. It didn't work for me, FWIW. Feel free to adapt it. It's not debug output, as it outputs received protocol data.

src/devices/somfy_iohc.c
mtdcr: Using a struct is both clean and easy to initialize to zero. There's no way it would add maintenance burden per se, only if you insist on adding documentation. As you'd accept the struct members as standalone variables without documentation, adding it to the struct seems rather pointless. It's not in a shared header file anyway.

mtdcr:
Thanks for you review! I added another commit reintroducing the use of the MIN macro. I'll stop here, because I'm not interested in discussing style preferences of yet another project just to contribute a single patch. You can modify it to suit your needs, if you like to.

src/devices/somfy_iohc.c
merbanan: The rule we try to follow is that data on the air should be translated as straight as possible to what is output. So we want it all. It is up to the consumer/presentation layer to select what data to use.

src/devices/somfy_iohc.c
zuckschwerdt: API changes. `bitbuffer_search()` does not guarantee any particular return value.

src/devices/somfy_iohc.c
zuckschwerdt: Ok, but `bitbuffer->bits_per_row[0] - offset >= 0` is checked by excluding `offset >= bitbuffer->bits_per_row[0]` 3 lines above. Modern compilers will infer that and bug us with "condition is constant".

src/devices/somfy_iohc.c
zuckschwerdt: Operator, yes, that's why `sizeof (x)` is prefered to `sizeof(x)`. The `sizeof x * y` just looks sketchy at a quick glance.
Newer and changed code should preferably use braces, even for single-line blocks -- the compiler and linter will warn on mishaps, still  it's just an accident waiting to happen.

zuckschwerdt:
> not interested in discussing style preferences of yet another project

Fair enough. The work analyzing and coding the protocol is appreciated.

TBH style preferences here are mostly "same dumb simple in all decoders" -- with ~200 decoders any "smart" or "neat" will cause headache when e.g. refactoring with some regex is needed.

Velocet:
@mtdcr / @zuckschwerdt You both should have a look at [iown-homecontrol](https://github.com/velocet/iown-homecontrol)
The protocol is mainly deciphered and the „encryption“ is broken and documented. You will find all relevant info there.

Issue #2275, opened 2022-12-13
BRESSER 8 Channel Thermo-/Hygro-Sensor (variant of efth800?)


andrewjw:
Hi,

I have some Bresser 8 channel thermo/hydro sensors ([link](https://www.bresseruk.com/Weatherstations/BRESSER-8-Channel-Thermo-Hygro-Sensor.html)) which I'm struggling to receive. I very occasionally receive them as [efth800](https://github.com/merbanan/rtl_433/issues/1375)s, but mostly they are not detected.

![front](https://user-images.githubusercontent.com/1055976/207445578-aeffc11b-b11d-4bf3-8686-bab0b6ce5b0b.jpeg)
![back](https://user-images.githubusercontent.com/1055976/207445595-ef9c73a9-6001-43ba-9394-bd5ba594a51a.jpeg)
![inside](https://user-images.githubusercontent.com/1055976/207445613-29a2b2e3-e243-4f07-84f4-6034f5727815.jpeg)

I have captured some signals, which when I decode them seem to be four packets in a ABAB format. I'm not sure what A is, but B is a valid efth800 packet.

[cu8 files.zip](https://github.com/merbanan/rtl_433/files/10222231/cu8.files.zip)

Here's [a link to a decoding](https://triq.org/pdv/#AAB00F060101E403D002DC00F412F827648155+AAB00F060301E403D002DC00F412F827649155+AAB00F060101E403D002DC00F412F827649455+AAB053060101E403D002DC00F412F82764A2A2A2A28383B083B083B0B0838383B0B0B0B083B083B083B08383B0B08383B083838383B083B083B083B0B0B0B083B08383B0B0B08383838383B0B08383B083B083B083B455+AAB043060101E403D002DC00F412F82764A2A2A2A2B083B083B083B0B0838383B0B0B0B0B0B083B0B08383B0B0B083B0B083B0B0B0B083B0B083B0B083B083B0B0B0B0B083B455+AAB053060101E403D002DC00F412F82764A2A2A2A28383B083B083B0B0838383B0B0B0B083B083B083B08383B0B08383B083838383B083B083B083B0B0B0B083B08383B0B0B08383838383B0B08383B083B083B083B455+AAB043060101E403D002DC00F412F82764A2A2A2A2B083B083B083B0B0838383B0B0B0B0B0B083B0B08383B0B0B083B0B083B0B0B0B083B0B083B0B083B083B0B0B0B0B083B555) of one of them.

[...]

PR #2278, opened 2022-12-15
Add support for longer EFTH800 messages

andrewjw:
This adds support for the longer EFTH800 messages in #2275. It passes all the tests, but I've yet to run it for real so probably not ready to be merged yet.

### DIFF
--- a/src/devices/efth800.c
+++ b/src/devices/efth800.c
@@ -62,7 +62,7 @@ static int eurochron_efth800_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     row = bitbuffer_find_repeated_row(bitbuffer, 2, 48);
 
     if (row < 0) // repeated rows?
-        return DECODE_ABORT_EARLY;    
+        return DECODE_ABORT_EARLY;
 
     if (bitbuffer->bits_per_row[row] > 49) // 48 bits per row?
         return DECODE_ABORT_LENGTH;
--- a/src/devices/efth800.c
+++ b/src/devices/efth800.c
@@ -50,7 +50,10 @@ static int eurochron_efth800_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     if (row < 0) // repeated rows?
         return DECODE_ABORT_EARLY;
 
-    if (bitbuffer->bits_per_row[row] > 49) // 48 bits per row?
+    while (row < bitbuffer->num_rows && (bitbuffer->bits_per_row[row] < 48 || bitbuffer->bits_per_row[row] > 49)) { // 48 bits per row?
+        row++; 
+    }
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Another idea to mostly keep the row search intact:
After the `row = bitbuffer_find_repeated_row(bitbuffer, 2, 48);` check if the row is not the one we are interested in.
If so set `bitbuffer->bits_per_row[row] = 0; // remove this row from candidates` and again run `row = bitbuffer_find_repeated_row(bitbuffer, 2, 48);`

zuckschwerdt:
Sorry, more changes. I think this is the most obvious way to clear out the long rows and keep the logic of one simple `bitbuffer_find_repeated_row()`

Just this block as first code in the decoder, no other changes needed then.

```
    // Remove long rows with unknown data
    for (row = 0; row < bitbuffer->num_rows; ++row) {
        if (bitbuffer->bits_per_row[row] > 49) {
            bitbuffer->bits_per_row[row] = 0; // cancel row
        }
    }
```

zuckschwerdt:
And add this comment between: `(0-bit) 250 us pulse, 480 us gap.` and `Data layout:`

```

There might be an alternative (longer) packet interleaved, e.g.:

    {65} 2B 1E A9 90 AB D3 83 2A 8
    {49} AB 1F B3 B7 B6 BE 80
    {65} 2B 1E A9 90 AB D3 83 2A 8
    {49} AB 1F B3 B7 B6 BE 8

```

andrewjw:
Thanks for the suggestions, I've applied them and will test them out for a few days. Seems to be working fine though.

With the earlier code one of my four devices dropped out for about 24h, but then reappeared. Unfortunately I didn't get time to capture unknown signals in that time. Might just have been a reception issue though. If it happens again I'll try to investigate.

merbanan:
@zuckschwerdt will this work ? are we not destroying the bitbuffer for other possible decoders with lower priority?

zuckschwerdt:
Yes perfectly fine ;) Each decoder gets a fresh bitbuffer always. E.g. `bitbuffer_invert()` would be a problem otherwise.

merbanan:
True. LGTM then :)

zuckschwerdt:
Ideally we would understand and parse the longer messages. They should contain the same data really.

Maybe @andrewjw can be persuaded to log the long and short codes side-by-side for a few days and share a list?
A text file with (unique) lines similar to `{65} 2B 1E A9 90 AB D3 83 2A 8    {49} AB 1F B3 B7 B6 BE 8`

andrewjw:
Don't worry, I'm easily persuaded. I'll see what I can do over the Christmas period 😃

PR #2289, opened 2022-12-25
Change rtl_433_mqtt_hass to enable passing custom topics

unverbraucht:
I like to set up my rtl_433 to disregard id, so that when I change the batteries on my sensors they don't turn into new devices in Home Assistant or get new MQTT topics. To this end I use a MQTT device topic of `rtl_433/<hostname>/devices[/type][/model]/C[channel:0]`.

rtl_433_mqtt_hass expects the default device topic, so I refactored it to accept a device topic as CLI argument, using the rtl_433 default as it's own default. It also derives device identifiers from the same topics, so if you exclude id it will also not use it as part of the device id.

It works well in my test so I thought I'd upstream it. I didn't test the secret_knock (lacking the hardware) but looks fine to me. I'm happy to


zuckschwerdt:
Thanks. Niche but LGTM. @rct can you have a look?

gdt:
Appears conceptually ok to me.   Would like to see comment with rationale that not passing -T gets the old behavior, which I think it does but am not entirely sure.

unverbraucht:
Thanks! @gdt the default of -T is also the default way in which rtl_433 creates topic paths: `devices[/type][/model][/subtype][/channel][/id]`. This way not passing anything for -T should get you the default paths, which it does in my testing.

gdt:
Thats great but it would be nice to have a comment that explains the intent.   I realize I am +1 to 2 sdev on asking for design comments...

rct:
I need to take a closer look tomorrow, but it seems like this would only be useful if you are only modifying the behavior for one model.

I have a bunch of different models, so I've wound up with hand edited names/identifier strings and hand edited topics for Hass to subscribe to.

I originally went down this path with the Ambient Weather F007TH because ID is volatile and there is a DIP switch for settings channel number 1-8.  For a while I used `/+/` for ID.   I eventually removed the wild cards because it let too much bad data through.

For some of my Acurite devices, I've needed to use `/+/` in the place of the A/B/C channel number. because moisture would get in the physical switch and the weather station would flop from A to C and back.

So now when I change batteries, I edit the MQTT definitions.  For me it is better than allowing bad data.

For me a useful implementation to have model specific overrides for which keys are used.

unverbraucht:
@rct for my sensors it works well. I have tested the following:

- Bresser 6-in-1 weather station.
- Auriol AHFL
- Lacrosse TX
- Acurite 606TX
- Infactory PT 310
- Nexus TH

In the week of running without ids in my topics I have not noticed random data, at least not in the HA charts. So for me this implementation is very useful, it fixes the issue without forcing me to add and manage all MQTT data sources to the HA config.

I agree that managing this per model would be better, but for me this doesn't fix anything that the current PR wouldn't fix. And wouldn't this also require changes to the rtl_433 config? What I like about this approach is that it takes an option that rtl_433 already offers (customizing topic paths) and brings it also to the rtl_433 mqtt hass binding. Currently a custom topic path makes the bindings unusable, so to me this PR is more of a fix than a feature.

Regarding the Acurite channel situation you had: this would not be affected by users removing ids from the topic path, so while this change here wouldn't fix it, it also wouldn't make it worse.

unverbraucht:
> Thats great but it would be nice to have a comment that explains the intent. I realize I am +1 to 2 sdev on asking for design comments...

@gdt I'm not sure I follow, do you mean the intent behind opening this PR? Or the intent behind the default for -T being the default for rtl_433?

gdt:
No, I mean that there are assumptions about the new code's default for -T matching the default for something else, but when one reads the code that is not explained.   Just to add a comment (that will live with code, not in the PR) that explains that big picture.

gdt:
I think we should merge this.  I am guessing rct@ would agree that this is incrementally ok even if it doesn't implement a feature he wants, but I'm not 100% sure I read that right.  I'll wait 48h for objections and then hit merge.

rct:
I don't object to this being merged, though I haven't tested it.

I wasn't happy to see the explicit definition of NAMING_KEYS removed. 

However, I think the cases where this would be useful is pretty limited. @unverbraucht mentions using this with a number of sensors including LaCrosse TX.  The only identifier that device has is the volatile ID that changes at start up.  So unless I'm missing something. the only case where this could be useful is if you only have a **single**LaCrosse TX?   (For reference I've have about 4 of those currently active, down from 6 as two have died.)

If this change went further and allowed the naming/topic string to be customized _on a per-model basis_, that would potentially be extremely useful. 

Note: I haven't done nay more work on this script, because I this it needs a major overhaul. It has too many limitations working on a per field, rather than model basis.

zuckschwerdt:
Please always squash on merge! I'm going to force push a squash now.

gdt:
Sorry, forgot that this project is different.  I'm used to a world where projects expect that PRs have been separated into logical commits (1 being ok if that's how it is) with no self-merges and rebased to clean things up.

zuckschwerdt:
Yes, rebase is the important thing. There are very few PRs that benefit for multiple commits, but those would be okay if they are rebased as fast-forward merge (most times easier to use multiple PRs though).

The idea is to present a nice and compact changelog with the commit history. There is next to no useful information in when/where a branch was started and how many changes it took to finish a feature.

PR #2290, opened 2022-12-26
CMakeLists.txt: use pkg-config to detect openssl when possible

ffontaine:
In order to take into account the libraries used by openssl when building statically, using pkg-config is recommended. This patch therefore improves the CMakeLists.txt to use pkg-config to detect openssl when pkg-config is available. This will avoid the following static build failure when openssl needs to link with `-latomic`:

```
/home/autobuild/autobuild/instance-13/output-1/host/lib/gcc/sparc-buildroot-linux-uclibc/10.4.0/../../../../sparc-buildroot-linux-uclibc/bin/ld: /home/autobuild/autobuild/instance-13/output-1/host/sparc-buildroot-linux-uclibc/sysroot/usr/lib/libssl.a(ssl_cert.o): in function `ssl_cert_free': ssl_cert.c:(.text+0x384): undefined reference to `__atomic_fetch_sub_4'
```

Fixes:
 - http://autobuild.buildroot.org/results/417c86963ffe038aa052ea3cf19fd52c3e9b7396

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


zuckschwerdt:
That's interesting, I've seen atomics trouble (in SoapySDR) before, on ARM IIRC. Not really clear when `-latomic` is needed.
But if OpenSSL can let us know using pkg-config then that's simple enough ;)

Not sure, but don't we need the `find_path`/`find_library` path magic usually in a cmake module?

zuckschwerdt:
I looks like pkg-config is already used in the official cmake module? https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake#L109

And we should get "All OpenSSL libraries and their dependencies."

ffontaine:
`-latomic` is needed on architectures such as sparc v8 and some flavors of ARM.

Unfortunately, official cmake modules don't really handle static building, instead of just using what is returned by `pkg-config --static --libs openssl` , it tries (and sometimes fails) to outsmart pkg-config. For example, cmake will only handle the standard / usual openssl dependencies such as `-lz`: https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake#L128.

zuckschwerdt:
We'd also need to change to `OPENSSL_LINK_LIBRARIES` in `list(APPEND SDR_LIBRARIES ${OPENSSL_LIBRARIES})`
or add the path(s) form `OPENSSL_LIBRARY_DIRS`, afaics?
(FindOpenSSL.cmake's OPENSSL_LIBRARIES is absolute and [FindPkgConfig](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html)'s is names only).

ffontaine:
`pkg_check_modules` will indeed only return the libraries and the not the full path in `OPENSSL_LIBRARIES`. The full library paths will be set in `OPENSSL_LINK_LIBRARIES`. I'll update the PR and check if it works well on buildroot.

zuckschwerdt:
There might be more incompatibilities still.

There is `OPENSSL_INCLUDE_DIR` in FindOpenSSL but pkg_check_modules sets
`OPENSSL_INCLUDE_DIRS` and `OPENSSL_INCLUDEDIR` only.

I guess we could use e.g. `set(OPENSSL_LIBRARIES ${OPENSSL_LINK_LIBRARIES})` and
`set(OPENSSL_INCLUDE_DIR ${OPENSSL_INCLUDEDIR})` (not sure if we might rather want `OPENSSL_INCLUDE_DIRS`?)

And then pkg-config sets extra static variants, e.g. I get
`OPENSSL_LIBRARIES = ssl;crypto`
but `OPENSSL_STATIC_LIBRARIES = ssl;crypto;z`
Does `OPENSSL_(LINK_)LIBRARIES` really work for you, is it the same as `OPENSSL_STATIC_(LINK_)LIBRARIES`?

ffontaine:
In the context of buildroot, the pkg-config wrapper will add `--static` to each pkg-config call when building statically. So `OPENSSL_(LINK_)LIBRARIES` work for me.

Adding include and library directories are also handled by buildroot through `CFLAGS` and `LDFLAGS` but but I'll update the PR as you suggested.

zuckschwerdt:
Bare pkg-config doesn't work well with cross builds. Refactored with ad33087.

PR #2295, opened 2022-12-31
Add support for selecting input channel if an SDR has more than one

l1berat0r:
Some SDRs (like LimeSDR) have more than one channel and it is convinient to be able to switch between them.
The change adds support for the command line argument that allows to select the input port.

`[-N <channel number> If the device supports more than one channel, then it may be sellected (default 0)`

I needed the change for myself so I implemented it and if it's possible I would also like to merge it to the main fork. 

This is my first contribution to this project. I tried to follow all the conventions that I was able to observe when reading the code. With some of my changes I needed to modify few `soapysdr_` function signatures but it didn't affect any tests (all are passing).
If I missed anything please let me know and I will fix my code.

Kind Regards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
   [-H <seconds>] Hop interval for polling of multiple frequencies (default: 600 seconds)
   [-p <ppm_error>] Correct rtl-sdr tuner frequency offset error (default: 0)
   [-s <sample rate>] Set sample rate (default: 250000 Hz)
+  [-N <channel number>] If the device supports more than one channel, then it may be sellected (default 0)
 		= Demodulator options =
   [-R <device> | help] Enable only the specified device decoding protocol (can be used multiple times)
        Specify a negative number to disable a device decoding protocol (can be used multiple times)
--- a/include/rtl_433.h
+++ b/include/rtl_433.h
@@ -10,6 +10,7 @@
 #include <time.h>
 #include <signal.h>
 
+#define DEFAULT_CHANNEL         0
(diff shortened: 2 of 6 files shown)


zuckschwerdt:
Interesting (and useful) idea, thanks! I'm not sure I would put the channel along with rate and freq. It's not something we control ongoing. I would see it with `dev_query`, i.e. "once on setup" parameter.

And maybe we could put this in `settings_str` / `-t` along with e.g. antenna and bandwidth. Likely we'd need to extract it before opening the device though.

l1berat0r:
You are probably right. One scenario I can think of is, that someone would like to use different antennas with different bands, but then the same would apply to the `antenna` parameter which is indeed configured with the `settings_str`. 
Regardless of the future ideas then, the first step should be to have this two parameters in the same place.

In the change I commited, I modified the original:
`int sdr_open(sdr_dev_t **out_dev, char const *dev_query, int verbose);`
with:
`int sdr_open(sdr_dev_t **out_dev, uint32_t channel, char const *dev_query, int verbose);`

Following your suggestion, it should rather be:
`int sdr_open(sdr_dev_t **out_dev, char const *sdr_settings, char const *dev_query, int verbose);`

Then the `sdr_open_*` functions can extract whichever parameters they need. Am I correct?

zuckschwerdt:
Yes. But I'm also poking another idea: the `dev_query` could include the channel arg as Soapy modules usually ignore extra args. It all depends on how we think about the channel: is it something that (in our application) is integral with the device selection? I lean towards: a setting like antenna that, in theory, we could want to change.

merbanan:
No update for several years. Closing.

Issue #2288, opened 2022-12-23
New TFA 14.1504.V2 sensor not supported by any decoders


joel-bourquard:
**EDIT:** **Turns out that this sensor is a V2 and due to that, has actually nothing to do with the Maverick, so it will likely require a new decoder. So the first few posts contain some erroneous assumptions.**
---

I bought a new TFA 14.1504 BBQ sensor that reportedly uses the same protocol as "[71]  Maverick ET-732/733 BBQ Sensor", according to this discussion:
https://github.com/RFD-FHEM/RFFHEM/issues/167

The TFA manual that comes with my device confirms that it's 433Mhz, and I have several other 433Mhz sensors working with rtl_433 already.

I built rtl_433 from HEAD (master). Yet, nothing gets picked up from the TFA.

I tried some fiddling such as:
`./rtl_433-s 1024k -Y autolevel -Y magest -M noise -M level -Y minmax`

which picks up more faraway sensors from my neighborhood, but nothing from the TFA that is a few meters from my antenna.

The discussion above also links to a commit in a different project that adds support from TFA 14.1504 to their Maverick ET-732/733 decoder, maybe that could give us *very* good hints?
https://github.com/RFD-FHEM/RFFHEM/commit/b50c779be083dded778397f55238eb0500091f15

In rtl_433, I tried adding some logs to maverick_et73x.c:

```
static int maverick_et73x_callback(r_device *decoder, bitbuffer_t *bitbuffer)
{
    data_t *data;
    bitbuffer_t mc = {0};

    if (bitbuffer->num_rows != 1)
        return DECODE_ABORT_EARLY;

    //check correct data length
    if (bitbuffer->bits_per_row[0] != 104) // 104 raw half-bits, 52 bits payload
        return DECODE_ABORT_LENGTH;

    //check for correct preamble (0x55666a)
```

From what I saw, the data length is never equal to 104 and thus the function exits with DECODE_ABORT_LENGTH.  I don't understand how it can abort so early, if the protocol is really similar to Maverick ET-73x.

I would be happy to help but I don't know what to do.

zuckschwerdt:
Can you grab a sample and upload as zip here? https://triq.org/rtl_433/ANALYZE.html

Your debugging is spot on. Use `bitbuffer_print(bitbuffer);` (or `bitbuffer_debug`) or e.g. even `decoder_log_bitbuffer(decoder, 0, __func__, bitbuffer, "before lenght check")` to get some insight.

[...]

PR #2296, opened 2023-01-02
Add support for TFA Dostmann 14.1504 Radio-controlled grill and meat thermometer (#2288)

joel-bourquard:


src/devices/tfa_14_1504.c
```diff
@@ -0,0 +1,199 @@
...
+    Copyright (C) 2022-2023 Joël Bourquard <joel.bourquard@gmail.com>
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+/**
+Decoder for temperature sensor model no. TPX305
```
zuckschwerdt, line 14: This should include (replicate) line 2 infos.

zuckschwerdt:
That checksum looks accidental. A wide enough CRC can always be produced for just a few bits.
Are you sure it's a CRC? Maybe it's a digest (LFSR)? Can you share the list of captured codes you derived this from?

src/devices/tfa_14_1504.c
joel-bourquard: Done! Thanks

joel-bourquard:
> That checksum looks accidental. A wide enough CRC can always be produced for just a few bits. Are you sure it's a CRC? Maybe it's a digest (LFSR)? Can you share the list of captured codes you derived this from?

I know, I can tell you that I tried a few exhaustive searches with the 16-bit LFSR functions from util.c (rtl_433), more precisely the lfsr_digest16() function with all possible combinations for "gen" and "key", and applied that to all possible substrings... with no success there. I was very surprised/frustrated too, but maybe I missed something obvious.

Sure, here are all samples along with the tool to compare the computed CRCs against them. You can simply clone this and run "./build.sh":
https://github.com/joel-bourquard/tfa_tpx305_reveng

zuckschwerdt:
Did you try our [revdgst16](https://github.com/triq-org/revdgst/)? It finds gen 8810 key 93ad final xor d17c when you omit the preamble.
0x8810 is *the* gen used everywhere, which sounds reassuring, the final xor could mean that we picket up some sync that shouldn't be used -- not sure.

joel-bourquard:
> Did you try our [revdgst16](https://github.com/triq-org/revdgst/)? It finds gen 8810 key 93ad final xor d17c when you omit the preamble. 0x8810 is _the_ gen used everywhere, which sounds reassuring, the final xor could mean that we picket up some sync that shouldn't be used -- not sure.

Aah awesome, thanks a lot! No I didn't know that tool, works great indeed. The reason my prior brute force attempts didn't work out, was that final XOR.

I've tried a few things but no zero final XOR so far - - so I've updated my PR with the LSFR and the values that you mentioned - which I think are the most suitable so far.

src/devices/tfa_14_1504.c
```diff
@@ -0,0 +1,132 @@
...
+        return DECODE_FAIL_SANITY;
+    }
+
+    uint16_t calc_mic = lfsr_digest16(data, OFFSET_MIC, 0x8810, 0x93ad) ^ 0xd17c;
+    uint16_t data_mic = (data[OFFSET_MIC] << 8) + data[OFFSET_MIC+1];
+    if (calc_mic != data_mic) {
+        return DECODE_FAIL_MIC;
+    }
+
+    uint16_t raw_temp_c = ((data[1] & 0xf) << 6) + (data[2] >> 2);
```
zuckschwerdt, line 99: Are you shifting off the last 2 bits to avoid a div 4? Are those never used, i.e. is the precission not 1/4 deg C?

zuckschwerdt:
If possible I wouldn't use the full preamble but just a tail portion. This will improve reception if the signal isn't immediatly demodulated perfectly, how typical receivers work to "tune-in". That scheme needs a "sync" at the end to known when data starts, could be a single bit, but is usually aligned. So if `5c` is hard coded to be matched anyway then you can use a robust preamble of e.g. `aaaa5c`.

src/devices/tfa_14_1504.c
joel-bourquard: > Are you shifting off the last 2 bits to avoid a div 4? Are those never used, i.e. is the precission not 1/4 deg C?

Yes exactly: for each degree C, the value increases by 4. The two lower bits are always zero, as far as I've seen (and it's also seen in the samples.txt file). Also the display on the device itself does not have a fractional part.
For context, it's a barbecue thermometer with a high temperature probe which goes up to 300 degrees C, but is probably not precise enough to justify having a fractional part...

zuckschwerdt: Ok, good. Add a `// stripping off last 2 bit as those are always zero` or such ;)  But a float div 4 shoould not be expensive either, as it likely only subs 2 from the exponent, no mantissa manipulation needed.

joel-bourquard:
> If possible I wouldn't use the full preamble but just a tail portion. This will improve reception if the signal isn't immediatly demodulated perfectly, how typical receivers work to "tune-in". That scheme needs a "sync" at the end to known when data starts, could be a single bit, but is usually aligned. So if `5c` is hard coded to be matched anyway then you can use a robust preamble of e.g. `aaaa5c`.

Good point thanks, I will do that!

Question about the Probe status: currently I set it to "Connected" / "Disconnected", is this good? Or should I use a different upper/lowercase convention, or an int (0/1) for that?

zuckschwerdt:
> Question about the Probe status: currently I set it to "Connected" / "Disconnected", is this good? Or should I use a different upper/lowercase convention, or an int (0/1) for that?

I would use a 0/1 int for easier transfer into databases (int supports aggregation), like the common `battery_ok`, so `probe_ok`. Or perhaps even invert the output and name it `probe_fail`? We don't have a common scheme for these "event like" data keys.

src/devices/tfa_14_1504.c
joel-bourquard: Done!

joel-bourquard:
> > Question about the Probe status: currently I set it to "Connected" / "Disconnected", is this good? Or should I use a different upper/lowercase convention, or an int (0/1) for that?
> 
> I would use a 0/1 int for easier transfer into databases (int supports aggregation), like the common `battery_ok`, so `probe_ok`. Or perhaps even invert the output and name it `probe_fail`? We don't have a common scheme for these "event like" data keys.

Good point! I think probe_fail is the best option, and pushed the change now!
(user-friendly text is "Probe failed").

joel-bourquard:
Battery flag implemented and works well. No more changes planned, so if you can consider it for inclusion, it would be great 😉
If any other changes are advised or if I should squash, please let me know...

joel-bourquard:
Hi @zuckschwerdt,
I've been run-testing the decoder for a few days now, under challenging SNR conditions, and it's working great so far. No other changes planned for me at this point.

What do you think about merging my PR?

I'm also looking forward to getting feedback from other owners of this device, to validate whether all LSFR params are identical or not. If they are not, then I have some ideas on how to address it.

zuckschwerdt:
Yes, looks good. The "mic" should be "CRC" as it's the same strength.
And "model" should be simpler, e.g. "TFA-141504v2" (only one dash in the middle, no other special chars).

joel-bourquard:
> Yes, looks good. The "mic" should be "CRC" as it's the same strength. And "model" should be simpler, e.g. "TFA-141504v2" (only one dash in the middle, no other special chars).

Great, thanks! I've implemented the exact changes that you suggested and squashed my commits. Ready to merge for me 😉

joel-bourquard:
Hi @zuckschwerdt ,
Great, thanks! I've implemented the exact changes that you suggested and squashed my commits. Ready to merge for me 😉

joel-bourquard:
Hi @zuckschwerdt,
Do you know when my PR can be merged? Or should I ping someone?
Thanks 😉

src/devices/tfa_14_1504_v2.c
```diff
@@ -0,0 +1,141 @@
...
+#define NUM_BITS_DATA     (NUM_BYTES_DATA * 8)
+#define NUM_BITS_TOTAL    (NUM_BITS_PREAMBLE + NUM_BITS_DATA)
+#define NUM_BITS_MAX      (NUM_BITS_TOTAL + 12)
+
+static int tfa_14_1504_v2_callback(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    uint8_t const preamble[] = {
+            0xaa,
+            0xaa,
+            0x5c,
```
merbanan, line 57: Less lines if all bytes are in a row.

src/devices/tfa_14_1504_v2.c
```diff
@@ -0,0 +1,141 @@
...
+    uint16_t calc_mic = lfsr_digest16(data, OFFSET_MIC, 0x8810, 0x0d42) ^ 0x16eb;
+    uint16_t data_mic = (data[OFFSET_MIC] << 8) + data[OFFSET_MIC+1];
+    if (calc_mic != data_mic) {
+        return DECODE_FAIL_MIC;
+    }
+
+    // we discard the last 2 bits as those are always zero
+    uint16_t raw_temp_c = ((data[0] & 0xf) << 6) + (data[1] >> 2);
+    unsigned is_probe_connected = (raw_temp_c != 0x1c0);
+    double temp_c = ((int)raw_temp_c) - 532;
```
merbanan, line 108: Should be float, besides that

merbanan:
LGTM, but I think we we need docs and config changes also. And then some sample signals.

merbanan:
@joel-bourquard 
Did you upload any signal samples to the test repo ? I can not see any pull request.

https://github.com/merbanan/rtl_433_tests/pulls

merbanan:
In this commit you can see the needed config and documentation changes.

https://github.com/merbanan/rtl_433/commit/cd90897b7f7359423058f392c9e786d96ca4c3b8

src/devices/tfa_14_1504_v2.c
joel-bourquard: Hi @merbanan , thank you!   Change applied: https://github.com/merbanan/rtl_433/pull/2296/commits/d145440875ebbc91f26b5ceb329072f81a18f51a

src/devices/tfa_14_1504_v2.c
joel-bourquard: Change applied: https://github.com/merbanan/rtl_433/pull/2296/commits/d145440875ebbc91f26b5ceb329072f81a18f51a

joel-bourquard:
> In this commit you can see the needed config and documentation changes.
> 
> [cd90897](https://github.com/merbanan/rtl_433/commit/cd90897b7f7359423058f392c9e786d96ca4c3b8)

Should be all done, thanks!  => https://github.com/merbanan/rtl_433/pull/2296/commits/d145440875ebbc91f26b5ceb329072f81a18f51a

joel-bourquard:
> @joel-bourquard Did you upload any signal samples to the test repo ? I can not see any pull request.
> 
> https://github.com/merbanan/rtl_433_tests/pulls

@merbanan not yet, but I will definitely do it!  I'll fork that repo and do a PR.

joel-bourquard:
Hi @zuckschwerdt , @merbanan ,
All ready on the rtl_433 side, I think :-)

So I have created a PR with documentation+samples on the rtl_433_tests side:
https://github.com/merbanan/rtl_433_tests/pull/445

joel-bourquard:
Rebased on latest master.

joel-bourquard:
Hi @zuckschwerdt ,
Rebased on latest master.  Would it be possible for you to merge it please? :)

zuckschwerdt:
Just a few other (~3) merges first.

PR #2305, opened 2023-01-09
fix: some Cotech-367959 clones do not have the light sensor installed.

obones:
I have a NicetyMeter 0366 weather station here that gets picked up as Cotech-367959 just fine.
The only thing is that Light intensity and UV index values are never changing and completely bogus which is quite expected as this model does not have a light sensor.
To avoid confusion, this PR removes the values from the output if it finds them to match what has been observed with my device.

If need be, I can create a PR on the tests projects with a sample signal final.


zuckschwerdt:
Thanks, this is a good sanity check.
Can you rewrite the change to simply
`
"uv",               "UV Index",         DATA_COND, ((uv & 0xf0) == 0), DATA_FORMAT, "%u", DATA_INT, uv,
`
Easier to see that it's a single output with one optional field then.

obones:
I've done just that. My initial version was also changing the model field value, but I guess it's not desirable and I can live with it being static anyway.

PR #2308, opened 2023-01-11
Fix HCS200 serial decoding

proger10:
Seems like most significant bits of serial (reverse8(b[7] & 0xf0) must be shifted by 24 (not 20) to get correct 28-bit serial


zuckschwerdt:
Ohh, I see, the `reverse8()`. Ok.

Issue #2298, opened 2023-01-06
Could not find procedure/function entry point "rtlsdr_cancel_async" from rtl_433-win-x64-22.11.zip


mdeweerd:
I extracted the official release rtl_433-win-x64-22.11.zip and try to run

```
rtl_433 -d ""
```
to use my HackRF One on Windows 11 with rtl_433.  The static build does not throw this exception but it does not provide soapysdr.

![image](https://user-images.githubusercontent.com/1504752/210924704-e162d177-f4e6-4600-b901-57119401a4f2.png)

CubicSDR from PothosSDR finds the device:
![image](https://user-images.githubusercontent.com/1504752/210925111-81ff3cb5-d8b6-4c58-93e1-aebd02222167.png)

I tried with the libraries from PothosSDR-2021.07.25-vc16-x64.exe as well with no success (I checked that those were loaded by tracking the files used by rtl_433.exe during runtime).

mdeweerd:
I tried with [rtl_433-win-msvc-x64-22.11.zip](https://github.com/merbanan/rtl_433/releases/download/22.11/rtl_433-win-msvc-x64-22.11.zip) which does not produce this error.  So I'll stick with that.

zuckschwerdt:
Thanks for the report. There was similar trouble in #2251. So far not sure what the actual problem is, sorry.

The verbose message problem is addressed in the next release.

mdeweerd:
FYI: I tried the new binaries posted in #2251 : no error popup about rtlsdr_cancel_async.
So that is better.
However, hackrf is not found there neither.

I also tried to build with (updated) Visual Studio's 2017 and 2019 but I get this message:
![image](https://user-images.githubusercontent.com/1504752/210998945-849398a4-0fd8-4215-819c-4fe1e2dfd650.png)

Should I use another PothosSDR version than the one in BUILDING.md ?

zuckschwerdt:
We have no file `sched.h` but I guess this originates with `THREAD_RETURN` maybe?

Re HackRF: you did install Pothos and e.g. `SoapySDRUtil` can find the SDR?

mdeweerd:
This shows where sched.h is located (PothosSDR) :
![image](https://user-images.githubusercontent.com/1504752/211023620-133fb1c1-4d2b-43ea-a994-441f1f54ed9e.png)

Checking the headers, I find:
c:\Program Files\PothosSDR\include\pthread.h:#include <sched.h>


```
[...]

PR #2310, opened 2023-01-13
Fix MSVC build

Mikayex:
PR #2291 mistakenly include pthread for all platforms instead of only Unix platforms.
On Visual Studio 2022, the following error occurs when including mongoose.h:
```
C:\Program Files\PothosSDR\include\sched.h(129,13): error C2040: 'HANDLE': 'int' differs in levels of indirection from 'void *'
```
It tries to use pthread.h from PothosSDR but there is a conflict on the definition of pid_t.

This PR restore the previous behavior: only include pthread on Unix platforms and thus, fix Windows build.

### DIFF
--- a/include/mongoose.h
+++ b/include/mongoose.h
@@ -4657,7 +4657,9 @@ size_t mg_fwrite(const void *ptr, size_t size, size_t count, FILE *f);
 #endif /* MG_ENABLE_FILESYSTEM */
 
 #if MG_ENABLE_THREADS
+#if CS_PLATFORM == CS_P_UNIX
 #include <pthread.h>
+#endif
 /*
  * Starts a new detached thread.
  * Arguments and semantics are the same as pthead's `pthread_create()`.


zuckschwerdt:
Oh, nice find. This will address #2298 also.

But the fix is possibly wrong. We don't actually use `MG_ENABLE_THREADS` but it is is force-enabled for `_WIN32` to support `mg_start_process()` which I'm not sure is used at all?

So the guard here should be `#ifndef _WIN32` -- but we would want pthreads on Win32.
Best to uncomment the include and add what you found:
`// possibly needed but disabled because of conflict with PothosSDR on the definition of pid_t.`

Mikayex:
I didn't saw #2298 (the error was in an image so can't find it).

For the `#if CS_PLATFORM == CS_P_UNIX` I used the exact same condition that was before #2291. pthread is actually not needed on Windows (for Mongoose at least). See the definition of `mg_start_process()`: https://github.com/merbanan/rtl_433/blob/2cd5623c9ac65d49a542420ce91ac33ea645f32d/src/mongoose.c#L10557-L10580
Also, there is no other `#include <pthread.h>` in the header file, only the one that would be for Unix platform.

If I understand correctly your suggestion, you want something like that ?
```c
#if MG_ENABLE_THREADS
#ifndef _WIN32
#include <pthread.h> // possibly needed but disabled because of conflict with PothosSDR on the definition of pid_t.
#endif
/*
 * Starts a new detached thread.
 * Arguments and semantics are the same as pthead's `pthread_create()`.
 * `thread_func` is a thread function, `thread_func_param` is a parameter
 * that is passed to the thread function.
 */
void *mg_start_thread(void *(*thread_func)(void *), void *thread_func_param);
#endif
```

zuckschwerdt:
Thanks for the explanation. You are right, this guard is what was there before #2291. Not sure what this is about then? Doesn't matter though :)
https://github.com/merbanan/rtl_433/blob/2070798d5e9bfc37e47e350475aaa466282281c5/include/mongoose.h#L3650-L3656

PR #2312, opened 2023-01-13
Added additional digit for rain gauge

matthias-bs:
As pointed out by  @aanban in https://github.com/matthias-bs/BresserWeatherSensorReceiver/issues/22, there is an additional digit for the rain gauge in the high nibble of msg[24].

This has been confirmed with [WeatherSensor.cpp](https://github.com/matthias-bs/BresserWeatherSensorReceiver/blob/main/src/WeatherSensor.cpp).


zuckschwerdt:
Thanks @matthias-bs for porting back ;)

PR #2316, opened 2023-01-15
Fix json keys for wmbus to be unique

jedi7:
* Key inst_volume_0 is now unique

Fast fix. Will be nice to find a better solution to provide variable arguments to the append_str.

Previous keys:
"inst_volume_0" : "6.257 m3", "inst_volume_0" : "6.116 m3", "inst_volume_1" : "5.764 m3", "inst_volume_2" : "5.494 m3", 
...

New keys:
"inst_volume_0" : "6.257 m3", "inst_volume_m1_0" : "6.116 m3", "inst_volume_m2_1" : "5.764 m3", "inst_volume_m3_2" : 
...

### DIFF
--- a/src/devices/m_bus.c
+++ b/src/devices/m_bus.c
@@ -280,18 +280,24 @@ static char *unit_names[][3] = {
 static double pow10_table[8] = { 0.001, 0.01, 0.1, 1, 10, 100, 1000, 10000 };
 
 
-static data_t *append_str(data_t *data, enum UnitType unit_type, uint8_t value_type, uint8_t sn, char const *extra, char const *value)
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks, looks great. It might be better to safe-guard
`if (key_extra[0] == '\0')` and `if (pretty_extra[0] == '\0')`
as `if (!key_extra || !*key_extra)` and `if (!pretty_extra || !*pretty_extra)`
because we do allow at least pretty_key to be NULL safely.
(this `!x || !*x` is the idiom we typically use for "is empty".)

jedi7:
updated :) thanks for review

PR #2319, opened 2023-01-16
Add support for CED7000 timer

ppapadeas:
Test data in https://github.com/merbanan/rtl_433_tests/pull/441/files

Signed-off-by: Papadeas Pierros <pierros@papadeas.gr>

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [230]  Vauno EN8822C
     [231]  Govee Water Leak Detector H5054
     [232]  TFA Dostmann 14.1504.V2 Radio-controlled grill and meat thermometer
+    [233]* CED7000 Shot timer
 
 * Disabled by default, use -R n or a conf file to enable
 
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -454,6 +454,7 @@ stop_after_successful_events false
   protocol 230 # Vauno EN8822C
   protocol 231 # Govee Water Leak Detector H5054
   protocol 232 # TFA Dostmann 14.1504.V2 Radio-controlled grill and meat thermometer
+# protocol 233 # CED7000 Shot timer
 
 ## Flex devices (command line option "-X")
 
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -240,6 +240,7 @@
     DECL(vauno_en8822c) \
     DECL(govee_h5054) \
     DECL(tfa_14_1504_v2) \
+    DECL(ced7000) \
 
     /* Add new decoders here. */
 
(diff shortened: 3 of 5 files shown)


zuckschwerdt:
Looks good, notes:
- you should loop all rows and find a / the long one, not hardcode `1`
- you should search the sync word, not assume it's at a fixed pos
- use one of `reverse8(), reflect_bytes(), reflect4(), reflect_nibbles()`
- you need to extract values as numbers not strings

E.g.
```
    uint8_t const sync_pattern[3] = {0xaa, 0x4d, 0x5e};

    // Find row
    int row = bitbuffer_find_repeated_row(bitbuffer, 1, 6*16+3*8);
    if (row < 0)
        return DECODE_ABORT_EARLY;

    // search for 24 bit sync pattern
    bitpos = bitbuffer_search(bitbuffer, row, bitpos, sync_pattern, 24) + 24;

    if ((bitpos >= bitbuffer->bits_per_row[row])
        return DECODE_ABORT_EARLY;

    bitbuffer_invert(bitbuffer);

    // Check and decode the Manchester bits
    ret = bitbuffer_manchester_decode(bitbuffer, row, bitpos, &decoded, NUM_BITS_DATA);

    ....

    int rfid = (b[1] & 0xF) * 1000 + (b[1] >> 4) * 100 + (b[0] & 0xF) * 10 + (b[0] >> 4);
```

ppapadeas:
Thanks for the review @zuckschwerdt ! Should I amend the commit or push another one on top, based on your suggestions?

zuckschwerdt:
As you like, adding commits is likely easier. We always squash on merge anyway.

ppapadeas:
@zuckschwerdt Amended following your 4 comments. Thanks for the pointers!

ppapadeas:
@zuckschwerdt let me know when adding this is next-in-line so I can rebase properly (to make your life easier ;) )

zuckschwerdt:
Rebase and correct the two line with sticky braces (`if …){`).
Good for merge then.

ppapadeas:
@zuckschwerdt done and done!

zuckschwerdt:
Perfect, thanks!
I just noticed: "rfid" is that an ID? It should be named just "id" then. All decoders have "model", "id", …

ppapadeas:
> "rfid" is that an ID?

Not really. It is a user-set parameter from the device menu to be able to "sync" the device with accessories and use multiple ones on the field.

zuckschwerdt:
Ok, so more like a "channel" settings on temperature sensors.
It would be nice if this was a random number by default, but I guess it's simply "0000"?

We really always want an "id" key to be present, this is the best we get. I kind of answers the "which (local) device is this" question. Can you document that RFID is used as ID and change the key then?

ppapadeas:
@zuckschwerdt gotcha. I changed it to ID now. Thanks!

PR #2331, opened 2023-01-20
Decode the radio clock packet (if available) for Eurochron-EFTH800 devices. (#2275)

andrewjw:


src/devices/efth800.c
```diff
@@ -41,6 +41,8 @@ Data layout:
 The sensor sends messages at intervals of about 57-58 seconds.
 */
 
+#include <stdlib.h>
```
zuckschwerdt, line 44: What for? Please remove.

andrewjw: That was left from an earlier version of the code. I've removed it.

Issue #2318, opened 2023-01-16
pretty vs raw values


jedi7:
Hi,
is there a design plan for pretty versus raw values?

Context:
For example I receive data saying power is "1000 kW". This is nice for user as visualization, but for computer processing (json, mqtt output) it is not good. There we will need raw value 1000000.

One solution is to add both keys. One key with pretty number and second key with raw value. But user will see many redundant information.

Or somehow extend the storage interface to contain these informations:
* pretty value // visible only in console
* pretty measurement unit  // visible only in console
* raw value   // visible only in json,mqtt, databases, etc

Or somehow allow visibility change by configuration parameter.

I'm able to do PR for this, at least for the WMBUS , but I will need your solution proposal :)

zuckschwerdt:
We aim to output only a raw value but converted to sensible units.
If the report is in kW or W then that scale. If a majority of devices use a specific one then we should convert. (Currently wind km/h vs m/s is undecided but should be normalized).

It might be possible to offer a conversion, e.g. `-C pretty`, but keys include the scale and unit and we should not changes those dynamically.

merbanan:
Is it only power that has this issue? To me the native unit of power is only watts.

jedi7:
It is for more units, like J, W, m3/h, etc. (see "unit_names" in m_bus.c)
Problem is, when I convert it to base unit, then I can get very hard to read numbers. 
Some of the numbers can have exponent 9 (or -6). So you will then see 1000000000 J instead of 1 GJ.

zuckschwerdt:
Agreed, we might see a kW/h report but not kW, only W (I mean "1000 kW"? is that an industrial application?)
The mbus units are defined by the spec I take it? We are just talking about the KV output then, and only for metric SI keys, and likely only a short hand-picked list? I would rather see consumers implementing this.

What is your specific use-case? Are you looking at the KV output for more than setup/debugging? Or do you want this in another output?

jedi7:
better example is with the Joule. When for heaters is normal to use GJ, but for other applications it can be really just J.
The wmbus have some predefined units and possible exponents from -6 to 9.

I like the idea of "-C pretty". So maybe we can store the base measure unit and on a common place recompute it to more readable value?

zuckschwerdt:
Ok so, for now, only "J". With J it won't be a negative scale, right? As said it would be good not to change the keys, so this should be located in the KV output at `print_kv_double()` / `print_kv_int()`.
[...]

PR #2336, opened 2023-01-22
[wmbus] use base units for json

jedi7:
This change uses format for pretty printing of the values with metric prefixes.
Values for computer processing (json, cvs,..) are kept in base units.

### DIFF
--- a/src/devices/m_bus.c
+++ b/src/devices/m_bus.c
@@ -16,6 +16,7 @@ Wireless M-Bus protocol. Will return a data string (including the CI byte)
 for further processing by an Application layer (outside this program).
 */
 #include "decoder.h"
+#include <math.h>
 
 #define BLOCK1A_SIZE 12     // Size of Block 1, format A
 #define BLOCK1B_SIZE 10     // Size of Block 1, format B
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Discussed in #2318

Best not to just do this for one single decoder. Wrap this around 
https://github.com/merbanan/rtl_433/blob/e37fc5dd3854ca7e95099a9cf6410357701c3089/src/output_file.c#L364

Match the format (of e.g. ...`_J`) then extract an exponent and rewrite `data` and `format`.
(e.g. have it result in ...`_kJ`, `data / 1e3` or  ...`_MJ`, `data / 1e6`)

jedi7:
I wanted to do it more common, but it means too many changes in code around and changes in design. Problems like where to store the measurement unit; how to deal with the variadic parameters for it, etc.

Anyway we can take this PR as step before common changes.
I tried to add unit_type into data_t, but it means more changes everywhere.

jedi7:
With the parsing key string it si not so easy. For example km3 does not have multiplier 1000. I prefer to find a more robust way.

jedi7:
I got one idea, I can base it on this PR

zuckschwerdt:
Should be
```
char format_copy[20] = {0};
if (format) {
    strncpy(format_copy, format, sizeof(format_copy) - 1);
    data = pretty_double_format(format_copy, sizeof(format_copy), data);
    format = format_copy;
}
kv->column += fprintf(kv->file, format ? format : "%.3f", data); 
```
and
```
static double pretty_double_format(char *format, int format_len, double value)
{
    if (!format || !*format)
        return value;
    if (ends_with(format, "_J")) { // implement "ends_with()"
        // decide how to scale
        // change format and value
    }
    return value;
}
```

jedi7:
I'm more thinking about to add DATA_UNIT_TYPE for setting the unit. And then have common table of the units (not just SI, but also combined one)

zuckschwerdt:
Yes, that is for another PR. We need other changes first -- don't bother with adding units right now.

PR #2339, opened 2023-01-23
Fix TFA-141504v2 decoder length check

joel-bourquard:
The length check was unfortunately broken in this commit:
https://github.com/merbanan/rtl_433/commit/73061ff018921d401f92dc7c3525115959e77b06

This is the proposed fix.


zuckschwerdt:
Oops. Thanks!

PR #2346, opened 2023-01-25
Fix Emax LUX decoding

ProfBoc75:
- LUX decoding is not done properly.
- When LUX data reached the maximum = 0x0000 , this creating a false gap and the data are never decoded.
Removing the gap setting resolve the issue, taking into account only the first 32 Bytes from the signal.


zuckschwerdt:
I don't follow. There are three repeats, spaced 1372 µs apart. If you remove the gap settings then those won't be recognized and this wont work:
```
    for (unsigned row = 0; row < bitbuffer->num_rows; ++row) {
```

Are you seeing a PCM code / bitbuffer with a run of more than 7 0-bits? Is 0x00 even valid, isn't 0x01 the smallest value?
Can you post such a code?

ProfBoc75:
> 
Hi Christian,

The situation is that the Lux max is reached too fast (327.6KLux) and yet the UV index still increasing and other useful weather sensor information will be dropped, just because of this limit.

Based on what already done for other devices, I can arrange the analysis of the three repeats from a single row, searching the preamble at the 3 different bit positions to get the same as with the 3 rows, I need little time for that, so later this week. I will push an interim update because I found other errors in the code, like filtering the bytes then remove 1 (wind direction, uv index, lux ..  ) , must be reverse, remove 1 then filter the data.

To be continued ...

zuckschwerdt:
Yes, please do that. Without a gap to divide packets we need a loop to search preamble/sync.

ProfBoc75:
> Yes, please do that. Without a gap to divide packets we need a loop to search preamble/sync.

Done, I thought it was more complicated, but it's fine. What I did :
- replace all around "row" by 0,
- add one control to check if expected only one row to allow decode
- replace all pos = 0 by string pos into a new while loop command instead of the for each row command.
- replace the debug information about row (no more relevant) by pos number
- add increment of pos to check all the 3 packets.

I tested it, I properly found 3 packets for Temp/Hum and for Weather Station, sounds good for me.

src/devices/emax.c
```diff
@@ -104,31 +104,38 @@ Decoded example:
...
 
+    // Because of a gap false positive if LUX at max for weather station, only single row to be analyzed with expected 3 repeats inside the data.
+    if (bitbuffer->num_rows != 1) {
+        return DECODE_ABORT_EARLY;
+    }
+
     int ret = 0;
-    for (unsigned row = 0; row < bitbuffer->num_rows; ++row) {
-        unsigned pos = bitbuffer_search(bitbuffer, row, 0, preamble_pattern, sizeof(preamble_pattern) * 8);
+    uint16_t pos = 0;
```
zuckschwerdt, line 120: just `int` please

src/devices/emax.c
```diff
@@ -174,12 +181,14 @@ static int emax_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
             int wind_raw      = ((b[7] - 1) << 8 ) | (b[8] -1);   // all default is 01 so need to remove 1 from all bytes.
             float speed_kmh     = wind_raw * 0.2f;
-            int direction_deg = (((b[9] & 0x0f) - 1) << 8) | (b[10] - 1);
+            int direction_deg = (((b[9] - 1) & 0x0f) << 8) | (b[10] - 1);
             int rain_raw      = ((b[11] - 1) << 8 ) | (b[12] -1);
             float rain_mm     = rain_raw * 0.2f;
-            int uv_index      = (b[13] & 0x1f) - 1;
-            int lux_multi     = (((b[14] - 1) & 0x80) >> 7);
-            int light_lux     = ((b[14] & 0x7f) - 1) << 8 | (b[15] - 1);
+            uint8_t uv_index  = ((b[13] - 1) & 0x1f);
```
zuckschwerdt, line 187: All these should just be `int` please.

zuckschwerdt:
Thanks, this looks very good. Change the types and then ready for merge.

ProfBoc75:
Hi Christian, 
Done, it was because of the 0x00 -1 the result was 0xFFFFFF instead of 0xFF but with int &0xFF filter I have the same result.
Thx

PR #2349, opened 2023-01-29
docs: Add GNU Radio integration for real-time signal processing.

i3Craig:
I was struggling to get `rtl_433` to decode any data from my HackRF. I think, since the DC spike was more powerful than the data itself, that `rtl_433` was "distracted" by the DC offset and was unable to detect the data being received. Perhaps the auto-noise detection was thrown off by this. Regardless, the signal wasn't great since the HackRF can sample at 2MHz minimum, which means that more bandwidth was covered than what should be processed. Thus, I used GNU Radio Companion to shift the DC spike off-center, then used a low-pass filter to remove the DC spike. Finally, I decimated the signal to reduce the processing load for `rtl_433`. I later rewrote the flowgraph to just use a "Frequency Xlating FIR Filter" to make it simpler. Now, I can receive and decode signals with my HackRF! However, I wasn't able to get real-time processing working in any way other than by writing to a named pipe in GNU Radio Companion and then telling `rtl_433` to read from that pipe. Ideally, there would be some sort of "fake SDR" node that GNU Radio could output to. It may be possible to re-implement the rtl_tcp protocol in GNU Radio and then connect to that using SoapySDR. But, this method works, so I went with this method. I created this pull request to share my findings. Hopefully this can be helpful for others. This also opens the door to more advanced preprocessing in GNU Radio and signal real-time signal visualizations while receiving data.

Let me know if you have any suggestions or changes you would like me to make.


zuckschwerdt:
Very interesting stuff! HackRF users always seemed to have limited success using rtl_433, glad you now maybe solved that.

I'm not sure I can support this use-case in the official docs, the Wiki might be more suited to this kind of howto?

Not much we can do about the lack of output options in GNU radio, but we can support another input like rtl_tcp if something standard is available there?

And if you get around to implement this shifting and banding we'd be interested to have that in rtl_433. Basically should be a complex multiplication to shift the band, then re-sample or decimate and low-pass filter.

i3Craig:
1. I agree. This would make more sense in the WIKI. I will put it there.

2. This program (rtl_433) already supports the `rtl_tcp` protocol according to the docs:
```
[-d rtl_tcp[:[//]host[:port]] (default: localhost:1234)
	Specify host/port to connect to with e.g. -d rtl_tcp:127.0.0.1:1234
```
However, when I looked at the code for the `rtl_tcp` program itself, it seemed that there was some metadata sent as well as the signal (so the client could remotely control the tuned frequency or bandwidth of the `rtl_tcp` server). When I tried a TCP Sink in GNU Radio, I connected to it using `gqrx` it just crashed. It turns out that SoapySDR also has a remote protocol, but it is seemingly also not just a raw stream of IQ samples. I tried the option listed in the docs (`-d "driver=remote,remote=tcp://localhost:1234"`) with no luck. Hence, a proper `rtl_sdr` or `SoapySDR` "server" implementation would be needed in GNU Radio. Hence, I fell back to using the named pipe option.

3. I don't think I have time currently to implement the frequency translation, filtering, and decimation in `rtl_433`, but having that feature built-in would be nice, I agree.


Thanks for your time reading and replying to my pull request!

i3Craig:
I added this to the wiki. I will close my Pull Request now.
https://github.com/merbanan/rtl_433/wiki/Signal-Preprocessing

zuckschwerdt:
Yes, regarding rtl_tcp. I meant to say …like our existing rtl_tcp input… ;) It's just one TCP connection with a few bytes header and then control commands in the other direction, as you describe. Easily adapted to other such protocols, though [TCP Sink](https://wiki.gnuradio.org/index.php/TCP_Sink) is very raw. TCP isn't great here though, flow-control, buffering and missing framing will quickly mess things up. The [UDP Sink](https://wiki.gnuradio.org/index.php?title=UDP_Sink) could be interesting?

SoapyRemote is more complex, different channels, UDP, auto discovery… but rtl_tcp is also available there https://github.com/pothosware/SoapyRTLTCP/ if needed.

zuckschwerdt:
Be sure to also link your page on the Wiki home/start page.

i3Craig:
I added that. Thanks!
image

PR #2351, opened 2023-01-30
It's really the gcc compiler that defines __restrict, not the fact that it is used to target a particular platform

obones:
Before this change, any other platform than ESP32 that uses GCC would fail compiling because of the `restrict` keyword being undefined.
I found it while working on the RFLink32 project that also happens to target the D1 mini (ESP8266)

### DIFF
--- a/include/r_util.h
+++ b/include/r_util.h
@@ -17,8 +17,8 @@
 #include <time.h>
 #include "compat_time.h"
 
-#if defined _MSC_VER || defined ESP32 // Microsoft Visual Studio or ESP32
-    // MSC and ESP32 have something like C99 restrict as __restrict
+#if defined _MSC_VER || defined __cplusplus // Microsoft Visual Studio or C++ compilers (G++ is used by ESP32, ESP8266...)
+    // MSC and C++ compilers have something like C99 restrict as __restrict
     #ifndef restrict
     #define restrict  __restrict
     #endif


zuckschwerdt:
> any other platform than ESP32 that uses GCC would fail

What platform/GCC version? Does it support C99? Did you mean "any platform that uses **old** GCC"?
GCC 4.5 should have full C99 support and restricted pointers support was initially added with GCC 2.95.

zuckschwerdt:
Are you maybe using a C++ compiler? Afaik only `__restrict` is available then,

obones:
Yes, you are right, *old* gcc is more appropriate.
With the project I'm testing, it's using `xtensa-esp32-elf-g++` which gives this as its `--version` result:

    xtensa-esp32-elf-g++ (crosstool-NG crosstool-ng-1.22.0-97-gc752ad5d) 5.2.0

I'm a bit stuck on this version right now and it's indeed a C++ compiler.
What would you suggest I use as a `#define` symbol to test for then?

obones:
> The ESP-IDF also has a gcc, is that maybe an option?

I could not find a way to tell Platform IO to use a different compiler depending on the extension, so it's C++ for all in my case.

But using `__cplusplus` as the symbol is working just fine, so I have changed the commit to this instead.

PR #2352, opened 2023-01-30
Make all r_device declarations const to place them in the readonly section of the final image

obones:
When encountering global variables, the compiler places them in different sections of the MAP file for the linker to assemble into the final image. It merges all declarations of the same type into one single "segment" that then gets placed into RAM upon starting the image.
On the ESP32, there is very limited RAM (320KiB) and what's even more limiting is that only 120KiB are contiguous. This means that should the linker produce a segment of RAM that is bigger than those 120KiB, the build process will fail.
Up until 2021, this wasn't much of a problem, but with the devices added since then, I have reached a state where I can't build rtl_433 anymore for this device.
Looking a the map file, I saw that all global `r_devices` declarations, those at the end of each device file, were placed in the `.bss` segment, which takes up a lot of RAM.
But if I add a `const` modifier to all `r_devices` declarations, they end up in the `.rodata` section which is a different segment, not even in the preciously rare RAM, and I can now build again and it runs just fine.
I have also tested under Windows, I could not see any ill effect of that change there.

To me, it even makes more sense to have those declarations `const` because they are not meant to be modified at runtime anyway.

### DIFF
--- a/include/decoder_util.h
+++ b/include/decoder_util.h
@@ -29,7 +29,7 @@
 #endif
 
 /// Create a new r_device, copy from dev_template if not NULL.
-r_device *create_device(r_device *dev_template);
+r_device *create_device(r_device const *dev_template);
 
 /// Output data.
 void decoder_output_data(r_device *decoder, data_t *data);
--- a/src/decoder_util.c
+++ b/src/decoder_util.c
@@ -16,7 +16,7 @@
 
 // create decoder functions
 
-r_device *create_device(r_device *dev_template)
+r_device *create_device(r_device const *dev_template)
 {
     r_device *r_dev = malloc(sizeof (*r_dev));
     if (!r_dev) {
--- a/src/devices/abmt.c
+++ b/src/devices/abmt.c
@@ -88,7 +88,7 @@ static char *output_fields[] = {
         NULL,
 };
 
-r_device abmt = {
+r_device const abmt = {
         .name        = "Amazon Basics Meat Thermometer",
         .modulation  = OOK_PULSE_PCM,
         .short_width = 550,
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -1869,7 +1869,7 @@ static char *acurite_rain_gauge_output_fields[] = {
         NULL,
 };
 
-r_device acurite_rain_896 = {
+r_device const acurite_rain_896 = {
         .name        = "Acurite 896 Rain Gauge",
         .modulation  = OOK_PULSE_PPM,
         .short_width = 1000,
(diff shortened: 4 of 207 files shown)


obones:
I'm learning something new here, as I have always written declarations like this: `const typeName variableName = defaultValue;`
I went looking for documentation and both syntaxes are valid and come down to a matter of taste when no pointers are involved (ie no `*` in there)
And I tested, it does not have an impact on the section used in the map file.

What I'm wondering though, is about consistency because I'm used to `const typeName variableName` and I see it has been used like this in most devices codes:

    const uint8_t preamble_pattern[4] = { 0x55, 0x5F, 0xF0, 0x04 };
    const uint8_t preamble_bit_length = 30;
    const int row = 0; // we expect a single row only.

But then again, there are other places where I see this kind of declarations:

    unsigned const hdr_len = 4;
    unsigned const hdr_bits = hdr_len * 8;
    uint8_t const zeroes[8] = { 0 };
    uint8_t const aaes[5] = { 0xaa, 0xaa, 0xaa, 0xaa, 0xaa };

For the last two, I believe it has an importance because an array is basically a pointer in disguise, but for the others, I'm a bit confused.

Note that I don't mind changing the `r_device` declarations and the symbolizer regular expression, it's just that I'd rather not do it if I misunderstood your reply.

merbanan:
> (and then there are all those `%.01f` format strings -- how did that catch on?)

Whats wrong with them ?

zuckschwerdt:
I wonder what the `0` in the precision field is meant to convey? Only a `0` leading the width field (before the dot) has meaning, right?

obones:
Yeah, maybe the intent was to indicate "round to 0.01" but that would be the `%.2f` specifier.
https://stackoverflow.com/questions/15060286/any-differences-with-1f-and-01f

merbanan:
int main()
{
    float x4 = 1.1234;
    float x5 = 1.0;

    printf("Test1 x4 = %f\n", x4);
    printf("Test1 x4 = %.01f\n", x4);
    printf("Test1 x4 = %.1f\n", x4);
    printf("Test1 x5 = %f\n", x5);
    printf("Test1 x5 = %.01f\n", x5);
    printf("Test1 x5 = %.1f\n", x5);

}

Test1 x4 = 1.123400
Test1 x4 = 1.1
Test1 x4 = 1.1
Test1 x5 = 1.000000
Test1 x5 = 1.0
Test1 x5 = 1.0

merbanan:
I think I introduced it. Lets leave it. If someone "steals" the code we can clearly trace it because of this.

zuckschwerdt:
You missed the `create_device()` in `decoder_util.{c,h}` :)

obones:
Good catch.
This time it should be good.

PR #2357, opened 2023-02-01
Add Flowis protocol decoder

merbanan:


src/devices/flowis.c
```diff
@@ -0,0 +1,121 @@
...
+static char *output_fields[] = {
+        "model",
+        "id",
+        "type",
+        "amount",
+        "mic",
+        NULL,
+};
+
+r_device flowis = {
```
zuckschwerdt, line 113: r_device **const** flowis
;)

merbanan: fixed

merbanan:
Any similarities with marlec_solar.c is purely coincidental.

merbanan:
@zuckschwerdt should be ready to commit now.

zuckschwerdt:
Looks good. Should "volume" maybe have a suffix "volume_m3", and should "timestamp" indicate it's device related and not metadata, e.g. something like "device_time" or "sensor_clock"?

zuckschwerdt:
Sorry, I just merged two decoders, so I pushed a fix here.

merbanan:
I'll rebase and fix when we are happy with the keys.

merbanan:
We have volume_gallon or something like that. Yes it should be _m3.

merbanan:
And device_time sounds fine.

PR #2362, opened 2023-02-04
Fix out-of-bounds access in klimalogg, which causes frequent segfaults

drehscheibe:
A better fix would have been to write _reflect_bytes(b, sizeof(b))_ but this should then also be done in the next line with _crc8_, and since that isn't related to the fix, I went for minimalism and consistency.


zuckschwerdt:
Thanks! Constants are okay if the code is easy to review.

PR #2374, opened 2023-02-10
Fix Emax units

Mikayex:
The temperature field is named Temperature_F for Emax devices. This is confusing, especially if the `-C si` option is used as it would display `Temperature_F: 19.6` but it is in Celsius !

I also added humidity and temperature units, like what's done on other weather stations.


zuckschwerdt:
Thanks!

PR #2378, opened 2023-02-14
Question mark precedence

obones:
This fixes a warning about question mark operator precedence as discussed here: #2360


zuckschwerdt:
I have to think about this, not sure if I like using macros. I would say `(x + 7) / 8` is pretty obvious and not a burden to type.
(Also in my tests `/8` can be optimized by the compiler (best if the type is unsigned) but `>>3` might stop optimizations.

obones:
Well, to me the `(x + 7) / 8` code does not convey the "round up to number of bytes" immediately, I had to look it up to be sure of what it does.
As to optimization, shifting three bits to the right is what the compiler is doing when it sees an unsigned division by 8, at least from what I have experienced in the past.

zuckschwerdt:
I want to (and often quickly have to) look at code blocks and grasp what is going on, same as @obones. Difference is that plain instructions don't bother me much but macros are always suspect and without documentation hints macros are obfuscating more than they help (me at least). If only we could have inlined functions across compilations units...

gdt:
We can have inline functions in headers, but I am not sure it fits with the build environment rules.   I don't really see macros and functions as all that different, and it seems easy to have  comment explaining the macro.    But really the top-level point is to make progress either merging  or rejecting this; we have too many things just long-term open.

merbanan:
So should we merge or reject this?

zuckschwerdt:
I'm still not in favor of using more macros. Typical decoders currently don't use any macros.

E.g. to me in complex code it's easier to read the block of operations here
https://github.com/merbanan/rtl_433/blob/2d63f30681dd527955542edea4b7a570bacf8a6f/src/devices/flex.c#L43-L46
then having another layer with a macro call in there
https://github.com/merbanan/rtl_433/blob/8b4376197c1d032823a9e461d525c9a8fdea089c/src/devices/flex.c#L43-L46

But if this is seen as a useful simplification then I won't argue.
We should find a more descriptive name though, `BITLEN_TO_BYTELEN`, `TO_BYTELEN`, `BYTELEN_OF`  or such (to convey we are not converting bits or bytes but lenghts).

obones:
As I have already voiced it earlier, I find the +7/8 code really hard to understand for a newcomer, hence my suggestion to use a macro.
Using a longer name for that macro is not an issue for me, I'd go with the `BITLEN_TO_BYTELEN` option. If that's good for everyone here I'll make the change in my PRs

merbanan:
I agree with less macros. Closing.

PR #2379, opened 2023-02-14
Change literals to string const for strict discarded-qualifiers warnings

obones:
After an upgrade from the maintainers of the project where I use rtl_433 as a library, the compiler has started to emit warnings about const qualifiers being discarded in all device files, on every string value used a field name.
This comes from the `discarded-qualifiers` warning that is also activated for CMake but does not seem to emit it on the regular project.
To avoid thousands of warnings, I thus changed all `static char *XXX_fields[] =` declarations to `static char const *XXX_fields[] =` which removes the warnings.
It leads in turn to a few auxiliary changes which I believe are for the better as it makes it clear that the source of a copy is considered constant by the method that works with it.

Before this is merged, I'm wondering if I should add a new test in `symbolizer.py` to make sure there won't be any more `static char* XXX` declared in the future. I have used the following regular expression to find them, so it should not be too hard to add it in this script:

    static char \*([a-z0-9_]+)\[\]

As always, I'm opened to discussion and suggestions.

### DIFF
--- a/include/optparse.h
+++ b/include/optparse.h
@@ -86,7 +86,7 @@ double arg_float(char const *str, char const *error_hint);
 /// also "//localhost", "//localhost:514", "//:514".
 /// Host or port are terminated at a comma, if found.
 /// @return the remaining options
-char *hostport_param(char *param, char **host, char **port);
+char *hostport_param(char *param, char const **host, char const **port);
 
 /// Convert a string to an unsigned integer, uses strtod() and accepts
 /// metric suffixes of 'k', 'M', and 'G' (also 'K', 'm', and 'g').
--- a/src/data_tag.c
+++ b/src/data_tag.c
@@ -172,8 +172,8 @@ data_tag_t *data_tag_create(char *param, struct mg_mgr *mgr)
     int gpsd_mode = strncmp(tag->val, "gpsd", 4) == 0;
     if (gpsd_mode || strncmp(tag->val, "tcp:", 4) == 0) {
         p          = arg_param(tag->val); // strip scheme
-        char *host = gpsd_mode ? "localhost" : NULL;
-        char *port = gpsd_mode ? "2947" : NULL;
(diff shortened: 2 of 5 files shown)


zuckschwerdt:
A very welcome clarification. I'll review and test shortly.

obones:
I have found a few others in devices, all related to local `char *` variables that should really be `char const *` because they only deal with hard coded strings.
You will also see in the latest commit that I have modified `r_api.c` because there are cases like this:

```
    char *host = "localhost";
    char *port = "514";
```

As they are constant strings, they raise the warning. But changing them to `char const *` as is done in my commit triggers a `incompatible-pointer-types` warning because `hostport_param` requires a `char **` and not a `char const **`.
This made me stop there because I can't see how I could modify this method to work with a `char const **` when all it receives is a `char*`

And finally, I have not created a commit for the following constructs in the same `r_api.c` file:

```
    list_push(&field_list, "time");
```

Clearly, the `const` qualifier coming from the `"time"`  hard coded string is discarded because `list_push` takes in a `void *` and not a `void const *`.
I initially tried to change the declaration, but it meant quite a few changes to the entire list struct, and I hit a roadblock with calls to `elem_free` further down the road.

So in the end, I'm not sure I should have modified `r_api.c` but I'm not too keen on shutting down the warning for the entire file. Should I add a "hard cast" in places where I get a `discarded-qualifiers` warning?

zuckschwerdt:
The `hostport_param()` change will actually work, I've pushed the change on top. No warnings anymore.

The `const` does not guarantee the value won't change, it just restricts that site to effect a change.
I.e. we can extract a `char const` from arbitrary memory. We thereby "promise" not to change the value.

zuckschwerdt:
If you have the time to add checks to `symbolizer.py` or `style-check` action/test to ensure `const` on output fields and `r_device` then please do :)

obones:
The one on `r_device` was in my previous PR, I'll see what I can do about output fields.

PR #2386, opened 2023-02-18
Add support for LaCrosse TX31U-IT

agspoon:
Add support for Lacrosse TX31U-IT weather sensor

This is the sensor/transmitter used with this product,
https://www.lacrossetechnology.com/products/ws-1910twc-it

It uses a variable length message format, and can report temperature, humidity, wind speed/direction/gust, and rain.

Testing this code for a couple of months, and it seems to be working well.


src/devices/lacrosse_tx31u.c
```diff
@@ -0,0 +1,227 @@
...
+
+    decoder_output_data(decoder, data);
+
+    return 1;
+}
+
+static char const *output_fields[] = {
+        "model",
+        "id",
+        "low_battery",
```
zuckschwerdt, line 208: "battery_ok" and  "temperature_C" ;)

zuckschwerdt:
Thanks! The protocol looks intricate and must have been a lot of work.

src/devices/lacrosse_tx31u.c
```diff
@@ -0,0 +1,227 @@
...
+    }
+    else {
+        decoder_logf(decoder, 1, __func__, "Packet truncated: received %d bytes, expected %d bytes", msg_bytes, expected_bytes);
+        return DECODE_ABORT_LENGTH;
+    }
+
+    /* clang-format off */
+    // what we know from the header
+    data_t *data = data_make(
+	    "model",            "",             DATA_STRING, "LaCrosse-TX31U-IT",
```
zuckschwerdt, line 137: "model" needs to be "LaCrosse-TX31UIT", we promise a single dash in the middle.

agspoon:
I can't figure out why the symbol error check is failing.  It doesn't fail in my local repo, and I can't see the error.

src/devices/lacrosse_tx31u.c
```diff
@@ -0,0 +1,227 @@
...
+        "humidity",
+        "wind_avg_km_h",
+        "wind_max_km_h",
+        "wind_dir_deg",
+        "mic",
+        NULL,
+};
+
+// Receiver for the Lacrosse TX31U-IT
+r_device lacrosse_tx31u = {
```
zuckschwerdt, line 219: r_device **const** lacrosse_tx31u

agspoon: Ahh, OK.  Fixed that.  Thanks!

Issue #2389, opened 2023-02-19
Add support for Celsia CZC1 Thermostat


libanp:
Hello,

Thanks for the help so far on the mailing list. Moving here as suggested.

The current flex decoder is:

`rtl_433 -X n=CZC1,m=OOK_PCM,s=1220,l=1220,r=4880,preamble=cccccccccccccccccccc`

I've created an annotated bitbench from the unique samples I've managed to get from the 3 thermostats I have running.

There are 2 packet types, one for pairing and one for commands. Pairing packets are 5 bytes, command packets are 6 bytes.

Pairing packet:
1. 0
2. ID
3. ID
4. Type
5. CRC

Command packet:
1. 0
2. ID
3. ID
4. Type
6. Heat level (0-255 LSB)
7. CRC

As shown on the mailing list, pre-padding the data reveals a known CRC (still valid with the extra samples):

```
./reveng -w 8 -s 000000002750f00017 000000002750f0ecab 000000002750f0d6b5 000000002750f08598 000000002750f0ab01 000000002750f0ffbb 000000002750f0c050 00000000ec46f00087 00000000ec46f08c80 00000000f0d3f0c0f4 00000000f0d3f000b3 00000000f0d3f0ec0f 00000000f0d3f0d611 00000000f0d3f0853c 00000000f0d3f0aba5 00000000f0d3f0ff1f 

width=8  poly=0x31  init=0xff  refin=false  refout=false  xorout=0x00  check=0xf7  residue=0x00  name="CRC-8/NRSC-5"
```

There are some samples that decode wrong, is this an issue with the timing specified in the flex decoder? See samples g2150, g363

[...]

PR #2391, opened 2023-02-21
Add support for Celsia CZC1

libanp:
Initial working code

### DIFF
--- a/src/devices/celsia_czc1.c
+++ b/src/devices/celsia_czc1.c
@@ -1,0 +1,166 @@
+/** @file
+    Celsia CZC1 Thermostat.
+
+    Copyright (C) 2023 Liban Hannan <liban.p@gmail.com>
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "decoder.h"
+
+/** @fn int celsia_czc1_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+Celsia CZC1 Thermostat.
+
+A PID thermostat compatible with various manufacturers' heaters.
+
+demod: OOK_PCM
+short: 1220
+long: 1220
+reset: 4880
+
+A packet starts with a preamble of {40}cccccccccccccccccccc, followed by a sync
+of {32}55555555 signalling the start of the data symbols. The packet is
+terminated with {8}f0.  Each symbol is 4 'raw' bits long: 0101(5) = 0, 1010(a)
+= 1. Command packets have 5 bytes of data, pairing packets have 4.
+
+```
+rtl_433 -X n=CZC1,m=OOK_PCM,s=1220,l=1220,r=4880,preamble=cccccccc55555555
+```
+
+Data layout:
+
+Command packet (5 bytes)
+
+- ID:   {16} ID
+- Type: {8}  type
+- Heat: {8}  heating level 0-255 (little endian unsigned integer)
+- CRC:  {8}  CRC-8, poly 0x31, init 0xd7
+
+Pairing packet (4 bytes)
+
+- ID:   {16} ID
+- Type: {8}  type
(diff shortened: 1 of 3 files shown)


libanp:
Follows discussion at #2389

This decoder works, so I thought I'd start the PR with what I've got. Aside from whatever else comes up, there are two issues with it I'd appreciate direction for:

- Are there better utility functions for converting the raw bits to symbols? I didn't have much luck finding them
- My use of a 'type' json field doesn't fit with fields described in [DATA_FORMAT.md](https://github.com/merbanan/rtl_433/blob/master/docs/DATA_FORMAT.md), should I be using 'subtype' instead?

src/devices/celsia_czc1.c
```diff
@@ -0,0 +1,177 @@
+/** @file
+    Bresser Weather Center 5-in-1.
```
zuckschwerdt, line 2: "Celsia CZC1 Thermostat." ;)

src/devices/celsia_czc1.c
```diff
@@ -0,0 +1,177 @@
...
+*/
+
+#include "bitbuffer.h"
+#include "decoder.h"
+#include "decoder_util.h"
+#include "r_device.h"
+#include <bits/stdint-uintn.h>
+
+/** @fn int celsia_czc1_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+Celsia CZC1 Thermostat
```
zuckschwerdt, line 19: Add a dot `.` at the end

src/devices/celsia_czc1.c
```diff
@@ -0,0 +1,177 @@
...
+
+A PID thermostat compatible with various manufacturers' heaters.
+
+demod: OOK_PCM
+short: 1220
+long: 1220
+reset: 4880
+
+A packet starts with a sync of {40}cccccccccccccccccccc, followed by a preamble
```
zuckschwerdt, line 28: cc... is the preamble, 55.. is the sync

src/devices/celsia_czc1.c
```diff
@@ -0,0 +1,177 @@
...
+        return DECODE_FAIL_MIC;
+    }
+    int id = (b[0] << 8) | b[1];
+
+    data_t *data;
+    if (b[2] == 0x00) { // pair packet
+        /* clang-format off */
+        data = data_make(
+            "model",    "",               DATA_STRING, "Celsia CZC1",
+            "id",       "ID",             DATA_FORMAT, "0x%X", DATA_INT, id,
```
zuckschwerdt, line 129: use `"%x"` please

zuckschwerdt:
Remove the "type" key you have. There is no need for that info, it's either packets with "heat" or empty packets.

I can rewrite the bit decoder later, basically I would use the fact that when aligned a byte will hold exactly 2 bits, i.e. it can only ever be 55, 5a, a5, aa. Otherwise using `extract_bits_symbols()` will also work https://github.com/merbanan/rtl_433/blob/master/include/util.h#L84

src/devices/celsia_czc1.c
libanp: Doh!

libanp:
I've updated the code with the corrections. The decoder matches against byte values now, but I'm happy for the it to be rewritten :)

src/devices/celsia_czc1.c
```diff
@@ -0,0 +1,170 @@
...
+    Copyright (C) 2023 Liban Hannan <liban.p@gmail.com>
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "bitbuffer.h"
+#include "decoder.h"
```
zuckschwerdt, line 13: You should only need "decoder.h"

zuckschwerdt:
Looks good. In the long term we might want to put the bits into a `uint8_t[]` instead of the (large) bitbuffer_t. But that's an optimization for later (e.g. manchester_decode() also needs that refactor some day).

Issue #488, opened 2017-02-26
Accurite 6045M Lightning Detector Bad Temp


n7qnm:
Using an Acurite 6045M Lightning Detector it's supposed to provide Temp and Humidity - humidity looks OK but temp appears to be decoding incorrectly  - I have it in a room that is between 65 and 75F; RTL decodes as 35 to 258.   Would be happy to help debug with some assistance.

clayj@nwlink.com

rct:
I wasn't able to figure out how the temperature was encoded in my 6045M lightning detector and eventually gave up, suspecting that my sensor was bad or I'm missing something obvious. 

If you can provide output with some clear observations maybe we can get this resolved.

rct:
Fixed by #500.   This should be closed.

rct:
@zuckschwerdt @merbanan @n7qnm the temperature decoding was fixied 13 months ago. I believe you can close this issue.

PR #2403, opened 2023-03-02
Add conf-file for SWETUP garage door remote

FoxWhiskey:
Here's a conf-file to decode the "SWETUP" garage door remote, available on AMAZON.
Please note [PR #488](https://github.com/merbanan/rtl_433_tests/pull/448) for test data

### DIFF
--- a/conf/SWETUP-garage-opener.conf
+++ b/conf/SWETUP-garage-opener.conf
@@ -1,0 +1,29 @@
+# SWETUP garage door opener
+# (c) FoxWhiskey 2023
+#
+# A four-button remote control device most likely based on EV1527 chip.
+# The device is available on AMAZON and a no-name-product branded for the german market.
+# See https://www.amazon.de/-/en/Handheld-Transmitter-Control-Universal-Wireless/dp/B09HWY1KH7
+#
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Great, thanks!
There is a problem with copyright annotated files with no clear licensing information. Would you be ok to change the second line to `by FoxWhiskey 2023` and drop the explicit ownership claim?

@merbanan can you update your two conf's in the same way?
Pretty sure Tommy and @psa-jforestier also meant to donate the conf without claims, but we should ask for a change also or delete those.
The only other way would be to explicitly license at least GPL 2.0-or-later or some compatible CC maybe.

FoxWhiskey:
Absolutely! Sorry for that -  just wanted to leave a piece of author information - no copyright claims intended...

zuckschwerdt:
No worries, the whole thing is complicated. And we don't dispute your authorship and implicit copyright. But since we are redistributing the file we need a license to do so. There an explicit copyright note with no licensing information is suspect.
Hopefully correct we assume that files given to the project without explicit claims are meant to be licensed with the project's GPL 2.0-or-later.

Issue #947, opened 2019-01-07
RFC Holman iWeather WH5029


fnarfbargle:
G'day all,

I've been working on support for an Australian Holman iWeather device. I have it working, although I haven't been able to sus out the checksum yet. The low nybble is a straight xor but I haven't figured out the high yet. As a result, this file contains extra debugging output and isn't suitable for inclusion yet, but I wanted to get it looked at to see if I'm on the right track for feature and format when I get it finished.

[holman.c.txt](https://github.com/merbanan/rtl_433/files/2731245/holman.c.txt)

The rain units are still a bit handwavy, although it looks pretty close to each count being 1/32 of an inch. I need to push a heap more water through the unit to refine the multiplier, but it's currently good over 50mm.

I have sample recordings and all the other requirements, but I thought I'd get an early version of the code looked at for comment.

Brad.

merbanan:
Look at the use of thebitbuffer_search()  and bitbuffer_find_repeated_row() function. Other then that the code looks fairly good. Regarding the checksum a check for the 3 byte preamble and a 4bit xor is enough to reliably verify the message content. Add a "mic" to the output.

fnarfbargle:
Much obliged.

I've used find_repeated_row, but I couldn't see how bitbuffer_search would simplify the code as I haven't yet seen an improperly aligned bitbuffer.

I did get a couple of corrupted messages through the previous version where the corruption was in the id byte and the xor passed, so looking for at least 3 identical rows should nail that down.

```C
    // Test for 3 identical rows (we get 6)
    int r = bitbuffer_find_repeated_row(bitbuffer, 3, 96);
    if (r < 0 || bitbuffer->bits_per_row[r] != 96)
        return 0;

    // Test for preamble
    if (memcmp(bb[r],header,3))
        return 0;

    // Test Checksum.
    if (xor_bytes(bb[r],11) & 0xF)
        return 0;
```
I really would like to nail the whole checksum, if only because it's annoying me, so I'll keep working on it for a while before putting it in git and making a pull request.

merbanan:
Then you need to post a few (50?) bit bufffers so we might be able to help out. If the bitbuffer is aligned then there is no need to align it. But only bitbuffer_invert(bitbuffer); after checking the bits per row and that the preamble + header matches.

zuckschwerdt:
The wind direction should preferably be in degrees.
[...]

PR #2419, opened 2023-03-12
Improve Holman ws5029, Add support for AOK-5056 and correction for Emax

ProfBoc75:
Hi,

I propose an improvement from the current Holman ws5029 decoding to take into account the xor checksum, partially with the lower nibble but better than nothing.
And also add the support for AOK-5056, same family product with UV & Lux sensor.

[My samples](https://github.com/ProfBoc75/rtl_433_tests/tree/patch-1/tests/aok_5056)

[BitBench](https://triq.net/bitbench#c=aaaaaaaaaaaa98f3a5c7670982b9080280111e47ca7a0b000000000&c=aaaaaaaaaaaa98f3a5c7670972b90803d0f36747cb9109000000000&c=aaaaaaaaaaaa98f3a5c7670972b21007a1e6ce8f972212000000000&c=aaaaaaaaaaaa98f3a5c7670982b90901805ffa87cc3409000000000&c=aaaaaaaaaaaa98f3a5c76709d2b909024397c2c7ce760f000000000&c=aaaaaaaaaaaa98f3a5c76709d2b90902409c8947d74c0b000000000&f=ID%3A16h%20TEMP_C%3A12s%20HUM_%25%3A8d%20RAIN_MM%3A12d%20WIND_SPEED_KMH%3A8d%20WIND_DIRECTION%3A4h%20UV_INDEX%3A5d%20LUX%3A17h%20BAT%3A2h%20COUNTER%3A12h%20XOR%3A8h%204h%20PREV_DIR%3A4h%2036x&a=Preamble&m=aaaaaaaaaaaa98f3a5&cw=4)

### DIFF
--- a/src/devices/emax.c
+++ b/src/devices/emax.c
@@ -185,10 +185,10 @@ static int emax_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             int temp_raw      = ((b[4] & 0x0f) << 8) | (b[5]); // weird format
             float temp_f      = (temp_raw - 900) * 0.1f;
             int humidity      = b[6];
-            int wind_raw      = ((b[7] - 1) << 8 ) | (b[8] -1);   // all default is 01 so need to remove 1 from all bytes.
+            int wind_raw      = (((b[7] - 1) & 0xff) << 8) | ((b[8] - 1) & 0xff);   // need to remove 1 from byte , 0x01 - 1 = 0 , 0x02 - 1 = 1 ... 0xff -1 = 254 , 0x00 - 1 = 255.
             float speed_kmh   = wind_raw * 0.2f;
-            int direction_deg = (((b[9] - 1) & 0x0f) << 8) | (b[10] - 1);
-            int rain_raw      = ((b[11] - 1) << 8 ) | (b[12] -1);
+            int direction_deg = (((b[9] - 1) & 0x0f) << 8) | ((b[10] - 1) & 0xff);
+            int rain_raw      = (((b[11] - 1) & 0xff) << 8) | ((b[12] - 1) & 0xff);
             float rain_mm     = rain_raw * 0.2f;
 
             if (b[29] == 0x17) {                               // with UV/Lux, without Wind Gust
@@ -250,7 +250,7 @@ static int emax_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     return ret;
 }
 
-static char const *output_fields[] = {
+static char const *const output_fields[] = {
         "model",
         "id",
         "channel",
--- a/src/devices/holman_ws5029.c
+++ b/src/devices/holman_ws5029.c
@@ -306,10 +306,10 @@ static int holman_ws5029pwm_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             "model",            "",                 DATA_STRING, "Holman-WS5029",
             "id",               "",                 DATA_INT,    id,
             "battery_ok",       "Battery",          DATA_INT,    !battery_low,
-            "temperature_C",    "Temperature",      DATA_FORMAT, "%.01f C",     DATA_DOUBLE, temp_c,
+            "temperature_C",    "Temperature",      DATA_FORMAT, "%.1f C",     DATA_DOUBLE, temp_c,
             "humidity",         "Humidity",         DATA_FORMAT, "%u %%",       DATA_INT,    humidity,
-            "rain_mm",          "Total rainfall",   DATA_FORMAT, "%.01f mm",    DATA_DOUBLE, rain_mm,
-            "wind_avg_km_h",    "Wind avg speed",   DATA_FORMAT, "%.01f km/h",  DATA_DOUBLE, speed_kmh,
+            "rain_mm",          "Total rainfall",   DATA_FORMAT, "%.1f mm",    DATA_DOUBLE, rain_mm,
+            "wind_avg_km_h",    "Wind avg speed",   DATA_FORMAT, "%.1f km/h",  DATA_DOUBLE, speed_kmh,
             "wind_dir_deg",     "Wind Direction",   DATA_INT,    (int)(wind_dir * 22.5),
             "mic",              "Integrity",        DATA_STRING, "CHECKSUM",
             NULL);
--- a/src/devices/holman_ws5029.c
+++ b/src/devices/holman_ws5029.c
@@ -57,20 +57,24 @@ Payload format: With UV Lux sensor
     Fixed Values 0x  : AA AA AA AA AA AA 98 F3 A5
 
     Byte position    : 00 01 02 03 04 05 06 07 08 09 10 11 12 13 14 15 16 17 18
-    Payload          : II II CC CH HR RR WW DU UL LL BN NN SS 0D 00 00 00 00 0
+    Payload          : II II CC CH HR RR WW |         | NN SS 0D 00 00 00 00 0
+                                +-----------+         +-------------+
+                                |                                   |
+                                |   07       08       09       10   |
+                  bits details : DDDDUUUU ULLLLLLL LLLLLLLL LLBBNNNN
 
-- IIII        station ID (randomised on each battery insertion)
-- CCC         degrees C, signed, in multiples of 0.1 C
-- HH          humidity %
-- RRR         cumulative rain in mm
-- WW          wind speed in km/h
-- D           wind direction (0 = N, 4 = E, 8 = S, 12 = W)
-- UU          Index UV
-- LLLB        Lux
-- B           Batterie
-- NNN         Payload number, increase at each message 000->FFF but not always, strange behavior. no clue
-- SS          XOR bytes checksum, lower nibble properly decoded, not the upper, unknown calcul.
-- D           Previous Wind direction other values
+- I     station ID (randomised on each battery insertion)
+- C     degrees C, signed, in multiples of 0.1 C
+- H     humidity %
+- R     cumulative rain in mm
+- W     wind speed in km/h
+- D     wind direction (0 = N, 4 = E, 8 = S, 12 = W)
+- U     Index UV
+- L     Lux
+- B     Battery
+- N     Payload number, increase at each message 000->FFF but not always, strange behavior. no clue
+- S     XOR bytes checksum, lower nibble properly decoded, not the upper, unknown calcul.
+- D     Previous Wind direction
 - Fixed values to 9 zeros
 
 To get raw data
@@ -146,7 +150,7 @@ static int holman_ws5029pcm_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         decoder_output_data(decoder, data);
         return 1;
     }
-    else if (bits < 220) {                         // model with UV LUX
+    else if (bits < 221) {                         // model with UV LUX
         float rain_mm    = rain_raw * 1.0f;
         int uv_index     = ((b[7] & 0x07) << 1) | ((b[8] & 0x80) >> 7);
         int light_lux    = ((b[8] & 0x7F) << 10) | (b[9] << 2) | ((b[10] & 0xC0) >> 6);
@@ -171,9 +175,7 @@ static int holman_ws5029pcm_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         decoder_output_data(decoder, data);
         return 1;
     }
-    else {
-        return 0;
-    }
+    return 0;
 }
 
 static char const *output_fields[] = {
--- a/src/devices/holman_ws5029.c
+++ b/src/devices/holman_ws5029.c
@@ -16,7 +16,7 @@ AOK Electronic Limited weather station.
 Known Rebrand compatible with:
 - Holman iWeather Station ws5029. https://www.holmanindustries.com.au/products/iweather-station/
 - Conrad Renkforce AOK-5056
-- Optex Electronique 99018 SM-018 5056
+- Optex Electronique 990018 SM-018 5056
 
 Appears to be related to the Fine pos WH1080 and Digitech XC0348.
 
@@ -195,7 +195,7 @@ static char const *output_fields[] = {
 };
 
 r_device const holman_ws5029pcm = {
-        .name        = "AOK Weather Station rebrand Holman Industries iWeather WS5029, Conrad AOK-5056, Optex 99018",
+        .name        = "AOK Weather Station rebrand Holman Industries iWeather WS5029, Conrad AOK-5056, Optex 990018",
         .modulation  = FSK_PULSE_PCM,
         .short_width = 100,
         .long_width  = 100,
--- a/src/devices/holman_ws5029.c
+++ b/src/devices/holman_ws5029.c
@@ -129,7 +129,7 @@ static int holman_ws5029pcm_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     float temp_c      = (temp_raw >> 4) * 0.1f;
     int humidity      = ((b[3] & 0x0f) << 4) | ((b[4] & 0xf0) >> 4);
     int rain_raw      = ((b[4] & 0x0f) << 8) | b[5];
-    int speed_kmh     = b[6];
+    float speed_kmh   = (float)b[6];
     int direction_deg = wind_dir_degr[(b[7] & 0xf0) >> 4];
 
     if (bits < 200) {                 // model without UV LUX
@@ -138,11 +138,11 @@ static int holman_ws5029pcm_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         /* clang-format off */
         data = data_make(
                 "model",            "",                 DATA_STRING, "Holman-WS5029",
-                "id",               "StationID",        DATA_FORMAT, "%04X",     DATA_INT,    device_id,
-                "temperature_C",    "Temperature",      DATA_FORMAT, "%.01f C",  DATA_DOUBLE, temp_c,
-                "humidity",         "Humidity",         DATA_FORMAT, "%u %%",    DATA_INT,    humidity,
-                "rain_mm",          "Total rainfall",   DATA_FORMAT, "%.01f mm", DATA_DOUBLE, rain_mm,
-                "wind_avg_km_h",    "Wind avg speed",   DATA_FORMAT, "%u km/h",  DATA_INT,    speed_kmh,
+                "id",               "StationID",        DATA_FORMAT, "%04X",       DATA_INT,    device_id,
+                "temperature_C",    "Temperature",      DATA_FORMAT, "%.1f C",     DATA_DOUBLE, temp_c,
+                "humidity",         "Humidity",         DATA_FORMAT, "%u %%",      DATA_INT,    humidity,
+                "rain_mm",          "Total rainfall",   DATA_FORMAT, "%.1f mm",    DATA_DOUBLE, rain_mm,
+                "wind_avg_km_h",    "Wind avg speed",   DATA_FORMAT, "%.1f km/h",  DATA_DOUBLE, speed_kmh,
                 "wind_dir_deg",     "Wind Direction",   DATA_INT, direction_deg,
                 NULL);
         /* clang-format on */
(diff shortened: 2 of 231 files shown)


zuckschwerdt:
Thanks!
Can you run `maintainer_update.py` so `README.md` and `rtl_433.1` also change?

zuckschwerdt:
I *really* like your "bits details" drawing!

ProfBoc75:
Need a little correction for the wind speed. I have errors into Domoticz because of bad value to 128.9 kmh instead of 1 kmh,
Because the current speed is int , must be float value, I checked other weather stations, they are all using float values.

zuckschwerdt:
> the current speed is int , must be float value,

How does Domoticz read the value? From MQTT, JSON, or CSV? For JSON and MQTT at least the type int is the same as float.

ProfBoc75:
Domoticz from json , I saw a conversion km/h to  m/s  by dividing the value by 3.6f, (from domoticz/hardware/Rtl433.cpp , line 236). 

```
wind_speed = ((float)atof(data["wind_avg_km_h"].c_str())) / 3.6F;
```

And I'm facing to this error lot of times, like 1 is 129 ,  I need to correct manually the value into Domoticz db.

In parallel, I monitor the weather station value from another rtl-sdr dongle, without error from rtl_433, but error into Domoticz. So for me, I deduct that the behavior is from Domoticz , some round value not done properly, may be because of int value.

ProfBoc75:
> But `atof("1")` is the same as `atof("1.0")` and why would it be `129`? Something else is wrong. (because in a string any int is just a float with not decimals.)

Agree, I'll try another test, to get the value from the current dongle I'm using into Domoticz, to check if it could from it.
I was expecting some improvement now that the xor is checked, but still partially check, since 1 and 129 the gap is at the upper nibble, and only the lower nibble from the xor is checked, this could be the reason.

ProfBoc75:
Well, more than 24 hours later, still no error, I'm monitoring the json output. I will continue to monitor.

I have [code data](https://github.com/ProfBoc75/rtl_433_tests/blob/patch-1/tests/aok_5056/test/aok.json) that can be used to found the good algorithm for the checksum / crc.

Into the code data, the 13th byte sounds to be the xor sum from the 12th first bytes, for sure for the lower nibble and I added it into the device improvement, but not the upper nibble, and I don't know how to find it ?   
Is there a way to find the xor calcul (or checksum ?) for the upper nibble ? I need some help on this.

merbanan:
Maybe revdgst can help.

ProfBoc75:
Hi, thanks for help, but unfortunately no one give me a good result except it confirmed the XOR by shift up =  0x18 = 24 means the exact number of nibbles. But still no luck with the upper nibble.

In another hand, I monitored the messages for more than one full day, and no error, then it's running with Domoticz, same without errors, this means that the decoder is not so bad, acceptable for me.

So from my side, it's ok to merge it.

ProfBoc75:
Hi,

I have enough messages to work on this xor checksum and I found that the Upper Nibble is an xor between the upper xor and lower xor of the message with a kind of shift at the lower nibble (lfsr ?) .

I was able to extract all the 256 possibilities, see the table, still working on the good algorithm to make the map.  For sure the result will be something related to lfsr.

xor_bytes of the 12 bytes message --> upper_result XOR  ( LFSR xxx ??? of lower_result )

Here the table between the calculated XOR from the message (12 bytes) and the value of the 13th Byte checksum:
XOR   Message | 13th Byte
-- | --
00 | 00
01 | 31
02 | 62
03 | 53
04 | c4
05 | f5
06 | a6
07 | 97
08 | 88
09 | b9
0A | ea
0B | db
0C | 4c
0D | 7d
0E | 2e
0F | 1f
10 | 10
11 | 21
12 | 72
13 | 43
14 | d4
15 | e5
16 | b6
17 | 87
18 | 98
19 | a9
1A | fa
1B | cb
1C | 5c
1D | 6d
1E | 3e
1F | 0f
20 | 20
21 | 11
22 | 42
23 | 73
24 | e4
25 | d5
26 | 86
27 | b7
28 | a8
29 | 99
2A | ca
2B | fb
2C | 6c
2D | 5d
2E | 0e
2F | 3f
30 | 30
31 | 01
32 | 52
33 | 63
34 | f4
35 | c5
36 | 96
37 | a7
38 | b8
39 | 89
3A | b0
3B | eb
3C | 7c
3D | 4d
3E | 1e
3F | 2f
40 | 40
41 | 71
42 | 22
43 | 13
44 | 84
45 | b5
46 | e6
47 | d7
48 | c8
49 | f9
4A | aa
4B | 9b
4C | 0c
4D | 3d
4E | 6e
4F | 5f
50 | 50
51 | 61
52 | 32
53 | 03
54 | 94
55 | a5
56 | f6
57 | c7
58 | d8
59 | e9
5A | ba
5B | 8b
5C | 1c
5D | 2d
5E | 7e
5F | 4f
60 | 60
61 | 51
62 | 02
63 | 33
64 | a4
65 | 95
66 | c6
67 | f7
68 | e8
69 | d9
6A | 8a
6B | bb
6C | 2c
6D | 1d
6E | 4e
6F | 7f
70 | 70
71 | 41
72 | 12
73 | 23
74 | b4
75 | 85
76 | a2
77 | e7
78 | f8
79 | c9
7A | 9a
7B | ab
7C | 3c
7D | 0d
7E | 5e
7F | 6f
80 | 80
81 | b1
82 | e2
83 | d3
84 | 44
85 | 75
86 | 26
87 | 17
88 | 08
89 | 39
8A | 6a
8B | 5b
8C | cc
8D | fd
8E | ae
8F | 9f
90 | 90
91 | a1
92 | f2
93 | c3
94 | 54
95 | 65
96 | 36
97 | 07
98 | 18
99 | 29
9A | 7a
9B | 4b
9C | dc
9D | ed
9E | be
9F | 8f
A0 | a0
A1 | 91
A2 | 92
A3 | f3
A4 | 64
A5 | 55
A6 | 06
A7 | 37
A8 | 28
A9 | 19
AA | 4a
AB | 7b
AC | ec
AD | dd
AE | 8e
AF | bf
B0 | b0
B1 | 81
B2 | d2
B3 | e3
B4 | 74
B5 | 45
B6 | 16
B7 | 27
B8 | 38
B9 | 09
BA | 5a
BB | 6b
BC | fc
BD | cd
BE | 9e
BF | af
C0 | c0
C1 | f1
C2 | a2
C3 | 93
C4 | 04
C5 | 35
C6 | 66
C7 | 57
C8 | 48
C9 | 79
CA | 2a
CB | 1b
CC | 8c
CD | bd
CE | ee
CF | df
D0 | d0
D1 | e1
D2 | b2
D3 | 83
D4 | 14
D5 | 25
D6 | 76
D7 | 47
D8 | 58
D9 | 69
DA | 3a
DB | 0b
DC | 9c
DD | ad
DE | fe
DF | cf
E0 | e0
E1 | d1
E2 | 82
E3 | b3
E4 | 24
E5 | 15
E6 | 46
E7 | 77
E8 | 68
E9 | 59
EA | 0a
EB | 3b
EC | ac
ED | 9d
EE | ce
EF | ff
F0 | f0
F1 | c1
F2 | 92
F3 | a3
F4 | 34
F5 | 05
F6 | 56
F7 | 67
F8 | 78
F9 | 49
FA | 1a
FB | 2b
FC | bc
FD | 8d
FE | de
FF | ef

zuckschwerdt:
Great success! This can be solved with a reverse Galois algorithm then (key = 0x31).
But there are 3 errors(?) in your table:
- `3a b0` should be `3a da`
- `76 a2` should be `76 d6`
- `a2 92` should be `a2 c2`

Can you double check that?
Do you have a file with all full message codes (before XOR)?

zuckschwerdt:
Oh and it looks like a broken try at LFSR -- the "F" is missing ;)
For an LFSR this would be characteristic:
- `01` -> `31`
- `02` -> `62`
- `04` -> `c4`
- `08` -> `89`
- `10` -> `13`
- `20` -> `26`
- `40` -> `4c`
- `80` -> `98`

not this:
- `01` -> `31`
- `02` -> `62`
- `04` -> `c4`
- `08` -> `88`
- `10` -> `10`
- `20` -> `20`
- `40` -> `40`
- `80` -> `80`

ProfBoc75:
I'm updating the code to add the good checksum control and commit it very soon.

```
    uint8_t chk_calc = xor_bytes(b, 12);
    int chk_expected = lfsr_digest8_reflect(&chk_calc, 1, 0x00, 0x31);
    if (... 
```

merbanan:
@zuckschwerdt what is missing from revdgst to figure out this?

Besides that LGTM. This change will most likely be valid for the PWM model also.

merbanan:
@ProfBoc75 this PR https://github.com/merbanan/rtl_433/issues/947 seems to have all the details/data to complete the pwm version of the protocol. Can you look into that also?

ProfBoc75:
> @ProfBoc75 this PR #947 seems to have all the details/data to complete the pwm version of the protocol. Can you look into that also?

Yes, sure, I will check that asap.

merbanan:
> Or like suggested here we could xor all bytes first, then run (any) LFSR -- that could be a more general case perhaps?

Basically this is xor on all but the last message byte that instead uses a non linear mapping for the last byte. The question is if that has a good rf property regarding message validation.

I cant really see any other reason then to make the message unique. By doing this last step we can have a lot of protocols that have the exact same rf parameters but uses different values for the sbox. The hardware implementation would be simple. 12 xor and then one 256 bytes lookup table.

So I would not say it is a broken implementation but rater the tiniest step-up from xor. The sbox design is the tricky part so you need to generate it.

Anyway I think this is a worthy addition to revdgst. I do remember we had some other decoder where we also only where able to verify 4 bits of data.

ProfBoc75:
I found PWM Samples here [Holman-WH5029-samples.zip](https://github.com/merbanan/rtl_433/files/4030361/Holman-WH5029-samples.zip) from the subject PR https://github.com/merbanan/rtl_433/issues/947 , sounds good to start from here, since the other files are truncated , missing end of the message which contains the checksum or whatever we can call it 😄 ... to be continued ...

zuckschwerdt:
@merbanan I don't think it's a deliberate design. It clears the lower bits because the key is 0x31 (just the identity bit in the lower nibble) and that we don't rotate but shift the bits off. There is no LUT, it's just a mask that shifts.
Doing it this way has no useful properties (except obfuscation maybe). A digest (LFSR) would be robust, a CRC (0x31 is common choice there) would be, but 1 round of linear S-Box looks like a fail ;)
Anyway, yes, I'll add that to revdgst and run it on the some mystery checksums.

ProfBoc75:
Hi, 

For the PWM, it's not the same algorithm, same approach, XOR of the bytes but the last, but not the same result as for the PCM.
Several XOR values are common to several expected results and the reverse too, so not possible to exactly map or reverse the control from the expected to XOR value with this method.
And it's not related to RF , the samples are good and the message is repeated 6 times the same into the frame. 

data | XOR | 11th Byte
-- | -- | --
aaa5984c00e42401704a80 | 20 | 80
aaa5984c00ea2601704deb | 2B | eb
aaa5984c00ec2501704fbc | 2C | bc
aaa5984c00ec2601704ded | 2D | ed
aaa5984c00f5240170479c | 3C | 9c
aaa5984c00ef2401705c9d | 3D | 9d
aaa5984c00ea2301702271 | 41 | 71
aaa5984c00ed2401702dee | 4E | ee  --> Duplicate 11th Byte  EE, issue here, XOR 4E or 7E
aaa5984c00e72501703cc4 | 54 | c4
aaa5984c00e42601703d95 | 55 | 95
aaa5984c00f12301702f67 | 57 | 67 --> Duplicate XOR 57, issue here, 11th byte 67 or C7 
aaa5984c00e32501703bc7 | 57 | c7 --> Duplicate XOR 57, issue here, 11th byte 67 or C7 
aaa5984c00ed2401703bf8 | 58 | f8
aaa5984c00ed24017039fa | 5A | fa
aaa5984c00ee2601703d9f | 5F | 9f
aaa5984c00ef24017001c0 | 60 | c0
aaa5984c00df2301703d5b | 6B | 5b
aaa5984c00e72401701fd6 | 76 | d6
aaa5984c00e82401701ed8 | 78 | d8
aaa5984c00ed2401701dde | 7E | de --> Duplicate XOR 73, issue here,  11th byte DE or EE
aaa5984c00ed2501701cee | 7E | ee --> The worse case, duplicate for XOR and for 10th Byte ... 

FYI :  rtl_433 -r g*.cu8 -X "n=TEST,m=FSK_PWM,s=488,l=976,g=2000,r=6000,invert" to get the above values, I removed the last byte, not relevant, (always 00 or 80).

zuckschwerdt:
I have an old test algorithm that I don't quite follow anymore, but it seems this is the same pattern but a key of 0x18?

ProfBoc75:
> I have an old test algorithm that I don't quite follow anymore, but it seems this is the same pattern but a key of 0x18?

0x18 doesn't match, galois = lfsr_digest8_reflect(&xor, 1, 0x00, 0x18)

holman_ws5029pwm_decode: 11th byte 5b xor 6b galois e8
holman_ws5029pwm_decode: 11th byte 71 xor 41 galois 18
holman_ws5029pwm_decode: 11th byte 67 xor 57 galois c8
holman_ws5029pwm_decode: 11th byte c0 xor 60 galois 00
holman_ws5029pwm_decode: 11th byte d8 xor 78 galois 40
holman_ws5029pwm_decode: 11th byte 9d xor 3d galois 38
holman_ws5029pwm_decode: 11th byte 80 xor 20 galois 00
holman_ws5029pwm_decode: 11th byte ee xor 4e galois 90
holman_ws5029pwm_decode: 11th byte 9c xor 3c galois 20
holman_ws5029pwm_decode: 11th byte fa xor 5a galois 70
holman_ws5029pwm_decode: 11th byte d6 xor 76 galois d0
holman_ws5029pwm_decode: 11th byte de xor 7e galois 10
holman_ws5029pwm_decode: 11th byte f8 xor 58 galois 40
holman_ws5029pwm_decode: 11th byte c4 xor 54 galois e0
holman_ws5029pwm_decode: 11th byte c7 xor 57 galois c8
holman_ws5029pwm_decode: 11th byte c4 xor 54 galois e0
holman_ws5029pwm_decode: 11th byte bc xor 2c galois a0
holman_ws5029pwm_decode: 11th byte ee xor 7e galois 10
holman_ws5029pwm_decode: 11th byte ed xor 2d galois b8
holman_ws5029pwm_decode: 11th byte 9f xor 5f galois 08
holman_ws5029pwm_decode: 11th byte 95 xor 55 galois f8
holman_ws5029pwm_decode: 11th byte eb xor 2b galois e8

zuckschwerdt:
Actually something more like this:

Add all odd rows (indexes 0, 2, ...) to result0, add all even rows (indexes 1, 3, ...) to result1.
Add result0, result1, result0 << 4, result0 << 5.
I.e. the shifting doesn't need to be in the loop. It effectivly just folds half the (final) lower bits up.
```
uint8_t xor_shift_bytes(uint8_t const message[], unsigned num_bytes, uint8_t shift_up)
{
    uint8_t result0 = 0;
    for (unsigned i = 0; i < num_bytes; i += 2) {
        result0 ^= message[i];
    }
    uint8_t result1 = 0;
    for (unsigned i = 1; i < num_bytes; i += 2) {
        result1 ^= message[i];
    }
    uint8_t resultx = 0;
    for (unsigned j = 0; j < 7; ++j) {
        if (shift_up & (1 << j))
            resultx ^= result0 << (j + 1);
    }
    return result0 ^ result1 ^ resultx;
}
```

ProfBoc75:
> Actually something more like this:
> 
> Add all odd rows (indexes 0, 2, ...) to result0, add all even rows (indexes 1, 3, ...) to result1. Add result0, result1, result0 << 4, result0 << 5. I.e. the shifting doesn't need to be in the loop. It effectivly just folds half the (final) lower bits up.
> 
> ```
> uint8_t xor_shift_bytes(uint8_t const message[], unsigned num_bytes, uint8_t shift_up)
> {
>     uint8_t result0 = 0;
>     for (unsigned i = 0; i < num_bytes; i += 2) {
>         result0 ^= message[i];
>     }
>     uint8_t result1 = 0;
>     for (unsigned i = 1; i < num_bytes; i += 2) {
>         result1 ^= message[i];
>     }
>     uint8_t resultx = 0;
>     for (unsigned j = 0; j < 7; ++j) {
>         if (shift_up & (1 << j))
>             resultx ^= result0 << (j + 1);
>     }
>     return result0 ^ result1 ^ resultx;
> }
> ```

Good job ! this one is working , so I will update the PWM part accordingly, and enrich also the comments that are very weak for PWM information (no bitbench, no byte position detail ... ) . Then all will be properly check, not only the lower nibble !

```
chk_calc = xor_shift_bytes(b, 10, 0x18);

holman_ws5029pwm_decode: 11th byte 5b chk_calc 5b
holman_ws5029pwm_decode: 11th byte 71 chk_calc 71
holman_ws5029pwm_decode: 11th byte 67 chk_calc 67
holman_ws5029pwm_decode: 11th byte c0 chk_calc c0
holman_ws5029pwm_decode: 11th byte d8 chk_calc d8
holman_ws5029pwm_decode: 11th byte 9d chk_calc 9d
holman_ws5029pwm_decode: 11th byte 80 chk_calc 80
holman_ws5029pwm_decode: 11th byte ee chk_calc ee
holman_ws5029pwm_decode: 11th byte 9c chk_calc 9c
holman_ws5029pwm_decode: 11th byte fa chk_calc fa
holman_ws5029pwm_decode: 11th byte d6 chk_calc d6
holman_ws5029pwm_decode: 11th byte de chk_calc de
holman_ws5029pwm_decode: 11th byte f8 chk_calc f8
holman_ws5029pwm_decode: 11th byte c4 chk_calc c4
holman_ws5029pwm_decode: 11th byte c7 chk_calc c7
holman_ws5029pwm_decode: 11th byte c4 chk_calc c4
holman_ws5029pwm_decode: 11th byte bc chk_calc bc
holman_ws5029pwm_decode: 11th byte ee chk_calc ee
holman_ws5029pwm_decode: 11th byte ed chk_calc ed
holman_ws5029pwm_decode: 11th byte 9f chk_calc 9f
holman_ws5029pwm_decode: 11th byte 95 chk_calc 95
holman_ws5029pwm_decode: 11th byte eb chk_calc eb

```

zuckschwerdt:
> I also tried for PCM with 0x31, but not the good result, does xor_shift_bytes() universal (to put into util.c ?) or only for this PWM use case ?

It's special for now. I'm not sure how they calculate something like this. It easy but very strange:
You xor all bytes. Then all odd bytes (0, 2, ...) are xored again but shifted by 4 and also by 5.
This sounds so random! It could be that they "fold" not bytes but 16-bit words, with a key of 0x1800 then.

ProfBoc75:
I propose to call this "mic" "Integrity" method "XORSHIFT" , instead of CHECKSUM or any other suggestion before my commit ?

zuckschwerdt:
We only want to indicate the approximate strength. The strength of Sum/Xor and such is "CHECKSUM". We only care if the check is significantly better (Digest/CRC) then it's "CRC".

src/devices/holman_ws5029.c
```diff
@@ -86,6 +87,25 @@ To get raw data
 
 #include "decoder.h"
 
+// see #2419 for more details about the xor_shift_bytes , used by PWM device
```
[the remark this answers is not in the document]
ProfBoc75: Done, but I moved the xor_shift_bytes function to the bottom and enrich both doc-comments for PCM and PWM.
Let me know.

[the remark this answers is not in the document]
zuckschwerdt: Very good, thanks. One thing: `%.01f` is not a sensible format ;) It is often in our code but it should be `%.1f`.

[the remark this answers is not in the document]
ProfBoc75: It's was there from the initial author, I just commit these "minor" corrections.

ProfBoc75:
Hi , 

Please notice a little correction the emax device, I was facing an error with the byte conversion (for wind and rain) , 0x01 is the 0 default value. So we need to remove 1 for each byte to get the good value, but the 0x00 was not properly calculate, expected 255 and it was 65536 because of "int" is 2 bytes data. So I added the 0xff mask to get 1 byte data and the expected 255.
The same approach was already done for the Lux value, that is already properly decoded.

No error since the improvement to AOK with the full checksum. So both Aok and Emax are ready to be merged.

Thanks.

src/devices/emax.c
```diff
@@ -250,7 +250,7 @@ static int emax_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     return ret;
 }
 
-static char const *output_fields[] = {
+static char const *const output_fields[] = {
```
zuckschwerdt, line 253: This is already checked in. Can you rebase?

src/devices/holman_ws5029.c
```diff
@@ -9,31 +11,38 @@
     the Free Software Foundation; either version 2 of the License, or
     (at your option) any later version.
  */
-/**
-Decoder for Holman Industries WS5029 weather station,
-a.k.a. Holman iWeather Station.
-https://www.holmanindustries.com.au/products/iweather-station/
+/** int holman_ws5029pcm_decode(r_device *decoder, bitbuffer_t *bitbuffer)
```
zuckschwerdt, line 14: This should be just `/**` no function link.

src/devices/holman_ws5029.c
```diff
@@ -141,29 +211,61 @@ r_device const holman_ws5029pcm = {
         .fields      = output_fields,
 };
 
-/**
+/** int holman_ws5029pwm_decode(r_device *decoder, bitbuffer_t *bitbuffer)
```
zuckschwerdt, line 214: A function link is `/** @fn int holman…`

ProfBoc75: Done, thx

PR #2421, opened 2023-03-13
fix: make r_device.fields a constant pointer 

obones:
This allows for static "output_fields" to be stored in readonly memory on embedded architectures saving the very limited RAM available.
This saves about 7kB, representing 2.3% of the total installed RAM, which may not sound much but makes quite a difference for other parts of the program.

### DIFF
--- a/include/r_device.h
+++ b/include/r_device.h
@@ -71,7 +71,7 @@ typedef struct r_device {
     struct r_device *(*create_fn)(char *args);
     unsigned priority; ///< Run later and only if no previous events were produced
     unsigned disabled; ///< 0: default enabled, 1: default disabled, 2: disabled, 3: disabled and hidden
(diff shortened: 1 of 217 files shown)


zuckschwerdt:
The `clang-format` wants the asterisk to stick, e.g. `static char const *const output_fields[] = {`

obones:
Bummer, I did not notice any issue from the 5 actions that have run, I'll fix that

zuckschwerdt:
I did once try to check/enforce clang-format, but we have too much "special formatting" in the code still. It's not checked for now. But the idea is to have it automated once some problems with `.clang-format` have been solved (i.e. clang-format should pass more code through unmodified and not e.g. change so many linebreaks and indents).

obones:
Thanks for the explanation about clang-format, I have learnt a new thing here.
I have amended the last commit, I believe this is now correct.

PR #2427, opened 2023-03-17
Fix CM160 energy calculation factor

diepeterpan:
…eadings


diepeterpan:
This is to improve the kWh readings to correlate much closer to the values from the LCD.

PR #2428, opened 2023-03-17
Change CM160 output to double

diepeterpan:
Rounding problem, the display will show e.g. 435, and rtl_433 decoded will show 434

ROUND the number before INT fixed it.


zuckschwerdt:
I'm afraid we don't want to support quirks like rounding for values.
Correction factor to get the same (physical) values as best as possible -> yes; Rounding, truncating -> no.

You might need to round yourself when displaying values.

diepeterpan:
> I'm afraid we don't want to support quirks like rounding for values. Correction factor to get the same (physical) values as best as possible -> yes; Rounding, truncating -> no.
> 
> You might need to round yourself when displaying values.

All good, changed the code to provide the fraction and consumer can round themselves.

PR #2430, opened 2023-03-18
Improve validation for Acurite 896 rain gauge, set priority to 10, and re-enable by default

homeautomaton:
I have both the rain gauge and multiple Oregon sensors, which are the source of the frequent misattributions by the old Acurite 896 rain gauge code. Checking bitbuffer->num_rows clears up the issue described in #353.


zuckschwerdt:
What is the nominal count of repeats for the Acurite 896 rain gauge? Is it 12?

homeautomaton:
> What is the nominal count of repeats for the Acurite 896 rain gauge? Is it 12?

16

zuckschwerdt:
Can you comment that before the `< 12` test, e.g.
```
// The nominal repeat count is 16, require a minimum of 12 rows
```

PR #2434, opened 2023-03-19
Add support for Silvercrest Doorbell T (STKK 16 B1)

tnixeu:
This PR adds support for the Silvercrest Doorbell T. 

The samples for it can be found here: https://github.com/merbanan/rtl_433_tests/pull/451

It seems that just one byte of data is sent from the emitter. I am not sure if this the value I got is the same for all devices, since I have only one unit, 

Is there a way to make this code configurable, in case it varies between devices?

I am not exactly sure about the `reset_limit` and the `gap_limit`.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -251,7 +251,7 @@
     DECL(tpms_eezrv) \
     DECL(baldr_rain) \
     DECL(celsia_czc1) \
-
+    DECL(silvercrest_doorbell_t) \
     /* Add new decoders here. */
 
 #define DECL(name) extern r_device name;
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
Sorry, decoders for simple doorbell-style devices should be written as conf files. See https://github.com/merbanan/rtl_433/tree/master/conf

PWM 25/75 sounds way to fast. We usually expect 24 bits (25 with sync) for x1527 style devices. This looks more like PCM though. The recordings are right on DC you need to offset, e.g. `-f 433.90M` and perhaps try a low gain value or increase the distance.

PR #2446, opened 2023-03-22
changed get_os_temperature() to accommodate the AWR129, which has an extra decimal digit for >100˚ C

homeautomaton:
As a BBQ thermometer, the AWR129 reads higher temperatures than a weather thermometer ever would. The hundreds digit appears, right in order, in the low half of the fifth byte of the message. Since I don't know whether the THR228N and other similar devices handled by the protocol decoder might have something besides "0" in that spot, I've made the code conditional, so that only the AWR129 will be affected.

I tested this in my oven above 200˚C, and the readings increased smoothly at each multiple of 100, rather than wrapping back to zero as it did before this change.

### DIFF
--- a/src/devices/oregon_scientific.c
+++ b/src/devices/oregon_scientific.c
@@ -22,7 +22,7 @@
 #define ID_RGR968   0x2d10
 #define ID_THR228N  0xec40
 #define ID_THN132N  0xec40 // same as THR228N but different packet size
-#define ID_AWR129   0xec41 // same as THR228N
+#define ID_AWR129   0xec41 // similar to THR228N, but an extra 100s digit
 #define ID_RTGN318  0x0cc3 // warning: id is from 0x0cc3 and 0xfcc3
 #define ID_RTGN129  0x0cc3 // same as RTGN318 but different packet size
 #define ID_THGR810  0xf824 // This might be ID_THGR81, but what's true is lost in (git) history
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Are you sure it is `(message[5] & 0x0f) * 100.0F` and not just  `(message[5] & 0x07) * 100.0F` -- because the upper but is already the sign bit!
Then this can be the general case. No `if (hundreds_digit)` needed.

Otherwise in line 378 put this and don't change anything else:
```
float temp_c = get_os_temperature(msg);
if (sensor_id == ID_AWR129) {
    // The AWR129 BBQ thermometer has another digit (maybe it's always zero for other devices, and this conditional isn't needed?)
        temp_c += (message[5] & 0x07) * 100.0F;
}

        /* clang-format off */
```

homeautomaton:
> the upper but is already the sign bit! 

Thanks for the heads up. You'd think I could read the very next line of code! Yes, I'll change to 0x7 for the bit-and operation.

> Then this can be the general case. 

Then I'll remove the conditional, too.

zuckschwerdt:
It seems there are THR228N test samples in #1525 -- if you want to check.

And you need to rebase this PR as the AWR129 is already merged.

zuckschwerdt:
It's complicated to work from a changes "master". In the future us a "feature" branch. For now just force push your master here again.

homeautomaton:
I was able to download and test with the THR228N samples from #1525. It is as we expect, for a positive temperature, the low half of byte 5 is all zeros. Hopefully the case with the many other Oregon Scientific products that share this protocol.

homeautomaton:
I had already pushed this:

https://github.com/merbanan/rtl_433/pull/2446/commits/146b40c134a4890857e829636c168cd16d7d7824

PR #2448, opened 2023-03-23
Added support for Fineoffset WS90

jpochmara:
Update to add support for Fineoffset/Ecowitt WS90.

Have been using this code for about 2 weeks now, seems to be working well.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -67,13 +67,13 @@ typedef union data_value {
 } data_value_t;
 
 typedef struct data {
+    struct data *next; /**< chaining to the next element in the linked list; NULL indicates end-of-list */
     char        *key;
     char        *pretty_key; /**< the name used for displaying data to user in with a nicer name */
-    data_type_t type;
     char        *format; /**< if not null, contains special formatting string */
     data_value_t value;
+    data_type_t type;
     unsigned    retain; /**< incremented on data_retain, data_free only frees if this is zero */
-    struct data *next; /**< chaining to the next element in the linked list; NULL indicates end-of-list */
 } data_t;
 
 /** Constructs a structured data object.
--- a/src/devices/fineoffset_ws90.c
+++ b/src/devices/fineoffset_ws90.c
@@ -98,7 +98,7 @@ static int fineoffset_ws90_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int uv_index    = (b[13]);
     int unknown     = (b[14] << 8) | (b[15]);
     int rain_raw    = (b[19] << 8 ) | (b[20]);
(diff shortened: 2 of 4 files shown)


jpochmara:
Updated code to correct code style issues.

zuckschwerdt:
Thanks. So basically a WS80 with some additional fields that still need to be identified?
Any other noteworthy changes? Perhaps you could add a paragraph before "Also sold by EcoWitt." and compare to the WS80.

jpochmara:
That is what it looks like to me. The bytes 1-13  convey the same information between the WS80 and WS90.
Bytes 14 & 15 are still unknown, but seem to have a different value between the two.
Bytes 15-29 are new, with some of the values known.

The only change I made beside adding support for the new data was that I made the "battery_ok" a status, instead on a level.  Now battery_ok is either a 0 or 1, depending on the state of the battery.

And I can add more detail, but I am not sure where you mean to added it....

zuckschwerdt:
A paragraph above line 21 would be nice.

`battery_mv < 1400 ? 1 : 0;` seems the wrong way around? We should pass through what data is available, a precentage gauge is prefered. See https://triq.org/rtl_433/DATA_FORMAT.html#common-device-data

The `"supercap_v"` key needs a capital `V`;)

jpochmara:
The reason I got the battery_ok backwards is that I am using weewx with the SDR driver,  and the SDR driver
negates the battery_low from Fineoffset devices (!battery_low).  And I did not notice that before.

The battery_lvl or battery_ok is a "made up" data point for the WS80 and WS90, both of these devices just provide a battery voltage reading.   So should the battery_lvl and battery_ok data points just be removed for the these devices??

I will work on adding detail to the file and fix the capitalization issue.  :-)

zuckschwerdt:
The data is normalized, not made up ;) Why not output that info if we have it and it's useful? (I wouldn't want a voltage level or worse just a binary good/bad indicator on any cordless device).

jpochmara:
I am fine with the level, I am not sure the range is quite right.  Right now my WS90 battery voltage reads 3.3V, which produced a battery_lvl of 110%.  The lithium batteries I have in it seem to have a voltage of ~1.7V when new.  Should we adjust the range for this? Or is  a battery_lvl  like 1.10 okay?

zuckschwerdt:
Good point, we likely should also clip at 1.00.

jpochmara:
Just wondering if there is some other work you would like done before this get merged in??

Thanks,

jpochmara:
Code style issues should be fixed now.

jpochmara:
Second time is the charm....

PR #2463, opened 2023-03-29
Added variant to fineoffset_wh1050.c, which contains time data every 2h

blackholefinder:
The "fine offset wh1050" decoder works in principle also with station TFA STRATOS 35.1077.

See the product page here: 
https://www.tfa-dostmann.de/en/product/wireless-weather-station-with-wind-and-rain-gauge-stratos-35-1077/

The only issue is, that every two hours some packages with wrong data is reported.

The test data which belongs to this pull request can be found here:
https://github.com/blackholefinder/rtl_433_tests/commit/976e27372b0dac5f3fa1858b2078f7c6c51cb6f7

### DIFF
--- a/src/devices/fineoffset_wh1050.c
+++ b/src/devices/fineoffset_wh1050.c
@@ -34,15 +34,26 @@ I.e., data packet, wait 48 seconds, two data packets, wait 48 seconds, data pack
 
 The 'Total rainfall' field is a cumulative counter, increased by 0.3 millimeters of rain each step.
 
-Message layout and example:
+The station is also known as TFA STRATOS 35.1077
+See the product page here: https://www.tfa-dostmann.de/en/product/wireless-weather-station-with-wind-and-rain-gauge-stratos-35-1077/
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Great code and well documented, thanks!
This should be exactly the case for a "subtype": different sensors in the same sender. I need to check if adding a subtype is a breaking change though. Not really sure.

blackholefinder:
OK, understand. Braking change is always bad idea.

At the moment I see several options:
- keep old code almost as is, but omit messages which contains the time information (e.g. field B != 0x5). This would lead to consistent weather data, but time information would not be decoded.
- use subtype only for the time information frames (e.g. field B == 0x6)
- only add msg_type field to be able to detect possible wrong weather data (which is in reality time data) in consuming application.
- create new model for this weather station device. But this maybe lead to double decoding, as telegrams would match to both models.
- not support this weather station at all and keep it in my fork only.

Do you have further ideas?
Please tell whats your preference to proceed, than I can adapt the code accordingly.

zuckschwerdt:
I was thinking to just remove the subtype and keep everything else in this PR the same. Consumers will then usually get the weather fields and sometimes a message with just a time information.

Issue #1375, opened 2020-05-07
Add device support for Eurochron EFTH-800 weather station


thorfae:
Hi

I have a request to add support for a new weather station, __Eurochron EFTH-800__. I have tried with the protocol already implemented in RTL_433 named Eruochron (129) but it does not give me any results. I have also tried `rtl_433 -G 4` but did not find any data that looked like the Eurochron sensors.

I have a few pictures of the sensors:
<img src="https://user-images.githubusercontent.com/64894314/81279519-8d930b00-9057-11ea-88a1-caf5c3fcf43d.jpg" alt="back" width="100"/> <img src="https://user-images.githubusercontent.com/64894314/81280318-9506e400-9058-11ea-8047-5316bfec8f4f.jpg" alt="back" width="100"/> <img src="https://user-images.githubusercontent.com/64894314/81280341-9d5f1f00-9058-11ea-980f-fbe818555a39.jpg" alt="back" width="100"/>

I already have RTL_433 running with succes collecting data from Ambientweather-F007TH (protocol number 20) so I know my setup is working.

Using `rtl_433 -S` I have recorded the below signals to cu8-files. I disconnected the antenna from the DVB dongle to avoid any noise from other sensors and it seems to work fine.

[cu8 files.zip](https://github.com/merbanan/rtl_433/files/4592195/cu8.files.zip)

Number  | Temperature<br />Celcius   | Temperature<br />Farenheit  | Relative<br />Humidity | Channel
:-- | :-: | :-: | :-: | :-:
001 |  22.8 | 73.0 | 43 | 1 
002 |  22.8 | 75.9 | 42 | 2
003 |  22.8 | 76.2 | 40 | 5
004 |  22.8 |  -   | 39 | 8
005 |  22.8 |  -   | 38 | 1
006 |  22.8 |  -   | 44 | 2
009 | -18.2 | -0.7 | 35 | 1
011 | -12.9 |  8.7 | 43 | 1
012 | -11.2 | 11.8 | 45 | 1
014 |  -9.7 | 14.5 | 48 | 1
015 |  -5.0 | 23.0 | 57 | 1
024 |   1.5 | 34.7 | 63 | 1
029 |   4.1 | 39.3 | 63 | 1

When I use the pulse analyser (-A) I get different results. The pulse look the same but for some I don't get a suggestion for a flex decoder and for some I get one. For g001 I get an output as well.

[...]

PR #2526, opened 2023-06-08
Fix Eurochron-EFTH800 radio_clock hour decoding

### LINKED ISSUE #2525: Eurochron-EFTH800 bad radio_clock hour readings
Hi, I'm consistently seeing (with rtl_433 git master version) bad radio clock values from a Eurochron-EFTH800 type device. E.g.:
```
{"time" : "2023-06-08 13:32:52", "model" : "Eurochron-EFTH800", "id" : 2496, "channel" : 1, "battery_ok" : 1, "temperature_C" : 45.300, "humidity" : 15, "mic" : "CRC", "radio_clock" : "2023-06-08T47:32:53"}
```
(radio_clock: `2023-06-08T47:32:53` at 15:32 local time (CEST, UTC+2), receiver machine is using UTC)

The time hour value is always 32 higher. According to the source code, 1 bit wider mask is used than necessary.
(In this case, those higher 3 bits are always 001, maybe DST and/or TZ info?)

(Unfortunately, I don't own or even see the specific device, so am not able to provide more info.)

mykhal:
This fixes #2525. Related: #1375, #2331.

### DIFF
--- a/src/devices/efth800.c
+++ b/src/devices/efth800.c
@@ -54,8 +54,9 @@ static int eurochron_efth800_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         uint8_t *b = bitbuffer->bb[row];
 
         // 0         1      2       3       4       5    6         7
-        // ?1b CH:3d ID:12d 2b H?6d 2b M:6d 2b S:6d Y?7d D:5d M:4d CHK?8h 1x
-        int dcf77_hour = (b[2] & 0x3f);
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Maybe you can locally track those extra bits for a year and confirm if it's a TZ flag? There could also be leap-year and leap-second flags?

mykhal:
I'll try to observe this long-term. I can also upload the signals (did not want to do it yet, because I don't know exact model of the device, as I mentioned).

zuckschwerdt:
Thanks, no rush. Signals are not that interesting. If you someday happen to find a pattern in those bits let us know, we are always keen to have most accurate protocol descriptions.

Issue #1024, opened 2019-03-26
New Device - Nissan Leaf TPMS


gcormier:
Test data
https://github.com/merbanan/rtl_433_tests/pull/269

Figured I would bring the discussion here.

Second guessing myself here - did I waste an evening decoding the raw bitstream? Or will the -X output the DMC decoded values? I feel like it's the former... :(

[...]

PR #2536, opened 2023-06-16
 Add support for Nissan Leaf TPMS

zuckschwerdt:
Code from https://github.com/alwilson/rtl_433

S.a. #1024 and discussion #2157


zuckschwerdt:
Updated and manually merged with d34e48d

Issue #2556, opened 2023-07-06
Analyzing Shenzhen EGQ TPMS Samples


js-3972:
I have this TPMS from Amazon that can be found at https://www.amazon.com/dp/B0BK8SHDRZ?psc=1&ref=ppx_yo2ov_dt_b_product_details

When I run rtl_433, it recognizes the sensor as the Abarth-124Spider, however, none of the values (pressure, temperature, etc.) appear to be matching what the sensor is displaying. 

I have ran "rtl_433 -S all" to grab samples of some of the received frames. I have attached some of the cu8 files I captured below but I'm unsure if the data I have captured is even valid. For example, I'm confused at to why the modulation scheme varies between the files even though they were taken with the same device or why some packets are longer than others. It would be much appreciated if someone couId provide some guidance on what the next steps are with the data I have captured.
[ShenzenEGQ.zip](https://github.com/merbanan/rtl_433/files/11972054/ShenzenEGQ.zip)

ProfBoc75:
Hi @js-3972 ,

The length of the capture is not an issue, it depends on the rf conditions, are you isolate, or living in big city, for the last, you may capture a lot of rf signals.

Can you try with rtl_433 -R 123 , since all your cu8 files sounds to be properly decoded with this decoder disabled by default.

For all the 4 cu8 files :
![image](https://github.com/merbanan/rtl_433/assets/62882637/e8a2d9b4-414d-4107-90c8-825e6e06dea8)

with -C customary :
![image](https://github.com/merbanan/rtl_433/assets/62882637/c09391b4-f27a-45e8-80d9-a807268a7e7f)

Check if it's ok with -R 123. Also , what are the values and the unit on the console ? If you capture again, for each cu8 file, put the information you have, with the unit (C / F , PSI / Bar / Kpa .. )

js-3972:
Hi @ProfBoc75 ,

The device that's being recognized is the Abarth-124Spider, not the Jansite TPMS, so I ran "-R 156" instead. I have attached a few cu8 files that were generated. As you can see, the rtl software displays a value of 31 c and 48 kpa, but the device's console displays 15 psi and 26 c.
![image](https://github.com/merbanan/rtl_433/assets/135895245/6da40701-d170-4fa3-bfa1-a1b5eb523ab2)

Attached are all of the cu8 files generated with the previous values. 
[15psi_26c.zip](https://github.com/merbanan/rtl_433/files/11984688/15psi_26c.zip)

[...]

PR #2558, opened 2023-07-09
Add support for Shenzhen EGQ Q85 and improve Abarth 124 Spider

ProfBoc75:
Hi,

This is a PR related to #2556 Analyzing Shenzhen EGQ TPMS , similar to Abarth 124 Spider but not properly decoded.
Issue reported by @js-3972.

3 majors gaps have been identified and added into this PR:
- EGQ Q85 data layout is common to Abarth 124 Spider (72 bits) plus a CRC 16 CCITT-FALSE data (96 bits).
- Temperature (°C) offset is 55 C instead 50
- Pressure (KPa) is by 3 instead of 1.38

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [153]  Cotech 36-7959, SwitchDocLabs FT020T wireless weather station with USB
     [154]  Standard Consumption Message Plus (SCMplus)
     [155]  Fine Offset Electronics WH1080/WH3080 Weather Station (FSK)
-    [156]  Abarth 124 Spider TPMS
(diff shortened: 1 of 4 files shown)


gdt:
This PR is in a conflicting state and needs to be rebased/cleaned up.

merbanan:
@ProfBoc75 can you look into the PR? Parts of it looks good from the logic and just needs some cleanup.

ProfBoc75:
> @ProfBoc75 can you look into the PR? Parts of it looks good from the logic and just needs some cleanup.

I will cancel this pr and open another one, as fineoffset_wh1050.c commits should not be there, probably issue with my github when I created this PR as I was working the fineoffset just before this tpms decoder.

PR #2565, opened 2023-07-14
Example hass typo fixes

DigiH:
I think this is a typo/legacy definition in the hass example, as only ever **rain_rate_mm_h** is used in the actual decoders.


zuckschwerdt:
Good find about the `rain_rate_mm_h`. But the other two have conversions applied and are thus reported in `mm` not `in`.

DigiH:
True ;) reverted them back to mm and mm/h

PR #2578, opened 2023-07-27
README.md: set long=short for FSK_MC_ZEROBIT

atlanticn:
add documentation for FSK_MC_ZEROBIT, long should be set to same value as short

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,6 +392,8 @@ PPM     short: Nominal width of '0' gap [us]
 PWM     short: Nominal width of '1' pulse [us]
          long: Nominal width of '0' pulse [us]
          sync: Nominal width of sync pulse [us] (optional)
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks. Yes, this might need documentation.

It is implied by this
```
PCM/RZ  short: Nominal width of pulse [us]
         long: Nominal width of bit period [us]
```
Because MC is just a decoding step after PCM.
The long width isn't really used, MC could be on top of most things, e.g. RZ, and only then long would be needed.

Anyway, source is actually here https://github.com/merbanan/rtl_433/blob/master/src/devices/flex.c#L363

atlanticn:
Ok I'll close the issue, it's good to be documented somewhere that short=long for FSK_MC

PR #2581, opened 2023-07-31
Add battery_mV to rtl_433_mqtt_hass, change moisture class

fallenpegasus:
Moisture should be "moisture" class, not "humidity" class.
Some devices have battery_mv in addition to battery_ok.

### DIFF
--- a/src/devices/revolt_nc5462.c
+++ b/src/devices/revolt_nc5462.c
@@ -65,6 +65,9 @@ static int revolt_nc5462_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     uint8_t *b = bitbuffer->bb[0];
 
     int sum = add_bytes(b, 11);
+    if (sum == 0) {
+        return DECODE_FAIL_SANITY;
+    }
     int chk = b[11];
     if ((sum & 0xff) != chk) {
         return DECODE_FAIL_MIC;


zuckschwerdt:
Thanks. We don't output `battery_mv` but `battery_mV` :)
Yes, class of `moisture` is valid according to https://www.home-assistant.io/integrations/sensor/ (I guess it wasn't before?)

noahwilliamsson:
Re: fix moisture class.  I'm thinking that the `"object_suffix": "H"` should be changed to `"object_suffix": "M"`.  It's used when computing the MQTT publishing path.  I assume it's in place to avoid name conflicts when a device has multiple sensors.

- https://github.com/merbanan/rtl_433/blob/master/examples/rtl_433_mqtt_hass.py#L751-L759

```
def publish_config(mqttc, topic, model, object_id, mapping, value=None):
    """Publish Home Assistant auto discovery data."""
    global discovery_timeouts

    device_type = mapping["device_type"]
    object_suffix = mapping["object_suffix"]
    object_name = "-".join([object_id, object_suffix])

    path = "/".join([args.discovery_prefix, device_type, object_id, object_name, "config"])
```

For reference, from https://www.home-assistant.io/integrations/sensor/#device-class

>  **humidity**: Percentage of humidity in the air in %
>  **moisture**: Percentage of water in a substance in %

zuckschwerdt:
@fallenpegasus can you include the `object_suffix` change so we can merge this?

fallenpegasus:
I'm going to resubmit this as two pull requests, incorporating the requested changes.

PR #2585, opened 2023-08-02
misc fixes for Ecowitt WS90

jktjkt:
Thanks a lot for an excellent SDR SW, and to @jpochmara for this driver. Just some minor cleanups.

## WS90: docs: fix an error in field description

The elementary unit is 10 lux, not 10'000 lux. The expressoin and the reported value are correct, just the English text was wrong.

Fixes: f8489e9b Add support for Fineoffset WS90 (#2448)

## WS90: include correct bytes in the "extra-data" field

This got garbled in the commit which added FW version discovery.

Fixes: c94ee770 Change Fineoffset-WS90 to add firmware version (#2517)

## WS90: put all unknown fields into a single extra-data blob

Let's get rid of the "unknown" field which is only conditionally returned, and instead put that data into "Extra Data".

The buffer size was arbitrary, but sufficient. This change makes the buffer exactly sized for all the unknown data and the trailing `\0`.

## WS90: don't report supercapacitor voltage through "Extra Data"

`b17` is already used for the supercapacitor voltage, so let's use a visual placeholder instead of just dumping that field.

## WS90: mask out the known rain sensor data in "Extra Data"

### DIFF
--- a/src/devices/fineoffset_ws90.c
+++ b/src/devices/fineoffset_ws90.c
@@ -105,16 +105,15 @@ static int fineoffset_ws90_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int wind_dir    = ((b[7] & 0x20) << 3) | (b[11]);
     int wind_max    = ((b[7] & 0x40) << 2) | (b[12]);
     int uv_index    = (b[13]);
-    int unknown     = (b[14] << 8) | (b[15]);
     int rain_raw    = (b[19] << 8 ) | (b[20]);
     int supercap_V  = (b[21] & 0x3f);
     int firmware    = b[29];
-    char extra[30];
+    char extra[31];
 
     if (battery_lvl > 100) // More then 100%?
         battery_lvl = 100;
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! Can you check if the "unknown" field is `0x3fff` in normal conditions for you? Maybe document that finding if so.

jktjkt:
> Thanks! Can you check if the "unknown" field is `0x3fff` in normal conditions for you? Maybe document that finding if so.

Yup, I haven't seen any different value so far.

For me, the extra data right now are `3fffa0a8d9------8c0bffaffa0000`. That first part slowly changed from `3fff20a8d9` over night to the current `3fffa0a8d9`, so it *could* be either some 16bit LE rain sensor data, or something related to the solar cell. I do not see a pattern (yet) for the other fields.

Pushed an update for these two fixed bytes. The rest might follow if I see something useful. It's a shiny toy, both the HW, the SDR stick as well as the code :).

PR #2586, opened 2023-08-03
Influxdb configurable prefix

jktjkt:
I'm feeding data from various weather sensors to VictoriaMetrics (and Grafana). I did not like metric names that are based on free-text model names because they make it harder to plot, say, temperature across multiple sensor types. Without this patch, I might get multiple metrics about, say, temperature:
- `Fineoffset-WS90_temperature_C`,
- `Fineoffset-WN34_temperature_C`.

Since this project already has a concept of unified property names, I would prefer to see metrics named like `rtl433_temperature_C` which would be further parametrized ("labels" in PromQL/Grafana terms) by the protocol (`-M proto`), sensor ID, channel #, etc.

This patch implements just that. I ended up making the pattern format string customizable a la MQTT  which is one step further than what I needed, but it came essentially free, so why not. That means that I can now run `rtl_433` this like:
```
rtl_433 -f 868.35M -M time:utc -s 300k -F 'influx://[::1]:8428/write?.,metric=rtl433' \
  -M protocol -M level -R 142 -R 221 -R 244 -R 113 -Y autolevel
```
...and I will get nice, unified metric names which enable me to plot a graph like this with a single metric, despite the fact that it's aggregating battery health from three different HW types:

image

Here's how this PR implements that:
1) fix `-Wstringop-truncation` which would silenty truncate URLs longer than 399 characters (because my compiler bugs me about this)
2) make the MQTT formatter reusable
3) use the MQTT formatter by adding the `metric` option to the `inlfux://` reporter
4) add some additional metadata like `protocol` to PromQL/MetricsQL labels


src/output_influx.c
```diff
@@ -89,11 +94,11 @@ static void influx_client_event(struct mg_connection *nc, int ev, void *ev_data)
     }
 }
 
-static influx_client_t *influx_client_init(influx_client_t *ctx, char const *url, char const *token)
+static influx_client_t *influx_client_init(influx_client_t *ctx, char const *url, char const *token, const char *metric_format)
 {
-    strncpy(ctx->url, url, sizeof(ctx->url));
-    ctx->url[sizeof(ctx->url) - 1] = '\0';
+    strncpy(ctx->url, url, RTL433_INFLUX_URL_LENGTH);
```
zuckschwerdt, line 99: I'm not sure about getting rid of the zero termination. We kind of know that the influx_client_t should be zero filled. But glancing at a strncpy without accompanying termination it looks like a bug.

src/output_influx.c
```diff
@@ -327,13 +337,25 @@ static void R_API_CALLCONV print_influx_data(data_output_t *output, data_t *data
...
+        } else if (!strcmp(data->key, "model") && !influx->metric_format[0]) {
+            // non-configurable metric format uses "model" for metric name -> skip
+        } else if (!strcmp(data->key, "time")) {
             // skip
         }
         else if (!strcmp(data->key, "type")
                 || !strcmp(data->key, "subtype")
+                || !strcmp(data->key, "protocol")
                 || !strcmp(data->key, "id")
+                || !strcmp(data->key, "dst_id")
```
zuckschwerdt, line 357: `dst_id` is not metadata it's part of the Somfy remote commands, there is no `src_id`?

src/output_influx.c
```diff
@@ -309,15 +314,20 @@ static void R_API_CALLCONV print_influx_data(data_output_t *output, data_t *data
             data_time = d;
     }
 
-    if (!data_model) {
+    if (influx->metric_format[0])  {
+        char metric_name[1000];
+        mbuf_reserve(buf, sizeof (metric_name) + 1000);
+        expand_topic_string(metric_name, influx->metric_format, data, influx->hostname, influx_sanitize_tag);
+        mbuf_snprintf(buf, "%s", metric_name);
+    } else if (!data_model) {
```
zuckschwerdt, line 322: I'm not sure about this? `influx->metric_format` is a user setting but `data_model` is an event property, is there a logic to mixing those within a condition?
Also do we guarantee `influx->metric_format` is always present? Otherwise it should be `if (influx->metric_format && *influx->metric_format)`

include/string_expand.h
```diff
@@ -0,0 +1,16 @@
...
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "data.h"
+
+typedef char *(expand_string_sanitizer)(char *, char *);
+
+char *expand_topic_string(char *topic, char const *format, data_t *data, char const *hostname, expand_string_sanitizer sanitizer);
```
zuckschwerdt, line 16: I think we could run the sanitizer after each expand_topic_string call (it would be more efficient for multiple expansions and less if there was no expansion, but overall neglible). Less nesting/passing of calls and a clear order of expand then sanitize in the calling code. What do you think?

zuckschwerdt:
Great stuff! This was already requested IIRC.

src/output_influx.c
```diff
@@ -89,11 +94,11 @@ static void influx_client_event(struct mg_connection *nc, int ev, void *ev_data)
...
 }
 
-static influx_client_t *influx_client_init(influx_client_t *ctx, char const *url, char const *token)
+static influx_client_t *influx_client_init(influx_client_t *ctx, char const *url, char const *token, const char *metric_format)
 {
-    strncpy(ctx->url, url, sizeof(ctx->url));
-    ctx->url[sizeof(ctx->url) - 1] = '\0';
+    strncpy(ctx->url, url, RTL433_INFLUX_URL_LENGTH);
     snprintf(ctx->extra_headers, sizeof (ctx->extra_headers), "Authorization: Token %s\r\n", token);
+    strncpy(ctx->metric_format, metric_format, RTL433_INFLUX_METRIC_LENGTH);
```
zuckschwerdt, line 101: Clang is right, this is a bug ;)

src/output_influx.c
jktjkt: You're right that these are not metadata, but to me, the real question is whether this is something that I, as a consumer of these data, should be filtering on, or whether it's some runtime data that I want to plot. In InfluxDB terms (cf. [line protocol](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/)), these "filterable" items correspond to tags, in [Prometheus](https://prometheus.io/docs/practices/naming/) these are called labels.

I don't use Somfy-IOC myself (it's just some neighbor's roller shutters that I hear clearly in this apartment building), but the data looked pretty static and I was thinking that it makes more sense to produce separate time series based on the (source, destination) pairs. That way users can track communication of the individual device pairs as separate series (if they choose to).

I'll move this to a separate patch, though, and if you think that it's still a wrong idea, I'm OK with dropping it as well.

zuckschwerdt: It just struck me as pretty arbitrary to filter this one random key. It might look similar to the "id" key but is really just a random data field like all the others. So I assumed an understandable mixup. Yes, let's discuss (universal) handling in another PR.

include/string_expand.h
zuckschwerdt: Sorry, good as it is. We need to run the sanitizer on each trailing part so we don't remove the e.g. slashes in MQTT topics.

src/output_influx.c
jktjkt: Meanwhile, this has changed upstream and now it's done via `snprintf` in the master branch, so I followed that pattern in this patch respin.

src/output_influx.c
jktjkt: OK, let's drop this part of the patch for now.

src/output_influx.c
jktjkt: changed to `snprintf` (as it's currently done with the URL)

src/output_influx.c
jktjkt: I thought so. If the user has provided a metric name, probably it makes sense to use that rather than the hardcoded, arbitrary value of `rtL_433_<hostname>`. But maybe you would prefer to swap the order of these two if statements, so that (e.g.) the statistics reports still go to the hardcoded metric name? I have no use for these so I don't really care.

(Also, sorry for a "small" delay of two years :). )

merbanan:
@zuckschwerdt any update on this? Is it still valid ?

zuckschwerdt:
I have not looked at this in a long time, and I can't readily test it. Let's just merge and deal with fixes later.

PR #2595, opened 2023-08-12
Add the "battery" key to the home-assistant MQTT script

triplepoint:
The Inkbird sensor in `src/devices/inkbird_ith20r.c` reports "battery", but it doesn't exist in the MQTT field mapping in `examples/rtl_433_mqtt_hass.py`.  This PR fills in that field mapping.


gdt:
Your commits could use squashing or rebasing, it looks like from the messages.  That's just me; other people are fine with that happening on merge.

If this is the first battery, I think it's important to figure out how to align rtl_433s's ideas of battery and HA's, so that these are well-behaved battery entities when they land.  Perhaps you did that and I'll have a look.   Historically there have been multiple kinds of representations in rtl_433, much more cleaned up now.

merbanan:
Why not change inkbird_ith20r.c instead ? I think that is the correct change to do.

zuckschwerdt:
Yes, please fix `inkbird_ith20r.c` (you can reuse this PR if you like). I don't know how this slipped through the key needs to be `battery_ok` (https://triq.org/rtl_433/DATA_FORMAT.html#common-device-data) and the range is `0`–`1`, I think here we need to divide by `100.0`?

triplepoint:
ok, let's abandon this PR in favor of this one https://github.com/merbanan/rtl_433/pull/2596

PR #2596, opened 2023-08-12
move the ith20r inkbird sensor to use battery_ok and the correct float range for the battery value

triplepoint:
The inkbird ith20r use the wrong name and format for the battery data.

Based on feedback from abandoned PR https://github.com/merbanan/rtl_433/pull/2595

### DIFF
--- a/src/devices/inkbird_ith20r.c
+++ b/src/devices/inkbird_ith20r.c
@@ -102,7 +102,7 @@ static int inkbird_ith20r_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     uint32_t subtype = (msg[3] << 24 | msg[2] << 16 | msg[1] << 8 | msg[0]);
     int sensor_num = msg[4];
     uint16_t word56 = (msg[6] << 8 | msg[5]);
-    float battery = msg[7] * 0.01f;
+    float battery_ok = msg[7] * 0.01f;
     uint16_t sensor_id = (msg[9] << 8 | msg[8]);
     float temperature = ((int16_t)(msg[11] << 8 | msg[10])) * 0.1f;
     float temperature_ext = ((int16_t)(msg[13] << 8 | msg[12])) * 0.1f;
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Looks good, thanks! One minor change:
`"battery_ok", "Battery", DATA_FORMAT, "%.1f %%", DATA_DOUBLE, battery,`
isn't really in percent when displayed. Use just `"%.1f"`

src/devices/inkbird_ith20r.c
```diff
@@ -115,12 +115,12 @@ static int inkbird_ith20r_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     data = data_make(
             "model",            "",             DATA_STRING, "Inkbird-ITH20R",
             "id",               "",             DATA_INT,    sensor_id,
-            "battery",          "Battery",      DATA_INT,    battery,
+            "battery_ok",       "Battery",      DATA_FORMAT, "%.1f %%", DATA_DOUBLE, battery,
```
gdt, line 118: how does this build, since the variable is still battery?

src/devices/inkbird_ith20r.c
```diff
@@ -102,7 +102,7 @@ static int inkbird_ith20r_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     uint32_t subtype = (msg[3] << 24 | msg[2] << 16 | msg[1] << 8 | msg[0]);
     int sensor_num = msg[4];
     uint16_t word56 = (msg[6] << 8 | msg[5]);
-    int battery = msg[7];
+    float battery = msg[7] * 0.01f;
```
gdt, line 105: I would suggest renaming to battery_ok as it is confusing to fill battery_ok with battery.    At least I at first thought you had a bug on review.

Also, it seems you believe that the 7-bit field has values from 0 to 100, but you are publishing eg only 0.1 and not 0.11, which is surprising.   I would add a comment that says what's in the field and why the precision is as it is.

Note that I like things explained to a far higher degree than others seem to ask for, but I stand by my opinion!

gdt: The name is fixed -- thanks!   And the comment about comments, well I know I'm an outlier....

gdt (approved):
I'm fine with this being  merged.

src/devices/inkbird_ith20r.c
zuckschwerdt: I do want good documentation too. Hard to enforce, sometimes too bold comments about the nominal case can hide corner cases. E.g. here we have seen some values 0-100, I guess. But can't for certain claim to know there are no other special values. Maybe someday someone could map out all possible values with a variable voltage power supply?

PR #2602, opened 2023-08-16
Fix id allow list mistake in MQTT+HASS example

ajacques:
It looks like the `--ids` argument is broken right now and all devices are being exposed to HomeAssistant. The previous commit 315b98b104612bee7f898f225dde837d82c07243 tried to fix a bug. `id in data` doesn't check if the `data` object has a key called `id`, it's checking to see if the `data` dict has the built-in function [id](https://docs.python.org/3/library/functions.html#id).


zuckschwerdt:
You are right, good catch. Thanks!

PR #2607, opened 2023-08-20
Improve handling of units within HA UI

lmilius:
By changing the device_class attribute to be "wind_speed" or "precipitation", this allows the lovelace UI in HomeAssistant to automatically do unit conversions through the UI. I tested these changes locally and they worked after removing the device from the MQTT integration and running this python script to re-preform the autodiscovery.

Valid device classes can be found here: https://www.home-assistant.io/integrations/sensor/#device-class


zuckschwerdt:
Looks good to me.

Issue #2531, opened 2023-06-14
Add support for Landis+Gyr G5 Integrated FOCUS AXe Power Meter


MidwestTechnologies:
The power company installed new power meters recently so I decided to embark on yet another rtl_433 project.  I've been going it alone for more than a week and decided to have the experts take a stab at it.  I've not been able to come up with anything resembling useful data.

FCC ID: R7PEG1R1S6
https://fccid.io/R7PEG1R1S6

I believe mine is using mode 1 (see page 3 of the test report) at 9.6 kbps (104.17 uS per pulse).

Test report: https://fccid.io/R7PEG1R1S6/Test-Report/Test-Report-2789309

If this unit is sending the LAN ID, this should be 0x900F418E as printed on the meter sticker.  This has been what I have been searching for to no avail.

Sample: [g024_916.15M_250k.cu8.zip](https://github.com/merbanan/rtl_433/files/11746721/g024_916.15M_250k.cu8.zip)

Close (within about 10 seconds) to the time this sample was recorded, my meter was  reading 182 kW delivered, 00000 recorded, "KW" 6.336 kW, 245.031 volts, and "IKW" 0.897 kW.  I can't get all of the readings at the time of transmission because I have to wait for the meter to scroll through.

Throwing this into bit bench, I've gotten close to the classic sync word of  0xD391 but found 0xD351 and 0xD371 at times.  I've been using PCM and short and long = 104 uS in pdv.

GFSK or FSK may be in use according to the information found in the links above.

Bit bench for the above sample: <a href="http://triq.net/bitbench#c=55%2055%2055%2055%2055%2055%2040%200F%20FF%202A%2055%2050%2005%2089%2006%205F%20F7%20FD%20FF%207F%20DF%20F7%20FC%20FF%2004%20DE%2012%2014%2043%2000%2043%20B0%2004%20C1%2001%20D3%2051%202D%20A1%2004%20DE%2012%2014%2043%2040%2040%2016%2044%2021%2060%204A%2016%20BC%20FD%2007%2059%2017%20DC&f=hh%20ID%3Ahh%20b%20CH3d%20TEMP_C%3A12d%20HUM%3Ad%20CRC%3A8h%20%7C%208h%2016h%2016h%20&a=Preamble&m=&r=true&cw=4"> here </a>

There may be two or more packets in one transmission.  The above transmission was recorded without an antenna connected.  Let me know if different sample rates are needed.  I have recorded samples at 1M with worse results.  Thanks.

[...]

PR #2616, opened 2023-09-02
Add support for Gridstream RF protocol from Landis & Gyr meters

krvmk:
This PR is to add initial support for the Gridstream RF protocol as discussed in [#2531](https://github.com/merbanan/rtl_433/issues/2531).

### DIFF
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -497,9 +497,9 @@ convert si
   protocol 268 # Bresser Thermo-/Hygro-Sensor Explore Scientific ST1005H
   protocol 269 # DeltaDore X3D devices
 # protocol 268 # Quinetic
-  protocol 270 # Landis & Gyr Gridstream Power Meters 9.6k
-  protocol 271 # Landis & Gyr Gridstream Power Meters 19.2k
-  protocol 272 # Landis & Gyr Gridstream Power Meters 38.4k
+  protocol 271 # Landis & Gyr Gridstream Power Meters 9.6k
+  protocol 272 # Landis & Gyr Gridstream Power Meters 19.2k
+  protocol 273 # Landis & Gyr Gridstream Power Meters 38.4k
 
 ## Flex devices (command line option "-X")
 
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -497,6 +497,9 @@ convert si
   protocol 268 # Bresser Thermo-/Hygro-Sensor Explore Scientific ST1005H
   protocol 269 # DeltaDore X3D devices
 # protocol 268 # Quinetic
+  protocol 270 # Landis & Gyr Gridstream Power Meters 9.6k
+  protocol 271 # Landis & Gyr Gridstream Power Meters 19.2k
+  protocol 272 # Landis & Gyr Gridstream Power Meters 38.4k
 
 ## Flex devices (command line option "-X")
 
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -278,6 +278,9 @@
     DECL(bresser_st1005h) \
     DECL(deltadore_x3d) \
     DECL(quinetic) \
+    DECL(gridstream96) \
+    DECL(gridstream192) \
+    DECL(gridstream384) \
 
     /* Add new decoders here. */
 
--- a/src/devices/gridstream.c
+++ b/src/devices/gridstream.c
@@ -66,7 +66,7 @@ Subtype 0x55 with a data length of 0x23 can be used for this.
 Known CRC init values can be added to the code via PR when they have been identified.
 
 */
-static struct const crc_init known_crc_init[] = {
+static const struct crc_init known_crc_init[] = {
         {0xe623, "Kansas City, MO", "Evergy-Missouri West"},
         {0x5fd6, "Dallas, TX", "Oncor"},
         {0xD553, "Austin, TX", "Austin Energy"},
@@ -76,7 +76,10 @@ static struct const crc_init known_crc_init[] = {
         {0x2C22, "Santa Barbara, CA", "Southern California Edison"},
         {0x142A, "Washington", "Puget Sound Energy"},
         {0x47F7, "Pennsylvania", "PPL Electric"},
-        {0x22c6, "Long Island, NY", "PSEG Long Island"}};
+        {0x22c6, "Long Island, NY", "PSEG Long Island"},
+        {0x8819, "Alameda, CA", "Alameda Municipal Power"},
+        {0x4E2D, "Milwaukee, WI", "We Energies"},
+        {0x1D65, "Phoenix, AZ", "APS"}};
 
 static int gridstream_checksum(int fulllength, uint16_t length, uint8_t *bits, int adjust)
 {
@@ -131,7 +134,7 @@ static int gridstream_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     char destaddress_str[9]     = "";
     int srcwanaddress           = 0;
     uint32_t uptime             = 0;
-    int clock;
+    int clock                   = 0;
     int subtype;
     unsigned offset;
     int protocol_version;
--- a/src/devices/gridstream.c
+++ b/src/devices/gridstream.c
@@ -1,0 +1,277 @@
+/** @file
+    Decoder for Gridstream RF devices produced by Landis & Gyr.
+
+    Copyright (C) 2023 krvmk
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+ */
+
+/** @fn int gridstream_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+Landis & Gyr Gridstream Power Meters.
+
+- Center Frequency: 915 Mhz
+- Frequency Range: 902-928 MHz
+- Channel Spacing: 100kHz, 300kHz
+- Modulation: FSK-PCM (2-FSK, GFSK)
+- Bitrates: 9600, 19200, 38400
+- Preamble: 0xAAAA
+- Syncword v4: 0b0000000001 0b0111111111
+- Syncword v5: 0b0000000001 0b11111111111
+
+This decoder is based on the information from: https://wiki.recessim.com/view/Landis%2BGyr_GridStream_Protocol
+Datastream is variable length and bitrate depending on type fields
+Preamble
+Bytes after preamble are encoded with standard uart settings with start bit, 8 data bits and stop bit.
+Data layouts:
+    Subtype 55:
+        AAAAAA SSSS TT YY LLLL KK BBBBBBBBBB WWWWWWWWWW II MMMMMMMM KKKK EEEEEEEE KKKK KKKKKK CCCC KKKK XXXX KK
+    Subtype D2:
+        AAAAAA SSSS TT YY LL K----------K XXXX
+    Subtype D5:
+        AAAAAA SSSS TT YY LLLL KK DDDDDDDD EEEEEEEE II K----------K CCCC KKKK XXXX
+- A - Preamble
+- S - Syncword
+- T - Type
+- Y - Subtype
+- L - Length
+- B - Broadcast
+- D - Dest Address
+- E - Source Address
+- M - Uptime (time since last outage in seconds)
+- I - Counter
+- C - Clock
+- K - Unknown
+- X - CRC (poly 0x1021, init set by provider)
+
+*/
+
+#include "decoder.h"
+
+struct crc_init {
+    uint16_t value;
+    char const *location;
+    char const *provider;
+};
+
+/*
+Decoder will iterate through the known values until checksum is validated.
+
+In order to identify new values, the reveng application, https://reveng.sourceforge.io/,
+can determine a missing init value if given several fixed length packet streams.
+Subtype 0x55 with a data length of 0x23 can be used for this.
+
+Known CRC init values can be added to the code via PR when they have been identified.
+
+*/
+static struct const crc_init known_crc_init[] = {
+        {0xe623, "Kansas City, MO", "Evergy-Missouri West"},
+        {0x5fd6, "Dallas, TX", "Oncor"},
+        {0xD553, "Austin, TX", "Austin Energy"},
+        {0x45F8, "Dallas, TX", "CoServ"},
+        {0x62C1, "Quebec, CAN", "Hydro-Quebec"},
+        {0x23D1, "Seattle, WA", "Seattle City Light"},
+        {0x2C22, "Santa Barbara, CA", "Southern California Edison"},
+        {0x142A, "Washington", "Puget Sound Energy"},
+        {0x47F7, "Pennsylvania", "PPL Electric"},
+        {0x22c6, "Long Island, NY", "PSEG Long Island"}};
+
+static int gridstream_checksum(int fulllength, uint16_t length, uint8_t *bits, int adjust)
+{
+    uint16_t crc_count = 0;
+    int crc_ok         = 0;
+    uint16_t crc;
+
+    if ((fulllength - 4 + adjust) < length) {
+        return DECODE_ABORT_LENGTH;
+    }
+    crc = (bits[2 + length + adjust] << 8) | bits[3 + length + adjust];
+    do {
+        if (crc16(&bits[4 + adjust], length - 2, 0x1021, known_crc_init[crc_count].value) == crc) {
+            crc_ok = 1;
+        }
+        else {
+            crc_count++;
+        }
+    } while (crc_count < (sizeof(known_crc_init) / sizeof(struct crc_init)) && crc_ok == 0);
+    if (!crc_ok) {
+        return DECODE_FAIL_MIC;
+    }
+    else {
+        return crc_count;
+    }
+}
+
+static int gridstream_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    data_t *data;
+    uint8_t const preambleV4[] = {
+            0xAA,
+            0xAA,
+            0x00,
+            0x5F,
+            0xF0,
+    };
+    uint8_t const preambleV5[] = {
+            0xAA,
+            0xAA,
+            0x00,
+            0x7F,
+            0xF8,
+    };
+    /* Maximum data length is not yet known, but 256 should be a sufficient buffer size. */
+    uint8_t b[256];
+    uint16_t stream_len;
+    char found_crc[5]           = "";
+    char destwanaddress_str[13] = "";
+    char srcwanaddress_str[13]  = "";
+    char srcaddress_str[9]      = "";
+    char destaddress_str[9]     = "";
+    int srcwanaddress           = 0;
+    uint32_t uptime             = 0;
+    int clock;
+    int subtype;
+    unsigned offset;
+    int protocol_version;
+    int subtype_mod = 0;
+    int crcidx;
+    int decoded_len;
+    offset = bitbuffer_search(bitbuffer, 0, 0, preambleV4, 36);
+    if (offset >= bitbuffer->bits_per_row[0]) {
+        offset = bitbuffer_search(bitbuffer, 0, 0, preambleV5, 37);
+        if (offset >= bitbuffer->bits_per_row[0]) {
+            return DECODE_FAIL_SANITY;
+        }
+        else {
+            decoded_len      = extract_bytes_uart(bitbuffer->bb[0], offset + 37, bitbuffer->bits_per_row[0] - offset - 37, b);
+            protocol_version = 5;
+        }
+    }
+    else {
+        decoded_len      = extract_bytes_uart(bitbuffer->bb[0], offset + 36, bitbuffer->bits_per_row[0] - offset - 36, b);
+        protocol_version = 4;
+    }
+    if (decoded_len >= 5) {
+        switch (b[0]) {
+        case 0x2A:
+            subtype = b[1];
+            if (subtype == 0xD2) {
+                stream_len  = b[2];
+                subtype_mod = -1;
+            }
+            else {
+                stream_len = (b[2] << 8) | b[3];
+            }
+            crcidx = gridstream_checksum(decoded_len, stream_len, b, subtype_mod);
+            if (crcidx < 0) {
+                decoder_log(decoder, 1, __func__, "Bad CRC or unknown init value. ");
+                if ((stream_len == 0x23) && (subtype = 0xAA)) {
+                    /* These data types can be used to find new init values. See comment block on line 67. */
+                    decoder_log_bitrow(decoder, 1, __func__, &b[4], decoded_len * 8, "Use RevEng to find init value.");
+                }
+                return DECODE_FAIL_MIC;
+            }
+            sprintf(found_crc, "%04x", known_crc_init[crcidx].value);
+            switch (subtype) {
+            case 0x55:
+                sprintf(destwanaddress_str, "%02x%02x%02x%02x%02x%02x", b[5], b[6], b[7], b[8], b[9], b[10]);
+                sprintf(srcwanaddress_str, "%02x%02x%02x%02x%02x%02x", b[11], b[12], b[13], b[14], b[15], b[16]);
+                srcwanaddress = 1;
+                sprintf(srcaddress_str, "%02x%02x%02x%02x", b[24], b[25], b[26], b[27]);
+                uptime = ((uint32_t)b[18] << 24) | (b[19] << 16) | (b[20] << 8) | b[21];
+                break;
+            case 0xD5:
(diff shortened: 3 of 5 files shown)


merbanan:
Can you add relevant information from here:
https://usermanual.wiki/Landis-Gyr-Technology/IWRS4/html

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
+/** @file
+    Decoder for Gridstream RF devices produced by Landis & Gyr.
+
+    Copyright (C) 2016 Benjamin Larsson
```
merbanan, line 4: I doubt that I wrote this decoder in 2016. Change so that attribution is correct.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+- Center Frequency: 915 Mhz
+- Modulation: FSK-PCM
+- Bitrates: 9600, 19200, 38400
+- Preamble: 0xAAAA
+- Syncword v4: 0b0000000001 0b0111111111
+- Syncword v5: 0b0000000001 0b11111111111
+
+Datastream is variable length and bitrate depending on type fields
+Preamble
+Bytes after preamble are encoded with 8N1
```
merbanan, line 24: Elaborate that this is standard uart settings and that the actual bit stream is start bit, data and stop bit.

merbanan: Add a link to the protocol wiki: https://wiki.recessim.com/view/Landis%2BGyr_GridStream_Protocol and a note that the decoder is based on information from there.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+    bool crc_ok        = false;
+    char destwanaddress_str[13];
+    char srcwanaddress_str[13];
+    char srcaddress_str[9];
+    char destaddress_str[9];
+    uint32_t uptime;
+    time_t clock;
+    char clock_str[80];
+    int subtype;
+    unsigned offset = bitbuffer_search(bitbuffer, 0, 0, preamble, 24) + 16;
```
merbanan, line 85: Eventually we want to know if it is a V4 or V5 packet. You should use the sync words from the wiki and then align the bits by adjusting the offset manually depending on the version. But the comment below does not clearly indicate if this decoder also supports V5 packets.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+
+                /* clang-format off */
+                        data = data_make(
+                                "model",       "", DATA_STRING,  "LandisGyr GridStream",
+                                "id",          "Source Meter ID", DATA_STRING,  srcaddress_str,
+                                "subtype",     "", DATA_INT,     subtype,
+                                "wanaddress",  "", DATA_STRING,  srcwanaddress_str,
+                                "destaddress", "", DATA_STRING,  destwanaddress_str,
+                                "uptime",      "", DATA_INT,     uptime,
+                                "mic",         "Integrity", DATA_STRING,  "CRC", // CRC, CHECKSUM, or PARITY
```
merbanan, line 129: Remove the comment.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+                    return DECODE_FAIL_MIC;
+                }
+
+                sprintf(destwanaddress_str, "%02x%02x%02x%02x%02x%02x", b[7], b[8], b[9], b[10], b[11], b[12]);
+                sprintf(srcwanaddress_str, "%02x%02x%02x%02x%02x%02x", b[13], b[14], b[15], b[16], b[17], b[18]);
+                sprintf(srcaddress_str, "%02x%02x%02x%02x", b[26], b[27], b[28], b[29]);
+                uptime = ((uint32_t)b[20] << 24) | (b[21] << 16) | (b[22] << 8) | b[23];
+
+                /* clang-format off */
+                        data = data_make(
```
merbanan, line 122: Packet version should be added (maybe later) and I want the unknown data to be in the standard output in hex format so that we can begin to correlate the output to real values. As the CRC is the network id, add that to the output.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+                } while (crc_count < (sizeof(known_crc_init) / 2) && crc_ok == false);
+                if (!crc_ok) {
+                    decoder_log(decoder, 1, __func__, "Bad CRC or unknown init value. ");
+                    return DECODE_FAIL_MIC;
+                }
+
+                /* clang-format off */
+                        data = data_make(
+                                "model",      "", DATA_STRING, "LandisGyr GridStream",
+                                "id",         "", DATA_STRING,    "n/a",
```
merbanan, line 156: Not sure we should output this key.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+                                "subtype",    "", DATA_INT,    subtype,
+                                "mic",        "Integrity", DATA_STRING, "CRC", // CRC, CHECKSUM, or PARITY
+                                NULL);
+                /* clang-format on */
+
+                break;
+            case 0xD5:
+                stream_len = (b[4] << 8) | b[5];
```
merbanan, line 164: This code is duplicated 3 times, move it before the switch.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+        "destaddress",
+        "uptime",
+        "srclocation",
+        "destlocation",
+        "timestamp",
+        "mic",
+        NULL,
+};
+
+r_device const gridstream = {
```
merbanan, line 249: Duplicate this for the other rates and change the name to match the rate.

merbanan:
Looks good, make sure to run the maintainer update script also.

src/devices/gridstream.c
krvmk: I was not certain either. It was not referenced in the wiki, but it is referenced in the gnuradio gridstream decoder. I have seen these broadcasted, but I'm not certain if the timing is different or something else, because I have not seen any pass the CRC check. None of the data is obvious like the others.

merbanan: We if data is present in the bitstream then we should output it. In this case it is hard coded to "n/a".

src/devices/gridstream.c
krvmk: I was contemplating adding a struct around known_crc_init to record the company that is using that CRC init value, but wasn't sure if there could be duplicate and what the implication of that would be.

src/devices/gridstream.c
krvmk: The code is different for packet D2, as it is only a single byte instead of two for 55 and D5.

src/devices/gridstream.c
```diff
@@ -0,0 +1,258 @@
...
+                }
+
+                sprintf(destwanaddress_str, "%02x%02x%02x%02x%02x%02x", b[7], b[8], b[9], b[10], b[11], b[12]);
+                sprintf(srcwanaddress_str, "%02x%02x%02x%02x%02x%02x", b[13], b[14], b[15], b[16], b[17], b[18]);
+                sprintf(srcaddress_str, "%02x%02x%02x%02x", b[26], b[27], b[28], b[29]);
+                uptime = ((uint32_t)b[20] << 24) | (b[21] << 16) | (b[22] << 8) | b[23];
+
+                /* clang-format off */
+                        data = data_make(
+                                "model",       "", DATA_STRING,  "LandisGyr GridStream",
```
zuckschwerdt, line 123: The `model` should be shorter and the only non-alphanum is one dash in the middle. `LandisGyr-GS` maybe?

src/devices/gridstream.c
krvmk: All of the data is unknown in this subtype. Without a captured packet stream that passes CRC, it is hard to go further. I'm okay with either dumping the decoded frame, provided it passes CRC, or just omitting it altogether.

src/devices/gridstream.c
merbanan: That would be nice to know. Try to serve as much info as possible.

src/devices/gridstream.c
merbanan: The CRC check could be factored at least.

src/devices/gridstream.c
merbanan: I want the complete frame dumped in all cases as this will aid RE work.

krvmk: Would it be appropriate to convert it to char with bitrow_asprint_code and output under data, or just use decoder_log_bitrow with level 0?

zuckschwerdt: Do not print anything, it has to be `decoder_log_` so we can redirect to e.g. a web UI.

src/devices/gridstream.c
krvmk: Addressed in [7edd93a](https://github.com/merbanan/rtl_433/pull/2616/commits/7edd93a5aead3c669f4bd8e5dbf2d6d32e9ede96)

src/devices/gridstream.c
krvmk: Added the other rates, and they appear to work, at least with my saved samples. However, I cannot determine a way to output the bitrate in the data table.

src/devices/gridstream.c
krvmk: Addressed in https://github.com/merbanan/rtl_433/pull/2616/commits/7edd93a5aead3c669f4bd8e5dbf2d6d32e9ede96

src/devices/gridstream.c
krvmk: Addressed in https://github.com/merbanan/rtl_433/pull/2616/commits/7edd93a5aead3c669f4bd8e5dbf2d6d32e9ede96

src/devices/gridstream.c
krvmk: Addressed in https://github.com/merbanan/rtl_433/pull/2616/commits/7edd93a5aead3c669f4bd8e5dbf2d6d32e9ede96

src/devices/gridstream.c
krvmk: https://github.com/merbanan/rtl_433/pull/2616/commits/7edd93a5aead3c669f4bd8e5dbf2d6d32e9ede96 added handling for both V4 and V5 and verified with testing that both are being seen and decoded, in live and saved samples.

src/devices/gridstream.c
krvmk: Addressed in https://github.com/merbanan/rtl_433/pull/2616/commits/7edd93a5aead3c669f4bd8e5dbf2d6d32e9ede96

src/devices/gridstream.c
krvmk: Addressed in https://github.com/merbanan/rtl_433/pull/2616/commits/7edd93a5aead3c669f4bd8e5dbf2d6d32e9ede96

gdt:
There are changes to README and man page that seem like they have crept in vs being intended.

include/rtl_433_devices.h
```diff
@@ -254,7 +254,9 @@
     DECL(fineoffset_ws90) \
     DECL(thermopro_tx2c) \
     DECL(tfa_303151) \
-
```
gdt, line 257: Blank line should be left.   While it doesn't matter if it's there or not in the grand scheme of things, commits should not make unintended changes.

src/devices/gridstream.c
```diff
@@ -0,0 +1,323 @@
...
+        {0x5fd6, "Dallas, TX", "Oncor"},
+        {0xD553, "Austin, TX", "Austin Energy"},
+        {0x45F8, "Dallas, TX", "CoServ"},
+        {0x62C1, "Quebec, CAN", "Hydro-Quebec"},
+        {0x23D1, "Seattle, WA", "Seattle City Light"},
+        {0x2C22, "Santa Barbara, CA", "Southern California Edison"},
+        {0x142A, "Washington", "Puget Sound Energy"},
+        {0x47F7, "Pennsylvania", "PPL Electric"},
+        {0x22c6, "Long Island, NY", "PSEG Long Island"}};
+
```
gdt, line 72: Would be good to have a comment explaining if you have to search, or what the grand plan is.   Looks like search, but there is wisdom in somebody's head not in comments right now.

src/devices/gridstream.c
```diff
@@ -0,0 +1,323 @@
...
+            0xF0,
+    };
+    uint8_t const preambleV5[] = {
+            0xAA,
+            0xAA,
+            0x00,
+            0x7F,
+            0xF8,
+    };
+    uint8_t *b = malloc(bitbuffer->bits_per_row[0] / 8);
```
gdt, line 116: Maybe this is normal, but I'm nervous about not storing b's length for bounds checking.

src/devices/gridstream.c
```diff
@@ -0,0 +1,323 @@
...
+                                "destaddress",  "",                 DATA_STRING,    destwanaddress_str,
+                                "uptime",       "",                 DATA_INT,       uptime,
+                                "protoversion",  "",                 DATA_INT,       protocol_version,
+                                "mic",          "Integrity",        DATA_STRING,    "CRC",
+                                NULL);
+                /* clang-format on */
+
+                decoder_output_data(decoder, data);
+                break;
+            case 0xD2:
```
gdt, line 194: spacing is odd here, and that makes it a tiny bit harder to read.

src/devices/gridstream.c
```diff
@@ -0,0 +1,323 @@
...
+            crcidx = gridstream_checksum(decoded_len, stream_len, b, subtype_mod);
+            if (crcidx < 0) {
+                decoder_log(decoder, 1, __func__, "Bad CRC or unknown init value. ");
+                if ((stream_len == 0x23) && (subtype = 0xAA)) {
+                    decoder_log_bitrow(decoder, 1, __func__, &b[4], decoded_len * 8, "Use RevEng to find init value.");
+                }
+                free(b);
+                return DECODE_FAIL_MIC;
+            }
+            sprintf(found_crc, "%04x", known_crc_init[crcidx].value);
```
gdt, line 168: I get what's happening here, but given that location/provider are not precomputed it made me think there was  a bug.  I would prefer to see all three handled the same way so that the reader is led to the right conclusion the first time.

src/devices/gridstream.c
```diff
@@ -0,0 +1,323 @@
...
+                                "mic",          "Integrity",    DATA_STRING, "CRC",
+                                NULL);
+                /* clang-format on */
+
+                decoder_output_data(decoder, data);
+                break;
+            case 0xD5:
+                sprintf(destaddress_str, "%02x%02x%02x%02x", b[5], b[6], b[7], b[8]);
+                sprintf(srcaddress_str, "%02x%02x%02x%02x", b[9], b[10], b[11], b[12]);
+                if (stream_len == 0x47) {
```
gdt, line 213: There are conditional "include this item if this condition is true" constructs.   That would perhaps make this easier to understand, if that fits the situation.

conf/rtl_433.example.conf
```diff
@@ -471,7 +471,9 @@ stop_after_successful_events false
 # protocol 242 # Baldr / RainPoint rain gauge.
   protocol 243 # Celsia CZC1 Thermostat
   protocol 244 # Fine Offset Electronics WS90 weather station
-# protocol 245 # ThermoPro TX-2C Thermometer
+# protocol 245 # ThermoPro TX-2C Thermometer and Humidity sensor
+  protocol 246 # TFA 30.3151 Weather Station
+  protocol 247 # Gridstream decoder
```
gdt, line 476: "decoder" has no content .   Perhaps
```
   protocol 247 # Gridstream (power meters)
```
if that's fair.

src/devices/gridstream.c
krvmk: The decoder attempts all of the known CRC init values until it finds one that works. If they all fail, either the data is actually bad, or the CRC init is not known. In the case of the latter, using the reveng program with several fixed length bytestreams with CRC terminated values can output a previously unknown CRC init value, which is how the current values in the struct were derived.  This struct would continue to be expanded (through PRs) as new ones are identified.

src/devices/gridstream.c
krvmk: As the maximum size of the gridstream data stream is not known (to my knowledge), and the bitbuffer length can be arbitrarily sized, there was no way to set a maximum size of b without the possibility of overflow issues.

src/devices/gridstream.c
krvmk: found_crc is the only one that needs a transform to string from the struct in order to be displayed properly in the output data. The location/provider is already in a string.

src/devices/gridstream.c
gdt: That's irregular, but makes sense.  What I was asking for was for the PR to be amended so that this wisdom would be in the code, so that people reading it, long after the pull request is closed, would see it.  (My review is that when we merge, there should be no reason for anyone to look at the PR history.)

src/devices/gridstream.c
gdt: It would still be nice to regularize and/or comment, to make a reader who is competent but not expert quickly come to the right conclusion.

src/devices/gridstream.c
gdt: See fineoffset_ws80.c and look for DATA_COND.   There is a way to have what is basically an if statement and if true, the item is added, and if not, it isn't.  At least that's my understanding.

doctorkb:
I think this might solve my need -- my water company just swapped out the previous meter for a Landis + Gyr GridLinx Interpreter which supposedly connects to the GridStream network.

Is there anything I can do to help get this out of development and into production? It looks like it maybe just failed a single test on code style...

src/devices/gridstream.c
zuckschwerdt: The last offset you are reading seems `b[37]`? Then clamp the input length from `bitbuffer->bits_per_row[0] - offset - 37` to `38*8` and use a fixed 38 byte buffer. Or any sane upper length really.

zuckschwerdt:
The CRLF problems need to be fixed, indentation fixed, some comments added, style updated (as per the review) and then it should be in good shape to merge. Maybe we need to manually merge with those fixes added?

krvmk:
> The CRLF problems need to be fixed, indentation fixed, some comments added, style updated (as per the review) and then it should be in good shape to merge. Maybe we need to manually merge with those fixes added?



> The CRLF problems need to be fixed, indentation fixed, some comments added, style updated (as per the review) and then it should be in good shape to merge. Maybe we need to manually merge with those fixes added?

I had made most of the recommendations to a separate branch here, but I have not had time to move forward with finishing it up.
https://github.com/krvmk/rtl_433/blob/merge/src/devices/gridstream.c

src/devices/gridstream.c
krvmk: The CRC location is based on an offset value that is stored in the datastream and there have been some discovered streams that were much longer than 38 bytes, but are still unknown at this time on how to interpret. Arbitrarily limiting the size would reduce the streams that would captured and/or potentially cause memory exception.

zuckschwerdt: Assume a sane upper bound, say 50(?) bytes, and abort with a warning if a longer message is encountered. That will additionally gain insight into the protocol. Potentially accepting multiple kB of message does not sound proper. If length is a byte then it's bounded anyway ;)

krvmk: We don't know for certain, but it is believed length is a double byte, as the byte before has always been 0x00. But then again, if it is always 0x00, that is effectively a single byte.

krvmk:
Reset and repushed, but the maintainer_update changed a bunch of other stuff in the README. I'm not sure why.

krvmk:
Should be good to go now.

gdt:
What's the status of this?  I'm looking at #2531 and it sort of seems like if this were merged we could close the issue and be down two items.

merbanan:
LGTM, we need some samples though and a rebase.

krvmk:
> LGTM, we need some samples though and a rebase.

What do you need for samples? There are some in the issue discussion #2531.

gdt:
stdlib.h is described at https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdlib.h.html

It is surprising to me that use of it in rtl_433 is novel.  But
```
./output_mqtt.c:#include <stdlib.h>
./output_influx.c:#include <stdlib.h>
./bitbuffer.c:#include <stdlib.h>
./pulse_data.c:#include <stdlib.h>
./output_file.c:#include <stdlib.h>
./baseband.c:#include <stdlib.h>
./data_tag.c:#include <stdlib.h>
./output_trigger.c:#include <stdlib.h>
./sdr.c:#include <stdlib.h>
./pulse_slicer.c:#include <stdlib.h>
./optparse.c:#include <stdlib.h>
./write_sigrok.c:#include <stdlib.h>
./pulse_detect.c:#include <stdlib.h>
./decoder_util.c:#include <stdlib.h>
./r_util.c:#include <stdlib.h>
./pulse_analyzer.c:#include <stdlib.h>
./rtl_433.c:#include <stdlib.h>
./r_api.c:#include <stdlib.h>
./term_ctl.c:#include <stdlib.h>
./samp_grab.c:#include <stdlib.h>
./mongoose.c:#include <stdlib.h>
./mongoose.c:#include <stdlib.h>
./am_analyze.c:#include <stdlib.h>
./bit_util.c:#include <stdlib.h>
./data.c:#include <stdlib.h>
./confparse.c:#include <stdlib.h>
./getopt/getopt.c:/* Don't include stdlib.h for non-GNU C libraries because some of them
./getopt/getopt.c:# include <stdlib.h>
./getopt/getopt.c:#include <stdlib.h>
./compat_paths.c:#include <stdlib.h>
./output_rtltcp.c:#include <stdlib.h>
./list.c:#include <stdlib.h>
./output_log.c:#include <stdlib.h>
./output_udp.c:#include <stdlib.h>
./pulse_detect_fsk.c:#include <stdlib.h>
./devices/flex.c:#include <stdlib.h>
./devices/blueline.c:#include <stdlib.h>
./fileformat.c:#include <stdlib.h>
```

zuckschwerdt:
> It is surprising to me that use of it in rtl_433 is novel. But

Decoders (src/devices) should not use stdlib because alloc() should not be used. It appears you don't alloc, why do you need stdlib?

gdt:
I guess it's easy enough to remove it and see if there is a warning.

dustMason:
Any chance this could be merged? I'd love to use this functionality! Thanks

zuckschwerdt:
Yes, this is nearly good for merge. We need to find out if `strftime()` could be replaced, move the static crc_init var to a const, and attach the doc-comment with `@fn`.

krvmk:
> Yes, this is nearly good for merge. We need to find out if `strftime()` could be replaced, move the static crc_init var to a const, and attach the doc-comment with `@fn`.

Without adding a local function to convert unix time to replace strftime(), I can only think to output the timestamp as int instead. 

Added the rest and sync with current.

src/devices/gridstream.c
```diff
@@ -0,0 +1,277 @@
...
+        {0xe623, "Kansas City, MO", "Evergy-Missouri West"},
+        {0x5fd6, "Dallas, TX", "Oncor"},
+        {0xD553, "Austin, TX", "Austin Energy"},
+        {0x45F8, "Dallas, TX", "CoServ"},
+        {0x62C1, "Quebec, CAN", "Hydro-Quebec"},
+        {0x23D1, "Seattle, WA", "Seattle City Light"},
+        {0x2C22, "Santa Barbara, CA", "Southern California Edison"},
+        {0x142A, "Washington", "Puget Sound Energy"},
+        {0x47F7, "Pennsylvania", "PPL Electric"},
+        {0x22c6, "Long Island, NY", "PSEG Long Island"}};
```
dustMason, line 79: one more entry:
```
{0x8819, "Alameda, CA", "Alameda Municipal Power"}};
```

zuckschwerdt:
Thanks! One last question, `subtype` will split the output on e.g. MQTT messages. It is meant for something like compound devices where there are multiple very distinct functions, e.g. a fictional PIR motion and temperature sensor, or a meter that can be configured to electric or water. Is this the case here? Should the output be seen as multiple different devices depending on subtype? Or is it more like a `msg_type` where there are alternating values for the same device, e.g. a weather station that sends temp/hum and wind data in different messages.

krvmk:
Honestly, I don't really know. The protocol is not documented and there is limited knowledge on what each subtype is for. These are broadcasts and relays from utility meters in a mesh network, and the only relatively useful data that can be determined at this point is meter uptime. 

If it makes sense to be merged as disabled by default, I'm good with that.

zuckschwerdt:
It's okay to merge as-is. Just beware that MQTT topic might look different than expected and that we might change `subtype` to `msg_type` later to make it more useable and standard-looking.

The error is from `static struct const crc_init known_crc_init`. It should be `static struct crc_init const known_crc_init` I guess (`struct crc_init` is the type name).

krvmk:
That's what I get for committing before a test compile. Should be fixed now. 

I'm without a suitable SDR at the moment, so I can't test with a live signal, but my old captures still work.

osullivryan:
Just wanted to say how awesome this PR is - over a year in the making! Thanks so much to the folks who did this.

PR #2632, opened 2023-09-22
Add SigMF file format support

zuckschwerdt:
Some very basic support for SigMF file reading and writing.

Supports `-r`, `-w`, and `-S`, e.g. `-w myfile.sigmf` or `-r myfile.sigmf` or `-S sigmf:unknown`.

### DIFF
--- a/include/fileformat.h
+++ b/include/fileformat.h
@@ -27,6 +27,7 @@ enum file_type {
     F_FLOAT    = 1 << 1,
     F_1CH      = 1 << 4,
     F_2CH      = 2 << 4,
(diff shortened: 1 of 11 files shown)


merbanan:
Looks ok but it is too much for an in-depth review.

zuckschwerdt:
Thanks, yes, needs some seroius refactoring to integrate smoother. Very much work-in-progress.

adulau:
Thanks a lot for your work! We will also do some tests.

merbanan:
Well I've now had a look and I think we can merge this. Looks very modular.

Issue #2631, opened 2023-09-21
Enable also Protocol 200 # Auriol temperature/rain sensor


MatejsLab:
I understand I can use `rtl_433 -R 200` to enable the * disabled protocol, but I would also like to listen in to other temperature sensors.

I need protocol 200 for the Auriol 4-LD5661 (actually LD5972) rain sensor, and it seems to work fine with `rtl_433 -R 200` in short-term testing.

How can I set the conf file properly as the [provided conf file](https://github.com/merbanan/rtl_433/blob/master/conf/rtl_433.example.conf) with removed # at protocol 200 seems to crash rtl_433

I put the file into `/home/m/.config/rtl_433/rtl_433.conf`, the file below has .txt for GitHub to approve the upload.
[rtl_433.example.conf.txt](https://github.com/merbanan/rtl_433/files/12678771/rtl_433.example.conf.txt)

Is the an issue with my config, why protocol 200 is being disabled by default, and how to enable it properly?

In essence, I would like to enable all standard protocols and protocol 200, 

Thanks in advance
Matej

MatejsLab:
What is that bussines with so much Acurite stations found, they seem fake or **is there one that is changing id** all the time.

I could not belive that all my neighbours have Acurrite rain sensors, i have Auriol that luckily reads fine.

![image](https://github.com/merbanan/rtl_433/assets/28715097/89c92d3e-a9aa-46fd-a115-0f8bc9865bd3)

zuckschwerdt:
We have no reports of false positive with Acurite. Watch them for a longer time to see if the rain_mm looks good and distinct between all.

IIRC, the Fitipower FC0012 tuner you use has problems with gain and thus noise. Use a conf with `pulse_detect autolevel` (default) or use `-Y autolevel` on the command line.

MatejsLab:
I will have to try this **autolevel** feature, those Acurite stations that I don't have, seem to be bogus, there are over 20 of them on different IDs.

What is the best way to install RTL_433 on the Peppermint 10 Linux that does not have RTL in its repo?
I have openhab there on a physical machine that hopefully will be more reliable than VM.

I don't feel like building from a source. Is there a link I can get the compiled version and install manually?

gdt:
This looks like a support request, not a bug, and I think it has run its course and the issue can be closed.  If not, please explain the bug in rtl_433 :-)

MatejsLab:
[...]

Issue #1837, opened 2021-10-17
Help decoding Auriol LD5661 (LIDL)


Gollam:
Hi,
I've been trying to figure out how to decode these samples for a while now. The device reports outside temperature, amount of rain and I guess low battery signal. Once I think I saw some hex data indicating some decoded bits, but I could never produce them again. Probably I'm not doing the analyzing part right. Can someone please help me on how to start working with these samples?
[samples_ld5661_1.zip](https://github.com/merbanan/rtl_433/files/7359755/samples_ld5661_1.zip)

zuckschwerdt:
Try the files on https://triq.org/pdv/ to get a spectrogram. The second file is empty, the other three files only have a tail portion of a signal. I.e. the frame is picked up to late. The last file is picked up as `Acurite-01185M` strangely.

Try to add `-Y autolevel -Y minmax -Y magest -M level -M noise` if that doesn't help to get good sample (with `-S unknown`) try to record fixed time files (`-w myname_433.92M_250k.cu8 -T 10`) and pick one containing the full signal.

It looks like PWM with 1000/2000 µs. Ideally this should work: `-X 'n=name,m=OOK_PWM,s=1000,l=2000,r=3000'`

Gollam:
Thanks, it went much better this time. Attaching some samples, in which I am able to see the temperature bits with the following parameters: `rtl_433 -r *.cu8 -X 'n=name,m=OOK_PPM,s=960,l=1936,g=1952,r=1952'`

d1, d2, d4 is the temperature, but I couldn't able to figure out the rain measurement. It increments by 0.3, in the last sample, it shall be up to 3.9mm.
[samples_ld5661_2.zip](https://github.com/merbanan/rtl_433/files/7362620/samples_ld5661_2.zip)

Gollam:
This is what I have now with rounded up values `-X 'n=name,m=OOK_PPM,s=1000,l=2000,g=2000,r=2000'` :

https://triq.net/bitbench#c=1a80d4f00008a%20%5B0.3mm%5D&c=1a80d4f00008c%20%5B0.9mm%5D&c=1a80d4f00008d%20%5B1.2mm%5D&c=1a80d4f000093%20%5B3.0mm%5D&f=hh%20ID%3Ahh%20TEMP_C%3A8d%20hhhhhh&cw=4

My guess is that the last part contains the rain data, but I don't yet know how.

Gollam:
[...]

PR #2633, opened 2023-09-22
Add Lidl Auriol 4-LD6313 and 4-LD5972 temperature/rain sensor

soosp:
This code is heavily based on auriol_4ld5661.c. It contains minor modifications to support Lidl Auriol 4-LD6313 sensor.

Communication of Lidl Auriol 4-LD6313 temperature/rain sensor (https://ugyfelszolgalat.lidl.hu/SelfServiceHU/s/product2?productId=01t6N000002QZ4PQAW&q=ian%20415894) is very similar to Lidl Auriol 4-LD5661 temperature/rain sensor but the ID and the tips to rain gauge conversion ratio is different.


zuckschwerdt:
Isn't this the exact same protocol (ID is variable anyway) and you just went with a different rain conversion?
With the Auriol 4-LD5661 it was found that the display (at ~ 0.3mm /tip) is way off.

soosp:
Yes this is same protocol. The conversion ratio was got on the following way:
Make 100 tip ba hand and see the increasement of rain gauge the display. In this case I got 24.2 mm at every 100 tips.
But if you doubt it should be better to report the tip counter (too), and user can use his/him own conversion ratio.

zuckschwerdt:
> tip by hand and see the increasement of rain gauge the display. In this case I got 24.2 mm at every 100 tips.

Yes, that is roughly the display on the Auriol 4-LD5661 as well. But it was found to be bad. Can you test with actual water?
See the discussion in #1837

> But if you doubt it should be better to report the tip counter (too), and user can use his/him own conversion ratio.

That's the usual way rtl_433 handles values: report just the data and let the user do advanced calculations. But I with simple values we try to give a standard units output.

soosp:
> > tip by hand and see the increasement of rain gauge the display. In this case I got 24.2 mm at every 100 tips.
> 
> Yes, that is roughly the display on the Auriol 4-LD5661 as well. But it was found to be bad. Can you test with actual water? See the discussion in #1837

Yes, I can test but I had to buy an analog meter before :). And have to wait for rain :).

> > But if you doubt it should be better to report the tip counter (too), and user can use his/him own conversion ratio.
> 
> That's the usual way rtl_433 handles values: report just the data and let the user do advanced calculations. But I with simple values we try to give a standard units output.

OK, but as @merbanan [wrote](https://github.com/merbanan/rtl_433/issues/1837#issuecomment-1149561126):
> If the display unit uses 0.3 then that is the "correct" value. But we can also report more keys, like steps and custom values based on command line options. Then a user can supply its own multiplier and select another key. But presenting the same value as the display unit is sort of the rule we follow. All devices deviate but we can only do so much to correct it.

So what do you think about merge these drivers (4-LD5661 and 4-LD6313) into a common driver and report the tip counter, the rain gauge based on factory conversion factor, and optionally the corrected value? The last two can be based on the device ID.

soosp:
Meanwhile I measured the capacity of the spoon of seesaw in the sensor: it tips at 7.2 ml gauge. The upper diameter of the collector funnel is 96 mm, so the calculations of @Gollam seems can be apply for this sensor too.

zuckschwerdt:
Yes, we usually follow the conversion or standards of a display /head unit. But in this special case we found with good evidence that the display is really broken.

Thanks for doing the measurements and tests to confirm the findings. I hope this helps you to get more accurate values from your device. They should have recalled the displays.

Gollam:
@soosp by the way, I had 4 products of the other type I made the driver for. Those counted the gauge tips very hectically, two did not work at all, one counted only the beginning of rain (the measure on display stayed on 0.3 when I knew for sure that at least 5 mm had fell), and one just randomly missed counts. So I'm just warning you, it's not a very accurate product :) I sold all of them, but writing the decoder was a fun task.

soosp:
@Gollam Thank you for the warning :). I'll make some comparsion with an analog meter, and longer time tests too. We'll see ;)

soosp:
Hi, I updated the PR. Merged the two sensors into a single file. Rain tips are reported now. The conversion factor was not touched, it remained 1 as discussed above.

zuckschwerdt:
Thanks. Are we sure the ID is a fixed type code? Seems it could be a random factory code.

Gollam:
@zuckschwerdt is correct, it is just a factory hardcoded random ID, you cannot decide the type based on it.

soosp:
So it seems better to refer these model by a generic name (e.g.: Auriol-4LDxxxx or Auriol-4LD). What do you think?

soosp:
After some Google search it seems that there are a lot of Auriol weather station with product code 4-LDxxxx. Maybe 4-LD is acronym of "For Lidl"? :) So the driver name should be Auriol-4LD-Rain or Auriol-4LDxxxxx-Rain.

zuckschwerdt:
At least for the model we need to stick with `Auriol-4LD5661` for now. There are people using this and we should not break that name. We document that also 4-LD6313 is supported and wait if other similar sensors are reported.

soosp:
OK, I'll keep the name and add 4-LD6313 to the documentation and comments. BTW the 4-LD5972 seems to be same device. See issue #2631

zuckschwerdt:
Good catch on the other sensor type, thanks!

zuckschwerdt:
Can you change the ID docs (line 25) to something like "- I: ID, 8 bit: factory hardcoded random ID"

soosp:
Sure

soosp:
I have updated the PR as discussed above. Reporting of the rain tips for further processing is left in the driver. It maybe do not break anything.

zuckschwerdt:
Looks good to merge to me.

soosp:
@MatejsLab Could you confirm that Auriol 4-LD5972 sensor works with `rtl_433 -R 200`. Just for documentation purposes :)

soosp:
> Looks good to merge to me.

Thanks! May us wait for confirmation 4-LD5972 and add both to the documentation?

MatejsLab:
I see some serious activity here 

Yes, my Auriol 4-LD5972 seems to work with protocol 200. 
My only observation is that the **rain_mm** counts by 2 mm/tip on RTL_433, but it should be by 1 mm/tip.

Also, the original clock station from LIDL is mismatched with the sensor, it increments by 0.5mm/tip it should be 1 mm.

It has a dual 7ml tipping cup (I measured 7g on a scale).
It has a 95mm capture diameter, so 7088 mm2 capture area, 1mm rain x area = 7088 mm3 = 7ml.

Which converts 1mm rain = 7ml captured = 1 tip
I guess I am correct.

soosp:
@MatejsLab Thank you for the quick confirmation. Special thanks for the calculations confirms the previous ones.

Added 4-LD5972 to the PR.

soosp:
> My only observation is that the rain_mm counts by 2 mm/tip on RTL_433, but it should be by 1 mm/tip.
Maybe your sensor is bouncing?

MatejsLab:
Maybe is not debounced, I am not too bothered if it shows 2x as much and I can compensate in openhab code. 

I like this RTL reading is cumulative and way more responsive than that LCD station. 
**The LCD station is so lazy**, we got a bunch of rain, and shows 5mm. the RTL shows 23mm increase which means 11,5mm for sure.

soosp:
OK, thanks! It seems we are ready :)

zuckschwerdt:
Yes, looks good! One minor thing: the "s" in sensors is missing from the line in `conf/rtl_433.example.conf`

soosp:
Fixed.

zuckschwerdt:
So sorry, I missed another thing: the new "rain" key needs to be registered in "output_fields", add below line 93 ("rain_mm").

PR #2636, opened 2023-09-24
ci: Add pre-commit

mdeweerd:
I propose to add pre-commit.

This PR only enables the hooks that do not detect issues.
I created other PRs that enable other hooks one by one.
Except for cpplint, the issues reported by these other hooks are now fixed and merged.

pre-commit can be run locally to check the code and make corrections before committing.
This PR also adds what is needed to run pre-commit in the CI-workflow.

It helps detects bugs and improve code quality.

Local usage:
```shell
# install pre-commit on the system
pip install pre-commit

# run certain hooks
pre-commit run -a codespell
pre-commit run -a cpplint

# Setup git to run pre-commit on every commit (to run in local clone).
pre-commit install
```

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,48 @@
+---
+name: pre-commit
+on:
+  pull_request:
+  push:
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    env:
+      RAW_LOG: pre-commit.log
+      CS_XML: pre-commit.xml
+    steps:
+      - run: sudo apt-get update && sudo apt-get install cppcheck
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          cache: pip
+          python-version: 3.12.1
+      - run: python -m pip install pre-commit
+      - uses: actions/cache/restore@v4
+        with:
+          path: ~/.cache/pre-commit/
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
I had a test run with some of the tools but the changes are too severe. I would like to have the tools optional and I'd like to see only occasional change suggestions.

I propose to take this much slower, test and tune one tools after the other before adoption.

mdeweerd:
Thanks for checking it out.  I've changed the clang-format hook to manual.

I think the whitespace adjustments (end-of-line/file) and shell script formatting steps are reasonable - but I can also change those to manual.
The spelling and python code message are easy to fix and fairly limited - but again, those can be set to manual as well.

mdeweerd:
I have added a git workflow as well.

I'll make sure that the git workflow passes once the pending "issues"/PRs are decided on - there are some PRs that fix some isues.

- cpplint complains about 'long' : "Use int16/int64/etc, rather than the C type long".  Should these messages be ignored in existing code or should long be changed to a (u)int64_t or (u)int32_t type.

mdeweerd:
I was working on getting cpplint installed ...  Will let you know when that is ok.

mdeweerd:
Except for codespell (fix proposed in #2730),
all pre-commit hooks resulting in changes have been set to "manual" so that they can be enabled one by one.

So this PR will become green once rebased to the merge of #2730.

PR #2644, opened 2023-09-26
doc: Add rtl_sdr driver pointer

### LINKED ISSUE #892: Portable Android signal recorder
Someone ported rtl-sdr to Android and run it via loop back as a rtl_tcp instance (RTL2832U driver). It would be nice if we could add support for rtl_tcp and then write an android app that runs rtl_433 for signal grabbing purposes.

gdt:
This adds a pointer to an Android package that implements rtl_tcp.

Fixes #892


gdt:
Is self-merging allowed, or should I wait, or ?  This seems low risk.

merbanan:
Code logic changes should have an external review. Non logic changes are fine to merge directly.

gdt:
Thanks.   That's a totally reasonable approach!

PR #2648, opened 2023-09-29
Add support for Watts WFHT-RF thermostat

### LINKED ISSUE #2230: Help decoding Watts Thermostat
I have a 12 zone underfloor heating system with wireless thermostats which each have a wheel for setting temperature and a thermistor for sensing the current temp. [MASTER-RAIL NY.PDF](https://github.com/merbanan/rtl_433/files/9950889/MASTER-RAIL.NY.PDF)

The data is PWM encoded with a duty cycle of approx. 1/3, which is correctly detected by the analyzer. Here shown in pulseview:
image

Another sample: https://triq.org/pdv/#AAB1041850025C0108F3208291A291A1A291A291A1A2929291A291A1A2929291A29291A1A1A1A1A2929291A1A1A1A1A29292929291A291A1A291A291A1A1A1A1A1A355

I have started to analyze the data received from the sensors and found the ID and the ANNOUNCE flag (used for pairing with the controller). After power on the thermostat quickly sends 5 packets, then settles down to a slower pace, I think perhaps the pace is also dependent on the current temp versus the set-point, but I'm not sure. 

I expect at least the current temp and the set point, but maybe not in real units like C or F, since the controller does not really need it. Also maybe some checksum. The following bitbench is a power on of a single thermostat and then a gradual sweep of the set-point wheel from min to max. All performed within around 30 seconds, so the sensor value probably did not change much.
image

http://triq.net/bitbench#c=%7B54%7D5ab24970f799e4&c=%7B54%7D5ab24978f79be4&c=%7B54%7D5ab24978f79be4&c=%7B54%7D5ab24978f79be4&c=%7B54%7D5ab249f8f79814&c=%7B54%7D5ab249f8f79814&c=%7B54%7D5ab249f8f686ec&c=%7B54%7D5ab249f0f01620&c=%7B54%7D5ab249ff707408&c=%7B54%7D5ab249f8f07508&c=%7B54%7D5ab249f8f0cd58&c=%7B54%7D5ab249f0f0ce58&c=%7B54%7D5ab249f0f25f14&c=%7B54%7D5ab249f8f65e94&f=ID%3Ahhhh%20ANNOUNCE%3Ab%20bbbbbbbbbbbbbbbbbbbbbbbbbbbbb%0A&a=Preamble&m=5a&i=true&co=24&cl=22&cw=2

aadnehovda:
closes #2230

Added the chksum check. 

Test output:
```
C:\Users\ahovda\source\repos\rtl_433_build>src\Debug\rtl_433.exe -s1M -R0 -R247
rtl_433 version nightly-113-g237c8bc4 branch watts_thermostat at 202309291130 inputs file rtl_tcp RTL-SDR SoapySDR
Use -h for usage help and see https://triq.org/ for documentation.
Trying conf file at "C:\Users\ahovda\source\repos\rtl_433_build\src\Debug\rtl_433.conf"...
Trying conf file at "C:\Users\ahovda\AppData\Local\rtl_433\rtl_433.conf"...
Trying conf file at "C:\ProgramData\rtl_433\rtl_433.conf"...
Disabling all device decoders.
[Protocols] Registered 1 out of 247 device decoding protocols [ 247 ]
Found Rafael Micro R820T/2 tuner
[SDR] Using device 0: Realtek, RTL2838UHIDIR, SN: 00000001, "Generic RTL2832U OEM"
Exact sample rate is: 1000000.026491 Hz
Allocating 15 (non-zero-copy) user-space buffers
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2023-09-30 11:19:06
Model     : Watts-WFHTRF ID        : 28082
Pairing   : 1            Temperature: 25.6 C       Setpoint  : 22.1 C        Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2023-09-30 11:19:06
Model     : Watts-WFHTRF ID        : 28082
Pairing   : 1            Temperature: 25.5 C       Setpoint  : 22.1 C        Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2023-09-30 11:19:06
Model     : Watts-WFHTRF ID        : 28082
Pairing   : 1            Temperature: 25.6 C       Setpoint  : 22.1 C        Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2023-09-30 11:19:07
Model     : Watts-WFHTRF ID        : 28082
Pairing   : 1            Temperature: 25.6 C       Setpoint  : 22.1 C        Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2023-09-30 11:19:07
Model     : Watts-WFHTRF ID        : 28082
Pairing   : 0            Temperature: 25.6 C       Setpoint  : 22.1 C        Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2023-09-30 11:19:10
Model     : Watts-WFHTRF ID        : 28082
Pairing   : 0            Temperature: 25.6 C       Setpoint  : 22.1 C        Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2023-09-30 11:19:13
Model     : Watts-WFHTRF ID        : 28082
Pairing   : 0            Temperature: 25.5 C       Setpoint  : 22.1 C        Integrity : CHECKSUM
```


docs/LINKS.md
```diff
@@ -34,3 +34,6 @@
 
 - [rtl_snr](https://github.com/hdtodd/rtl_snr): 
   snr is a pair of equivalent C and Python programs that catalog and analyze signal-to-noise ratios from devices seen by RTL_SDR dongles and logged in JSON format by rtl_433.
+
+- [rtl_sdr driver](https://f-droid.org/en/packages/marto.rtl_tcp_andro/):
+  rtl_sdr driver is a port of rtl_tcp to Android.  See also the [source code](https://github.com/signalwareltd/rtl_tcp_andro-).
```
gdt, line 39: This appears to be an unintended change.  It might be good to rebase into one squashed commit or a series of clean changesets, to make review easier.    Note that it' is my personal opinion that PRs should be rebaesd into clean changesets, even though it is not project doctrine.   (Project doctrine, I try to remind myself, is to squash on merge.)

src/devices/govee.c
```diff
@@ -198,12 +198,21 @@ static int govee_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
     // Figure out what event was triggered
     if (event == 0xafa) {
         event_str = "Button Press";
+        // The H5054 water sensor does not send a message when it transitions from wet to dry nor does it have a
+        // dedicated message to indicate that it is not wet. However, the sensor only sends a "button press" message if
+        // the button is pressed while the device is dry (no button press message is sent if the button is pressed while
+        // the sensor is wet). Since we know the sensor is dry when a "button press" message is received, "detect_wet:0"
+        // is included in the output when the button is pressed as a workaround to allow the user to transition the
+        // device back to the dry state.
+        wet = 0;
```
gdt, line 211: seems like a merge error.  not sure what's going on.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+#include "decoder.h"
+/**
+Watts WFHT-RF Basic Thermostat
+
+This code is based on a slightly older OEM system created by ADEV in France which
+later merged with Watts. The closest thing currently available seems to be
+https://wattswater.eu/catalog/regulation-and-control/radio-wfht-thermostats/electronic-room-thermostat-with-rf-control-wfht-rf-basic/,
+but it is not known whether they are protocol compatible.
+
+https://github.com/merbanan/rtl_433/issues/2230
```
gdt, line 23: My inclination is not to have that sort of history link, and not to have analyzer output.  Rather, the info about how it works woudl be cleaned up and summarized and just presented.  I think we want to leave the messy path to understanding (not a criticism here -- it is *always* messy) behind.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+    {54}5ab249fb8f9acc
+
+    https://tinyurl.com/2z5jtfuu
+
+    Format string:
+    ID:^16d FLAGS:^4b TEMP:^9d SETP:^9d CHK:^8d
+
+Decoded example:
+
+    ID:28205 FLAGS:0001 TEMP:265 SETP:048 CHK:214
```
gdt, line 86: This isn't really like normal decoders, which describe the format and modulation and have a stylized comment with nibbles.    As such it is harder to understand.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+        // we expect 54 bits
+        if (bitbuffer->bits_per_row[row] != 54) {
+            decoder_log(decoder, 2, __func__, "Length check fail");
+            ret = DECODE_ABORT_LENGTH;
+            continue;
+        }
+
+        bitbuffer_invert(bitbuffer);
+
+        unsigned pos = bitbuffer_search(bitbuffer, row, 0, &preamble_pattern, 8);
```
gdt, line 112: If we are looking for a preamble, why are we strict on length?   Preamble tends to mean that once found we decode so some stray bits before are not such an issue.   I'm not saying this is wrong, but it would be good to have rationale.,

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+
+        if (pos >= bitbuffer->bits_per_row[row]) {
+            decoder_log(decoder, 2, __func__, "Preamble not found");
+            ret = DECODE_ABORT_EARLY;
+            continue;
+        }
+
+        pos += 8;
```
gdt, line 120: Is this 8 the length of the preamble?  Symbolic would be nicer if so, and if not it needs explaining.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+
+        uint8_t id_raw[2] = {0};
+        bitbuffer_extract_bytes(bitbuffer, row, pos, id_raw, 16);
+        unsigned id = (reverse8((id_raw[1])) << 8) | reverse8((id_raw[0]));
+
+        pos += 16;
```
gdt, line 126: This goes with the extract, so I would move it near it and not have space  which makes it look like a separate logical step.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+
+        uint8_t PAIRING = 1;
```
gdt, line 128: I would make these const up higher as part of defining the protocol, vs blurring decode variables with specs.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+        uint8_t flags   = {0};
+        bitbuffer_extract_bytes(bitbuffer, row, pos, &flags, 4);
+        flags = reverse8(flags);
+        unsigned pairing = flags & PAIRING;
+
+        pos += 4;
+
+        uint8_t temp_raw[2] = {0};
```
gdt, line 136: I realize it seems clean to init, but presumably we are extracting into it.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+
+        uint8_t id_raw[2] = {0};
+        bitbuffer_extract_bytes(bitbuffer, row, pos, id_raw, 16);
+        unsigned id = (reverse8((id_raw[1])) << 8) | reverse8((id_raw[0]));
+
+        pos += 16;
+
+        uint8_t PAIRING = 1;
+        uint8_t flags   = {0};
+        bitbuffer_extract_bytes(bitbuffer, row, pos, &flags, 4);
```
gdt, line 130: Is the code extracting 4 bytes into 1 byte of space?

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
...
+        bitbuffer_extract_bytes(bitbuffer, row, pos, setp_raw, 9);
+        unsigned setp = (reverse8((setp_raw[1])) << 8) | reverse8((setp_raw[0]));
+
+        pos += 9;
+
+        uint8_t chk_raw = {0};
+        bitbuffer_extract_bytes(bitbuffer, row, pos, &chk_raw, 8);
+        unsigned chk = reverse8(chk_raw);
+
+        uint8_t chksum = ((id >> 8) + (id & 0xFF) + flags + (temp >> 8) + (temp & 0xFF) + (setp >> 8) + (setp & 0xFF)) & 0xFF;
```
gdt, line 152: I am unclear on the norm, but this feels like extract to var and then recreate bits.  I don't follow why that's the approach vs summing the original bytes, if that's doable.

docs/LINKS.md
aadnehovda: Yes, not sure what happened there, I did not make that change. I agree a rebase would be better but in this case some change was merged in instead. I already made sure to put those changes that could conflict (registration in cmake and rtl_433_devices.h) as a separate commit so it would be easy to rebase. I can do the rebase push again, but perhaps @zuckschwerdt intended something else here.

src/devices/govee.c
aadnehovda: Same as above

src/devices/watts_thermostat.c
aadnehovda: Not really sure since this I'm pretty new to RF decoding. But is the preamble enough to be sure we detect the correct protocol? It's not like this is regulated in any way and any manufacturer is free to make a protocol that starts with those same bits. Checking the length in addition to the preamble is at least more unique. Also, since we check the length, we won't need to do any range checking at the rest of the `bitbuffer_search()` calls (have not checked whether that is necessary).

gdt: I may be confusing preamble with synchronization.   There are various ways for a receiver to identify the first bit.   The question is how that first bit of preamble is lined up; there may be sync first, and then it's a "identify the device type" word.

Certainly you should validate everything you can including length.  I was questioning "always N and preamble starts" vs "after you find the preamble, N bits including preamble to end".   So basically, either you ask "are the first bits the preamble" and expect fixed buffer size, or you search, and it's ok to have junk bits before but you need the right number afterwards.  I can't tell from the code what the plan is and why.   I, more than others, want the reader to be led to understand, partly because I try to leave such notes for future self.  Sometimes I have read code I wrote 20 years ago.

src/devices/watts_thermostat.c
aadnehovda: Please point me to a good example and I can try to follow that. The tinyurl points to a bitbench where those parameters are used (although they might not match exaclty). I took an existing decoder and adapted it so I didn't know exactly what to put where.

src/devices/watts_thermostat.c
aadnehovda: Good idea, will add constants for the field lenghts.

src/devices/watts_thermostat.c
aadnehovda: noted

src/devices/watts_thermostat.c
aadnehovda: Will add these magic numbers to start of decoder function

src/devices/watts_thermostat.c
aadnehovda: The flags field is 4 bits. There's no type for just a nibbe. Alternatively I could extract the (known) flags one bit at the time and ignore the others. I was thinking it would be nice to have them all for some debugging purposes if there are other similar Watts devices that use other flags. Hence the commented out "flags" field in the output below. Let me know if I should include the flags in the output.

src/devices/watts_thermostat.c
aadnehovda: I agree this is redundant. I'll rewrite this and calculate the chksum during decode.

src/devices/watts_thermostat.c
```diff
@@ -0,0 +1,199 @@
+/** @file
+    Watts WFHT-RF Thermostat
```
zuckschwerdt, line 2: Fist doc line (also row 16) needs to end with a dot.

src/devices/watts_thermostat.c
zuckschwerdt: Might be somewhat outdated but https://github.com/merbanan/rtl_433/blob/master/src/devices/new_template.c

src/devices/watts_thermostat.c
aadnehovda: I'll start with a search for the preamble and do a minimum length check from that position (any additional tail bits would then be ignored). However I'm thinking I can hardcode that it expects only one row, is that okay?

src/devices/watts_thermostat.c
gdt: You are probably right then.   But, I find extracting 4 bits to be odd.   I usually see pulling the bytes out to variables and then extracting with bit ops.   I personally like to see if all other bits are the expected values and add some diagnostic-type data elements when not, to help others figure out more.

src/devices/watts_thermostat.c
gdt: I personally like to decode all that is known and emit unknown bits as elements.  I am not sure everybody else likes that.

gdt:
Not sure where we are - I am waiting for a cleanup/rebase and force push.  I realize everybody is busy, and this is hard, so I don't mean to press you about time - just saying that so you aren't waiting for me while I wait for you!

aadnehovda:
> Not sure where we are - I am waiting for a cleanup/rebase and force push. I realize everybody is busy, and this is hard, so I don't mean to press you about time - just saying that so you aren't waiting for me while I wait for you!

Still have not found time to update the doc, which I believe is now the only thing missing. There are some other pending reviews that you can close.

aadnehovda:
@gdt, @zuckschwerdt, not sure how to improve the doc more at this point.

zuckschwerdt:
The line `bitbuffer_search(bitbuffer, row, 0, &preamble_pattern,` should be `bitbuffer_search(bitbuffer, row, 0, preamble_pattern,` and the file needs a newline at the end.

PR #2185, opened 2022-09-24
Add support for WEC-2103 temperature/humidity sensor

tthurnreiter:
This temperature/humidity transmitter came with a cheap weather station.

The only available information on it that I could find is here: https://fcc.report/FCC-ID/WEC-2103

PR #2662, opened 2023-10-06
Adds checksum and battery_ok for WEC-2103.

### LINKED ISSUE #2661: WEC-2103 checksum algorithm
https://github.com/merbanan/rtl_433/pull/2185 added support for WEC-2103, by @tthurnreiter. It never received a resolution about the checksum.

I've spent some time figuring out the protocol of my cheap temp/humidity sensor, and it seems the protocol is identical. Here's a decoder: https://github.com/tommie/tx07k-reverse/

However, I think I also managed to figure out the checksum, though I can't be sure the WEC-2103 and my TX07K-THC (as written on the PCB) are the same. Looking at the WEC-2103 FCC internal photo, it's labeled TX07Y-THC, so very likely the same.

The checksum is a [modified/broken CRC-4](https://github.com/tommie/tx07k-reverse/blob/3907833c7b46ba3116255916ddd04a8eebeb298a/tempdec.py#L28). I'd love to contribute that, but since I don't, technically, have a WEC-2103 to confirm this with, I'm not sure if it would be accepted. Seeking contribution guidance, or help confirming the checksum.

tommie:
It uses a somewhat mangled CRC-4 implementation with polynomial 3.

Confirmed this passes the recordings in https://github.com/merbanan/rtl_433/pull/2185#issuecomment-1257032110.

None of these recordings have batt-low set, but on my device (TK07K-THX, as opposed to WEC-2103's TK07Y-THX) it exists, and they look pretty much identical otherwise.

Closes #2661.


src/devices/wec2103.c
```diff
@@ -39,33 +39,73 @@ Example datagram:
...
+    int battery_low = (b[1] & 0x04) >> 3;
 
     /* clang-format off */
     data_t *data = data_make(
             "model",            "",             DATA_STRING, "WEC-2103",
             "id",               "ID",           DATA_INT,    device_id,
             "channel",          "Channel",      DATA_INT,    channel,
-            //"battery_ok",       "Battery",      DATA_INT,    !battery_low,
-            "flags",            "Flags",       DATA_INT,    flags,
+            "flags",            "Flags",        DATA_INT,    flags,
```
zuckschwerdt, line 103: flags and battery_ok should swap places in this list

src/devices/wec2103.c
```diff
@@ -39,33 +39,73 @@ Example datagram:
 - Example: 82.4F->824->1724->0x6bc
 */
 
+static uint8_t almost_crc4(uint8_t const message[], unsigned nBytes, uint8_t polynomial, uint8_t init)
```
zuckschwerdt, line 42: The doc comment now needs to start with
`/** @fn int wec2103_decode(r_device *decoder, bitbuffer_t *bitbuffer)`
as not to attach to this helper function.

zuckschwerdt:
Can you update the doc-comment with your findings about the alternate model name, the short info about the mangled CRC. And change the list to confirm
`- X: Checksum: mangled CRC-4`

src/devices/wec2103.c
tommie: That also means moving button, since it's also part of the flags? And I updated output_fields to match. Not sure if that's right.

src/devices/wec2103.c
tommie: Done. There's no precedent for this in the code, though. See e.g. blueline.c.

src/devices/wec2103.c
zuckschwerdt: Yes to both :)
We generally order the important/well-known fields first (but "mic" last). I.e.
"model", "id", "channel", "battery_ok", "temperature_F", "humidity", ..., "mic".

src/devices/wec2103.c
zuckschwerdt: Some of the doc-comments are broken, yes. PR if you have the time and like to fix some :)

PR #2663, opened 2023-10-09
Update tpms_eezrv.c

flyboy013:
Checksum and pressure bug fixes.  Add leak detection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
 
 ```
 
+  A "rtl_433.conf" file is searched in "./", XDG_CONFIG_HOME e.g. "$HOME/.config/rtl_433/",
+  "/usr/local/etc/rtl_433/", "/etc/rtl_433/", then command line args will be parsed in order.
 		= General options =
   [-V] Output the version string and exit
   [-v] Increase verbosity (can be used multiple times).
@@ -328,7 +330,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [242]* Baldr / RainPoint rain gauge.
     [243]  Celsia CZC1 Thermostat
     [244]  Fine Offset Electronics WS90 weather station
-    [245]* ThermoPro TX-2C Thermometer
+    [245]* ThermoPro TX-2C Thermometer and Humidity sensor
     [246]  TFA 30.3151 Weather Station
 
 * Disabled by default, use -R n or a conf file to enable
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -471,7 +471,8 @@ stop_after_successful_events false
 # protocol 242 # Baldr / RainPoint rain gauge.
   protocol 243 # Celsia CZC1 Thermostat
   protocol 244 # Fine Offset Electronics WS90 weather station
-# protocol 245 # ThermoPro TX-2C Thermometer
+# protocol 245 # ThermoPro TX-2C Thermometer and Humidity sensor
+  protocol 246 # TFA 30.3151 Weather Station
 
 ## Flex devices (command line option "-X")
 
--- a/man/man1/rtl_433.1
+++ b/man/man1/rtl_433.1
@@ -37,6 +37,10 @@ PlutoSDR, HackRF One (using SoapySDR drivers), as well as SoapyRemote.
 A summary of options is included below.
 Detailed information on some options follows.
 .\" body
+
+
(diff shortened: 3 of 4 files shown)


src/devices/tpms_eezrv.c
```diff
@@ -14,28 +14,28 @@
...
-    PRE CC IIIIII PP TT FF 00
+    PRE CC IIIIII PP TT FF FF
 
 - PRE : FFFF
 - C : 8 bit CheckSum
 - I: 24 bit ID
 - P: 8 bit pressure  P * 2.5 = Pressure kPa
 - T: 8 bit temperature   T - 50 = Temperature C
-- F: 8 bit battery (not verified) and deflation pressure (for sure) flags
+- F: 16  bit battery (not verified), deflation pressure (for sure), pressure modifier flags
```
zuckschwerdt, line 30: `640/2.5 = 256` ;) so please: "pressure MSB"

src/devices/tpms_eezrv.c
```diff
@@ -55,36 +55,50 @@ static int tpms_eezrv_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+    uint8_t b[7]  = {0};
     uint8_t cc[1] = {0};
     bitbuffer_extract_bytes(bitbuffer, 0, pos + 16, cc, sizeof(cc) * 8);
     bitbuffer_extract_bytes(bitbuffer, 0, pos + 24, b, sizeof(b) * 8);
+
     // verify checksum
-    if ((add_bytes(b, 6) & 0xff) != cc[0]) {
+    int computed_checksum = add_bytes(b, sizeof(b));
+    if (computed_checksum > 0xff) 
+        computed_checksum |= 0x80;
```
zuckschwerdt, line 73: Can you add a comment that the top bits fold around? That's what you mean here, right?
Did you observe a checksum bigger than 1FF ? I.e. what would 2FF do?

src/devices/tpms_eezrv.c
```diff
@@ -55,36 +55,50 @@ static int tpms_eezrv_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
         return DECODE_FAIL_MIC;
     }
     char id_str[7];
     sprintf(id_str, "%02x%02x%02x", b[0], b[1], b[2]);
     float pressure_kPa = b[3] * 2.5;
     int temperature_C  = b[4] - 50;
-    int flags          = b[5];
-    char flags_str[3];
-    sprintf(flags_str, "%x", flags);
+    int flags1          = b[5];
```
zuckschwerdt, line 83: are these the same indent as above? looks like one extra space?

src/devices/tpms_eezrv.c
```diff
@@ -55,36 +55,50 @@ static int tpms_eezrv_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+    int flags2          = b[6];
+    int fast_leak       = 0;
+
+    fast_leak = (flags1 & 0x10) && !(flags1 & 0x20);   // Check for fast leak flag set and resolved flag NOT set
```
zuckschwerdt, line 87: perhaps write as
```
int leak_flag = (flags1 & 0x10);
int resolved_flag = (flags1 & 0x20);
int fast_leak = leak_flag && !resolved_flag;
```

zuckschwerdt:
Rebase if possible (otherwise I can do that). Some minor comments, otherwise looks good!

src/devices/tpms_eezrv.c
flyboy013: If the checksum is greater than 0xFF then the MSB is set.  I occurs whether the bit is already set or not and was observed when checksum was in the 0x1FF and the 0x2FF range.  I did not observe a checksum higher than in the 0x200-0x2FF range.  I'm assuming this holds true for higher than 0x2FF, but I have not data to support it.

zuckschwerdt:
Did you overwrite your changes? Don't panic, git still has all the data. Read up on git reflog, e.g. https://stackoverflow.com/questions/10099258/how-can-i-recover-a-lost-commit-in-git

src/devices/tpms_eezrv.c
flyboy013: Changes made to add value in flag field (currently on 1 bit) to the pressure as the MSB.

src/devices/tpms_eezrv.c
flyboy013: Code updated to note value of fast_leak_detected and fast_leak_resolved and to then determine reported fast_leak output field base on these.

flyboy013:
PR accidentally deleted and resubmitted as PR  #2664

zuckschwerdt:
Note for posteriority: we prefer that branches be kept (force push into the same PR) and not resubmitted.

PR #2668, opened 2023-10-11
Fix id_rev portion of hcs200

Crsarmv7l:
Correctly display ID_rev portion. 

Current decoder uses & 0xf0, when tested with a Doorhan remote with id_rev of A29D6890, rtl_433 shows id_rev A9D68900 (attached).

Using & 0xff gives the expected A29D6890.

rtl


Crsarmv7l:
Disregard. b[7] is both part of the serial and the button id. Sorry!

Issue #2541, opened 2023-06-22
Security+ 2.0 Decoder not working


dasrue:
I'm having trouble getting the Security+2.0 decoder to work with my remotes. In the src/devices/secplus_v2.c file I have used that information to build a flex decoder, and run it using `-g 1 -X "n=vI3,m=OOK_PCM,s=230,l=230,t=40,r=10000,g=7400"`. 
This gets me codes from the remotes as multiple messages each with only 1 row. If I combine 2 messages together to 2 rows and give it to the decoder using the -y option it works. I have tried increasing the reset time to 1 second in the flex decoder, but it still always wants to decode as 2 messages. Any ideas why this would be happening?

zuckschwerdt:
Can you check a capture on the actual distances between packets? It should be only about 500 µs

dasrue:
Using `-M time:usec` with the same -X decoder shows that there is just over 100ms between the packets

zuckschwerdt:
Capture a sample with either `-S known` or save a longer stretch with `-w myfile_250k_433.92M.cu8 -T 20`
Then look at the timings in https://triq.org/pdv/

dasrue:
The output from PDV suggests a gap of 10004.0us. If I take the decoded bits from PDV and put them into the security+2.0 decoder using -y it works.
The output I get from PDV is

| Pulses  | 132× 223.9 ±24.0 µs | 27× 472.7 ±20.0 µs |                    |                      |
|---------|---------------------|--------------------|--------------------|----------------------|
| Gaps    | 129× 273.1 ±22.0 µs | 27× 520.7 ±20.0 µs | 2× 10004.0 ±0.0 µs |                      |
| Periods | 110× 497.0 ±18.0 µs | 39× 745.1 ±16.0 µs | 7× 994.3 ±14.0 µs  | 2× 10352.0 ±140.0 µs |
| Timings | 260× 248.4 ±46.0 µs | 54× 496.7 ±48.0 µs | 1× 196.0 ±0.0 µs   | 3× 10004.0 ±0.0 µs   |

zuckschwerdt:
What you'd want would be `-X "n=vI3,m=OOK_PCM,s=230,l=230,t=40,r=12000,g=900"` but I guess that's beyond what rtl_433 expects as sensible timing values.

dasrue:
I'm still getting split up packets using that decoder. I have sent you an email with my capture if that helps anything

zuckschwerdt:
[...]

PR #2673, opened 2023-10-16
Added context to secplus_v2 decoder to allow split packets

dasrue:
Solution for issue #2541

### DIFF
--- a/src/devices/secplus_v2.c
+++ b/src/devices/secplus_v2.c
@@ -19,6 +19,8 @@ Security+ 2.0  is described in [US patent application US20110317835A1](https://p
 */
 
 #include "decoder.h"
+#include "fatal.h"
+#include "compat_time.h"
 
 /**
 Security+ 2.0 rolling code.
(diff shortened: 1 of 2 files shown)


src/devices/secplus_v2.c
```diff
@@ -245,15 +257,16 @@ static int secplus_v2_callback(r_device *decoder, bitbuffer_t *bitbuffer)
...
-    bitbuffer_t fixed_1   = {0};
-    uint8_t rolling_1[16] = {0};
+    struct timeval tv_r1;
+    struct timeval tv_r2;
+    struct timeval tv_now;
 
-    //bitbuffer_t bits_2    = {0};
-    bitbuffer_t fixed_2   = {0};
-    uint8_t rolling_2[16] = {0};
+    struct secplus_v2_stateful_context *context = decoder->decode_ctx;
```
gdt, line 265: This is pretty complicated (which is how it is), but I am not finding comments that explain the grand plan, what info is stored, how the transmitter works, etc.  I am not finding an explanation of whether the two packets are tied together or whether a 1st packet from one can be used with a 2nd from another.   I'm not saying your code is wrong.  Rather, I'm asking for the reader to be led to understand all the hard parts and be able to follow an argument that it's correct.

zuckschwerdt:
The discussion is in #2314 I think. Closing as a dup, correct me if I missed a new concept or idea?

The problems remain: we can't assume a concept of linear (or even real) time. There might be delayed runs, replays, etc.
Also `bitbuffer_t` is not the proper type to store information, data should be parsed already.

And note that currently every decoder that is not static will give up it's protocol number. It will just be some anonymous instance of a decoder.

It's not just a patch of code that's missing but a bigger concept for (multi-stage?) data processing flow.

gdt:
We may wish to discuss this on the mailing list.  There are a lot of complicated issues, and I find that much better for real discussions.  But #2314 is ok if that's what others want.

PR #2681, opened 2023-10-22
shell script formatting enhancing readability

mdeweerd:
Formatted shell scripts improving readability.


zuckschwerdt:
These scripts capture a list of commands used to configure and build parts. They are written and intended to be read top to bottom, line by line with no structure. The few gaurds are just points where you can skip or resume. I don't oppose indentations but I don't think it adds any legibility.

mdeweerd:
IMHO indenting the commands inside the if/else clauses improves readability.  We indent in other languages so why not in a shell script?

gdt:
I think the lack of indentation around if makes it much harder to read.   It is also irregular relative to normal shell style.   I was thinking of submitting a PR to indent myself.

PR #2682, opened 2023-10-22
Do not use xor which is equivalent to ^

mdeweerd:
xor is an alternative for '^' and should be avoided.

The code was updated to use 'xor_byte' as the name for the variable.


zuckschwerdt:
Oh, `iso646.h` (C alternative spellings) baggage. Better safe than sorry, I guess.

Issue #1193, opened 2019-11-02
runtime error: left shift of 162 by 24 places cannot be represented in type 'int'


merbanan:
rtl_433 -R 130 -s 250k -Y 1 -r rtl_433_tests/tests/IKEA_Sparsnas/01/*.cu8

gives

Test mode active. Reading samples from file: rtl_433_tests/tests/IKEA_Sparsnas/01/g001_867.95M_250k.cu8
/home/benjamin/projects/rtl_433/src/devices/ikea_sparsnas.c:116:53: runtime error: left shift of 162 by 24 places cannot be represented in type 'int'

(k0 << 24) -> (((uint32_t)k0 << 24) gets rid of the error message but is it the correct way to do it? Is changing the k0 type better?

zuckschwerdt:
Yes, that's the right change. We have that in all other places like this. C integer promotion rules give us int (int32_t) for the right side, we need to force unsigned (uint32_t) if we use the uppermost bits. Thus k0 should trigger that. I'll push the change.

PR #2683, opened 2023-10-22
Improve code and annotations for cpplint

mdeweerd:
This PR includes:
- Annotations "// NOLINT" so that cpplint does not complain about the line;
- Some changes to the length provides in snprintf when sizeof could be used (fixes cpplint message).

Pending: changes needed for 'long' - specify it as (u)int32_t o (u)int64_t ?

### DIFF
--- a/src/decoder_util.c
+++ b/src/decoder_util.c
@@ -51,7 +51,7 @@ static char *bitrow_asprint_code(uint8_t const *bitrow, unsigned bit_len)
     row_bytes[0] = '\0';
     // print byte-wide
     for (unsigned col = 0; col < (unsigned)(bit_len + 7) / 8; ++col) {
-        sprintf(&row_bytes[2 * col], "%02x", bitrow[col]);
+        sprintf(&row_bytes[2 * col], "%02x", bitrow[col]); // NOLINT
     }
     // remove last nibble if needed
     row_bytes[2 * (bit_len + 3) / 8] = '\0';
@@ -67,7 +67,7 @@ static char *bitrow_asprint_code(uint8_t const *bitrow, unsigned bit_len)
         WARN_MALLOC("decoder_output_bitbuffer()");
         return NULL; // NOTE: returns NULL on alloc failure.
     }
-    sprintf(row_code, "{%u}%s", bit_len, row_bytes);
+    sprintf(row_code, "{%u}%s", bit_len, row_bytes); // NOLINT
 
     return row_code;
 }
@@ -248,7 +248,7 @@ void decoder_output_bitbuffer_array(r_device *decoder, bitbuffer_t const *bitbuf
         row_bytes[0] = '\0';
         // print byte-wide
         for (unsigned col = 0; col < (unsigned)(bitbuffer->bits_per_row[i] + 7) / 8; ++col) {
-            sprintf(&row_bytes[2 * col], "%02x", bitbuffer->bb[i][col]);
+            sprintf(&row_bytes[2 * col], "%02x", bitbuffer->bb[i][col]); // NOLINT
         }
         // remove last nibble if needed
         row_bytes[2 * (bitbuffer->bits_per_row[i] + 3) / 8] = '\0';
--- a/src/decoder_util.c
+++ b/src/decoder_util.c
@@ -51,7 +51,7 @@ static char *bitrow_asprint_code(uint8_t const *bitrow, unsigned bit_len)
     row_bytes[0] = '\0';
     // print byte-wide
     for (unsigned col = 0; col < (unsigned)(bit_len + 7) / 8; ++col) {
-        sprintf(&row_bytes[2 * col], "%02x", bitrow[col]); // NOLINT
+        sprintf(&row_bytes[2 * col], "%02x", bitrow[col]);
     }
     // remove last nibble if needed
     row_bytes[2 * (bit_len + 3) / 8] = '\0';
@@ -67,7 +67,7 @@ static char *bitrow_asprint_code(uint8_t const *bitrow, unsigned bit_len)
         WARN_MALLOC("decoder_output_bitbuffer()");
         return NULL; // NOTE: returns NULL on alloc failure.
     }
-    sprintf(row_code, "{%u}%s", bit_len, row_bytes); // NOLINT
+    sprintf(row_code, "{%u}%s", bit_len, row_bytes);
 
     return row_code;
 }
@@ -248,7 +248,7 @@ void decoder_output_bitbuffer_array(r_device *decoder, bitbuffer_t const *bitbuf
         row_bytes[0] = '\0';
         // print byte-wide
         for (unsigned col = 0; col < (unsigned)(bitbuffer->bits_per_row[i] + 7) / 8; ++col) {
-            sprintf(&row_bytes[2 * col], "%02x", bitbuffer->bb[i][col]); // NOLINT
+            sprintf(&row_bytes[2 * col], "%02x", bitbuffer->bb[i][col]);
         }
         // remove last nibble if needed
         row_bytes[2 * (bitbuffer->bits_per_row[i] + 3) / 8] = '\0';
--- a/src/devices/oregon_scientific.c
+++ b/src/devices/oregon_scientific.c
@@ -93,9 +93,9 @@ static unsigned int get_os_uv(unsigned char *message)
     return uvidx;
 }
 
-static unsigned short int cm180i_power(uint8_t const *msg, unsigned int offset)
+static uint16_t cm180i_power(uint8_t const *msg, unsigned int offset)
 {
-    unsigned short int val = 0;
+    uint16_t val = 0;
 
     val = (msg[4+offset*2] << 8) | (msg[3+offset*2] & 0xF0);
     // tested across situations varying from 700 watt to more than 8000 watt to
@@ -104,15 +104,15 @@ static unsigned short int cm180i_power(uint8_t const *msg, unsigned int offset)
     return val;
 }
 
-static unsigned long long cm180i_total(uint8_t const *msg)
+static uint64_t cm180i_total(uint8_t const *msg)
 {
-    unsigned long long val = 0;
+    uint64_t val = 0;
     if ((msg[1] & 0x0F) == 0) {
         // Sensor returns total only if nibble#4 == 0
-        val = (unsigned long long)msg[14] << 40;
-        val += (unsigned long long)msg[13] << 32;
-        val += (unsigned long)msg[12] << 24;
-        val += (unsigned long)msg[11] << 16;
+        val = (uint64_t)msg[14] << 40;
+        val += (uint64_t)msg[13] << 32;
+        val += (uint32_t)msg[12] << 24;
+        val += msg[11] << 16;
         val += msg[10] << 8;
         val += msg[9];
     }
@@ -124,9 +124,9 @@ static uint8_t swap_nibbles(uint8_t byte)
     return (((byte&0xf) << 4) | (byte >> 4));
 }
 
-static unsigned short int cm180_power(uint8_t const *msg)
+static uint16_t cm180_power(uint8_t const *msg)
 {
-    unsigned short int val = 0;
+    uint16_t val = 0;
     val = (msg[4] << 8) | (msg[3] & 0xF0);
     // tested across situations varying from 700 watt to more than 8000 watt to
     // get same value as showed in physical CM180 panel (exactly equals to 1+1/160)
@@ -134,16 +134,21 @@ static unsigned short int cm180_power(uint8_t const *msg)
     return val;
 }
 
-static unsigned long long cm180_total(uint8_t const *msg)
+static uint64_t cm180_total(uint8_t const *msg)
 {
-    unsigned long long val = 0;
+    uint64_t val = 0;
     if ((msg[1] & 0x0F) == 0) {
         // Sensor returns total only if nibble#4 == 0
-        val = (unsigned long long)msg[10] << 40;
-        val += (unsigned long long)msg[9] << 32;
-        val += (unsigned long)msg[8] << 24;
-        val += (unsigned long)msg[7] << 16;
+        val = (uint64_t)msg[10] << 40;
+        val += (uint64_t)msg[9] << 32;
+        val += (uint32_t)msg[8] << 24;
+<<<<<<< HEAD
+        val += (uint32_t)msg[7] << 16;
+        val += (uint16_t)msg[6] << 8;
+=======
+        val += msg[7] << 16;
         val += msg[6] << 8;
+>>>>>>> f6385da2 (fixup! long/long long to int32_t/int64_t, short int to int16_t and comment fix)
         val += msg[5];
     }
     return val;
@@ -839,8 +844,8 @@ static int oregon_scientific_v3_decode(r_device *decoder, bitbuffer_t *bitbuffer
         int id       = msg[2] << 8 | (msg[1] & 0xF0);
         int batt_low = (msg[3] & 0x1); // 8th bit instead of 6th commonly used for other devices
 
-        unsigned short int ipower = cm180_power(msg);
-        unsigned long long itotal = cm180_total(msg);
+        uint16_t ipower = cm180_power(msg);
+        uint64_t itotal = cm180_total(msg);
         float total_energy        = itotal / 3600.0 / 1000.0;
         if (valid == 0) {
             /* clang-format off */
@@ -871,13 +876,13 @@ static int oregon_scientific_v3_decode(r_device *decoder, bitbuffer_t *bitbuffer
         int id       = msg[2] << 8 | (msg[1] & 0xF0);
         int batt_low = (msg[3] & 0x40)?1:0; // 8th bit instead of 6th commonly used for other devices
 
-        unsigned short int ipower1 = cm180i_power(msg,0);
-        unsigned short int ipower2 = cm180i_power(msg,1);
-        unsigned short int ipower3 = cm180i_power(msg,2);
-        unsigned long long itotal= 0;
+        uint16_t ipower1 = cm180i_power(msg,0);
+        uint16_t ipower2 = cm180i_power(msg,1);
+        uint16_t ipower3 = cm180i_power(msg,2);
+        uint64_t itotal= 0;
         if (msg_len >= 140) itotal= cm180i_total(msg);
 
-        // per hour and in kilowat
+        // Convert `itotal` which is in Ws (or J) to kWh unit.
         float total_energy        = itotal / 3600.0 / 1000.0;
 
         if (valid == 0) {
--- a/src/devices/oregon_scientific.c
+++ b/src/devices/oregon_scientific.c
@@ -93,9 +93,9 @@ static unsigned int get_os_uv(unsigned char *message)
     return uvidx;
 }
 
-static uint16_t cm180i_power(uint8_t const *msg, unsigned int offset)
+static unsigned cm180i_power(uint8_t const *msg, unsigned int offset)
 {
-    uint16_t val = 0;
+    unsigned val = 0;
 
     val = (msg[4+offset*2] << 8) | (msg[3+offset*2] & 0xF0);
     // tested across situations varying from 700 watt to more than 8000 watt to
@@ -124,9 +124,9 @@ static uint8_t swap_nibbles(uint8_t byte)
     return (((byte&0xf) << 4) | (byte >> 4));
 }
 
-static uint16_t cm180_power(uint8_t const *msg)
+static unsigned cm180_power(uint8_t const *msg)
 {
-    uint16_t val = 0;
+    unsigned val = 0;
     val = (msg[4] << 8) | (msg[3] & 0xF0);
     // tested across situations varying from 700 watt to more than 8000 watt to
     // get same value as showed in physical CM180 panel (exactly equals to 1+1/160)
@@ -142,13 +142,8 @@ static uint64_t cm180_total(uint8_t const *msg)
         val = (uint64_t)msg[10] << 40;
         val += (uint64_t)msg[9] << 32;
         val += (uint32_t)msg[8] << 24;
-<<<<<<< HEAD
-        val += (uint32_t)msg[7] << 16;
-        val += (uint16_t)msg[6] << 8;
-=======
         val += msg[7] << 16;
         val += msg[6] << 8;
->>>>>>> f6385da2 (fixup! long/long long to int32_t/int64_t, short int to int16_t and comment fix)
         val += msg[5];
     }
     return val;
@@ -844,7 +839,7 @@ static int oregon_scientific_v3_decode(r_device *decoder, bitbuffer_t *bitbuffer
         int id       = msg[2] << 8 | (msg[1] & 0xF0);
         int batt_low = (msg[3] & 0x1); // 8th bit instead of 6th commonly used for other devices
 
-        uint16_t ipower = cm180_power(msg);
+        unsigned ipower = cm180_power(msg);
         uint64_t itotal = cm180_total(msg);
         float total_energy        = itotal / 3600.0 / 1000.0;
         if (valid == 0) {
@@ -876,9 +871,9 @@ static int oregon_scientific_v3_decode(r_device *decoder, bitbuffer_t *bitbuffer
         int id       = msg[2] << 8 | (msg[1] & 0xF0);
         int batt_low = (msg[3] & 0x40)?1:0; // 8th bit instead of 6th commonly used for other devices
 
-        uint16_t ipower1 = cm180i_power(msg,0);
-        uint16_t ipower2 = cm180i_power(msg,1);
-        uint16_t ipower3 = cm180i_power(msg,2);
+        unsigned ipower1 = cm180i_power(msg,0);
+        unsigned ipower2 = cm180i_power(msg,1);
+        unsigned ipower3 = cm180i_power(msg,2);
         uint64_t itotal= 0;
         if (msg_len >= 140) itotal= cm180i_total(msg);
 
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -146,7 +146,7 @@ static void data_append_exception(data_t* data, int exception, uint8_t* bb, int 
 
     rawp = (char *)raw_str;
     for (int i=0; i < browlen; i++) {
-        sprintf(rawp,"%02x",bb[i]);
+        sprintf(rawp,"%02x",bb[i]); // NOLINT
         rawp += 2;
     }
     *rawp = '\0';
@@ -458,7 +458,7 @@ static int acurite_6045_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsign
      */
     rawp = (char *)raw_str;
     for (int i=0; i < MIN(browlen, 15); i++) {
-        sprintf(rawp,"%02x",bb[i]);
+        sprintf(rawp,"%02x",bb[i]); // NOLINT
         rawp += 2;
     }
     *rawp = '\0';
@@ -811,7 +811,7 @@ static int acurite_atlas_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsig
     // having to enable debug for long running rtl_433 processes.
     rawp = (char *)raw_str;
     for (int i=0; i < MIN(browlen, 15); i++) {
-        sprintf(rawp,"%02x",bb[i]);
+        sprintf(rawp,"%02x",bb[i]); // NOLINT
         rawp += 2;
     }
     *rawp = '\0';
--- a/src/devices/acurite.c
+++ b/src/devices/acurite.c
@@ -146,7 +146,7 @@ static void data_append_exception(data_t* data, int exception, uint8_t* bb, int 
 
     rawp = (char *)raw_str;
     for (int i=0; i < browlen; i++) {
-        sprintf(rawp,"%02x",bb[i]); // NOLINT
+        sprintf(rawp,"%02x",bb[i]);
         rawp += 2;
     }
     *rawp = '\0';
@@ -458,7 +458,7 @@ static int acurite_6045_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsign
      */
     rawp = (char *)raw_str;
     for (int i=0; i < MIN(browlen, 15); i++) {
-        sprintf(rawp,"%02x",bb[i]); // NOLINT
+        sprintf(rawp,"%02x",bb[i]);
         rawp += 2;
     }
     *rawp = '\0';
@@ -811,7 +811,7 @@ static int acurite_atlas_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsig
     // having to enable debug for long running rtl_433 processes.
     rawp = (char *)raw_str;
     for (int i=0; i < MIN(browlen, 15); i++) {
-        sprintf(rawp,"%02x",bb[i]); // NOLINT
+        sprintf(rawp,"%02x",bb[i]);
         rawp += 2;
     }
     *rawp = '\0';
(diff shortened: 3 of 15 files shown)


src/decoder_util.c
```diff
@@ -51,7 +51,7 @@ static char *bitrow_asprint_code(uint8_t const *bitrow, unsigned bit_len)
     row_bytes[0] = '\0';
     // print byte-wide
     for (unsigned col = 0; col < (unsigned)(bit_len + 7) / 8; ++col) {
-        sprintf(&row_bytes[2 * col], "%02x", bitrow[col]);
+        sprintf(&row_bytes[2 * col], "%02x", bitrow[col]); // NOLINT
```
zuckschwerdt, line 54: don't nolint `sprintf()`, we should rather fix those. I already got the easy ones 8f6032a

zuckschwerdt:
Good catch about the `sizeof()`. The `sprintf()` etc. are genuine code-smell and need careful fixing.

The `long` are used with `struct timeval` and `%ld`, right? I'd say those are proper uses.

mdeweerd:
No, long is not always used with timeval (except for  src/compat_time.c I suppose).

These are the current occurrences:
```text
tests/baseband-test.c:86:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
tests/baseband-test.c:87:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
tests/baseband-test.c:138:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/pulse_data.c:156:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/pulse_data.c:158:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/compat_time.c:30:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/compat_time.c:31:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
tests/pulse-eval.c:176:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:107:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:109:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:112:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:113:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:114:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:115:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:137:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:139:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:142:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:143:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:144:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:145:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:843:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/devices/oregon_scientific.c:877:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:393:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:625:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:629:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:637:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:645:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:649:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:657:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:661:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:676:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:682:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:689:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:692:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:699:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:702:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:1888:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:1904:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
src/rtl_433.c:1918:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
```

src/decoder_util.c
mdeweerd: Ok, I'll remove the annotations.

I've made some easy fixes, but I wonder what you prefer for the next case: change the length of the variable to 3 and use sizeof, leave it as size 4 and use fixed value 3 or sizeof?

https://github.com/merbanan/rtl_433/pull/2683/files#diff-3deacdaee1d97c163db52aa550e5802735d32fd07170be2408627c2aadc585e6R104

mdeweerd:
In src/devices/oregon_scientific.c a "long" seems to be 32 bit and a "long long" is 64 bit.

src/decoder_util.c
zuckschwerdt: The buffer should be as close as possible (so we can easily check) and the size wouldn't matter, but 3 here.
```
    // Extract the id as hex string
    char id[3];
    snprintf(id, sizeof(id), "%02X", b[1]); // FIXME: id should be a plain int really
```

zuckschwerdt:
`unsigned long long` in oregon_scientific.c  should be `uint64_t` and the `short int` can reasonable be `int`, right?

src/decoder_util.c
mdeweerd: Should I remove the NOLINT on strcpy as well?

```c
        strcpy(topic, data->value.v_ptr); // NOLINT
```

mdeweerd:
> `unsigned long long` in oregon_scientific.c should be `uint64_t` and the `short int` can reasonable be `int`, right?

When the size matters I prefer to specify it exactly.  

In scientific... there are places where it clearly should be  a uint16_t.

However, I have an issue with:

        val += msg[6] << 8;

as msg[6] is a uint8_t and shifting it by 8 is going to result in 0.    I am fixing that with a uint16_t cast.

https://github.com/merbanan/rtl_433/blob/2c9cfba836c23f4748501d7fecd30a90ed46ab67/src/devices/oregon_scientific.c#L146

zuckschwerdt:
> When the size matters I prefer to specify it exactly.

and otherwise let the compiler optimize. There is always integer promotion anyways.

> is a uint8_t and shifting it by 8 which logically results in 0 (unless C casts to int before shifting).

Yeah, C is not obvious there. Integer promotion rules will (very broadly speaking) cast all operands to at least integer in any calculation. There is no problem with code like that.

mdeweerd:
I work on 8-bit CPUs sometimes, so I do not like 'msg[x] << 8' with uint8_t.

zuckschwerdt:
> I work on 8-bit CPUs sometimes, so I do not like 'msg[x] << 8' with uint8_t.

It's not `uint8_t` though, is it? It is implicit `(int)msg[x] << 8` -- C, esp. with optimizing compilers, is not assembler and far from WYSIWYG ;)

That's why I don't like to see specific types like `uint16_t` when it's actually all `int`s -- casual readers are tricked into assuming the wrong types.

mdeweerd:
If we go that way, all "short int" values "should" be converted to int unless the size is really important.

'int' depends on the architecture of the processor, and I avoid it unless the value fits in a byte or if the type is imposed by a library.
'int' is 'uint8_t' when it's compiled for an 8-bit processor.  The size of 'int' is not known even if we can suppose that it will be more than 8 bit nowadays.  And I consider that explicit typing is better than implicit.
Shifting only in a 16 bit value here can help the compiler/end result.

With regards to optimisations, modern processors are flexible enough for compilers to allow optimizing this kind of stuff.  Multiplying an uint16_t with another rather than an int (that is natively 32bit) will likely be less expensive than multiplying two 32bit values.
In terms of memory occupation, specifying the smaller size will use up less stack, etc.

So verify the impact of using uint16_t in the shift operation, I compiled the following C-code with a gcc compiler towards a 32bit architecture.

```c
#include <stdint.h>
volatile uint64_t total=0;
volatile uint8_t byte=0xFF;

volatile uint64_t total2=0;
volatile uint8_t byte2=0x23;

int main() {
   total+= byte << 8;
   total2+= (uint16_t)byte << 8;
   total+= byte << 8;
   total2+= (uint16_t)byte << 8;
}
```

I repeated the same operations two times as this avoided confusion when analysing the assembly output generated using `gcc -m32 -Os -Wa,-adhln -c -S test.c`.
This resulted in the following for the first two operations, so the output is exactly the same for gcc.

```
        movzbl  _byte, %eax
        movl    _total, %ecx
        movl    _total+4, %ebx
        sall    $8, %eax
        cltd
        addl    %ecx, %eax
        adcl    %ebx, %edx
        movl    %eax, _total
        movl    %edx, _total+4

        movzbl  _byte, %eax
        movl    _total2, %ecx
        movl    _total2+4, %ebx
        sall    $8, %eax
        cltd
        addl    %ecx, %eax
        adcl    %ebx, %edx
        movl    %eax, _total2
        movl    %edx, _total2+4
```

So, using the uint16_t on the value during the shift operation did not hinder the compiler.

This project is used on different architectures, and specifying the exact size that is needed is a good way to help ensure cross-platform compatibility.  Copying/assigning values is faster as well, and if the compiler can replace multiple assignments with a single "memcpy", then this memcpy will have to move less bytes/words. 

In my personal experience, not specifying the sizes has led to several cases leading to issues - including SW running on a Linux system.  It was developped with 32 bit in mind and when it was built for a 64bit system, it failed.  I had to convert *all* int values to explicit (u)int32_t.

If we want to optimize, rtl_433 offers quite a few opportunities to do that.

gdt:
> Good catch about the sizeof(). The sprintf() etc. are genuine code-smell and need careful fixing.
> 
> The long are used with struct timeval and %ld, right? I'd say those are proper uses.

Sorry about using edit vs quote reply, mouse-o.

POSIX says only that time_t is an integral type.  It is a bug to assume that it is `long`.   On NetBSD, it's `int64_t`, for example.   Unfortunately the best there is for printf is to cast to `intmax_t` and use `%j`.

gdt:
> In src/devices/oregon_scientific.c a "long" seems to be 32 bit and a "long long" is 64 bit.

That's a bug then.  C99 does not specify the size of such types.  If the code cares about size -- and it is likely that it does -- then it must use `int32_t` and `int64_t`, or unsigned versions.

mdeweerd:
> Unfortunately the best there is for printf is to cast to `intmax_t` and use `%j`.

https://www.gnu.org/software/libc/manual/html_node/Time-Types.html says that time_t can be a float as well... .  So technically, one would have to convert the time_t value to a better predictable type using one of the available functions such as difftime.

zuckschwerdt:
Using `volatile` is not what we talk about. We don't compile for microcontrollers. There is no way around Integer promotions. Re arithmetic and stack use: think archs with alignment restrictions. On ARM you introduce things like sign extend and masking:
```
        sxtw    x8, w0
        and     w9, w0, #0xff
        lsl     w10, w9, #9
```

"when it was built for a 64bit system, it failed. I had to convert all int values to explicit (u)int32_t." that does not sound right, the (u) int's were (u) int32_t before already. And on LP64 they will not be anything else.

Once an Integer promotion happened there is no good justification to perhaps spend cycles to down- and upcast in cicrcles.

printf format string support for inttypes is not great. If we can reasonably assume a smaller range then downcasting e.g. long will do. Upcasting will always work.

mdeweerd:
> Using `volatile` is not what we talk about.

I use volatile to check the compilation to ensure a fair comparison between methods and ensure that the operations are not optimized out as the values are not used.

>  On ARM you introduce things like sign extend and masking

I also checked the arm7 architecture on https://godbolt.org/ and it had one less instruction with uint16_t.

> Once an Integer promotion happened there is no good justification to perhaps spend cycles to down- and upcast in cicrcles

Clearly, the compiler did not care.  Again, IMHO being explicit is better than making implicit assumptions about the int size.  There's also uint_fast16_t in C99 which may be acceptable for you.

> printf format string support for inttypes is not great.

C99 introduced the PRI... constants.  https://en.cppreference.com/w/c/types/integer is a nice resource ti list those.  There are even macros to define integer constants with the correct post fix (LL/L).

> And on LP64 they will not be anything else.

When compiling for a 64bit architecture, an int is 64bit, not 32bit.  And it mattered for quite a few things.
For instance `0x80000000<<1` on a 32 bit architecture will result in 0, and on 64bit it's `0x100000000`.

The test case further below gives different results on a 64bit architecture than on 32bit architecture.
"After" indicates the value after a left shift by 1.

On 64 bit:
```
unsigned      Before: 0000000080000000 After: 0000000000000000
uint32_t      Before: 0000000080000000 After: 0000000000000000
uint32_fast_t Before: 0000000080000000 After: 0000000100000000
```

On 32bit:
```
unsigned      Before: 0000000080000000 After: 0000000000000000
uint32_t      Before: 0000000080000000 After: 0000000000000000
uint32_fast_t Before: 0000000080000000 After: 0000000000000000
```

Test case:
```c
#include <stdio.h>
#include <inttypes.h>
int main() {
  unsigned test_unsigned    = 0x80000000;
  uint32_t test_uint32      = UINT32_C(0x80000000);
  uint_fast16_t test_fast32 = UINT32_C(0x80000000);
  printf("unsigned      Before: %016X"           " After: %016x" "\n",
         test_unsigned, test_unsigned<<1);
  printf("uint32_t      Before: %016" PRIX32     " After: %016" PRIX32 "\n",
         test_uint32, test_uint32<<1);
  printf("uint32_fast_t Before: %016" PRIXFAST32 " After: %016" PRIXFAST32 "\n",
         test_fast32, test_fast32<<1);
  return 0;
}
```

zuckschwerdt:
> When compiling for a 64bit architecture, an int is 64bit, not 32bit.

Very much not true. It might be on ILP64 but that is esoteric, we are talking about (L)LP64.
The thing you show in your test code is just `uint32_fast_t == uint64_t` (on amd64 I guess)

mdeweerd:
> Very much not true. It might be on ILP64 but that is esoteric, we are talking about (L)LP64.

You are right, I just noticed that "LP64" defines it otherwise.  And my test case proves it.

mdeweerd:
Ok, now back to the core issue:
- I'll remove the uint16_t casts if that what you want;
- I'll replace all "short int" types that were changed to uint16_t to "int" if you want (in oregon_scientific.c).  Other than semantics for the developer and 8bit systems, there does not seem to be a need to keep those.
- I'll replace all ((u)int32_t to the unsigned/int if you want.  Other than semantics for the developer and lower then 32 bit systems there is no need to keep those.



```
unsigned     Before: 0000000000000080 After: 0000000000000100
uint8_t      Before: 0000000000000080 After: 0000000000000100
uint8_fast_t Before: 0000000000000080 After: 0000000000000100
```

```c
#include <stdio.h>
#include <inttypes.h>
int main() {
  unsigned test_unsigned    = 0x80;
  uint8_t test_uint8      = UINT8_C(0x80);
  uint_fast8_t test_fast8 = UINT8_C(0x80);
  printf("unsigned     Before: %016X"           " After: %016" PRIX64 "\n",
         test_unsigned, (uint64_t)(test_unsigned<<1));
  printf("uint8_t      Before: %016" PRIX8     " After: %016" PRIX64 "\n",
         test_uint8, (uint64_t)(test_uint8<<1));
  printf("uint8_fast_t Before: %016" PRIXFAST8 " After: %016" PRIX64 "\n",
         test_fast8, (uint64_t)(test_fast8<<1));
  return 0;
}
```

zuckschwerdt:
As rule of thumb: if it's a small-ish number and we don't care then `int` (integer promotion will prefer int anyway).
Yes, please change the over-specific `short int` to `int` and not `int16_t`.

Use `unsigned` only to flag a small-ish number that is never negative (e.g. for-loop).
Use `uint32_t` if the full 32 bit are needed (e.g. a 32 bit ID).
These should already be in place mostly.

gdt:
> > Unfortunately the best there is for printf is to cast to `intmax_t` and use `%j`.
> 
> https://www.gnu.org/software/libc/manual/html_node/Time-Types.html says that time_t can be a float as well... . So technically, one would have to convert the time_t value to a better predictable type using one of the available functions such as difftime.

POSIX says that time_t must be an integer type, and POSIX is the right thing to consult, not a particular libc implementation.   It's clear that time_t must be an integer type:  https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html

mdeweerd:
> I don't know if glibc supports a mode that isn't compatable with POSIX but if so I think we should ignore it.

The ISO99 standard says time_t is an arithmetic type and that refers to the collection of integers and floats.
Apparently it's a double on a IBM 370 system and can be on VAX/VMS systems.
So that's why gcc supports it.

mdeweerd:
I've update the c-code for the int types.

gdt:
> > I don't know if glibc supports a mode that isn't compatable with POSIX but if so I think we should ignore it.
> 
> The ISO99 standard says time_t is an arithmetic type and that refers to the collection of integers and floats. Apparently it's a double on a IBM 370 system and can be on VAX/VMS systems. So that's why gcc supports it.

Wow, but ok.   I realized that rtl_433 is supposed to have less-than-POSIX portability requirements, just C99.   Still, I have ~never seen code that assumes time_t is other than integer, and it's essentially always the same as either int32_t (in old systems) or int64_t (in up-to-date systems).

gdt:
Where are we on this?    it seems like there has been a wide-ranging discussion, but are there specific changes still on the table?   If so, could you rebase so it's clean and force-push and ask people to look again?    As ticket gardener I would like us to merge what we agree on and separate the rest, perhaps close it.

zuckschwerdt:
I don't wan to merge the `// NOLINT` and forget about the string operations. Those should rather be updated to better syntax (in another PR).

gdt:
So therefore we should close this?

mdeweerd:
The goal of the PR is to fix cpplint messages so that cpplint can be integrated in the CI flow.

Possible actions:
- Merge PR as-is;
- Merge, but after all or some of the following:
  - Update PR to remove `// NOLINT`;
  - Change all occurrences of `long` to a time to be chosen for long counts (size_t?).
  - Fix strtol usage ?
  - Add more '// NOLINT' on lines that are decided to leave as is (fo have no cpplint messages).
- Deny this PR, throw away all proposed changes.


Recap on this PR:
- Fixes some code to specify a better length in snprintf when identified by cpplint;
- `// NOLINT` annotations - applied IMHO on lines that are IMHO not worth the effort to rewrite as-is.  This helps to be able to run cpplint on the code and identify issues in new or changed code.
- "short int" changed to "int" (or unsigned if unsigned) - as per discussion
- "long long" changed to (u)int64_t
   - with one exceptions in the current code (`long long unsigned) cast) - related to a printf).

Not changed: several occurences of long that are (mostly) used as counters, not as time:
```plaintext
src/baseband.c:    unsigned long i;
src/baseband.c:    unsigned long i;
src/baseband.c:    unsigned long i;
src/baseband.c:    unsigned long i;
src/baseband.c:    unsigned long i;
src/baseband.c:    unsigned long i;
src/baseband.c:    for (unsigned long i = 1; i < len; i++) {
src/baseband.c:void baseband_demod_FM(uint8_t const *x_buf, int16_t *y_buf, unsigned long num_samples, uint32_t samp_rate, float low_pass, demodfm_state_t *state)
src/baseband.c:void baseband_demod_FM_cs16(int16_t const *x_buf, int16_t *y_buf, unsigned long num_samples, uint32_t samp_rate, float low_pass, demodfm_state_t *state)
src/devices/flex.c:static unsigned long compact_number(uint8_t *data, unsigned bit_offset, unsigned long mask)
src/devices/flex.c:    unsigned long val = 0;
src/devices/flex.c:static unsigned long extract_number(uint8_t *data, unsigned bit_offset, unsigned bit_count)
src/devices/flex.c:    unsigned long val = data[pos];
src/devices/flex.c:    unsigned long mask;
src/devices/flex.c:        unsigned long val;
src/devices/flex.c:        unsigned long key;
src/devices/secplus_v2.c:    snprintf(fixed_str, sizeof(fixed_str), "%llu", (long long unsigned)fixed_total);
src/pulse_data.c:        long mark  = strtol(p, &endptr, 10);
src/pulse_data.c:        long space = strtol(p, &endptr, 10);
src/rtl_433.c:    unsigned long n_samples;
src/rtl_433.c:        unsigned long out_len = n_samples * demod->sample_size;
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples * 2; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples * 2; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples * 2; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples * 2; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples * 2; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples * 2; ++n)
src/rtl_433.c:            for (unsigned long n = 0; n < n_samples; ++n)
src/rtl_433.c:            for (unsigned long n = 0; n < n_samples; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples; ++n)
src/rtl_433.c:                for (unsigned long n = 0; n < n_samples; ++n)
src/rtl_433.c:            unsigned long n_read;
src/rtl_433.c:                    for (unsigned long n = 0; n < n_read; n++) {
src/rtl_433.c:                        for (unsigned long n = 0; n < n_read; n++) {
tests/baseband-test.c:    long n_read;
tests/baseband-test.c:    unsigned long n_samples;
tests/baseband-test.c:    for (unsigned long i = 0; i < n_samples * 2; i++) {
tests/pulse-eval.c:    long n_read;
```

src/devices/oregon_scientific.c
```diff
@@ -871,13 +871,13 @@ static int oregon_scientific_v3_decode(r_device *decoder, bitbuffer_t *bitbuffer
...
-        unsigned short int ipower3 = cm180i_power(msg,2);
-        unsigned long long itotal= 0;
+        unsigned ipower1 = cm180i_power(msg,0);
+        unsigned ipower2 = cm180i_power(msg,1);
+        unsigned ipower3 = cm180i_power(msg,2);
+        uint64_t itotal= 0;
         if (msg_len >= 140) itotal= cm180i_total(msg);
 
-        // per hour and in kilowat
+        // per hour and in kilowatt (hence kwh)
```
gdt, line 880: kilowatt, even spelled correctly, is not a unit of energy.  I am not sure what's going on here, but it seems at least not quite right.

gdt:
I am somewhat of an outlier in opinions, but I'd say:
  1) personally, I like things to be rebased, so one can consider clean commits, even if the rtl_433 norm is to squash
  2) this PR bundles (trying to use a neutral term)
     a) fixes for things that lint identified (types), using sizeof vs repeated numbers for snprintf, and some random other fixes
     b) instructions not to lint

My interpretation and my own view is that  type 2a changes are in general agreeable and are very useful, and type 2b are not agreeable.
So yes, update to not add NOLINT sounds good as a plan.

As for 1, I have different norms, and would like to see it rebased and the clean commits not squashed, but that's not how this project works, and thus i expect someone else will push the merge button.

src/devices/oregon_scientific.c
mdeweerd: I did not want to change the original text too much, fix at least the obvious typo and remind that the result is kWh according to the data label further below.

Fixing this fishy thing is not the goal of the PR.  cm180i_total probably returns  Ws (so Joules), with 3600Ws = 1Wh, and to get kWh that's divided by 1000.

The comment could be:

```c
// Convert itotal which is in Ws (or J) to kWh unit.
```

mdeweerd:
> 1. personally, I like things to be rebased, so one can consider clean commits, even if the rtl_433 norm is to squash

- This PR was git-rebased recently;
- What is a "clean commit" for you: a commit for each type of fix, a commit for each file?
- As my goal is to have 0 items reported by cpplint so that it can be integrated in the CI flow, fixing all of that is a single operation.

> 2. this PR bundles (trying to use a neutral term)
>    a) fixes for things that lint identified (types), using sizeof vs repeated numbers for snprintf, and some random other fixes
>    b) instructions not to lint

As the goal is to have an "empty" cpplint report.
IMHO, splitting this into many small commits is not very useful here.

> My interpretation and my own view is that type 2a changes are in general agreeable and are very useful, and type 2b are not agreeable. So yes, update to not add NOLINT sounds good as a plan.

I agree that "type 2b" is not as-such agreeable, however the goal is to have cpplint in the flow so that it helps for new code.
Changing the legacy code is risky as it's not all unit-tested.
So its a tradeoff between what is an easy low risk fix, and a more risky fix.
Not applying NOLINT means that all of that needs to be fixed before this can be used in CI.

> As for 1, I have different norms, and would like to see it rebased and the clean commits not squashed, but that's not how this project works, and thus i expect someone else will push the merge button.

If it is agreed upon that this has to be many smaller commits and even more than one PR, I can do that.  However, I need to know what the agreed upon granularity has to be, and if it's a single PR, then it should not be squashed of course.

src/devices/oregon_scientific.c
zuckschwerdt: There was a [lengthy discussion](https://github.com/merbanan/rtl_433/pull/2418#discussion_r1133254997) about the CM160, IIRC the calculation works by chance only.

gdt: The new comment looks fine.   I am ok with 1) don't touch it or 2) leave it in a good state, and don't like the middle ground...

gdt:
When I looked, I thought I saw a commit that added annotations and another that removed them.  To me, it's "one logical change per commit" and certainly that's multiple files.    As I tried to be clear on, I'm an outlier and project doctrine is to squash -- but I think that's mostly because most things that are merged are messy, and are small enough that they are one change anyway.

I understand that your goal is to have cpplint produce no warnings.  But that doesn't mean that "fix things that cpplint identified that are actually problems" is not a worthy subgoal.    I think it's entirely reasonable to merge fixes that we agree on and leave the pieces that we don't agree on (yet, perhaps) for later.

We'll see what others say, but I suspect that if there's one commit that fixes things that can be clearly seen to be  wrong by  code reading, and don't add NOLINT, then we can all be happy enough.

mdeweerd:
> and don't add NOLINT, then we can all be happy enough.

I hear your arguments, and I've written a tool called CANDY for static code analysis of C-Code in 1992/1993 - fun fact: written in Pascal as it also had to run on IBM Mainframe where no C Compiler was available.


That said, if we do not annotate the code with "NOLINT", then the code has to be "fixed" with regards to cpplint's perspective. 
 Otherwise it's not usable in continuous integration.

A reminder about not using NOLINT in the document indicating that current occurences in the code are there to make cpplint useable in the continuous integration.   So that new code is "compliant".  "NOLINT" is still searchable throughout the code.

I think that using NOLINT is better than ignoring entire files.

gdt:
My point is that we should make incremental progress on the things we agree on, rather than deferring everything until some theoretical future agreement.    i am entirely avoiding the future discussion right now.

zuckschwerdt:
Ok, to get some results let's merge all changes. Improving the string operations can be done later.
(I'm not happy excluding the lines the linter is suspicious of, but frankly I did the same thing for the GCC static analysis. E.g. https://github.com/merbanan/rtl_433/blob/master/src/sdr.c#L958, i.e. this cast https://github.com/merbanan/rtl_433/blob/master/src/sdr.c#L982, is not broken but suspect still.)

src/devices/oregon_scientific.c
mdeweerd: I updated the commit where is made the first change.

src/devices/oregon_scientific.c
```diff
@@ -104,15 +104,15 @@ static unsigned short int cm180i_power(uint8_t const *msg, unsigned int offset)
...
+    uint64_t val = 0;
     if ((msg[1] & 0x0F) == 0) {
         // Sensor returns total only if nibble#4 == 0
-        val = (unsigned long long)msg[14] << 40;
-        val += (unsigned long long)msg[13] << 32;
-        val += (unsigned long)msg[12] << 24;
-        val += (unsigned long)msg[11] << 16;
+        val = (uint64_t)msg[14] << 40;
+        val += (uint64_t)msg[13] << 32;
+        val += msg[12] << 24;
```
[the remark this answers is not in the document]
mdeweerd: I understood the policy to assume "implicit integer promotion" where integer is a 32 bit and only write (u)int32_t when it HAS to be.

IMHO, it does not "have to be" uint32_t as we do not care about the exact size in memory and no bits will be shifted outside the 32bit range and integer promotion is supposed to ensure 32bit and msg[12] is already unsigned.

If it is needed here then it is also needed for "val += msg[11] << 16"

Also see #1193 .


My personal policy is to put (u)int16_t, (u)int32_t, etc. everywhere where I need more than 8 bit as on a microcontroller that is less than 32bit, int is only as big as the microcontroller's native register size.  And I use (u)int8_t as well to help ensure that I use as little memory as possible on microcontrollers.


Please confirm that the policy requires to cast this line, and if it is or not needed for the next line as well.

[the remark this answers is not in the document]
zuckschwerdt: Not about policy: `<<24` will fill an int and thus might get negative -- this might be ok with bit operations but not with `+=`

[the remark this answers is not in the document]
mdeweerd: It's worse than I thought:
```c
#include <stdio.h>
#include <inttypes.h>

int main() {
   uint8_t m[32];
   uint64_t sum=0;
   uint64_t or_op=0;
   uint64_t or_op_32cast=0;

   m[12]=0xCF;

   printf("%08X << 24 is %08X\n", m[12], m[12]<<24);
   sum += m[12]<<24;
   or_op |= m[12]<<24;
   or_op_32cast |= (uint32_t)m[12]<<24;
   printf("sum      %016" PRIX64 "\n", sum);
   printf(" or      %016" PRIX64 "\n", or_op);
   printf(" or cast %016" PRIX64 "\n", or_op_32cast);
   return 0;
}
```
```plaintext
000000CF << 24 is CF000000
sum      FFFFFFFFCF000000
 or      FFFFFFFFCF000000
 or cast 00000000CF000000
```

Which pleads for casting as strict as possible rather than relying on interpretations of what the compiler does by default.

Will update the code later.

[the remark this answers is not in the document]
mdeweerd: I fixed the relevant commit, rebased and forced pushed.

I also changed https://github.com/merbanan/rtl_433/blob/dcfd6cc43c4998cbca3fa232226f81965de63987/src/devices/oregon_scientific.c#L145 .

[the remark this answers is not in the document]
zuckschwerdt: The compiler isn't funny here, it's by spec for integer promotion and sign extend.
No need to cast on anything `<<23` or lower, we can't deal with `sizeof int < 32` anyway.

mdeweerd:
A follow up with regards to 32bit int typing.

I found one of the contexts in which I had an issue with int not being 32 bit [Arduino Atmega](https://www.arduino.cc/reference/en/language/variables/data-types/int/#:~:text=On%20the%20Arduino%20Uno%20(and,(2%2Dbyte)%20value).

So for portability, I continue to suggest using the int types constrainging the width in number of bits.

PR #2686, opened 2023-10-23
Add support for Schou 72543 Rain sensor

36jemini:
Hi,

I have tried to my best ability to write up a decoder for my Schou rain sensor (model: 72543).
The code is not 100% done, but it is at a stage where I would like to request some help.

I have added a pull request with test samples and a longer device description to the rtl_433_tests project,
which has not yet been merged in, so please find it here until then: [https://github.com/36jemini/rtl_433_tests/tree/master/tests/Schou](https://github.com/36jemini/rtl_433_tests/tree/master/tests/Schou)

I would like some guidance on,
- In which file should I "register" the _schou_72543_rain.c_ file?
- Have I made the right approach to extract the data from the bitbuffer?

Comments, feedback and guidance will be highly appreciated.

### DIFF
--- a/src/devices/schou_72543_rain.c
+++ b/src/devices/schou_72543_rain.c
@@ -68,16 +68,21 @@ static int schou_72543_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
     //Assuming option #1
 
+    printf("\nRestart from begining.c\n");
+
     // Full data is 3 rows, two are required for data validation
     if (bitbuffer->num_rows < 2){
         return DECODE_ABORT_LENGTH;
     }
 
+    printf(" 2) More than two rows in bitbuffer\n");
 
     int row = bitbuffer_find_repeated_prefix(bitbuffer, 2, 64);
     if (row < 0) {
         return DECODE_ABORT_EARLY;
     }
+
+    printf(" 3) Found at least two alike rows\n");
 
     uint8_t *b = bitbuffer->bb[row];
 
@@ -93,6 +98,10 @@ static int schou_72543_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         decoder_logf_bitrow(decoder, 1, __func__, b, 65, "Checksum error, expected: %02x calculated: %02x", micSum, calSum);
         return DECODE_FAIL_MIC;
     }
+
+    printf(" 4) MIC check passed\n");
+
+
     // <<<<<
     // Assuming from here on that b[9] = a1 f8 8c f6 ff 35 06 55 00
 
--- a/src/devices/schou_72543_rain.c
+++ b/src/devices/schou_72543_rain.c
@@ -62,32 +62,32 @@ static int schou_72543_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     }
 
     // Load bitbuffer data and validate checksum
-    uint8_t *b = bitbuffer->bb[row];
-    int micSum = b[7];                    // Checksum as read
-    int calSum = add_bytes(b, 7) & 0x0FF; // Checksum as calculated, accounting for the lowest 8 bit
+    uint8_t *b  = bitbuffer->bb[row];
+    int micsum = b[7];                    // Checksum as read
+    int calsum = add_bytes(b, 7) & 0x0FF; // Checksum as calculated, accounting for the lowest 8 bit
 
-    if (micSum != calSum) {
-        decoder_logf_bitrow(decoder, 1, __func__, b, 65, "Checksum error, expected: %02x calculated: %02x", micSum, calSum);
+    if (micsum != calsum) {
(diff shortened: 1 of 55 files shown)


36jemini:
Also, should this be a "draft" Pull request? I'm still new to both git and github

src/devices/schou_72543_rain.c
```diff
@@ -0,0 +1,156 @@
...
+    // Option #2:
+    //     As a slower but more roubist alternative I would extract each row in turn and account for
+    //     the differences per loop. This has the added benefit that only one rows data must be intact
+    //     I.e. If the MIC check is positive, then calculate values, exit loop and return values.
+    // >>>>>
+
+    //Assuming option #1
+
+    // Full data is 3 rows, two are required for data validation
+    if(bitbuffer->num_rows < 2){
```
zuckschwerdt, line 73: add space after keywords (if) and before braces.

zuckschwerdt:
>     * In which file should I "register" the _schou_72543_rain.c_ file?

`include/rtl_433_devices.h` then stage or commit and call `./maintainer_update.py`. Compile and call  `./maintainer_update.py` again.

src/devices/schou_72543_rain.c
```diff
@@ -70,7 +69,7 @@ static int schou_72543_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     //Assuming option #1
 
     // Full data is 3 rows, two are required for data validation
-    if(bitbuffer->num_rows < 2){
+    if (bitbuffer->num_rows < 2){
```
zuckschwerdt, line 72: and a space near the end :)

src/devices/schou_72543_rain.c
```diff
@@ -82,12 +81,13 @@ static int schou_72543_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
-    }
+    uint8_t micSum = b[7];              // Checksum as read
+    int    cal_sum = add_bytes(b, 7);   // Checksum as calculated
+    //uint8_t calSum = 0;               // Checksum as calculated
+    //
+    //for (i = 0; i<7; i++){
+    //    calSum = calSum + b[i];
+    //}
 
     if (micSum != calSum) {
```
zuckschwerdt, line 92: this now needs `if (micSum != calSum & 0xff) {`

src/devices/schou_72543_rain.c
```diff
@@ -111,18 +111,18 @@ static int schou_72543_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     deviceID        =   (b[0] << 8 ) | b[1];
     isBatteryLow    =   (b[2] & 0x80) == 1;                  // if one, battery is low
     isMessageRepeat =   (b[2] & 0x40) == 1;                  // if one, message is a repeat
-	messageCounter  =   (b[2] & 0x0e) >> 1;                  // 3 bit counter (rather than 4 bit incrementing by 2 each time
+    messageCounter  =   (b[2] & 0x0e) >> 1;                  // 3 bit counter (rather than 4 bit incrementing by 2 each time
     rain_mm         =  ((b[4] << 8 ) & b[3]) / 10.0f;
-	temp_F          = (((b[6] << 8 ) & b[5]) / 10.0f) - 90;
+    temp_F          = (((b[6] << 8 ) & b[5]) / 10.0f) - 90;
```
zuckschwerdt, line 116: this would be better written as `(((b[6] << 8 ) | b[5]) - 900) * 0.1f;
also above, and note to use `|` not `&`

src/devices/schou_72543_rain.c
```diff
@@ -111,18 +111,18 @@ static int schou_72543_rain_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
 
     /* clang-format off */
     data = data_make(
             "model",          "",             DATA_STRING, "Schou 72543 rain sensor",
```
zuckschwerdt, line 120: model needs to be just "Schou-72543"

36jemini: Thank you @zuckschwerdt for the comments. I seriously appreciate the support!
And I am sorry for the amount of commits. I still did not get the "squish" i think it is called of multiple- to one commit.

zuckschwerdt: Happy to help. No need to squash. We'll do it on merge anyway.

src/devices/schou_72543_rain.c
36jemini: moved to calSum

zuckschwerdt:
> 1. Rain count is 0.108 mm / each gauge flip (instead of 0.1mm) according to my tests -  can this be changed?

You can change that in the code and recompile. https://github.com/merbanan/rtl_433/blob/master/src/devices/schou_72543_rain.c#L80
We don't offer "calibration", maybe you can use a bridge, see https://github.com/merbanan/rtl_433/tree/master/examples

> 2. Now decoder is returning temperature in Fahrenheits and rain in mm. Could temperature to be reported in Celsius 

Use `-C si` or `convert si` in a conf file.

PR #2677, opened 2023-10-20
Add Carchet TPMS

kesuhm:
Hi, there! 

I characterized the Carchet TPMS sensor in May of last year and had wanted to perform a few additional tests before sharing.  Since my last post, I have made corrections to the checksum code and confirmed the flags for low battery and inflation detection.  I would like to share my work up to this point.

In testing the code this week, I see that the Carchet TPMS sensor is similar to the EezTire-E618 sensor added earlier this year.  The data format is the same and the flags seem to agree. Perhaps the low battery and inflate flags in the Carchet TPMS code will be useful.  Also, I characterized the sensor data in PSI and deg F, so the pressure and temperature readouts will compliment the readouts of the EezTire-E618.

This is my first pull request. I hope I have done everything correctly.

PR #2692, opened 2023-10-24
Merge Carchet TPMS with EezTire TPMS

kesuhm:
I merged the code for these two sensors because they appear to use the same protocol. I added low battery and inflate flags to the readout.  The low battery flag has been verified with the Carchet dashboard monitor, but not the EezTire monitor. During low battery flag testing, I was able to trigger the second flags byte (flags2) to a value of 0x20, but was unable to controllably reproduce this.

I also commented in operations which can be used to display pressure in PSI and temperature in deg F.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
 ```
 
   A "rtl_433.conf" file is searched in "./", XDG_CONFIG_HOME e.g. "$HOME/.config/rtl_433/",
-  "/usr/local/etc/rtl_433/", "/etc/rtl_433/", then command line args will be parsed in order.
+  SYSCONFDIR e.g. "/usr/local/etc/rtl_433/", then command line args will be parsed in order.
 		= General options =
   [-V] Output the version string and exit
   [-v] Increase verbosity (can be used multiple times).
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Thanks. Very good improvements!
- "8 bit little-endian"... only multi-byte can have endianess, right?
- "int temperature_C = b[4] - 50;//int temperature_F = b[4];" can you check with extreme temperatures if it's F or C?
- "//float pressure_PSI = b[6] * 0.363 - 0.06946;" that sounds like 2.5 kPa steps (nice round number)
- "low batt" must be "battery_ok", see other decoders.
- "EezTire E618 | Carchet" this should be "EezTire E618, Carchet TPMS"

zuckschwerdt:
Please stick with a PR. Don't just abandon to open new ones.
Note: this continues #2677

kesuhm:
I'm sorry. I didn't realize the green button would assign a new PR number every time I pushed it. I am now working back in #2677 -- I have merged the Carchet TPMS code with the EezTire code, corrected my terminology, and fixed my Fahrenheit math comment. In my first merge, I had missed the unit conversion I had done in the Carchet TPMS data_make code.

zuckschwerdt:
The code is manually merged with a7e5aaf

PR #2704, opened 2023-10-31
Bresser leakage: Fixed inverted startup flag

matthias-bs:
The startup flag was inverted. The alarm/no_alarm expressions were changed as a matter of style (no functional changes).


matthias-bs:
Done!

matthias-bs:
Thank you!

zuckschwerdt:
> Thank you!

Progress is built on all the user contributions, so rather: thank you for caring about good an maintainable code.

PR #2716, opened 2023-11-08
Add support for Aldi Sempre 4-AH0423-4 temperature/rain sensor

sorny:
See the discussion in https://github.com/merbanan/rtl_433/issues/1837

### DIFF
--- a/src/devices/auriol_4ld5661.c
+++ b/src/devices/auriol_4ld5661.c
@@ -1,5 +1,5 @@
 /** @file
-    Auriol 4-LD5661/4-LD5972/4-LD6313 sensors.
+    Auriol 4-LD5661/4-LD5972/4-LD6313, Sempre 4-AH0423-4 sensors.
 
     Copyright (C) 2021 Balazs H.
     Copyright (C) 2023 Peter Soos
@@ -19,12 +19,14 @@ See also issues #1857, #2631 and PR #2633
 
 Data layout:
 
-    II B F CC TTT F RRRRRR
+    1a       80       88       f0       00       14       5
+    00011010 10000000 10001000 11110000 00000000 00010100 0101
+    IIIIIIII B?CCTTTT TTTTTTTT FFFFRRRR RRRRRRRR RRRRRRRR RRRR
 
 - I: id, 8 bit: factory (hard)coded random ID
 - B: battery, 1 bit: 1=OK, 0=LOW
-- F: flag, 1 bit: unknown yet
-- C: channel, 2 bit: 0=CH1, 1=CH2, 2=CH3
+- ?: flag, 1 bit: seems to always 0
+- C: channel, 2 bit: seems to be a hardcoded channel, 0 on auriol, 3 on sempre
 - T: temperature, 12 bit: 2's complement, scaled by 10
 - F: 4 bit: seems to be 0xf constantly, a separator between temp and rain
 - R: rain sensor, probably the remaining 24 bit: a counter for every 0.3 mm (4-LD5661) or 0.242 mm (4-LD6313)
@@ -45,10 +47,9 @@ static int auriol_4ld5661_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
         uint8_t *b  = bitbuffer->bb[i];
         int id      = b[0];
-        int battery  = b[1] & 0x80;
-        int channel  = ((b[1] & 0x30) >> 4) + 1;
+        int batt_ok = b[1] >> 7;
 
-        if (b[3] != 0xf0) {
+        if (b[3] != 0xf0 || (b[1] & 0x40) != 0) {
             ret = DECODE_FAIL_MIC;
             continue;
         }
@@ -76,8 +77,7 @@ static int auriol_4ld5661_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         data_t *data = data_make(
                 "model",            "Model",        DATA_STRING, "Auriol-4LD5661",
                 "id",               "ID",           DATA_FORMAT, "%02x", DATA_INT, id,
-                "channel",          "Channel",      DATA_INT,    channel,
-                "battery_ok",       "Battery",      DATA_INT,    !!battery,
+                "battery_ok",       "Battery OK",   DATA_INT, batt_ok,
                 "temperature_C",    "Temperature",  DATA_FORMAT, "%.01f C", DATA_DOUBLE, temp_c,
                 "rain_mm",          "Rain",         DATA_FORMAT, "%.01f mm", DATA_DOUBLE, rain,
                 "rain",             "Rain tips",    DATA_INT, rain_raw,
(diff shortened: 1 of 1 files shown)


Gollam:
The changes you made look good for the Sempre type sensor, but will it continue to work with the original Auriol ones? As I understood, the Sempre sends more fields, so Auriol decoding will fail, isn't it?

sorny:
I checked with the available binary samples from https://github.com/merbanan/rtl_433/issues/1837 and they all work was expected.

General samples:
`rtl_433 -y {52}1a00f7f000000`
> Model     : Auriol-4LD5661                         ID        : 1a
> Channel   : 1            Battery   : 0             Temperature: 24.7 C       Rain      : 0.0 mm        Rain tips : 0

`rtl_433 -y {52}c6806bf00000f`
> Model     : Auriol-4LD5661                         ID        : c6
> Channel   : 1            Battery   : 1             Temperature: 10.7 C       Rain      : 15.0 mm       Rain tips : 15

`rtl_433 -y {52}9fb0bcf0000ef`
> Model     : Auriol-4LD5661                         ID        : 9f
> Channel   : 4            Battery   : 1             Temperature: 18.8 C       Rain      : 239.0 mm      Rain tips : 239

Low battery sample:
`rtl_433 -y {52}1a00faf000001`
> Model     : Auriol-4LD5661                         ID        : 1a
> Channel   : 1            Battery   : 0             Temperature: 25.0 C       Rain      : 1.0 mm        Rain tips : 1

src/devices/auriol_4ld5661.c
```diff
@@ -11,16 +13,18 @@
 */
 
 /**
-Lidl Auriol 4-LD5661/4-LD5972/4-LD6313 sensors.
+Lidl Auriol 4-LD5661/4-LD5972/4-LD6313 and Aldi Sempre 4-AH0423-4 sensors.
```
zuckschwerdt, line 16: Change also line 2

src/devices/auriol_4ld5661.c
```diff
@@ -11,16 +13,18 @@
...
 
 See also issues #1857, #2631 and PR #2633
 
 Data layout:
 
-    II B TTT F RRRRRR
+    II B F CC TTT F RRRRRR
```
zuckschwerdt, line 22: The chars are usually nibbles. Mark just `F` then detail what's in the flags in the list.

src/devices/auriol_4ld5661.c
```diff
@@ -41,9 +45,10 @@ static int auriol_4ld5661_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
         uint8_t *b  = bitbuffer->bb[i];
         int id      = b[0];
-        int batt_ok = b[1] >> 7;
+        int battery  = b[1] & 0x80;
```
zuckschwerdt, line 48: Don't change this line

src/devices/auriol_4ld5661.c
```diff
@@ -71,7 +76,8 @@ static int auriol_4ld5661_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         data_t *data = data_make(
                 "model",            "Model",        DATA_STRING, "Auriol-4LD5661",
                 "id",               "ID",           DATA_FORMAT, "%02x", DATA_INT, id,
-                "battery_ok",       "Battery OK",   DATA_INT, batt_ok,
+                "channel",          "Channel",      DATA_INT,    channel,
+                "battery_ok",       "Battery",      DATA_INT,    !!battery,
```
zuckschwerdt, line 80: don't change

zuckschwerdt:
Does the unit have a channel selector? What are the markings on that?

src/devices/auriol_4ld5661.c
sorny: Would be necessary to support the battery bit of both auriol and sempre

src/devices/auriol_4ld5661.c
sorny: Which line do you mean?

src/devices/auriol_4ld5661.c
zuckschwerdt: `>> 7` is the same bit as `0x80` just a different output place.

src/devices/auriol_4ld5661.c
zuckschwerdt: the second line in this file.

zuckschwerdt:
If there is no channel we should just add the docs and change the check, e.g.
```
// Flags are 0 for Auriol and 2 for Sempre
if (b[3] != 0xf0 || (b[1] & 0x70) != 0) || (b[1] & 0x70) != 0x2) {
```
(if 0x2 is the fixed value)

src/devices/auriol_4ld5661.c
sorny: Redid the current documentation to reflect the unknown bits 9, 10 and 11 of the payload.

src/devices/auriol_4ld5661.c
sorny: Removed the channel part as we are not sure if thats really the channel or not

sorny:
Reworked the check to evaluate the unknown, seemingly always `0` bit after the battery flag

src/devices/auriol_4ld5661.c
```diff
@@ -45,10 +47,9 @@ static int auriol_4ld5661_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 
         uint8_t *b  = bitbuffer->bb[i];
         int id      = b[0];
-        int battery  = b[1] & 0x80;
-        int channel  = ((b[1] & 0x30) >> 4) + 1;
+        int batt_ok = b[1] >> 7;
 
-        if (b[3] != 0xf0) {
+        if (b[3] != 0xf0 || (b[1] & 0x40) != 0) {
```
zuckschwerdt, line 52: This will ignore the two bits. But don't we want to check that it's 0 or 3? I.e. 1 or 2 are invalid.

zuckschwerdt:
You are getting a "channel" value of 3, right? With the Nexus logic of C+1 that would be CH4? So 0…2 as channel (as Nexus has) and then 3 for rain would be a good guess here?

src/devices/auriol_4ld5661.c
sorny: Depends, no idea if there are other brands of this rain meter out there that would use either ch1 or ch2 :/

sorny:
Yeps, I am getting channel 4 for my sempre rain meter, based on the nexus channel logic. The samples for Auriol result in channel 1 which could make sense that different brands use different channels for the rain meter.

devslo:
This PR also supports the Hofer Krontaler WS-0308-w-v, as it uses same protocol as Aldi Sempre ("Channel"  value== 3).

MattWestb:
Is it some technical reason for not merging  this pull request ?
I have both Lidl and Hofer / Aldi devices and only Lidl is working so its little. . . 

Thanks in advance !

merbanan:
@zuckschwerdt this looks good to me. Any comments?

PR #2725, opened 2023-11-25
Update rtl_433_mqtt_hass.py to add battery_mV

fallenpegasus:
In addition to battery_ok, many devices also surface the measured battery voltage.

This is a decomposed part of https://github.com/merbanan/rtl_433/pull/2581

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -190,6 +190,19 @@ mappings = {
         }
     },
 
+    "battery_mV": {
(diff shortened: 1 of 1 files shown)


gdt:
I more or less think it's a bug that we have battery_mV, instead of battery_V, SI base units and all that, but given that we do we might as well send a discovery message.

fallenpegasus:
I agree that it should be "battery_V" but that was a bigger patch than I wanted to write, especially since it would touch nearly every file in src/devices/*

PR #2728, opened 2023-11-26
Update fineoffset_ws90.c battery_ok can >100

fallenpegasus:
The ws90 driver capped battery_ok at 100, at 1.4V.
Lithium batteries go to 1.5V instead of 1.4V, so battery_ok >100 is fine.

### DIFF
--- a/src/devices/fineoffset_ws90.c
+++ b/src/devices/fineoffset_ws90.c
@@ -109,9 +109,6 @@ static int fineoffset_ws90_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int supercap_V  = (b[21] & 0x3f);
     int firmware    = b[29];
 
-    if (battery_lvl > 100) // More then 100%?
-        battery_lvl = 100;
-
     char extra[31];
     snprintf(extra, sizeof(extra), "%02x%02x%02x%02x%02x------%02x%02x%02x%02x%02x%02x%02x", b[14], b[15], b[16], b[17], b[18], /* b[19,20] is the rain sensor, b[21] is supercap_V */ b[22], b[23], b[24], b[25], b[26], b[27], b[28]);
 


src/devices/fineoffset_ws90.c
```diff
@@ -109,9 +109,6 @@ static int fineoffset_ws90_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int supercap_V  = (b[21] & 0x3f);
     int firmware    = b[29];
 
-    if (battery_lvl > 100) // More then 100%?
```
gdt, line 112: The comment about 1.5 vs 1.4 and the code do not line up at all.  The code is mapping 1.4 to 3.0 to be 0 to 100.  That doesn't make a lot of sense.   Also "lithium battery" is confusing, as there is lithium primary, lithion ion, and lithium iron phosphate.   I'm really not sure which kind we are talking about.

This comment may amount to "the code as it exists needs more help than your PR gives it" which might not be fair.

fallenpegasus:
It is recommend to use Lithium Metal batteries in any application where they may get well below freezing.  Such as weather stations.  LM chemistry is 1.5V when new, unlike 1.4 or 1.3 for ordinary alkaline batteries, so they can be more than 100%.

The "fineoffset_ws80.c" doesn't have this "cap to 100%".  No other device driver that I looked at has "cap to 100%".  Why should this one cap to 100%?

gdt:
> It is recommend to use Lithium Metal batteries in any application where they may get well below freezing. Such as weather stations. LM chemistry is slightly higher voltage than alkaline when new, so the battery_ok % can be more than 100%.

There's a lot going on here and I feel like things are blurred.   A normal alkaline AA will be at 1.6V when new (open circuit) and rapidly drop to a bit over 1.5V.  In most devices, is not workable below 1.3V.   I can't believe it would be workable at 0.7V.  Even 1.2V would be iffy.   However this does not correspond to a linear state of charge from 0.7 to 1.5V.    My impression is that lithium metal batteries start out quite a bit higher (1.8V) and I'm unclear on voltage vs state of charge.

It's clear that applying a voltage to SOC function without knowing the chemistry doesn't really work.   So we aren't really getting "battery level"; we're getting "some linearized representation of voltage that's useful to humans".

I don't understand if this "1.4 to 3.0" is per cell or for 2 cells.  Given `ambientweather_wh31e.c` it seems that it's for 2, but there's no comment.

> The "fineoffset_ws80.c" doesn't have this "cap to 100%". No other device driver that I looked at has "cap to 100%". Why should this one cap to 100%?

That's a perfectly good argument, but your commit message didn't say that.   Instead it has  a  statement about "instead of 1.4V" that to me does not make sense.

Could you add a comment that this is for 2 cells (if that's indeed true; presumably you have a WS90 and can measure the cells with a voltmeter and cross check with rtl_433 output)?    And change the commit message to just say that this clamping to 100% is irregular as no other driver does it, and perhaps that if the voltage is higher than 1.5V/cell (and if you want, that this can happen with lithium metal primary cells or fresh alkalines) it's still useful to see that implicitly in a >100% battery level?  Then I think future readers would be better served in terms of correctly understanding what is going on.


Sorry if I'm sounding cranky, but I've read and tried to figure out a lot of code, and to me the goal is to get the wisdom in the author's head into a form that a future reader can quickly absorb.    I've read my own code years later and been glad I wrote decent comments as I have had no memory of it.

PR #2729, opened 2023-11-26
mongoose.c: Fix SO_NOSIGPIPE change

gdt:
On NetBSD 10, this fails to build, with both sock and nopipe undefined, and code reading agrees with the compiler.
  - Define nopipe, used in the ioctl, guarded by the same ifdef, in 2 places.
    - Add ifdef guard in the 1 existing place
  - Use nc->sock instead of bare sock

### DIFF
--- a/src/mongoose.c
+++ b/src/mongoose.c
@@ -3877,11 +3877,14 @@ static int mg_is_error(void) {
 void mg_socket_if_connect_tcp(struct mg_connection *nc,
                               const union socket_address *sa) {
   int rc, proto = 0;
+#ifdef SO_NOSIGPIPE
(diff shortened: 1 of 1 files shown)


gdt:
I don't see that the build failure is related to building on NetBSD.  From code reading, this must fail to build on all platforms that support `SO_NOSIGPIPE`.

gdt:
rtl_433 builds ok on Raspberry Pi OS 12 and actual Debian 12, but both appear not to define SO_NOSIGPIPE.

This PR builds ok on Debian 12.

zuckschwerdt:
Thanks! I made quite some copy+paste errors there.

PR #2741, opened 2023-11-29
ci: Apply clang-format on c-code (with pre-commit) as seen in #2740

mdeweerd:
This applies clang-format on files that are not excluded from formatting in the pre-commit configuration.


zuckschwerdt:
Technically `sizeof` is not a function and there should be a space after it. But I take it nobody ever minded seeing it used as a function ;)

The first change, the multi-line `if` is more readable as-is -- do you see any possible change in the clang-format defs to make that possible?

mdeweerd:
> Technically `sizeof` is not a function and there should be a space after it. But I take it nobody ever minded seeing it used as a function ;)

Well, it can be understood as a function as it yields a value, where if, while, switch do not yield values (and never can).  It's a matter of choice, and then one needs to be consistent because a bit further there is `sizeof(msg)`.  IMHO th econsensus is not not put a space.

> The first change, the multi-line if is more readable as-is -- do you see any possible change in the clang-format defs to make that possible?

I do not think it is possible with clang - it may be with uncrustify where there are many more options.
The solution is to exclude that part of the code from clang-format.

zuckschwerdt:
> it can be understood as a function as it yields a value

It can. But the reasoning is not sound. `sizeof` is a unary operator, like e.g `-` (minus) as in `bar = -foo` -- one operand and usable as rvalue (yields a value).
And annoyingly this is an example where we wouldn't put a space after the operator (as opposed to the infix `-` as in `c = a - b`) -- yes, a matter of taste :) (incidentally the parens there are thus part of the expression and not delimiting arguments… I can't see how that matters but in C things like that tend to wreck havoc).
Agreed on consistency as the goal.

PR #2742, opened 2023-11-29
minor: apply clang-format

mdeweerd:
This is another change that is impacted by clang-format.


mdeweerd:
The original file looks ok after testing pre-commit on this modification - on my setup it reverts this change - not sure why it is reported in ci.

PR #2746, opened 2023-12-04
Add support for Grill Thermometer (RF-T0912)

eshaz:
* Adds support for the "Remote Grill Thermometer" (RF-T0912)
  * FCC ID: TXRFPT0912
  * Product Manual: 
     image

### DIFF
--- a/src/devices/grill_thermometer.c
+++ b/src/devices/grill_thermometer.c
@@ -19,17 +19,16 @@ Manufacturer:
 Supported Models:
 - RF-T0912 (FCC ID TXRFPT0912)
 
-0 - 255 F, frequency 434.052 MHz
+9 - 415 F, frequency 434.052 MHz
 
 Data structure:
 
 10 repetitions of the same 24 bit payload.
 
-    11111111 AAAAAAAA BBBBBBBB
+    AAAAAAAA AAAAAAAA BBBBBBBB
 
-- 1: 8 bit preamble. Always set to 0xff
-- A: 8 bit temperature in Fahrenheit. Calculated as (Temp = 255 - A)
-- B: Copy of 'A' presumably for message integrity
+- A: 16 bit temperature in Fahrenheit. Big Endian.
+- B: Checksum of A
 
 */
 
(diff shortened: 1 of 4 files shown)


src/devices/grill_thermometer.c
```diff
@@ -0,0 +1,106 @@
...
+                raw_value = current_value;
+                repeats   = 0;
+            }
+            else {
+                repeats++;
+            }
+        }
+    }
+
+    if (raw_value == -1 || repeats < 1)
```
zuckschwerdt, line 71: Always use `{ }`

src/devices/grill_thermometer.c
```diff
@@ -0,0 +1,106 @@
...
+        return DECODE_ABORT_EARLY;
+
+    data_t *data;
+    int id = 0, temp_f = 0xFF - raw_value;
+
+    /* clang-format off */
+    data = data_make(
+            "model",            "",             DATA_STRING, "Grill Thermometer",
```
zuckschwerdt, line 79: Name it `RF-T0912`

src/devices/grill_thermometer.c
```diff
@@ -0,0 +1,106 @@
...
+
+static char const *const output_fields[] = {
+        "model",
+        "id",
+        "temperature_F",
+        NULL,
+};
+
+r_device const grill_thermometer = {
+        .name        = "Grill Thermometer",
```
zuckschwerdt, line 97: `"RF-T0912 grill thermometer"`

src/devices/grill_thermometer.c
```diff
@@ -0,0 +1,106 @@
...
+        .modulation  = OOK_PULSE_PWM,
+        .short_width = 252,
+        .long_width  = 736,
+        .gap_limit   = 5000,
+        .reset_limit = 8068,
+        .sync_width  = 980,
+        .decode_fn   = &grill_thermometer_callback,
+        .fields      = output_fields,
+};
```
zuckschwerdt, line 106: End with a linefeed

zuckschwerdt:
> one subtraction operation at the end, vs. invert the entire data.

Maybe do an xor, it's easier to glance what's going on then.

zuckschwerdt:
The "model" needs to be"RF-T0912" -- our rule is `alphanum`+ `-` `alphanum`+

eshaz:
> The "model" needs to be"RF-T0912" -- our rule is `alphanum`+ `-` `alphanum`+

I was just looking at this. There are no labels on the device, other than the FCC ID which I had to look up in their database to get the model number. Wondering if it makes sense to allow more description here so it's easier to identify this device in the rtl_433 output?

merbanan:
The documentation line should have the "Remote Grill Thermometer" name and the short model name.

eshaz:
> The documentation line should have the "Remote Grill Thermometer" name and the short model name.

I'll make the change, but I still think could be confusing to omit this data from the event. Is there any place where I can add a "Remote Grill Thermometer" description on the output data for the decode events?

eshaz:
I did some more testing with the device and it appears to support a two byte temperature range (around 9F to 410F) and also uses checksum validation, which makes more sense than what I previously thought. I've added these changes.

src/devices/grill_thermometer.c
```diff
@@ -0,0 +1,109 @@
...
+
+    if (temp_f == -1029) {
+        temp_f   = 0;
+        overload = 1;
+    }
+
+    /* clang-format off */
+    data_t *data = data_make(
+            "model",            "",             DATA_STRING, "RF-T0912",
+            "temperature_F",    "Temperature",  DATA_FORMAT, "%i F", DATA_INT, temp_f,
```
zuckschwerdt, line 80: use `DATA_COND` to remove this key on overload

src/devices/grill_thermometer.c
```diff
@@ -0,0 +1,109 @@
...
+            "overload",         "Overload",     DATA_STRING, overload ? "true" : "false",
```
zuckschwerdt, line 81: boolean needs to be `DATA_INT`

src/devices/grill_thermometer.c
```diff
@@ -0,0 +1,109 @@
...
+        uint8_t *row_data = bitbuffer->bb[row];
+        checksum          = row_data[0] + row_data[1];
+
+        if (bitbuffer->bits_per_row[row] != 24 ||
+                checksum != row_data[2] ||
+                checksum == 0) {
+            continue;
+        }
+
+        short current_value = (row_data[0] << 8) + row_data[1];
```
zuckschwerdt, line 57: we usually write this as:
`int current_value = (int16_t)(row_data[0] << 8) | row_data[1]; // uses sign-extend`

merbanan:
@zuckschwerdt LGTM, any comments?

merbanan:
Code merged into main branch.

PR #2748, opened 2023-12-05
Update rtl_433_mqtt_hass.py to support storm_dist_km by WH31L

mailaenderli:
This adds support for the storm_dist_km field in WH31L


zuckschwerdt:
Thanks!

PR #2755, opened 2023-12-18
Memory leak when converting inches to mm - https://github.com/1technophile/OpenMQTTGateway/issues/1693

NorthernMan54:
Identified a memory leak that occurs when converting inches to mm.  This was identified by the rtl_433_ESP community - https://github.com/NorthernMan54/rtl_433_ESP/pull/122/files.


zuckschwerdt:
Bug has been introduced in Apr 2019 by #1031

zuckschwerdt:
Though `_inch` was deprecated since Mar 2019 and then removed with #1735 in Jun 2021. We should remove that conversion.

NorthernMan54:
Tks

PR #2765, opened 2023-12-26
Remove 100% cap for battery_lvl in fineoffset_ws90.c

fallenpegasus:
No other drivers cap the battery level, so this one should not either.

The WS90 backup power, used when the solar photovoltatic charged supercap is empty, is provided by 2 AA batteries.

The standard manufacturer specification is that the nominal voltage of a fresh alkaline battery is 1.5V.

The voltage of a new alkaline battery ranges from 1.50V to 1.65V, depending on the purity of chemistry and the quality of manufacturing.

The voltage of a new lithium iron disulphide battery is 1.72V.  The lithium iron disulphide chemistry is good down to -50C and has an operational life of 15 years, so it is often used for weather stations such as the WS90.

It is useful to the user to be able to see a Battery Level >100% to let them know that they have a pair of very high quality batteries installed and that they are fully fresh.


zuckschwerdt:
I second that we should not change correc values that are just slightly out of bounds.
Code is from #2448, @jpochmara do you agree?

zuckschwerdt:
Isn't this a dup of #2728? Why did resubmit the same patch?

jpochmara:
In this pull request  [#2448](https://github.com/merbanan/rtl_433/pull/2448) we talked about this issue, and you thought we should clip the results at 100% (1.00).

fallenpegasus:
it is a dupe of #2728 which i closed, and then redid

PR #2772, opened 2024-01-02
Add Ecowitt air quality sensors to rtl_433_mqtt_hass.py

ChrisRomp:
Updates to `examples/rtl_433_mqtt_hass.py` adds new sensor configurations for various measurements, including PM 2.5 concentration, PM 10 concentration, CO2 concentration, protocol used, and external power status.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -750,17 +750,6 @@ mappings = {
         }
     },
 
-    "protocol": {
-        "device_type": "sensor",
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
I don' think we'd want to add "protocol", otherwise looks ok.

ChrisRomp:
@zuckschwerdt Happy to remove it. I found it useful for debugging rogue entries showing up in HA and it’s disabled by default, but I can omit it if it gets the rest merged faster.

Issue #2786, opened 2024-01-11
Figuring out units for Ecowitt WS68 in ambientweather_wh31e.c


nitrogen76:
I have two wind reporting instruments.  (I am writing a set of scripts to automate setup and ingestion of weather data mostly for my own use) 
I have an acurite 5-n-1 and a Ecowitt WS68

the acuirite reports to influx as "wind_avg_km_h/wind_gust_hm_h" while the EcoWitt reports as "wind_avg_raw/wind_gust_raw"

What do i need to do to get the Ecowitt reporting with the same tags?
I'm not very good in C, but i see code in that module that would report this way, with tags about "clang-format on/off" and i'm not sure what that means.

My commandline:

rtl_433 -s 1024k -d:00000915 -M time:unix:usec:utc -f 915M -R 142 -R 40 -R 73 -R 113 -R 42  -M protocol -M level -C si -Y autolevel -F "influx://${influx-db-url}" 

Thanks for your help.

zuckschwerdt:
I added the WS68 in 2020 as preliminary for #1283. We still need a user to figure out units and scaling of
- battery_raw
- lux_raw
- wind_avg_raw
- wind_max_raw
- wind_dir_deg_raw

There was new interest in the WS68 / that issue in 2022 but it didn't conclude anything iirc.

It would be great if you can figure out things. Maybe by comparing both your units?

nitrogen76:
Ok, that's a GREAT reason!

I will try and do this.  Is placing my findings in this thread okay, or is there a way that's better for you?

zuckschwerdt:
Perfectly fine. Report anything you find. Thanks!

nitrogen76:
Ok, so the 5n1 doesnt report wind speed exactly like the WS68 does, but it APPEARS it's reporting in km/h, as any other 
measurement for my wind instrument doesn't make sense for the weather conditions.  They match close enough.

* Wind speed: km/h (avg and max) (max is gust)
* wind_dir_deg is degrees, with 0 being north
* battery_raw appears to be a arbitrarily figured percentage of battery used, but i'm not sure how accurate i am in that.

a brand new battery last night was at 90%.  A battery with 1.485 volts came in at 75%, and a completely used battery came up at 0%.

The funny thing is, ZERO battery just uses the last reported value (this thing as a solar cell in it.)

[...]

Issue #1283, opened 2020-01-31
Ecowitt WS68 Anemometer


tolip-ydob:
I'd like to point out that I tried many times to attach a .zip, .gz, or .tar file to this post.
I did manage to get github to format this post properly.

```
Ecowitt WS68
Lux, Battery, Wind speed, Wind direction, Gust speed,  and UV

I'm posting what I have so far here. I am willing to submit to git repository but I am ignorant, I never added files or submitted a pull request before.
Is there a cheat sheet for doing this? I do read docs but I'm netested too deep at the moment.

My command line
rtl_433 -d 0 -f 915e6 -g 14 -M newmodel -X "n=WS68_WAS_WH51_soil_sensor,m=FSK_PCM,s=58,l=58,r=5000,g=4000,preamble=aa2dd4" > X.68hex.stdxxx.out 2>&1

I found the Flex decoder in rtl_433/src/devices/fineoffset.c line # 516  as of 20200130 

I used low-ish gain because device is close to rx and it has too much juice otherwise.
I am also in a target rich environment and too much stuff spews forth otherwise. 


The fun begins :-)

680000c500004b0fffff00000000d0af104 ### WS68            4b-VDC                   00-Wind-Direction-North 
680000c500004b0fffff005a0000d0af104 ### WS68            4b-VDC                   5a-Wind-Direction-East
680000c500004b0fffff00b4000079b2102 ### WS68            4b-VDC                   b4-Wind-Direction-South
680000c500004b0fffff7ee0940075ec102 ### WS68            4b-VDC  7e-Wind-Speed    e0-Wind-Direction-SouthWest      94-Wind-Gust
680000c500004b2fffff000e00008033208 ### WS68            4b-VDC                  10e-wind-Direction-West
680000c5000f4b0fffff002e0000d395108 ### WS68   0f-Lux   4b-VDC                   2e-Wind-Direction-NorthEast-ish
680000c501074b0fffff002e0002a663100 ### WS68   107-Lux  4b-VDC                   2e-Wind-Direction-NorthEast-ish

The .cu8 files are labeled similar to the descriptions above, the filenames contain the hex value to expect. 

Use a small fixed width font and my ASCIIdrawing(below) will look correct on a 180 column terminal. They are capital "i" and capital "o" because I used sed LOL.

[...]

PR #2790, opened 2024-01-14
Improve WS68 LUX and UVI decoding and add units

ProfBoc75:


zuckschwerdt:
Improves #1283
S.a. #2786

Issue #2796, opened 2024-01-19
Error in Bresser Lightning Decoder


matthias-bs:
Decoding of the `count` value is wrong in https://github.com/merbanan/rtl_433/blob/master/src/devices/bresser_lightning.c

It has been assumed that `count` is encoded as a binary counter.

I tried to find the counter's maximum value before rollover by triggering the sensor with a [variant](https://github.com/matthias-bs/BresserWeatherSensorReceiver/tree/main/extras/hw_test/lightning/lightning_tester) of Michael Gasperi's [AS3953 Lightning Generator](https://sites.google.com/view/as3935workbook/#h.5x43kh1ifn4o).

This way, I found out that `count` is actually encoded as BCD, but with the most significant digit counting up to 15 (sic!). Thus, the maximum counter value is 1599.

A bugfix will follow.

PR #2797, opened 2024-01-19
Fix bresser_lightning - Fixed decoding of 'count'

matthias-bs:
See https://github.com/merbanan/rtl_433/issues/2796


zuckschwerdt:
Thanks for the thorough investigation!

PR #2801, opened 2024-01-22
Fix example rtl_433_mqtt_hass.py rain_in sensor value template

shampeon:
This PR fixes the `rain_in` sensor in `examples/rtl_433_mqtt_hass.py` to return the measured rainfall in inches.

Previously, the sensor would return the value in mm, causing inconsistencies when setting the units using the `-C` option.


zuckschwerdt:
S.a. Discussion #2782

PR #2809, opened 2024-01-30
Added ccache speedup compilation

GermanAizek:
More information: https://ccache.dev

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,12 @@ if(NOT WIN32)
     add_definitions(-DINSTALL_SYSCONFDIR=${CMAKE_INSTALL_FULL_SYSCONFDIR})
 endif()
 
+find_program(CCACHE "ccache")
+if(CCACHE)
+    set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE})
+    set(ENV{CCACHE_SLOPPINESS} pch_defines,time_macros)
+endif(CCACHE)
+
 ########################################################################
 # Use pkg-config
 ########################################################################


zuckschwerdt:
Thanks! Is Ccache something people use with small projects like rtl_433? Is it expected behaviour to automatically employ Ccache on some projects? Wouldn't a Ccache user just use `cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache ...`?

GermanAizek:
> Wouldn't a Ccache user just use `cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache ...`?

Do you suggest specifying -D flag every time for configuration project? Then write it in BUILDING.md documentation.

zuckschwerdt:
That's my question. I don't know anybody using Ccache. Is it more a niche or mainstream thing? Do people use it by choice (and know that with cmake the -D flag is needed)? Or are there installations/distros where Ccache is present but hidden from the user? Would they want / expect Ccache to automatically run?
(in my Distro most mentions (46+) of CCACHE in the package tree are disabling it -- it seems like a disliked/unwanted default.)

gdt:
I use ccache, to speed up repeat builds.   In general I like it.   However occasionally it causes trouble.

My view is that a package should not automatically find and use ccache.   The standard approach in ccache is to create a bin directory where gcc/g++/etc. are linked to cmake and stick it in the path.

As I understand things,  CMAKE_C_COMPILER_LAUNCHER is a generic cmake instruction, not about rtl_433.  It does not appear in rtl_433's sources.

Therefore my view is that rtl_433 should try to auto-enable this, and should not even document it, because that would lead to inlining the entire cmake manual.  If anything how to deal with cmake belongs in the ccache man page.

Issue #2816, opened 2024-02-04
Bresser Air Quality sensors, ignore first packet


Mart124:
Hello,

There's a tiny issue with the following 3 sensors :
Bresser CO2 sensor, PN 7009977
Bresser HCHO/VOC sensor, PN 7009978
Bresser PM2.5/PM10 sensor, PN 7009970
(I do not own other Bresser sensors, so can only report for these 3).

When they start, they all send a first packet with their values set to 15.
CO2, HCHO, VOC, PM2.5, PM10 : all set to 15.
Of course these are dummy values, certainly because these devices send a first packet before their sensors had time to compute a real value.

Could we then ignore these "15" values please ?
So that they do not mess-up graphs etc... at presentation level (Home Assistant for instance).

Thank you very much 👍

ProfBoc75:
Hi @Mart124 : 

Thanks to report this use case.
It's normal behavior, as such sensors require preheating to generate the correct calibrated value, so the first one is not good and/or not yet filled.

Yes, for me their is 2 possible approaches around this use case:

- First one, we want from rtl_433 to alter the result as less as possible, so we would like to let any filtering of values outside rtl_433.

- Second one, if we can identify a flag present only during this first message then we can take it into account and not provide any value. For that, we need you to capture the codes at this exact moment of the first signal is sent, then, we may found into the codes (raw MSG and de-whitened XOR) this specific flag (like startup flag is already existing) that could help us identify this 15 "wrong" value, then ignore it with a simple DATA_COND and provide only this init situation, like "init = 1".

if you captured already this moment, replay it and provide both MSG and XOR codes :
`rtl_433 *.cu8 -R 173:vv`

sample result :
![image](https://github.com/merbanan/rtl_433/assets/62882637/4ab10c54-f569-45b5-a459-26f801bf4ddf)

or redo it in live :
`rtl_433 -f 868.34M -R 173:vv`

In that situation, the sensor ID may change too, but it will definitely help us to get the good figures.

Thanks for your contribution.

zuckschwerdt:
I suspect that the indicator is a raw `0xF` in the BCD values. We can DATA_COND those values, e.g. `int co2_ok = msg[5] & 0x0f == 0x0f;`

ProfBoc75:
@zuckschwerdt : you're right , I forgot it was coded in BCD, so yes, easy to filter out.

Mart124:
[...]

Issue #2814, opened 2024-02-02
New device Bresser HCHO/VOC sensor, PN 7009978


Mart124:
Hello,

Bresser does a HCHO/VOC sensor, PN [7009978](https://www.bresser.de/en/Weather-Time/BRESSER-HCHO-VOC-Sensor.html), [user manual](https://www.bresser.de/out/media/8b90e446809ad4799c34f577ee2db342.pdf).

Would be really nice to support it, in addition to the (same form-factor) Bresser [PM2.5 / PM10](https://github.com/merbanan/rtl_433/issues/2693) sensor and Bresser [CO2](https://github.com/merbanan/rtl_433/issues/2813) sensor.

Here are then some samples from it : [7009978.zip](https://github.com/merbanan/rtl_433/files/14145559/7009978.zip) :
- g001_868.34M_250k.cu8 : 5 ppb / 5 leaves
- g002_868.34M_250k.cu8 : 5 ppb / 5 leaves
- g001_868.34M_1000k.cu8 : 5 ppb / 4 leaves
- g002_868.34M_1000k.cu8 : 5 ppb / 4 leaves
- g003_868.34M_1000k.cu8 : 8 ppb / 2 leaves
- g004_868.34M_1000k.cu8 : 4 ppb / 1 leaves

Could we then try to decode it please ?
Of course, feel free if needed !

Thank you very much 👍

ProfBoc75:
Hi @Mart124 

What I already found is that the signal is like other Bresser devices,
you can try with :

live
`rtl_433 -f 868.34M -X 'n=Bresser,m=FSK_PCM,s=124,l=124,r=1000,preamble=aaaa2dd4'`

or replay 
`rtl_433 -X 'n=Bresser,m=FSK_PCM,s=124,l=124,r=1000,preamble=aaaa2dd4' -r *.cu8`

Then the data needs to be decoded, looking into the other Bresser sensors, we need to fallow a specific data processing to get the figures. I'm working on it and let you know.

it's the same for #2813, I guess we can combine both in the same decoder.

ProfBoc75:
Good progress for me, I found that the Bresser 7in1 should match, it found s_type=10 (CO2 / 7009977) and 11 (HCHO/VOC / 7009978) not yet decoded, where s_type=8 is already a AIR_PM sensor.

To be continued ....

Mart124:
Wow, good news @ProfBoc75 👍
Really glad you took a look at this, and managed to get good progress !
[...]

PR #2817, opened 2024-02-05
Improve bresser_7in1.c excluding init wrong BCD values

ProfBoc75:
Related to the issue #2816 , first signal is sending 0x0F values, which is not possible for BCD that can't be > 0x09.

This PR is to exclude these values through a simple DATA_COND formatting.

### DIFF
--- a/src/devices/bresser_7in1.c
+++ b/src/devices/bresser_7in1.c
@@ -122,6 +122,9 @@ XOR / de-whitened :
 Data layout de-whitened :
     DIGEST:16h ID:16h PPB:16h 8x8x8x8x8x8x8x8x8x8x4x BATT:1b 3x8x8x8x8x8x4x VOC:4h TRAILER:16x
 
+#2816 Bresser Air Quality sensors, ignore first packet:
+    The first signal is not sending the good BCD values , all at 0xF and need to be excluded from result (BCD value can't be > 9) .
+
 First two bytes are an LFSR-16 digest, generator 0x8810 key 0xba95 with a final xor 0x6df1, which likely means we got that wrong.
 */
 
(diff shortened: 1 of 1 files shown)


Mart124:
@ProfBoc75 strangely enough, since this commit (and #2830), rtl_433 does not receive values from my VOC/HCHO sensor anymore.
No issue with PM2.5/PM10 and CO2 sensors (even better, as wrong values are now properly ignored).

At least I think there's a small bug here :
```
        int hcho_init = (msg[5] & 0x0f) == 0x0f;
        int voc_init  = voc == 0x0f;
```
I think we should only rely on `voc_init`, as VOC value of 15 can't exist, it's out of scale (5 max here).
On the contrary, a proper HCHO value of 15 can be expected.
So I think we should use `voc_init` in place of `hcho_init` in the HCHO `DATA_COND`.
Or we would miss proper HCHO values of 15...

But even with this, I can't see why rtl_433 does not receive values from this device anymore.
Did you manage to replay against samples from the initial device report (#2814) ?

Thank you again 👍

Mart124:
Mmmh I finally may be wrong with my `voc_init` vs `hcho_init` proposal above, as per what you stated a few lines above in the code of this PR :
```
#2816 Bresser Air Quality sensors, ignore first packet:
    The first signal is not sending the good BCD values , all at 0xF and need to be excluded from result (BCD value can't be > 9) .
```
So should not be an issue...

Mart124:
OK, found the culprit, @zuckschwerdt renamed the device in https://github.com/merbanan/rtl_433/commit/4c5081dbba8ac3a95a169425f8e28ca2c2e10746 :)
Sorry for the noise...

zuckschwerdt:
Oops sorry, yes. We have strict naming rule https://triq.org/rtl_433/DATA_FORMAT.html#message-data

Issue #2840, opened 2024-02-17
paho-mqtt 2.x is out, breaking compatibility with rtl_433_mqtt_hass.py


deviantintegral:
The current directions don't specify a version of paho-mqtt to install, which means users will now be getting the 2.x branch. I know for sure this breaks rtl_433_mqtt_hass.py, and it wouldn't surprise me if it [examples/rtl_433_mqtt_relay.py](https://github.com/merbanan/rtl_433/blob/5786c01c8ca89d689dbaf6ffba279dfd4812d19b/examples/rtl_433_mqtt_relay.py#L11) too.

```
Starting rtl_433_mqtt_hass.py...
[2024-02-17T10:53:24+0100] INFO:root:Enabling debug logging
[2024-02-17T10:53:24+0100] INFO:root:Discovering all devices
Exception ignored in: <function Client.__del__ at 0x7fb05d0900>
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/paho/mqtt/client.py", line 874, in __del__
    self._reset_sockets()
  File "/usr/local/lib/python3.12/site-packages/paho/mqtt/client.py", line 1133, in _reset_sockets
    self._sock_close()
  File "/usr/local/lib/python3.12/site-packages/paho/mqtt/client.py", line 1119, in _sock_close
    if not self._sock:
           ^^^^^^^^^^
AttributeError: 'Client' object has no attribute '_sock'
Traceback (most recent call last):
  File "/rtl_433_mqtt_hass.py", line 990, in <module>
    run()
  File "/rtl_433_mqtt_hass.py", line 917, in run
    rtl_433_bridge()
  File "/rtl_433_mqtt_hass.py", line 888, in rtl_433_bridge
    mqttc = mqtt.Client()
            ^^^^^^^^^^^^^
TypeError: Client.__init__() missing 1 required positional argument: 'callback_api_version'
```

https://eclipse.dev/paho/files/paho.mqtt.python/html/migrations.html has upgrading directions.

We fixed this at https://github.com/pbkhrv/rtl_433-hass-addons/pull/180 by pinning the old version for now.

zuckschwerdt:
Thanks! We should check if we can use 2.x by just using `mqttc = mqtt.Client(mqtt.CallbackAPIVersion.VERSION1)`

zuckschwerdt:
Closed by #2916

Issue #1964, opened 2022-01-25
RFC: Data sanity checking and filtering in rtl_433


rct:
For various reasons rtl_433 may output messages with bad data, such as humidity values that fall outside of the range of 0 - 100. For most protocols message integrity checks (MIC) based on checksums, CRCs, and/or parity bit should cause bad messages to be rejected.  However this doesn't always work. MIC are not 100% accurate. In some cases it is possibly hardware error with the device itself, a sensor might be bad or have a temporary glitch, but the device calculates the MIC correctly with the bad data. (There have been cases in some decoders identified where a bitbuffer of all zeros could pass the MIC checks like checksum. In a number of cases, those decoders have had checks added to prevent this.

This is a topic that at first glance seems somewhat trivial, but it isn't. Points of view may change after a year or more of operating a system with a variety of sensors. There is no one correct answer to this for all applications. I've been thinking about this for a while, but haven't tried to implement anything yet because I don't have a view yet for what is the correct thing. 

## The case for not filtering in rtl_433

This leads to the question of whether rtl_433 should do some sanity check of data, An argument against doing sanity checking in rtl_433 is an architectural view that rtl_433 should do its best to handle the RF and lower layers of the bit manipulation in a very general manner and then get out of the way to be of use to a wide range of consumers. 

Related to this is the architectural view that rtl_433 should not be maintaining any state for devices, that should be up to consumers and state requirements might be not only consumer specific, but could also be device specific.

Not doing any filtering/sanity checking makes sense when application specific code reads from rtl_433 and can make it's own decisions. For example:  

    rtl_433 -> application with filtering -> database/view applications

## rtl_433's ability to implement sanity checking is limited by not keeping state

The architectural view that rtl_433 shouldn't maintain any state limits the types of sanity checking that can be done within rtl_433. Data that is within range but still potentially bad can't be filtered by rules that need history.  For example rejecting values because the rate of change exceeds a threshold.

Not keeping state in rtl_433 does have many benefits for a clean design. It helps keep rtl_433 general and usable for a wide range of uses cases beyond just receiving weather and environmental data.  An application that uses rtl_433 to receive data from security sensors has different requirements than a weather console. 

## Common use cases that need filtering (for lack of sanity checking in other applications)

However in the last few years, rtl_433 has added generalized publishing for messaging (MQTT) and storage (InfluxDB). A common use case is:

    rtl_433 -> MQTT -> Home automation platform (Home Assistant). 

[...]

PR #2841, opened 2024-02-18
Read rtl_433_mqtt_hass.py configuration from rtl_433.conf

klattimer:
As the title suggests, a reasonably self contained change to load the configuration for HASS from the rtl_433.conf file if it exists.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -980,7 +980,7 @@ def bridge_event_to_hass(mqttc, topic_prefix, data):
 def rtl_433_bridge():
     """Run a MQTT Home Assistant auto discovery bridge for rtl_433."""
 
-    mqttc = mqtt.Client()
+    mqttc = mqtt.Client(mqtt.CallbackAPIVersion.VERSION1)
 
     if args.debug:
         mqttc.enable_logger()
@@ -1126,7 +1126,7 @@ if __name__ == "__main__":
             args.password = password
         if retain:
             logging.info(f"Using retain {retain} from config")
-            args.retain = retain
+            args.retain = int(retain)
         if devices:
             logging.info(f"Using device_topic_suffix {devices} from config")
             args.device_topic_suffix = devices
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -105,6 +105,26 @@ import re
 
 discovery_timeouts = {}
 
+CONFIG_PATHS = [
+    "/etc/rtl_433",
+    "/usr/local/etc/rtl_433",
+    os.path.expanduser("~/.config/rtl_433"),
+    os.path.dirname(os.path.realpath(__file__)),
+    os.path.join(os.path.dirname(os.path.realpath(__file__)), "data")
+]
+CONFIG_PATHS = [p for p in CONFIG_PATHS if os.path.exists(p)]
+
+CONFIG_FILENAME = "rtl_433.conf"
+
+def find_file_in_path(filename, paths, success_log=None):
+    for path in paths:
+        fullpath = os.path.join(path, filename)
+        if os.path.exists(fullpath):
+            if success_log:
+                logging.info(success_log % path)
+            return fullpath
+    raise FileNotFoundError(f"File {filename} not found in paths: {paths}")
+
 # Fields that get ignored when publishing to Home Assistant
 # (reduces noise to help spot missing field mappings)
 SKIP_KEYS = [ "type", "model", "subtype", "channel", "id", "mic", "mod",
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -1053,13 +1053,67 @@ if __name__ == "__main__":
     if not args.password and 'MQTT_PASSWORD' in os.environ:
         args.password = os.environ['MQTT_PASSWORD']
 
-    if not args.user or not args.password:
-        logging.warning("User or password is not set. Check credentials if subscriptions do not return messages.")
-
     if args.ids:
         ids = ', '.join(str(id) for id in args.ids)
         logging.info("Only discovering devices with ids: [%s]" % ids)
     else:
         logging.info("Discovering all devices")
+
+    found_config = False 
+    try:
+        with open(find_file_in_path(CONFIG_FILENAME, CONFIG_PATHS)) as f:
+            found = re.findall('^output (.*)$', f.read(), re.MULTILINE)
+            devices = None
+            events = None
+            retain = None
+            user = None 
+            password = None 
+            host = None 
+            port = None
+            for find in found: 
+                if "mqtt" not in find: continue
+                host = re.findall(r'mqtt://(.*?),', find)[0]
+                devices = re.findall(r'devices=(.*?)(?:,+|$)', find)
+                if len(devices) > 0: devices = devices[0]
+                events = re.findall(r'events=(.*?)(?:,+|$)', find)
+                if len(events) > 0: events = events[0]
+                retain = re.findall(r'retain=(.*?)(?:,+|$)', find)
+                if len(retain) > 0: retain = retain[0]
+
+                if '@' in host:
+                    user, password = host.split('@')[0].split(':')
+                    host = host.split('@')[1]
+                if ':' in host:
+                    host, port = host.split(':')
+                found_config = True
+                break 
+    except FileNotFoundError:
+        logging.warning("Config file not found")
+    except Exception as e:
+        logging.exception("Error reading config file")
+
+    if found_config:
+        if host:
+            logging.info(f"Using host {host} from config")
+            args.host = host
+        if port:
+            logging.info(f"Using port {port} from config")
+            args.port = int(port)
+        if user:
+            logging.info(f"Using user {user} from config")
+            args.user = user
+        if password:
+            args.password = password
+        if retain:
+            logging.info(f"Using retain {retain} from config")
+            args.retain = retain
+        if devices:
+            logging.info(f"Using device_topic_suffix {devices} from config")
+            args.device_topic_suffix = devices
+        if events:
+            logging.info(f"Using rtl_topic {events} from config")
+            args.rtl_topic = events
+    if not args.user or not args.password:
+        logging.warning("User or password is not set. Check credentials if subscriptions do not return messages.")
 
     run()
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -883,6 +883,27 @@ def rtl_433_device_info(data, topic_prefix):
     return (f"{topic_prefix}/{path}", id)
 
 
+def device_topic_from_data(data):
+    device_topic_suffix = ''+args.device_topic_suffix
+    keys = re.findall(r'\[([^\]]+)\]', device_topic_suffix)
+    for key in keys:
+        default = None
+        k = key
+        if key.startswith('/'):
+            k = key[1:]
+        if ':' in key:
+            k, default = key.split(':')
+        value = data.get(k, default)
+        if key.startswith('/') and value is None:
+            device_topic_suffix = device_topic_suffix.replace(f'[{key}]', '')
+            continue
+        elif key.startswith('/') and value is not None:
+            value = '/' + str(value)
+
+        device_topic_suffix = device_topic_suffix.replace(f'[{key}]', str(value))
+    return device_topic_suffix
+
+
 def publish_config(mqttc, topic, model, object_id, mapping, key=None):
     """Publish Home Assistant auto discovery data."""
     global discovery_timeouts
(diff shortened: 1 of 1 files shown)


gdt:
Currently, rtl_433_mqtt_hass is an example.  Even if it were promoted to first class and supported, I think it's a mistake to blur the programs together.   rtl_433_mqtt_hass should have its own config file, and it can be next to rtl_433.conf.

So I object to this PR being merged, at least for now.

What problem are you trying to solve?

gdt:
I don't really object to the hass program reading the config to rind out about rtl-433 config, but I don't think that any further hass/mqtt config belongs  in the main config file.   We have so far not crossed into configuring examples in the main config, and I don't think we should do that.

One might ask instead why the rtl_433 config is needed, view that as a bug, and try to resolve it.  I personally run rtl_433 without a config file.

rct:
> One might ask instead why the rtl_433 config is needed, view that as a bug, and try to resolve it. I personally run rtl_433 without a config file

That’s a pretty unhelpful statement in this context. I’m prone to saying things like that too. But let’s take a step back for some reflection on this.

——

This project has got to be getting close to 10 years old by now if it isn’t already. I’m pretty sure I have data files and maybe commits from 2015.  Things have changed since then.

The user community for rtl_433 has grown since the addition of Home Assistant and its add-ons with rtl_433 and the auto discovery script packaged in somewhat ready to run containers. It would be nice if there were some usage metrics to help measure this.

For this segment of the user community, for better or worse rtl_433 and the auto discovery script are conflated into one service. This comes up often when people ask for help. The technical barriers to entry have been lowered.  There is some expectation of an experience similar to using Zigbee networks.

From an rtl_433 developer’s perspective, the auto discovery script is just one of several examples and is a second or third class concern. While this is somewhat understandable, it can be frustrating to a segment of the user population. It might be time for some of these examples to be considered more like a _companion_ app rather than an example. 

So breaking the questions and concerns down:

**Example/Companion apps reading rtl_433 configuration**

I think it is pretty easy to agree that it makes sense, where possible, to avoid redundant configuration like MQTT connection information, topic strings, also possibly frequency, what protocols are enabled/disabled, etc. 

**Example/Companion apps storing their configuration information in the same place as rtl_433**

This is where any disagreement starts. While the title of this PR is “read”, the objection is on sharing the config file.  

From an rtl_433 developer’s perspective, it is easy to see an objection to having other configuration data stored together that might at some point cause some sort of conflict. 

However, from a user’s perspective, it is hard to understand and explain why there are two configuration files and how to figure out which one to put configuration information in.

If there was a design for the configuration syntax/grammar to allow clear delineation (namespaces?) then it might be possible to store the union of both sets of configuration data in one place and avoid conflict.

I think it would be useful to get @merbanan and @zuckschwerdt views on this. 

**Another alternative for apps that use/embed rtl_433**

An application that manages rtl_433 and other processes could just have it’s own single configuration source that is used to generate what ever files or command line arguments are needed for rtl_433 and the other processes.

**Problems with the hass auto discovery script**

It is probably worth mentioning here that some of these issues arise because the hass auto discovery script is somewhat of a quick hack rather than a well thought out application.  It winds up being useful enough that it doesn’t get replaced.  It probably works OK in 60-80% of the common cases. 

Also because it is primarily used when getting started, once things are working well enough it tends to be forgotten. It’s no longer needed if the configuration is retained and no new devices are added.

gdt:
Probably we should think about "is our stance that these are examples reasonable".   They are more than that, and I think if we elevate them to first-class, then we have a different situation.

What I don't like is merging things while out of the other side of our mouths saying that examples are 2nd/3rd class.

As for mqtt config, I sort of view it as a but that rtl_433 directly does mqtt, vs using a helper program.   Probably there should be a single program that receives json over udp and does all the mqtt including the config topics, optionally.

So I think we're already not in a great place and it would be better to dig out of it instead of blurring the primary program and examples.  I am not sure it would be that hard.

examples/rtl_433_mqtt_hass.py
```diff
@@ -1053,13 +1053,67 @@ def run():
...
+
+                if '@' in host:
+                    user, password = host.split('@')[0].split(':')
+                    host = host.split('@')[1]
+                if ':' in host:
+                    host, port = host.split(':')
+                found_config = True
+                break 
+    except FileNotFoundError:
+        logging.warning("Config file not found")
```
rct, line 1134: It would be helpful to be less ambiguous about what config file you are talking about in this and the other log messages. 

I would explicitly mention that it is rtl_433's config file and/or the path.

Some of the Home Assistant add-ons have their own config files for this script.

examples/rtl_433_mqtt_hass.py
```diff
@@ -1053,13 +1053,67 @@ def run():
...
+    except Exception as e:
+        logging.exception("Error reading config file")
```
rct, line 1136: Same as above, be explicit about the file.

examples/rtl_433_mqtt_hass.py
```diff
@@ -1053,13 +1053,67 @@ def run():
...
+
+    if found_config:
+        if host:
+            logging.info(f"Using host {host} from config")
```
rct, line 1140: Same as about, be more explicit about the source, "from config" is a bit ambiguous.

rct:
> As for mqtt config, I sort of view it as a but that rtl_433 directly does mqtt, vs using a helper program. Probably there should be a single program that receives json over udp and does all the mqtt including the config topics, optionally.

Given that this PR just adds the ability to read the MQTT configuration from the rtl_433 config file to eliminate redundant configuration that seems to be looking for a whole different battle.  

@zuckschwerdt added MQTT, HTTP, InfluxDB, etc. a while ago. If you want to change that, it's not going to happen by rejecting this PR.

If you consider the auto discovery script to be just part of the collection of examples, this PR should stand on it's own merits against that example. It doesn't seem like it is going to negatively impact rtl_433 architecturally.

examples/rtl_433_mqtt_hass.py
```diff
@@ -943,6 +964,8 @@ def bridge_event_to_hass(mqttc, topic_prefix, data):
     published_keys = []
 
     base_topic, device_id = rtl_433_device_info(data, topic_prefix)
```
rct, line 966: Minor nit, but it feels a bit unclean that you didn't remove the old assignment to `base_topic`. 

I know this whole script needs work, but If you don't want to refactor rtl_433_device_info, then just use:

 `_, device_id = rtl_433_device_info(data, topic_prefix)`.  

The underscore seems to be the prevailing convention for a value you don't care about when unpacking a tuple during assignment.

zuckschwerdt:
As it was mentioned here, rethinking/modularizing the architecture towards (external) modules that reformat/aggregate/... data before passing it out (e.g. by MQTT) is a goal. It seems useful with the complexities of matching MQTT to some eco-systems, or e.g. with problems in representation like Influx, where we now need to decide wether something is "data" or "label" and can easily configure it.

rct:
Note: the commit https://github.com/merbanan/rtl_433/pull/2841/commits/10e5ea9ece5bf2e73a7e035d0fec8d89bca33886 from this PR should fix issue #2840.

Right now it looks like there are two fix related commits in this PR, and two commits for adding the feature of pulling the MQTT connection details from the rtl_433 config file:  https://github.com/merbanan/rtl_433/pull/2841/commits/aa313f5d15c8d8ad76a6b346541ce0f0ce50eff8  (and the logic for finding the config file https://github.com/merbanan/rtl_433/pull/2841/commits/c3601900bf5c6c3b59753d724d58520681b88846 )

With regards to the fix related commits:

@klattimer do you have an example/test case of what https://github.com/merbanan/rtl_433/pull/2841/commits/f109051b984dedc32a870cef3e5c4a66a2d1eada fixes?

@deviantintegral - if you get a chance, it would be good if you could look over that change. 

With regards to the rest of the discussion here about coupling between rtl_433, the auto discovery script, and MQTT: 

This PR probably would have generated less notice if it had been clearer that the goal is read the MQTT connection information from rtl_433's config file and not any larger coupling between the two.

(I'd suggest renaming the PR to _rtl_433_mqtt_hass.py: Read MQTT connection configuration from rtl_433.conf_. )

Further discussion about architectural direction should probably move to #1964 and/or some other appropriate discussion.

deviantintegral:
I haven't done a line-by-line review, but at a high level:

1. It would be great to get the mqtt library changes in a separate PR. I think that could be quickly reviewed and merged.
2. In the current HA addon, we assume most users are running an MQTT broker inside of Home Assistant. If you're not running MQTT as an HA addon, then you're probably running with a manual docker-based setup in which case you don't need the HA addons at all - just use docker containers directly. Since we can make that assumption, by default we grab the mqtt connection details from Home Assistant's API. So, we wouldn't get much benefit from this feature, but since it's optional it wouldn't hurt us either (beyond any long-term code maintenance concerns).

I'll also say that I've been thinking a bit about what it would look like to connect to rtl_433 over a websocket instead of mqtt. zwavejs does that, and it's pretty great especially for new users. I'm nowhere near close enough to actually doing anything about it, but it's a thought!

rct:
@deviantintegral - Thanks for the info.  I'd like to see this PR merged at least for the fixes and I don't see it forcing some implicit architectural change. 

(I don't use the add-ons, my rtl-sdrs are attached to RPIs that are separate from my Home Assistant installation, and the discovery script only needs to be run on demand for new devices. I'm using the MQTT add on on my Hass instances, but have to configure connection details manually without access to the API.)

The rest of this belongs in a separate discussion:

> I've been thinking a bit about what it would look like to connect to rtl_433 over a websocket instead of mqtt. 

So what are you envisioning here in terms of connection topology for the various components and what would the benefits be?  

Are you thinking web socket just for the auto-discovery script or also the data path from rtl_433 into Home Assistant? 

FWIW, I run the ZwaveJS UI add on (Formerly known as ZwaveJS2MQTT). I'm only using the web socket data path for Home Assistant, but also have ZwaveJS UI publish to MQTT so I get access to some of the bits that the Home Assistant integration doesn't (or didn't) expose well (at the time) like Zwave lock events and parameters (user codes, etc.)

I also run Zigbee2MQTT rather than ZHA. I mention this because I think for rtl_433 there is probably more in common withZigbee2MQTT--It has to deal with a wider range of devices don't adhere to standards well.  It uses a library of "converters" for adding device support and can more easily handle new/ad-hoc devices.  https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_2-2-adding-converter-s-for-your-device 

Also ZwaveJS depends on the ZwaveJS integration running inside of Home Assistant.  Zigbee2MQTT relies completely on MQTT devices. Both are used by HA platforms other than just Home Assistant.

If I were a NodeJS developer, I'd be taking a look at how much of Zigbee2MQTT could be used to solve the gaps/desires around getting rtl_433 data into home automation platforms.

gdt:
I should have pointed out that I don't use the builtin mqtt support.

See also #3013.    Having thought more I'm ok with trying to find config in the rtl_433 config file, as long as there is a configf file possibel for the hass bridge, for things that don't belong in the rtl_433 file, or for people that want them separate.

zuckschwerdt:
It's a neat idea to load rtl_433 config in these helper scripts too. But there are no-config alternatives like https://rtl-433-hass.github.io/ now and we kind of don't actively develop this example script anymore.
There would be follow up questions, e.g. what if the script host and the rtl_433 host are not the same, or what if different credentials are used for the pieces. Let's close as not planned.

gdt:
The addon is not no-config.  Well, it lacks config so in that sense it is, but it behaves badly because you can't configure it -- based on the my 36 hours with it, there's no reasonable way to find new devices you want and reject hundreds of crud devices that don't really exist.    I don't really care if you close this PR or not as we are clearly in a clean-up-hard mood which is 95% good, but please don't point to something that while promising and useful is not fully baked as a reason to kill other things.

deviantintegral:
> you can't configure it

If you look at the addon startup logs, it should print a path where you can add your own configuration extras. I am realizing now that the path needs to be clearer - this is in the addon config directory, but the script inside the addon may not know what that is. If the additional config isn't being pulled in feel free to file an issue with details.

image

You can control discovery on the Home Assistant side in the Hub Settings. There's a screenshot of the toggle here:

https://rtl-433-hass.github.io/rtl_433/latest/configuration/#reconfigure-vs-configure

image

> there's no reasonable way to find new devices you want and reject hundreds

Hundreds? I'm in a suburban area, and I totally get invalid devices, but a handful only. If #3611 doesn't clean this up _and_ dozens or more of unusable devices in a short time are a normal situation, perhaps the integration should do an approval workflow where you discover all devices, you select specific ones to add, and the rest go away on the next HA restart.

gdt:
My setup is showing 413 devices.   I think about 30 of them are mine.  Maybe I'm off and it's 45, but I don't think so.   Assuming 350 extra devices, I think around a dozen or so are legit neighbor devices (temp sensor sort of thing).

I have not tried to write custom json yet, because the devices that are legit largely are ok and the ones that are not should just not show up, vs not showing up right.

I will be filing a concept of operations for approval workflow in the next day or two.   The integration is great in a lot of ways and I think it can be really usable by people who aren't expert with a few more things.

EDIT: After clearing stuff that seems wrong/not-mine after rebuilding rtl_433, I have 38 devices.  4 are dongles, and there are  dups (2 433 dongles), so that's more or less 24 legit transmitters of mine, but showing as 38.  I will let it run overnight and see how it goes.  So far picked up a T/H sensor that isn't mine, but the T/H match outside so I sort that into real.

PR #2849, opened 2024-02-23
Update README.md

HenningS81:
Added type names for devices No 250


zuckschwerdt:
Thanks! This needs to be added to https://github.com/merbanan/rtl_433/blob/master/src/devices/schou_72543_rain.c
(a longer desciption in the list around line 24 and a short name at the end in line 114. Then add your changes in git, compile, then run `maintainer_update.py`, and add those changes too.
Just keep adding to this PR please.

zuckschwerdt:
Manually merged with 66cb6ba

PR #2850, opened 2024-02-25
Update Fedora nightly builds info to docs/BUILDING.md

tvass:
Update Fedora nightly builds doc as RPM packages are back online at [copr](https://copr.fedorainfracloud.org/coprs/tvass/rtl_433/builds/) for 38, 39 and rawhide.
Build process fixed:
- Change path from `./do_build.sh` to `./.ci/scripts/do_build.sh`
- Update libusb-devel to  `libusb1-devel`

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -25,7 +25,7 @@ The usual update mechanism will now keep the rtl_433 version current.
 
 ### Fedora
 
-Fedora users (31, 32 and Rawhide) can add the following copr repository to get nightly builds:
+Fedora users (38, 39 and Rawhide) can add the following copr repository to get nightly builds:
 
     $ sudo dnf copr enable tvass/rtl_433
     $ sudo dnf install rtl_433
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
We have added a note for package maintainers on how to build: https://github.com/merbanan/rtl_433/blob/master/docs/BUILDING.md#package-maintainers

Note that we will remove the `do_build.sh` shortly -- it is not intended to be used.
Please update https://github.com/tvass/rtl_433_rpm

Thanks for providing these builds!

tvass:
Hello @zuckschwerdt 
I have updated the [build steps](https://github.com/tvass/rtl_433_rpm/commit/7cd65287f8389efdb24ad5cba727ef2779eef138), let me know if this is what you expect. 

Builds are [OK](https://copr.fedorainfracloud.org/coprs/tvass/rtl_433/build/7158997/).

zuckschwerdt:
Looks good. There is no requirement to use Ninja (but it's faster and recommended), you could drop `-GNinja` if that makes things easier -- but I guess Ninja is pretty standard to use these days.

I'm not sure if you would view the build output with color escape codes. It should be no harm though. Otherwise you can drop `-DFORCE_COLORED_BUILD:BOOL=ON` to remove unwanted escape codes and get plain text build output.

zuckschwerdt:
One more thing: You are building without SoapySDR support, which should be explicit (default is "auto"): 
`-DENABLE_SOAPYSDR=OFF`

And more importantly you are using OpenSSL by chance only, I guess it would be installed on default runners.
Choose to include or exclude it (default is "auto"): `-DENABLE_OPENSSL=ON`
(you may need `libssl-dev`)

Now your builds will reliably demand those features and break otherwise.

PR #2854, opened 2024-02-28
Fix Badger-ORION return on decode success

thetoivonen:
Changed the return code issued on decode success from a 0 to a 1.  When this is set to a 0, the "-E quit" flag fails and the system gets stuck forever decoding badger_water.  When this is set to a 1, the "-E quit" flag performs as expected.


zuckschwerdt:
Thanks!

Issue #2823, opened 2024-02-07
`rtl_443 --help` output barely readable on bright colour scheme terminals.


dreirund:
When I run `rtl_443 --help`, I get the following output (screenshot):

---

![rtl_433--help_colouroutput](https://github.com/merbanan/rtl_433/assets/1590519/115dba9a-1536-426d-86f9-ce2f9b5b69a0)

---

The bright colour on the white background is barely readable.

Please use either monochrome by default, or some colouring that works both on dark and bright terminals.

Regards!

zuckschwerdt:
This is a know shortcoming. We don't know how to determine if you use a bright or dark theme.

A workaround for now is e.g. `rtl_433 -h | cat`

Tested ideas would be appreciated, but for now this is not a todo for us.

dreirund:
On Wed, 07 Feb 2024 07:55:40 -0800, "Christian W. Zuckschwerdt"
***@***.***> wrote:

> A workaround for now is e.g. `rtl_433 -h | cat`

Interestingly, it works, but only for recognised argument (`-h`).

For output to standard error (e.g. `rtl_433 --help | cat` -- `--help`
is not recognised and so nothing is piped through `cat`) it does not
work.


> Tested ideas would be appreciated,

* Use no colouring by default (use a command line switch for colouring),
* use dark green instead of bright yellow.

Regards!

zuckschwerdt:
We could maybe use an env var like `RTL433_COLOR=[auto|never|always]`

zuckschwerdt:
I'm looking into this, and I remember explicitly using the non-bright colors with bold to suit light themes.
Does your terminal perhaps not support bold and converts dull bold to bright colors? I get e.g. these colors:
![rtl433_colors](https://github.com/merbanan/rtl_433/assets/12833/9db51924-af9a-4f74-91bd-6bc85abb0b68)

For comparison this is without using bold:
![rtl433_colors-nobold](https://github.com/merbanan/rtl_433/assets/12833/7d2b2c68-33c2-4f86-b23a-5ccb60a5acae)

You can test this by removing `;1` here https://github.com/merbanan/rtl_433/blob/2d63f30681dd527955542edea4b7a570bacf8a6f/src/term_ctl.c#L291

zuckschwerdt:
Thanks for the feedback. Lets merge #2864 -- your suggested change then is automatic for light backgrounds.

dreirund:
[...]

PR #2864, opened 2024-03-04
Add darker colors for light terminal backgrounds

zuckschwerdt:
This addresses hard to read colors on light terminal backgrounds.

We use the `COLORFGBG` environment variable to detect a light background and then remove the bright/bold attribute from the foreground color.

If your terminal emulates bright colors with bold characters the light "theme" can be used to get non-bold (lighter) output.

Can be selected/tested by
- for dark backgrounds `COLORFGBG="15;0" rtl_433 …`
- for light backgrounds `COLORFGBG="0;15" rtl_433 …`

Modern terminals will already set this appropriately, check with `echo $COLORFGBG`

### DIFF
--- a/src/term_ctl.c
+++ b/src/term_ctl.c
@@ -223,6 +223,38 @@ int term_get_columns(void *ctx)
 #endif
 }
 
+/**
+Returns wether the environment suggests a dark or light
+terminal background.
+
+Our default color theme is for dark backgrounds.
+
+Check the COLORFGBG environment variable, which should
+be 15;0 for a dark theme and 0;15 for a light theme.
+
+If the last value is 7 or 9 to 15 then assume a light theme.
+
+@return 1 if a light background was detected, 0 for a dark background otherwise
+*/
+static int term_get_bg(void)
+{
+    char const *colorfgbg = getenv("COLORFGBG");
+    if (!colorfgbg) {
+        return 0; // default dark theme
+    }
+    char *p = strrchr(colorfgbg, ';');
+    if (!p) {
+        return 0; // default dark theme
+    }
+
+    // Check if the last value is 7 or 9 to 15.
+    if (p[1] == '7' || p[1] == '9'
+            || (p[1] == '1' && p[2] != '\0')) {
+        return 1; // light theme
+    }
+    return 0; // dark theme
+}
+
 int term_has_color(void *ctx)
 {
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
This came up in #2823

@dreirund please review.

gvanem:
Can you also please make a case for this unreadable `time` field.
RTL_433-time

I had to patch it with:
```diff
--- a/src/output_file.c 2024-03-08 14:25:55
+++ b/src/output_file.c 2024-03-08 15:47:10
@@ -158,7 +158,7 @@
     if (!key || !*key)
         return TERM_COLOR_RESET;
     if (!strcmp(key, "tag") || !strcmp(key, "time"))
-        return TERM_COLOR_BLUE;
+        return TERM_COLOR_BRIGHT_WHITE;
     if (!strcmp(key, "model") || !strcmp(key, "type") || !strcmp(key, "id"))
         return TERM_COLOR_RED;
     if (!strcmp(key, "mic"))
```
*BTW*, the *TFA-TwinPlus* is totally out-of-wack. Another issue maybe later.

zuckschwerdt:
Do you have something like `COLORFGBG="15;4"` in that terminal? Is the theme a Windows default or a selection or user setting?

gvanem:
> Do you have something like COLORFGBG="15;4" in that terminal?

I do not understand. I'm on Windows and `COLORFGBG` is a Gnome thing AFAICS.

>  Is the theme a Windows default or a selection or user setting?

A user setting from the WinCon settings dialogue:
image

zuckschwerdt:
That `COLORFGBG` is just a plain env var that some terminals set and support for that mechanism is easy. (I don't dare look into the ANSI-escape magic to make terminals report color configuration.)

You can only set a background color in WinCon (sorry, is that the name? Is it Windows Console?)  and not the full color scheme, can you? Otherwise I'd recommend mapping the blue color to something like cyan maybe.

The idea is that you could set `COLORFGBG` fittingly and then rtl_433 could detect the bg color and remove that one color from the palette, e.g. replace it with white /fg.

gvanem:
> .. and not the full color scheme, can you?

I guess not. Have not tried [ColorTool](https://github.com/microsoft/terminal/tree/main/src/tools/ColorTool) and rtl_433.
But I have yet to try this patch here to see if it works on WinCon.

dreirund:
> *@dreirund please review.*

it seems to ignore `COLORFGBG` variable.

dreirund:
> > @dreirund please review.
> 
> *it seems to ignore `COLORFGBG` variable.*

Now it works, we had a race condition here, seemed to just have added it to the code a few minutes ago when I was compiling and testing.

Issue #2479, opened 2023-04-11
OOK_MC_ZEROBIT decoder behaves unexpectedly with a non-Manchester prefix, possibly leading to incorrect decodes


klohner:
In trying to work on improvements to protocol 70, the Honeywell 345 MHz decoder, I'm running into issues with how rtl_433 applies the OOK_MC_ZEROBIT decoder returns data.

For example, I'm looking at 
[rtl_433_tests/tree/master/tests/honeywell_5816)/g002_344.975M_250k.cu8](https://github.com/merbanan/rtl_433_tests/blob/75c5a508a942aa8c3a09dc57f7b335e03f3ea654/tests/honeywell_5816/g002_344.975M_250k.cu8).  

Granted, this is not a particularly clean signal with high bitrate, but it's a good example to illustrate this issue.  This signal is OOK, even though it appears ASK, (see [issue #2015](https://github.com/merbanan/rtl_433/issues/2015) for background).

Here's a URH representation of of one message in the demodulated signal
![honeywell_5816_g002_344 975M_250k](https://user-images.githubusercontent.com/8398098/231011760-9756e9a5-b1a1-4cce-b8e7-c537971d31ff.png)

I can ask rtl_433 to get that PCM data with
`rtl_433 -r tests/honeywell_5816/g002_344.975M_250k.cu8 -X "n=Honeywell,m=OOK_PCM,s=136,l=136,r=408"
`
and it'll tell me this is:
`{162}ffffffffeaaaaaacd54ad52ad52acd54cd2ad2d20`
or:
`0y111111111111111111111111111111111110101010101010101010101010110011010101010010101101010100101010110101010010101011001101010101001100110100101010110100101101001000`

This is fine and I wouldn't expect better when looking at this signal.  The part just before the Manchester data is being seen as pulses, but once the real pulses start, they all get decoded properly.

I can feed that data to bitbench and have it decode the useful Manchester data.  I just have it look for a preamble of `{2}8` and do a "Manchester (G.E. Thomas)" decode and invert. to get 
`y11111111111110100000111000011110000111101000001010011110011001` which is `{62}fffa0e1e1e829e64` (I have to be careful in bitbench with how it represents a final partial nibble in hex!).  This is data I'm trying to get rtl_433 to provide.

[...]

Issue #2015, opened 2022-03-19
Honeywell 5816 security contact decoding issues


rct:
I have a bunch of Honeywell 5800 series security contacts that decode reliably, but I have two 5816 door sensors that don't decode reliably.  If the sensor within 15 ft of the rtl-sdr it doesn't decode at all. If the distance is larger than that I'll get some decodes, but it's not as reliable as the others, only some of the repeats will decode. 

The analyzer (-A) isn't able to detect the pulses. Maybe it's saturating the receiver. Using `-Y autolevel` doesn't seem to help.

Frequency of both of the 5816 seems to be around 345.06 Mhz.  I've got rtl_433 set to 345.0M. So they seem to be within range and not falling on the DC spike.

I'll attach samples. Thanks for taking a look.

rct:
[honeywell-5816-1.zip](https://github.com/merbanan/rtl_433/files/8310340/honeywell-5816-1.zip)

```
rtl_433 version 21.05-111-gf30ee220 branch master at 202110220900 inputs file rtl_tcp RTL-SDR with
 TLS
Use -h for usage help and see https://triq.org/ for documentation.
Trying conf file at "rtl_433.conf"...
Trying conf file at "/home/rct/.config/rtl_433/rtl_433.conf"...
Trying conf file at "/usr/local/etc/rtl_433/rtl_433.conf"...
Trying conf file at "/etc/rtl_433/rtl_433.conf"...
Registered 168 out of 198 device decoding protocols [ 1-4 8 11-12 15-17 19-23 25-26 29-36 38-60 63
 67-71 73-100 102-105 108-116 119 121 124-128 130-149 151-161 163-168 170-175 177-197 ]
Test mode active. Reading samples from file: g070_345M_250k.cu8
baseband_demod_FM: low pass filter for 250000 Hz at cutoff 25000 Hz, 40.0 us
Detected OOK package    @0.076016s
Analyzing pulses...
Total count:    4,  width: 368.91 ms            (92227 S)
Pulse width distribution:
 [ 0] count:    4,  width: 21732 us [21728;21736]       (5433 S)
Gap width distribution:
 [ 0] count:    3,  width: 93992 us [93992;93996]       (23498 S)
Pulse period distribution:
 [ 0] count:    3,  width: 115724 us [115724;115728]    (28931 S)
Pulse timing distribution:
 [ 0] count:    4,  width: 21732 us [21728;21736]       (5433 S)
[...]

PR #2881, opened 2024-03-20
Added mappings for event, state, and contact_open for Honeywell 58XX devices in rtl_433_mqtt_hass.py

dmz006:
Author:David M. Zendzian (dmz@dmzs.com)
Date: Wed Mar 20 00:34:00 2024 -0500

Added mappings for event, state, and contact_open for Honeywell 58XX devices.

- Event is an event ID. I've not mapped them yet but I've seen 0, 4, 60, 57, 120, 128, 248 and others as the devices change state
- State is either "open" | "closed" and corresponds to the state of the door sensor
- contact_open appears to be tied to state

On the 5800 devices I tested alarm state did not correspond with device state or contact_open and doesn't appear to change with opening or closing door/window

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -452,25 +452,14 @@ mappings = {
         }
     },
 
-    "event": {
-        "device_type": "sensor",
-        "object_suffix": "event",
-        "config": {
-            "name": "Event",
-            "value_template": "{{ value|int }}",
-            "state_class": "measurement",
-            "entity_category": "diagnostic"
-        }
-    },
-
-    "state": {
+    "reed_open": {
         "device_type": "binary_sensor",
-        "object_suffix": "state",
+        "object_suffix": "reed_open",
         "config": {
             "device_class": "safety",
             "force_update": "true",
-            "payload_on": "open",
-            "payload_off": "closed",
+            "payload_on": "1",
+            "payload_off": "0",
             "entity_category": "diagnostic"
         }
     },
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -452,6 +452,41 @@ mappings = {
         }
     },
 
+    "event": {
+        "device_type": "sensor",
+        "object_suffix": "event",
+        "config": {
+            "name": "Event",
+            "value_template": "{{ value|int }}",
+            "state_class": "measurement",
+            "entity_category": "diagnostic"
+        }
+    },
+
+    "state": {
+        "device_type": "binary_sensor",
+        "object_suffix": "state",
+        "config": {
+            "device_class": "safety",
+            "force_update": "true",
+            "payload_on": "open",
+            "payload_off": "closed",
+            "entity_category": "diagnostic"
+        }
+    },
+
+    "contact_open": {
+        "device_type": "binary_sensor",
+        "object_suffix": "contact_open",
+        "config": {
+            "device_class": "safety",
+            "force_update": "true",
+            "payload_on": "1",
+            "payload_off": "0",
+            "entity_category": "diagnostic"
+        }
+    },
+
     "tamper": {
         "device_type": "binary_sensor",
         "object_suffix": "tamper",


zuckschwerdt:
"state" is legacy and should not be used. "event" is a code that is then broken down:
https://github.com/merbanan/rtl_433/blob/afb5b34709b699bf88aa95b6c577591235a5d591/src/devices/honeywell.c#L96-L103

dmz006:
Thank you for the info on the event.  Not sure if that is helpful or not; but it's not captured in mqtt_hass so not sure  

State may be legacy; but what about older devices?  The devices in my house do not use the alarm state (they never change);  only state or contact_open changes when magnet moves from sensor (or when motion sensor sees movement).

What do you suggest I do? I can't change out the devices and get non-legacy devices.  And the one new device I have appears to function the way the older devices do.

Is there any history on the state being legacy or not? I searched past PR and didn't find anything.

`{"time" : "2024-03-20T05:56:31.255725-0400", "model" : "Honeywell-Security", "id" : 390678, "channel" : 8, "event" : 124, "state" : "closed", "contact_open" : 0, "reed_open" : 1, "alarm" : 1, "tamper" : 1, "battery_ok" : 0, "heartbeat" : 1, "mic" : "CRC"}
`

In watching the data from RTL I never see alarm change value; in fact I have devices where alarm is 1 and some where alarm is 0; but the value for alarm never changes. However the value for state and contact_open changes.

What do you suggest I monitor? And is this functionality what is expected??

dmz006:
In looking at the event code you shared and the events from the last post:

Using this exampe: {"time" : "2024-03-20T05:52:15.564504-0400", "model" : "Honeywell-Security", "id" : 945266, "channel" : 8, "event" : 128, "state" : "open", "contact_open" : 1, "reed_open" : 0, "alarm" : 0, "tamper" : 0, "battery_ok" : 1, "heartbeat" : 0, "mic" : "CRC"}

Event code of 128 or in binary 10000000

to simplify the code shared above for event:

- 0x80 == 10000000
- 0x40 == 01000000
- 0x20 == 00100000
- 0x10 == 00010000
- 0x08 == 00001000
- 0x04 == 00000100

leaving 2 bits for future code??

So the event above 128 in hex is 10000000 which makes sense that the contact would be open as the state and contact show open.

However how come the alarm is 0? Shouldn't the alarm be 1 if the contact and state is open?

Other codes I've seen include:
- 0     == 00000000
- 4     == 00000100
- 28   == 00011100
- 36   == 00100100
- 52   == 00110100
- 124 == 01111100
- 128 == 10000000
- 188 == 10111100
- 132 == 10000100

{"time" : "2024-03-20T05:52:16.958049-0400", "model" : "Honeywell-Security", "id" : 945266, "channel" : 8, "event" : 0, "state" : "closed", "contact_open" : 0, "reed_open" : 0, "alarm" : 0, "tamper" : 0, "battery_ok" : 1, "heartbeat" : 0, "mic" : "CRC"}

Here the event is 0; but battery_ok is still 1; should't it be 0 with all bits to 0?

dmz006:
If a device is using the internal reed and magnet I am thinking maybe that is why we see the reed ones; and if it's using an external magnet connected to junction box then would that be the contact?

Just trying to grok this.

Also; if contact is part of the coding in rtl; shouldn't it be exposed in mqtt_hass? I would be OK with state being legacy if I can get the contact info as that appears to be the same for what I need.

Any idea what causes alarm to be tagged? It appears to me by some quick searches that alarm is tied to reed_open like state is tied to contact_open??

dmz006:
One last question; since I'm not using the reed switch I didn't notice that mqtt_hass wasn't publishing reed_open state - shouldn't it publish all active details, like contact and reed?

Should I update the PR to remove event (which would be good to capture but not necessary), remove state but keep contact_open and add reed_open?

zuckschwerdt:
Sorry, no idea. We see that "event" byte in the data transmission and we guess it decodes that way. I'm just saying you shouldn't use the raw "event" output. We have since added new information about decoding: the breakdown of event seen above.
"event" is undecoded, "state" is the old name for "contact_open", "battery_ok" is inverted. Click the source code I posted for more info.

Yes, remove "event" and "state" please.

zuckschwerdt:
Looks good to me.

dmz006:
Just curious; what is the normal process/workflow/timeline for PR?

zuckschwerdt:
We usually allow a week or two for people to chime in.

rct:
Sorry didn't see this earlier.  But, FYI, users of a number of the 5800 series sensors will need to do some hand configuring for which bit(s) they care about for each sensor. 

I thought more detail made it into the 5800 source but I guess not.

Background: Apparently when these devices get added to Honeywell panels, the installer selects which "loop" that zone is monitoring.  Many sensors use Loop 1 which is the 0x80 "contact" bit.  But some sensors use Loop 2 (0x20 reed).

It is possible on some of the 5800 series sensors to get 1-3 zones out of a single sensor.  For some since the contact terminals are separate from the reed switch, that sensor can cover 2 zones. 

Tamper 0x40 is considered Loop 4.

Some sensors send additional information on Loop 3 (0x10), like tilt or low temperature. 

So from my notes:

```
- 0x80 - Contact Open  (loop 1)
- 0x40 - Tamper (loop 4)
- 0x20 - Reed Open  (loop 2)
- 0x10 - Alarm (Loop 3 - Tilt, low temperature, etc.)
- 0x08 - Battery Low  (0 OK, 1 low)
- 0x04 - Heartbeat  (0 Event, 1 heartbeat)
- 0x02 - ??  (not used or not seen)
- 0x01 - ??  (not used or not seen)
	- From @klohner's decoder (might be related to RE220 repeater?)
		- 0x02 - Power (0 OK, 1 initialized?)  (maybe power up/reboot?)
		- 0x01  - Source (0 local, 1 Repeater) 
````

zuckschwerdt:
Great information! I always wondered why there are different contact/reed bits, as those contact would likely be "reed switch" type hardware too?

Do you want to PR this to the decoder comment text for posteriority?

rct:
Yeah, I'll try to clean that up at some point. Was waiting/hoping for some progress on the problematic 5816's #2015 (and the related #2479)

Issue #2861, opened 2024-03-03
No bits decoded from wmbus T1 Message


kralo:
The following capture is decoded quite nicely with 
`cat <file> | rtl_wmbus -d 3 -s` and yields one T1 Message starting with `0x294468506...`. However that message looks strange to me (some repeating fffffff inside...) so I wanted to decode it with rtl_433.

I tried `rtl_433 -r <file> -R 105 -vvv` or `-R 0 -A` but got no demodulation (nor with R 104).

I know this packet is potentially encrypted.  But I had no luck specifying the decoder with -X "n=a,m=FSK_PCM,short=10,long=10,reset=500"

[rtl_sdr_f868.625M_s2400000_onesampleMbusT1.cu8.zip](https://github.com/merbanan/rtl_433/files/14474650/rtl_sdr_f868.625M_s2400000_onesampleMbusT1.cu8.zip)

kralo:
The triq.org viewer seems to be sensible to the filename and gets slow, so this is another try `capt_rtlsdr_f868.625M_s2400000_onesample_larger_wMbusT1.cu8` 
[2861_sample2.cu8.zip](https://github.com/merbanan/rtl_433/files/14474688/2861_sample2.cu8.zip)

merbanan:
You need to set the correct frequency and sample rate also. (-f and -s)

klohner:
I took a look at this but I don't understand why -R 104 isn't picking this up either.  Am I doing something wrong?

Renaming either sample to something more normalized, I can do:

`rtl_433 -r MbusT1_g001_868.695M_2400k.cu8 -X n=MBUS,m=FSK_PCM,s=10,l=10,r=500,preamble={16}543d`

and get data like:

`{588}3a571c6ac6566a5b1935a6964daa56c5ca539965969693a968d2d659636c68d2d6596b1668d59658d5a5a4b5a65966a398d5964f48ce596b31d0b596972c5c5965a9c9959695a8e6800`

But this doesn't seem to return anything:

`rtl_433 -R 0 -R 104 -r MbusT1_g001_868.695M_2400k.cu8`

Even though, -R 104 is:

`    [104]  Wireless M-Bus, Mode C&T, 100kbps (-f 868.95M -s 1200k)`

and its code seems to look for that same modulation and symbol width:

https://github.com/merbanan/rtl_433/blob/afb5b34709b699bf88aa95b6c577591235a5d591/src/devices/m_bus.c#L1201-L1209

[...]

PR #2883, opened 2024-03-22
ignore 3of6 decoding errors, let CRC fail decoding if necessary, 

klohner:
Addresses [No bits decoded from wmbus T1 Message #2861](https://github.com/merbanan/rtl_433/issues/2861#issuecomment-2013663130)

### DIFF
--- a/src/devices/m_bus.c
+++ b/src/devices/m_bus.c
@@ -59,16 +59,19 @@ static uint8_t m_bus_decode_3of6(uint8_t byte)
 // Bad data must be handled with second layer CRC
 static int m_bus_decode_3of6_buffer(uint8_t const *bits, unsigned bit_offset, uint8_t* output, unsigned num_bytes)
 {
+    int successful_contiguous_bytes = -1;
     for (unsigned n=0; n<num_bytes; ++n) {
         uint8_t nibble_h = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset) >> 2);
         uint8_t nibble_l = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset+6) >> 2);
         if (nibble_h > 0xf || nibble_l > 0xf) {
             // return -1;  // fail at first 3of6 decoding error
             nibble_l &= 0x0F;  // assume logical 0 nibble if 3of6 decoding error, let CRC fail decoding if necessary
+            successful_contiguous_bytes > -1 || successful_contiguous_bytes = n;  // return count found until the first error
         }
         output[n] = (nibble_h << 4) | nibble_l;
     }
-    return 0;
+    successful_contiguous_bytes > -1 || successful_contiguous_bytes = num_bytes;  // if all data decoded successfully
+    return successful_contiguous_bytes;
 }
 
 // Validate CRC
--- a/src/devices/m_bus.c
+++ b/src/devices/m_bus.c
@@ -31,25 +31,25 @@ static unsigned bcd2int(uint8_t bcd)
 // Mapping from 6 bits to 4 bits. "3of6" coding used for Mode T
 static uint8_t m_bus_decode_3of6(uint8_t byte)
 {
-    uint8_t out = 0xFF; // Error
+    uint8_t out = 0xF0; // Error
     //fprintf(stderr,"Decode %0d\n", byte);
     switch(byte) {
-        case 22:    out = 0x0;  break;  // 0x16
-        case 13:    out = 0x1;  break;  // 0x0D
-        case 14:    out = 0x2;  break;  // 0x0E
-        case 11:    out = 0x3;  break;  // 0x0B
-        case 28:    out = 0x4;  break;  // 0x17
-        case 25:    out = 0x5;  break;  // 0x19
-        case 26:    out = 0x6;  break;  // 0x1A
-        case 19:    out = 0x7;  break;  // 0x13
-        case 44:    out = 0x8;  break;  // 0x2C
-        case 37:    out = 0x9;  break;  // 0x25
-        case 38:    out = 0xA;  break;  // 0x26
-        case 35:    out = 0xB;  break;  // 0x23
-        case 52:    out = 0xC;  break;  // 0x34
-        case 49:    out = 0xD;  break;  // 0x31
-        case 50:    out = 0xE;  break;  // 0x32
-        case 41:    out = 0xF;  break;  // 0x29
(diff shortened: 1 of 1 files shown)


src/devices/m_bus.c
```diff
@@ -63,7 +63,8 @@ static int m_bus_decode_3of6_buffer(uint8_t const *bits, unsigned bit_offset, ui
         uint8_t nibble_h = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset) >> 2);
         uint8_t nibble_l = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset+6) >> 2);
         if (nibble_h > 0xf || nibble_l > 0xf) {
-            return -1;
+            // return -1;  // fail at first 3of6 decoding error
+            nibble_l &= 0x0F;  // assume logical 0 nibble if 3of6 decoding error, let CRC fail decoding if necessary
```
zuckschwerdt, line 67: does this handle `nibble_h > 0xf`?
Shouldn't the goal be to return the number of successful decoded bytes instead of just an error (or hiding the error)?

zuckschwerdt: Ok, I see it does. We compose a partial byte filling the broken nibble with zero.
Should we end the decoding on first error or keep going?

klohner: I thought it redundant to do `nibble_h &= 0x0F;` since the high part of that byte is getting immediately shifted away in `output[n] = (nibble_h << 4) | nibble_l;` (thus hiding any error in nibble_h as a side effect)

I agree it'd be more useful to have the return code be the number of decoded bytes found before the end or first error detected (instead of just `0` on success or `-1` on failure), though that change would require further changes.

klohner: Are you thinking something like:

```
static int m_bus_decode_3of6_buffer(uint8_t const *bits, unsigned bit_offset, uint8_t* output, unsigned num_bytes)
{
    for (unsigned n=0; n<num_bytes; ++n) {
        uint8_t nibble_h = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset) >> 2);
        uint8_t nibble_l = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset+6) >> 2);
        if (nibble_h > 0xf || nibble_l > 0xf) {
            // return -1;  // fail at first 3of6 decoding error
            // nibble_l &= 0x0F;  // assume logical 0 nibble if 3of6 decoding error, let CRC fail decoding if necessary
            return n;  // return all successfully decoded bytes found until the first error
        }
        output[n] = (nibble_h << 4) | nibble_l;
    }
    return num_bytes;
}

```

It seems useful to me, but it looks like any positive result is then just categoriezed as success and then unused.  Should we change that behavior as well?

klohner: I think a case could also be made to return the number of successful bytes until the first decode error (or even all bytes, regardless), still decoding all available bytes, zeroing any decoding errors, and let the caller decide if it's worth looking beyond the first decode error for other packets of data or try some kind of error correction.  Though perhaps the returns on this effort would be too diminished to be useful.

zuckschwerdt: Since there are no dynamic lengths involved I'd just decode everything (while zeroing errors) and return the number of good bytes. (`int bytes_ok = 0; … ++ ; return bytes_ok`)
Then let the caller decide, yes. As far as I understood these broken bytes are past the payload and CRC anyway, right? That's where my thinking came from, the "abort" is too harsh right now as we might not need all the bytes, so let the caller know how many good bytes we got and maybe that is enough.

merbanan: In theory we can use the crc to error correct. This 6 to 4 mapping could hint what nibble to error correct.

One thing that is in the specs but not in the code is that the number of bit transitions in the 6 bit code word is always greater then 1.

Im not saying we should add error correcting code now but maybe we could add some logic towards that.

In transmission theory the symbols can have an euclidean distance compared to each other. These 6 bit codes should have this property that if the code does not match any of the table codes if should be easy to map it to one code that is close in a euclidean way instead.

So the 6 to 4 decoder should output what it thinks the value should be and still somehow keep track of the indexes that might be incorrectly decoded and in need of error correction.

klohner: Perhaps that kind of EC sounds like a job best left for rtl_wmbus?

What do you guys think of something like this:
```
static int m_bus_decode_3of6_buffer(uint8_t const *bits, unsigned bit_offset, uint8_t* output, unsigned num_bytes)
{
    int successful_contiguous_bytes = -1;
    for (unsigned n=0; n<num_bytes; ++n) {
        uint8_t nibble_h = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset) >> 2);
        uint8_t nibble_l = m_bus_decode_3of6(bitrow_get_byte(bits, n*12+bit_offset+6) >> 2);
        if (nibble_h > 0xf || nibble_l > 0xf) {
            // return -1;  // fail at first 3of6 decoding error
            nibble_l &= 0x0F;  // assume logical 0 nibble if 3of6 decoding error, let CRC fail decoding if necessary
            successful_contiguous_bytes > -1  || successful_contiguous_bytes = n;  // return count found until the first error
        }
        output[n] = (nibble_h << 4) | nibble_l;
    }
    successful_contiguous_bytes > -1  || successful_contiguous_bytes = num_bytes;  // if all data decoded successfully
    return successful_contiguous_bytes;
}

```

merbanan: Fine by me.

zuckschwerdt: I don't think 3of6 lends itself well to EC, e.g. assume a single bit error in `001111`, would this be `001101`, `001110`, or `001011`?

merbanan: Well the CRC will probably only pass with one of these values. But the complexity of doing EC 2 steps back will get quite complicated quickly. In practice all 6-bit codes that are invalid has a valid value that is most likely to be the correct one. So if one had lots of time to invest into this all kinds of clever heuristics could be developed.

merbanan: I do think keeping track of nibble indices that 6to4 decodes improperly and then brute force trying all 4-bit combinations and calculating the CRC over that would be just as effective at doing EC compared to some clever heuristics in the 6to4 decoder.

klohner: Do we need to implement EC as part of this change?

PR #2906, opened 2024-04-18
Update tpms_nissan.c

CppBaddy:
Fixed reported pressure_psi value.

### DIFF
--- a/src/devices/tpms_nissan.c
+++ b/src/devices/tpms_nissan.c
@@ -48,7 +48,7 @@ static int tpms_nissan_decode(r_device *decoder, bitbuffer_t *bitbuffer, unsigne
 
     // (PSI+THREE)*FOUR=8d
     int pressure_raw   = ((b[3] & 0x1F) << 3) | (b[4] >> 5);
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks!
The inversion of the documented `(PSI+THREE)*FOUR` would be `float pressure_psi = (float)pressure_raw / 4.0f - 3.0f;`, won't it?

CppBaddy:
I found by accident that the pressure value tracks my tire gauge better. Original formula gives much lower values. To verify, I am going to capture some data from my winter wheels tonight.

zuckschwerdt:
The decoder was merged from #2536 but I messed up the pressure by applying the factor twice.

I think only that error should be corrected. The difference in measurement you see is more likely due to operating temperature?

CppBaddy:
Ok, this makes sense, I'll make changes then.

zuckschwerdt:
Thanks! Sorry to be picky, could you use `4.0f` to be consistent with other float literals?

CppBaddy:
No worries, fixed.
Cheers!

PR #2915, opened 2024-05-02
The pressure unit should be pressure_PSI, not pressure_psi.

gson1703:
With this change, the pressure reported by a Nissan TPMS will be correctly converted to kPa when using the "-C si" option.


zuckschwerdt:
Thanks! Please also change line 101, just add a commit to this PR.

PR #2938, opened 2024-05-26
cache secplus_v2 message halves for when we dont get both in a single transmission

dgwynne:
This is modelled on the secplus_v1 handling, where messages are kept and timestamped so they can be assembled over a longer time period.

This makes the button presses from a Merlin M970E appear as fully decoded messages. The M970E operates around 433MHz, unlike other devices that this driver supports which look like they operate at 310MHz. Not sure if that matters.

I did this quickly, there is probably more elegant tweaks to the code that could have been made.

### DIFF
--- a/src/devices/secplus_v2.c
+++ b/src/devices/secplus_v2.c
@@ -19,6 +19,7 @@ Security+ 2.0  is described in [US patent application US20110317835A1](https://p
 */
 
 #include "decoder.h"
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
IIRC there is another PR with these changes.

merbanan:
Duplicate of #2314, closing. It is still unclear how we should handle this.

Issue #2943, opened 2024-05-27
New device Ecowitt water sensor, PN WN34D


Mart124:
Hello,

Ecowitt does a water temperature sensor, PN [WN34D](https://www.ecowitt.com/shop/goodsDetail/294), [user manual](https://osswww.ecowitt.net/uploads/20230919/WN34D%20Manual.pdf).

Here are then some samples from it : [WN34D.zip](https://github.com/merbanan/rtl_433/files/15458982/WN34D.zip)

g001_868.34M_250k.cu8 : 20.3°C
g001_868.34M_1000k.cu8 : 25.7°C
g002_868.34M_250k.cu8 : -4.9°C
g002_868.34M_1000k.cu8 : -9.0°C

It is decoded by protocol 221, but with a 40°C offset for positive values, and another one for negative values...

Could we have a look to properly support it please ?

Thank you very much 👍

ProfBoc75:
Hi @Mart124:

Yes, there is a gap at the data layout for the Temp.

The current decoder could be updated accordingly.

Current data layout:

    TYPE 8h ID 24h NOT_USED 5h TEMP_RAW 11s  NOT_USED 1b BAT 7d CRC 8h CHECKSUM 8h TRAIL_BYTE 12h 2x

New one:

    TYPE 8h ID 24h SUB_TYPE 4h TEMP_RAW 12s  NOT_USED 1b BAT 7d CRC 8h CHECKSUM 8h TRAIL_BYTE 12h 2x
 
    If SUB_TYPE = 0   = WN34L, Current decoder, temp = (Temp_raw - 400) * 0.1f
    If SUB_TYPE = 0x4 = WN34D, No Temp offset , temp = Temp_raw * 0.1f

[...]

PR #2944, opened 2024-05-28
Add Support for Ecowitt WN34D and improve FineOffset WN34

ProfBoc75:
Related to issue #2943 

Update the existing decoder to take into account this model WN34D, temperature without offset.
Few document updates and improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [218]  Microchip HCS200/HCS300 KeeLoq Hopping Encoder based remotes (FSK)
     [219]  Fine Offset Electronics WH45 air quality sensor
     [220]  Maverick XR-30 BBQ Sensor
-    [221]  Fine Offset Electronics WN34 temperature sensor
+    [221]  Fine Offset Electronics WN34S/L/D and Froggit DP150/D35 temperature sensor
     [222]  Rubicson Pool Thermometer 48942
     [223]  Badger ORION water meter, 100kbps (-f 916.45M -s 1200k)
     [224]  GEO minim+ energy monitor
--- a/src/devices/fineoffset_wn34.c
+++ b/src/devices/fineoffset_wn34.c
@@ -14,23 +14,26 @@
 /**
 Fine Offset Electronics WN34 Temperature Sensor.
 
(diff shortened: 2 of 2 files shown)


zuckschwerdt:
I think `subtype` might be a good choice here as it is a different "function" (coding of the value) in the same protocol.
But this will be a breaking change as the subtype is part of the MQTT key.

As a rare exception I would rather change the "model" key here to not break existing usage, i.e.
```
"model", "", DATA_COND, sub_type != 4, DATA_STRING, "Fineoffset-WN34",
"model", "", DATA_COND, sub_type == 4, DATA_STRING, "Fineoffset-WN34D",
```
and don't report `subtype`.

What do you think?

ProfBoc75:
> I think `subtype` might be a good choice here as it is a different "function" (coding of the value) in the same protocol. But this will be a breaking change as the subtype is part of the MQTT key.
> 
> As a rare exception I would rather change the "model" key here to not break existing usage, i.e.
> 
> ```
> "model", "", DATA_COND, sub_type != 4, DATA_STRING, "Fineoffset-WN34",
> "model", "", DATA_COND, sub_type == 4, DATA_STRING, "Fineoffset-WN34D",
> ```
> 
> and don't report `subtype`.
> 
> What do you think?

Yes, good approach, I'm updating the decoder. Instead, I will add the subtype as a verbose message.
Because we know the sub-type is 0x4 for WN34D, and it's 0x0 for WN34L ( cu8 files from this [PR at rtl_433_tests](https://github.com/merbanan/rtl_433_tests/pull/426)), but we don't know for other models, may be other values than 0 & 4, I would like to catch, the reason why I'm asking for some feedback in the issue.

In the same time, I was also wondering how to show this model, because the decoder is always reporting WN34, and yet it's able to decode other models WN34L/S and D, your approach answers to this concern, and I will enrich also the name accordingly.

Good point, thx.

Mart124:
Many thanks @ProfBoc75 for your work and @zuckschwerdt for your review 👍
Could we think about merging this patch if everything is OK with it ? So that we can begin using it :)
Thank you again !

zuckschwerdt:
Guess it doesn't matter too much, we can further enlarge the allowed length to 500. Do you want to PR the change?

PR #2946, opened 2024-05-29
"state" binary "ON"/"OFF" casing consistency for Waveman-Switch decoder

DigiH:
There are several decoders with the "state" key being a binary "ON"/"OFF" option.

To be able to create a single Home Assistant discovery routine for all of them, without having to create a separate one just for the Waveman-Switch decoder, brining the casing in line with all the others makes things more consistent and easier for discovery implementation.

See
Brennenstuhl RCS 2044
Nexa
Proove

### DIFF
--- a/src/devices/waveman.c
+++ b/src/devices/waveman.c
@@ -71,7 +71,7 @@ static int waveman_callback(r_device *decoder, bitbuffer_t *bitbuffer)
             "id",       "",     DATA_STRING,    id_str,
             "channel",  "",     DATA_INT,       (nb[1] >> 2) + 1,
             "button",   "",     DATA_INT,       (nb[1] & 3) + 1,
-            "state",    "",     DATA_STRING,    (nb[2] == 0xe) ? "on" : "off",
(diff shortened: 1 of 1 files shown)


DigiH:
0/1 would be fine as well, as long as there is some consistency, especially for commonly creating discovery messages for all such keys, with the payload_on and payload_off being the same. Then with assigning appropriate type device classes for different decoders the HA display would be converted nicely from ON/OFF - 1/0.

https://www.home-assistant.io/integrations/binary_sensor/

When there is no device class specified (None) the default binary sensor for HA is actually showing ON/OFF, so maybe it might be ok to keep. This would also mean less of a breaking change for existing users with different controllers, when only one decoders has a casing change from on/off to ON/OFF!?!

And yes, the key name "state" is also confusing, as it appears as binary ON/OFF 1/0 as above, but also as open/closed (again with inconsistencies throughout as "open", "Open" or "OPEN") or as completely arbitrary strings or an integer. 

But aligning all binary keys and generalising string casings would be a great help in making discovery messages a lot easier, without having to investigate every individual decoder.

PR #2698, opened 2023-10-27
Add support for Bresser Lightning Sensor and Bresser Air Quality Sensor

matthias-bs:
See https://github.com/merbanan/rtl_433/issues/2140 and https://github.com/merbanan/rtl_433/issues/2693

PR #2953, opened 2024-06-02
Update bresser_7in1.c mass concentration naming

bz177:
I noticed the mass concentration of air quality sensors are named different in the fineoffset_wh45 protocol and rtl_433_mqtt_hass.py script, this is to get the Bresser 7in1 protocol align to have the same name everywhere. 

(This is my very first PR / comment, please excuse any mistakes I'm probably unaware of)


zuckschwerdt:
You are right, this comes from #2698 and should not have slipped through. We may need a check to warn on uncommon keys.
This is a breaking change though and needs a warning in the commit log. I'll reword on merge.

"Change to pm_2_5_ug_m3, pm_10_ug_m3, BREAKING change to Bresser-7in1"

Let this PR sit for ~2 weeks so others can notice and comment if needed. Ping later if we miss to merge please.

PR #2960, opened 2024-06-07
Add nice_flor_s support for Nice One remote

Hellslicer:
This PR adds compatibility for Nice One remotes on the Nice Flor-s protocol


zuckschwerdt:
Thanks. Looks good.

Issue #2961, opened 2024-06-09
Add support for ThermoPro TP829B


AryehGielchinsky:
I'd like help adding the [ThermoPro TP829B](https://buythermopro.com/product/tp829/) meat thermometer. I'm somewhat new to radio decoding and RTL_433. I've read through https://triq.org/rtl_433/.

I have my rtl_sdr positioned an inch from the transmitter with the antennas unscrewed.

`rtl_433 -f 915.1M -S unknown` produces the attached files with probe1 reading 77F and the other 3 probes disconnected.
[77F_probe1.zip](https://github.com/user-attachments/files/15751685/77F_probe1.zip)

```
(base) aryeh@desktop:~$ rtl_433 -A g001_915.1M_1000k.cu8 -s 1000k
rtl_433 version 23.11-120-g8d35eb3e branch master at 202405210902 inputs file rtl_tcp RTL-SDR
[Input] Test mode active. Reading samples from file: g001_915.1M_1000k.cu8
Detected FSK package	@0.022479s
Analyzing pulses...
Total count:   74,  width: 26.67 ms		(26673 S)
Pulse width distribution:
 [ 0] count:    1,  width:    0 us [0;0]	(   0 S)
 [ 1] count:   52,  width:  102 us [93;106]	( 102 S)
 [ 2] count:   10,  width:  204 us [202;209]	( 204 S)
 [ 3] count:    5,  width:  307 us [306;312]	( 307 S)
 [ 4] count:    4,  width:  412 us [408;416]	( 412 S)
 [ 5] count:    1,  width:  719 us [719;719]	( 719 S)
 [ 6] count:    1,  width:   12 us [12;12]	(  12 S)
Gap width distribution:
 [ 0] count:    1,  width:   34 us [34;34]	(  34 S)
 [ 1] count:   59,  width:  103 us [98;107]	( 103 S)
 [ 2] count:    9,  width:  205 us [202;209]	( 205 S)
 [ 3] count:    2,  width:  563 us [508;618]	( 563 S)
 [ 4] count:    1,  width:  306 us [306;306]	( 306 S)
 [ 5] count:    1,  width: 6003 us [6003;6003]	(6003 S)
Pulse period distribution:
 [ 0] count:    1,  width:   34 us [34;34]	(  34 S)
 [ 1] count:   40,  width:  205 us [200;210]	( 205 S)
 [ 2] count:   18,  width:  307 us [305;314]	( 307 S)
[...]

PR #2964, opened 2024-06-10
Add support for ThermoPro TP829b and Improve rtl_433_mqtt_hass.py

ProfBoc75:
Related to #2961 

This decoder is decoding 4 temp probes, the reason why rtl_433_mqtt_hass.py is also updated accordingly (2 temp probes in previous version)

### DIFF
--- a/src/devices/thermopro_tp829b.c
+++ b/src/devices/thermopro_tp829b.c
@@ -1,0 +1,127 @@
+/** @file
+    ThermoPro TP829b Meat Thermometer four probes.
+
+    Copyright (C) 2024 Bruno OCTAU (\@ProfBoc75)
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "decoder.h"
+
+/**
+ThermoPro TP829b Meat Thermometer four probes.
+
+Issue #2961 open by \@AryehGielchinsky
+Product web page : https://buythermopro.com/product/tp829/
+
+Flex decoder:
+
+    rtl_433 -X "n=tp829b,m=FSK_PCM,s=102,l=102,r=5500,preamble=d2552dd4" *.cu8 2>&1 | grep codes
+
+    codes     : {164}082f2efeddeddedde8d2d2d2d2d20000000000000
+
+Data layout:
+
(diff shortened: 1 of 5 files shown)


src/devices/thermopro_tp829b.c
```diff
@@ -0,0 +1,127 @@
...
+    if ((msg_len - offset ) < 104 ) {
+        decoder_logf(decoder, 1, __func__, "Packet too short: %d bits", msg_len);
+        return DECODE_ABORT_LENGTH;
+    }
+
+    offset += sizeof(preamble_pattern) * 8;
+    bitbuffer_extract_bytes(bitbuffer, 0, offset, b, 9 * 8);
+
+    // checksum is a Galois bit reflect and byte reflect
+    // need first to reverse byte order then the galois bit reflect.
```
zuckschwerdt, line 74: Note that galois is a type of LFSR and bit/byte reflect is just input ordering.

src/devices/thermopro_tp829b.c
```diff
@@ -0,0 +1,127 @@
...
+    int checksum = lfsr_digest8(b_reflect, 8, 0x98, 0x55);
+    if (checksum != b[8]) {
+        decoder_logf(decoder, 1, __func__, "Checksum error, calculated %x, expected %x", checksum, b[8]);
+        return DECODE_FAIL_MIC;
+    }
+
+    decoder_log_bitrow(decoder, 2, __func__, b, 72, "MSG");
+
+    uint16_t id     = b[0] << 8 | b[1];
+    float p1_temp   = ((b[2] << 4 | (b[3] & 0xF0) >> 4) - 500) * 0.1f;
```
zuckschwerdt, line 87: It could be better to have
`int p1_raw = ...;`
`float p1_temp = (p1_raw - 500) * 0.1f;`
and then DATA_COND on the int (avoid `!=` on float)

src/devices/thermopro_tp829b.c
ProfBoc75: Yes, I did the finding with revdgst but I'm not able to find the corresponding function at rtl_433, I explorerd bit_util.c.

- I found lfsr_digest8() where the bit are reflect (bit from 8 to 1), but in normal byte order in the array.
- I found lfsr_digest8_reflect() where the bytes are reflect from last to first byte but bit are from bit 1 to bit 8.
- So missing a function to have both byte reflect and bit reflect, the reason why I need to reverse the order of byte first, then lfsr_digest8() as a workaround.

PR #2975, opened 2024-06-20
examples: mqtt_relay: Add config for individual and json topics

gdt:
Add variables to control publishing individual topics and the json topic.  Default both to on, so that there is no behavior change.

This eases use now, as one's local diff is cleaner.  It is also on the path to reading these variables from a config file.

### DIFF
--- a/examples/rtl_433_mqtt_relay.py
+++ b/examples/rtl_433_mqtt_relay.py
@@ -35,6 +35,8 @@ MQTT_USERNAME = None
 MQTT_PASSWORD = None
 MQTT_TLS = False
 MQTT_PREFIX = "sensor/rtl_433"
+MQTT_INDIVIDUAL_TOPICS = True
+MQTT_JSON_TOPIC = True
 
 def mqtt_connect(client, userdata, flags, rc):
     """Handle MQTT connection callback."""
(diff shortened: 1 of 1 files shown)


gdt:
I'm unclear on protocol, but think it's best if I recuse myself from merging my own PRs, even if this one is obviously ok and non-controversial.

zuckschwerdt:
Let it sit for a few days so everyone gets a chance to comment.
Looks good to me, merge when you see fit. Be sure to "Squash and merge" and set the commit message to e.g.
`Change HA script individual and json topics to optional`

gdt:
This is not an HA script.   It is mqtt_relay in general.   That's how i tried to label it.

Does squash and merge let you set the message?  Or are you saying I should edit it and force push first?

zuckschwerdt:
Oops, missed the nuance in the filename ;)
I guess this is one example script that people actually use as-is (I do), so still a thumbs up from me to improve it

The "Squash and merge" let's you change the message (and also suggests adding the issue number). It also rebases so the merge if fast-forward (no hidden merge changes).
I don't think changing the default squash merge message is possible with commits or force pushes, I think it's based on the issue title (never bothered to read up or find out).

For the actual message just order by importance even if it sounds odd. https://triq.org/rtl_433/CONTRIBUTING.html has the verbs and optional area of work. Though `examples:` could be a useful addition I guess.

gdt:
We should add examples:; to me the point is to orient to the area of the tree.

PR #2976, opened 2024-06-20
examples: Improve README by describing scripts

gdt:
Expand description of the directory, making it clearer that some programs have production use.

Declare that python scripts are python3.  It's clear that nobody is trying to use them with python 2.7, and if they did, we would tell them not to.

List the scripts with a one-line description, sorting them into production (none yet), generally usable, and true examples.  Leave some uncategorized.


zuckschwerdt:
Looks good! You can delete the sigrok-* entries. We once had external scripts for conversion but rtl_433 can now handle that internally. We should delete the deprecated scripts.

Issue #1612, opened 2021-01-15
Official Docker image?


gcormier:
I've done a search of the open and closed issues here for mention of Docker and couldn't find any discussions.

Would there be any interest in an official Docker image for rtl_433 to run from containers? There could be 2 paths

1. A simple Dockerfile in this repository, such that users could build their own image
2. An additional layer that integrates this repository with dockerhub to automatically build new images when the repository is updated

If there is interest, I am by no means a docker expert, but I could fork and give it a shot and send a PR.

zuckschwerdt:
rtl_433 itself is simple enough not to warrant a docker image. Basically it just needs a libc. There is a dep on libusb for error messages but that can be optional.
If someone wants a just-deploy-whatever-is-needed docker image it would need to contain librtlsdr and likely SoapySDR, and modules and then a mechanism to configure and run, possibly with supervision and esp. with output routing.
I guess consumers of rtl_433 output can benefit from a custom build. But just rtl_433 in itself? Not sure.

I'd rather offer precompiled static binaries and maybe a deb for, I guess most common, use on a Raspi. The user will need to configure stuff anyway, and that way without setting up e.g routing of MQTT through a network abstraction.

gcormier:
A docker build brings it into the container ecosystem. My home server is all docker, which includes a home automation ecosystem.

So it inherits all the benefits of docker, including my ability to blow away Ubuntu 20.04 and it's terrible forcing of snap, install another OS, and I have no reconfiguration at all to be done.

With rtl_433 that was broken - first off there is either the building process, or, finding some sort of repository that is up to date. Second, I then need rtl_433 to be able to connect to a MQTT container inside docker which was not being exposed, so I started looking at exposing this container such that the host would talk to it. I'd also need systemd/whatever to handle a crash of rtl_433.. something I've enjoyed not doing in ages 👍🏼 

Finally, it brings a single component outside of my perfect sphere of containers, and removes a level of automation I've come to enjoy.

I've got many other images that would not warrant a docker image as you suggested, other than the fact that it bundles them nicely and means I run an entire server from a single `docker-compose.yaml` :) It's a nice place to be.

zuckschwerdt:
I'm sure other users want that level of plugable services. At least I've seen such setups recommended. Still we'd need a base image with SDR libs or bring that into the build.

What host and distro is your automation server running? What is your process of bringing configs into containers?

gcormier:
[...]

PR #3006, opened 2024-07-24
docker: Add official local docker support

oliv3r:
Lets add our own docker container and use github actions to build and push it.

For now, only the alpine container is built and pushed, but we can very easily add a matrix entry for a debian container if there's enough demand.

```patch
@@ -25,21 +25,24 @@
         include:
           - container: ./dist/Containerfile
             autotag: auto
+          - container: ./dist/Containerfile.debian
+            autotag: false
+            suffix: -debian
```

But the build is done for multiple architectures, so there's that.

Best to reduce the amount of configurations we offer, as it also means more that can break.

Note, that there's a good chance tagged-builds will fail, because this repository is using so called 'lightweight' tags, instead of properly 'annotated' tags! See https://git-scm.com/book/en/v2/Git-Basics-Tagging for more details. This also could be the reason some additional complexity is added to the `git get tags` cmake stuff.

A nice addition could be to also run the tests as part of the build process, but I already saw that it's not as simple as `make test` so I figured better keep the current CI way of testing :)

Note, that I've pushed this code to my personal master branch and this shows how releases would look like here https://github.com/oliv3r/rtl_433/pkgs/container/rtl_433.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+examples/
--- a/.github/workflows/container-build.yaml
+++ b/.github/workflows/container-build.yaml
@@ -1,0 +1,73 @@
+name: Create and publish Container image
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '[0-9]+.[0-9]+*'
+  pull_request:
+    branches:
+      - master
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        include:
+          - container: ./dist/Containerfile
+            autotag: auto
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=edge
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+          flavor: |
+            latest=${{ matrix.autotag }}
+            suffix=${{ matrix.suffix }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          platforms: linux/amd64,linux/386,linux/arm64,linux/arm/v7,linux/arm/v6
+          context: .
+          file: ${{ matrix.container }}
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/README.md
+++ b/README.md
@@ -25,7 +25,24 @@ On FreeBSD, `pkg install rtl-433`.
 
 On MacOS, `brew install rtl_433`.
 
-Docker images with rtl_433 are available [on the github page of hertzg](https://github.com/hertzg/rtl_433_docker).
+## Docker
+
+Running the application through docker is easy with the offical docker image from this repository. Either use `latest` for the latest stable, a tagged version or `master` for the latest build from master. See https://github.com/merbanan/rtl_433/pkgs/container/rtl_433 for available releases.
+
+```console
+$ docker run \
+         --device '/dev/bus/usb:/dev/bus/usb' \
+         --interactive \
+         --rm \
+         --tty \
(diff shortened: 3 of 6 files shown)


zuckschwerdt:
Thanks! Did you see https://github.com/hertzg/rtl_433_docker/ ? It's well maintained and around for 4 years, so quiet stable I guess.

gdt:
I guess there is a big question if things should stay separate.  I see docker as one mechanism, and it appears not particularly portable.    I'd lean to letting others provide containers/packaging and not integrate it into the project and list containerization projects along with packaging.

oliv3r:
> t appears not particularly portable

What makes you say that? I think it's more portable then the binaries produced from this repo. One can run docker/podman with qemu (and binfmt-misc) and run any architecture binary on any host. Granted, docker/podman support is required, but I think on all major operating systems this is available. And in comparison to different distro's it's even more portable, because it doesn't matter. It doesn't need to be packaged, it doesn't need to be installed. So from a portability point of view, it's far more portable.

As for 'packaging for everything everywhere' I fully agree, distro's should be responsible for their own packaging. But this isn't packing per-say. This is just a binary build, in a different format if you will.

Anyway, Yes! I was fully aware of https://github.com/hertzg/rtl_433_docker/ repo, which should be visible in the actual code-change :p but there's multiple reasons for including it here. First, when looking 'hey, is there a docker container for this' (it wasn't packaged for archlinux, only as an aur), I found it 'on the hub' but was not sure if I could trust the source. It wasn't from `merbanan` or from `rtl_433` or something recognizable. So there's a trust issue.

After going to this repo, I found the link to the container, and figured 'ok this should be ok'. But since it was only a link to a git repo (which I followed) I only got confused. Why is there typescript files? Why are their build files? What's going on here? Where's the `Containerfile`, what is happening here?

Anyway, I pulled the container, and wanted to enter to container to run rtl_433 from within. This wasn't possible, because the container didn't adhere to the [official image consistency guideline]()https://github.com/docker-library/official-images#consistency). Granted, there's no rule that must be followed here, but from a users point of view, this was rather annoying (work-arounds certainly possible of course, e.g. `--entrypoint`).

Next, I wanted to build the container, or rather compile the source code using the container, and had to spend some time to find the actual `Dockerfile`, because it was hidden somewhere deep, under a root dir called `images` which I figured where about screenshots :p but finally having found the container, I was struck with some really really poor understanding on how to user Docker layering. Randomly put commands into place, or so it seemed.

Regardless, another 'issue' is, that the container is not automatically build upon a commit to master it seems (or it's not obvious and I've missed it?). The repo is untagged and unversioned.

So by bringing the container build into this project itself, we can at least build containers for every (master) commit, meaning it becomes super easy to run 'whatever is on master today'. Also (properly) tagged releases are also build, so version pinning is trivial and logical (though this is certainly also possible in the other repo). But there's also the trust issue, by having an official release from this repo, means people should be able to blindly trust things.

For me, my main trigger was, as I wrote above, I wanted to change the source, compile and run it; without having to setup a build environment and installing binaries that do not come from my package manager. This was now trivial (due to `docker build; docker run` from within the very same repo.

One thing I purposely didn't do, is do builds for many different kind of versions of operating systems (alpine-X, debian-Y) which I find not very useful. This is the same reason packaging for all distro combo's shouldn't be done here.

hertzg:
Hey, maintainer of hertzg/rtl_433_docker here. Thanks for the mention @zuckschwerdt!

@oliv3r - I appreciate the detailed feedback. Let me address a few points:

**On discoverability/trust:**
> "I found it 'on the hub' but was not sure if I could trust the source"

The rtl_433 repo already links to the project. This arrangement was discussed and agreed upon in #1612, where @zuckschwerdt said *"I'd also be okay with just making your build the recommended 'official docker solution'. Less bottleneck on our end"* and @Mindavi noted *"I'm in favor of keeping packaging and maintaining the tool separate as well."* I also added Debian/SoapySDR support at the maintainers' request as part of that discussion.

**On the repo structure:**
> "Why is there typescript files? Why are their build files? Where's the Containerfile?"

The Dockerfiles are at `images/{alpine,debian}/build-context/Dockerfile`. The TypeScript generates the CI matrix for building multiple rtl_433 versions × base image versions × architectures. I can improve the README to make navigation clearer - if you have suggestions, I'm open to PRs.

**On image consistency:**
> "the container didn't adhere to the official image consistency guideline"

The same documentation has a third case: *"If the image only contains the main executable and its linked libraries (ie no shell) then it is fine to use the executable as the ENTRYPOINT"*. This is a single-purpose tool container - the shell exists only because Alpine/Debian include one, not as an intended user interface. But FYI the official-images guideline is not a bible :)

**On Dockerfile quality:**
> "really poor understanding on how to use Docker layering. Randomly put commands into place"

The images use multi-stage builds - builder stage with compile-time deps, runtime stage copying only the compiled binary and runtime deps. Alpine image is ~3MB. Not sure what specifically you mean by "poor understanding" or "random commands" - I'm open to PRs with improvements.

**On automation/versioning:**
> "the container is not automatically built upon a commit to master... untagged and unversioned"

Images are built daily + on every rtl_433 release, for the last 3 versions + master, across multiple architectures. [Tag patterns are documented here](https://github.com/hertzg/rtl_433_docker#tags). Given the goal of keeping Docker separate from the main project, nightly builds are a reasonable tradeoff vs building on every upstream commit.

---

For context, this arrangement was discussed in #1612 where the maintainers decided to keep Docker/packaging separate from the main project.

That said, I'm not opposed to rtl_433 having its own official container if that's what the maintainers now prefer. Happy to help either way.

merbanan:
We have trouble tending to issues and PRs in a timely manner. As long as @hertzg maintains the docker actively and handles support then we should keep things like they are.

oliv3r:
> We have trouble tending to issues and PRs in a timely manner. As long as @hertzg maintains the docker actively and handles support then we should keep things like they are.

yeah sure no problem. ping me if you want me to pick this up again.

@hertzg I think it helps if you (with blessing) mark yours as like 'official' container?

Issue #2902, opened 2024-04-15
Self Powered Switch Device 433mhz Model: QX-302, QX-305


adminy:
here is some logs from running `rtl_433 -A` and pressing the button a couple of times. No json, so its probably not a registered device yet. let me know how I can collect info for it and / or decode. thanks.

QX-302: https://logpaste.com/l8qA3HsW
QX-305: https://logpaste.com/fPuVfdnJ

Wireless Frequency: 433.92Mhz

Manufacturer: Shenzhen QiChip Technology Co. Ltd.

zuckschwerdt:
If you look at the "count" fields the numbers are too low. We expect more data, like 32 bits or even much more.
From what we know of self-powered devices the transmissions are short and fast.

Re-run with a higher sampe rate, e.g. `-s 1024k`
Also grab one sample, verify it looks promissing in https://triq.org/pdv/ and upload here as zip, see https://triq.org/rtl_433/ANALYZE.html

adminy:
[scans.tar.gz](https://github.com/merbanan/rtl_433/files/14984137/scans.tar.gz)

looks promising on the verify step, but not totally sure. still feels like its not consistent across scans. this is data from QX-305 only.

zuckschwerdt:
Actually a very simple OOK signal with 25 bits at 33/100 µs. This will decode the data:
`rtl_433 -R 0 -X 'n=name,m=OOK_PWM,s=33,l=100,g=150,r=1500'`

If it only has one function (single press on single button) then use this to only accept that button:
`rtl_433 -R 0 -X 'n=name,m=OOK_PWM,s=33,l=100,g=150,r=1500,match={25}573ace8,countonly'`

See https://github.com/merbanan/rtl_433/tree/master/conf example on how to write a documented conf file.
PR or paste us the finished file if you like to.

adminy:
I'm not sure that's it though ... https://logpaste.com/3X0Y7K6j, Also the single press command literally gives back nothing.

adminy:
https://github.com/merbanan/rtl_433/compare/master...adminy:rtl_433:patch-1 Something like this then?

edco:
I have a QX-304 which works with qx-30x.conf. Unfortunately I just got the one, single gang button to test, so I can't compare the data from different devices/buttons, but would it be reasonable to assume this is a 1527 code? In that case I'm thinking we'd ignore the final bit, and split the remainder into 20 and 4 bits:
```
  bits       >= 24,
[...]

PR #3008, opened 2024-07-25
Change support for qx-30x to extract device and button ID

edco:
Per [my comment](https://github.com/merbanan/rtl_433/issues/2902#issuecomment-2094812165) on the initial support request for these devices, this PR adds ID and button extraction based on the EV1527 encoding.

@adminy would you be able to give this a try with your switches and share the ID and button values extracted? I'm hoping you've got a 2-gang QX-302, but I'm keen to see the output regardless.

### DIFF
--- a/conf/qx-30x.conf
+++ b/conf/qx-30x.conf
@@ -1,6 +1,10 @@
 # QX-30X
 #
-# This decoder reads button presses from QX-302, QX-305 self powered switches.
+# This decoder reads button presses from QX-302, QX-304 and QX-305 self powered
+# switches. These switches appear to use an EV1527 based encoding, where the
+# first 20 bits contain the device ID, the next 4 bits identify which button on
+# the device was pushed, and there may be a trailing sync bit. A sample rate of
+# around 1MHz is recommended to reliably detect these devices.
 #
 
 decoder {
@@ -10,6 +14,9 @@ decoder {
         long        = 100,
         gap         = 150,
         reset       = 1500,
-        bits       >= 25,
-	unique
+        bits       >= 24,
+        bits       <= 25,
+        unique,
+        get         = get=@0:{20}:id,
+        get         = get=@20:{4}:button
 }


zuckschwerdt:
These changes won't break anything, even if you get no test confirmation please ping back for merge in a few days.

edco:
> These changes won't break anything, even if you get no test confirmation please ping back for merge in a few days.

Ping.

adminy:
sure, I'll run a test tomorrow morning if that's okay.

edco:
Sounds good.

adminy:
So your config didn't work out of the box, I had to add extra fields, namely `name`, `modulation`, `short`
```sh
# QX-30X
#
# This decoder reads button presses from QX-302, QX-304 and QX-305 self powered
# switches. These switches appear to use an EV1527 based encoding, where the
# first 20 bits contain the device ID, the next 4 bits identify which button on
# the device was pushed, and there may be a trailing sync bit. A sample rate of
# around 1MHz is recommended to reliably detect these devices.
#

decoder {
	name        = QX,
	modulation  = OOK_PWM,
	short  	    = 33,
	long        = 100,
	gap         = 150,
	reset       = 1500,
	bits       >= 24,
	bits       <= 25,
	get         = get=@0:{20}:id,
	get         = get=@20:{4}:button,
        unique
}
```
Screenshot 2024-07-31 at 09 35 24

Also I added `unique` back and now the output looks a lot more readable:

Screenshot 2024-07-31 at 09 40 40

zuckschwerdt:
> So your config didn't work out of the box, I had to add extra fields, namely `name`, `modulation`, `short`

Did you perhaps copy from the diff? The file seems complete (click the dots to the very right of the file name, then View File) https://github.com/merbanan/rtl_433/blob/dd57e1f879231b8c13530389284425d147681b3b/conf/qx-30x.conf

adminy:
hahah, yeah, that's it.

edco:
> as for two gang switch, data is mostly the same:

It seems wrong that each one would have a different device ID. But then-again, I suppose it's entirely possible that they've put a completely separate transmitter board in each of the switches, rather than trying to wire up the generators from both switches to a single board and attempting to tell the board which button is being pushed while sending a brief pulse of power to it.

It seems to me like it's working.

PR #3013, opened 2024-07-29
examples: Add ability for mqtt_relay to use a config file

gdt:
examples: Add ability for mqtt_relay to use a config file

This commit adds the ability to read config values from
rtl_433_mqtt_relay.yaml (overriding defaults).  This read is skipped
if the yaml module can't be loaded or the config file can't be read,
so those without PyYAML or a config file see no change.

Those with a config file can set parameters for address/port to listen
for json-in-sylog, MQTT credentials, whether to post individual
values, a json dictionary, or both.

While rudimentary, this is a huge functional improvement, allowing
those using TLS and broker passwords (as arguably all should be) to
use the script without modifying it.

### DIFF
--- a/examples/rtl_433_mqtt_relay.py
+++ b/examples/rtl_433_mqtt_relay.py
@@ -24,19 +24,52 @@ import socket
 import json
 import paho.mqtt.client as mqtt
 
-# Syslog socket configuration
-UDP_IP = "127.0.0.1"
-UDP_PORT = 1433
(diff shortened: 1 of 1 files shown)


gdt:
This ran well overnight on three machines (1 mac and 2 Linux), so I am changing from 'this is a draft' to "I would like to merge this'.

zuckschwerdt:
LGTM, thanks!

PR #3018, opened 2024-08-02
examples: Add dup filtering to mqtt_relay

gdt:
Keep information about the previous value sent.  If it's been 5 seconds, or new value is different (ignorning keys like snr and frequency), then send it.  Otherwise, just don't.  This causes bursts of e.g. 4 transmissions to result in one MQTT message, on the theory that the 4 transmissions are not actually 4 messags, but a strategy to transmit one message more reliably.

Define a new configuration option to enable duplicate filtering, and default it to True.

Tonight, this is a draft with debug logging enabled.
But it's running on 3 systems, 1 macOS and 2 Debian, and seems to be working ok.

Obviously I should turn off debug logging, and I'm inclined to just change to debug = False, and not make it a config.   I lean to understandability and defensive code over beauty.

### DIFF
--- a/examples/rtl_433_mqtt_relay.py
+++ b/examples/rtl_433_mqtt_relay.py
@@ -20,10 +20,14 @@
 from __future__ import print_function
 from __future__ import with_statement
 
-import socket
 import json
+import logging
+import socket
+import time
+
 import paho.mqtt.client as mqtt
 
+
 # The config class represents a config object.  The constructor takes
 # an optional pathname, and will switch on the suffix (.yaml for now)
 # and read a dictionary.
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Very nice! Maybe change to `logging` and later just init to INFO level. S.a. https://github.com/merbanan/rtl_433/blob/master/examples/mqtt_filter.py#L179-L180

gdt:
Thanks; will steal the logging stuff.

Interesting to see two independent dedup implementations.  I already had been designing it in my head, and I did see your commit, but I avoided looking.   Your impl keeps track of multiple messages and mine doesn't.  Mine calls things a match if they differ in rx metadata, and yours doesn't.   That is probably fine if people aren't using -M.

All of this is making me think we need to move to having a real python module with installed scripts, that can use common code via import rtl_433.  But that's a lot of work and not urgent.

zuckschwerdt:
Don't look too hard at my code. It's just a quick proof of concept to get a basis for discussion. I haven't even tested anything…

gdt:
> Don't look too hard at my code. It's just a quick proof of concept to get a basis for discussion. I haven't even tested anything…

Thanks for the warning.   I am running what's in the PR and it worked overnight.  I'm seeing only 1 line per sensor report (well 2 for 433 from 2 computers in different places) and my sensors are reporting (modulo a sysadmin error now fixeed).

But I do want to improve logging before merging.

gdt:
I took the logging config from mqtt_filter.py, and made it configurable.  I also moved comments into docstrings.

This is now ready for actual review; modulo recent cosmetic changes, it ran overnight.

PR #3055, opened 2024-09-19
RollEaseAcmedia pulse time refinement

bdnstn:
Refine the pulse timing by exploring the limits of pulse timings that work and choosing the mid points. Tested an verified against 10 different blind motors on 2 different address and found improved (now 100%) reliability.

### DIFF
--- a/conf/oma-blind-remote.conf
+++ b/conf/oma-blind-remote.conf
@@ -1,0 +1,63 @@
+# Decoder for oma battery powered roller blind motor controller
+# Tested with blind motor with electronic limit switch. Mechanical limit 
+# switch motors not tested.
+# Instructions (LKC040902 ver 1.0) reference motors C-C16/C-C17.
+
+# The device tested is model 8C-C18-DC-WA-A101 5 channels
+# 433.92MHz
+# https://www.omaautomation.com/
+# FCC ID: Unknown
+
+# Note. There is another controller of similar apearance, which has 
+# programmable date/time functions, which uses a different (incompatable)
+# protocol. 
+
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
Appreciated, thanks! Can you document your findings about the timings ranges?

bdnstn:
Sure, in the conf file?
I built a transmitter using _Liligo LoRa ESP32_ to operate my blinds. I found the original timings were failing about 3% of operations. By experimentation, I found that the upper and lower limits of short pulses for successful operation were 460 and 350 us respectively. In all test I just made the long pulse timing 2*short pulse. 
I then chose the mid-point between upper and lower limits `(460+350)/2=405us` for short and `2*405=810us` for long. Subsequent testing of transitions with these timings has been 100% reliable over about 100 cycles.

zuckschwerdt:
Yes, a line of something like  "short pulses have been found to work between 350 to 460 us, with long pulses twice that."

Note that  for the simple case (no `tolerance` set) the slicer determines short/long at the mid-point between short and long, i.e. for s=352, l=725 it's 538.5 µs and for the new s=405, l=810 it's 607.5 µs. I'd say that both timings should work the same for the decoder, assuming there are no pulses in that mid range of 500-600 µs.

bdnstn:
Sorry, it looks like the new conf I have created for OMA controller has got muddled with the PR for the amendment to the rolleasacmedia.conf. I'm a NOOB with github :-(

zuckschwerdt:
I'll see if I can fix that with a manual merge.

zuckschwerdt:
manually merged, also changed the line endings from DOS (CR+LF) to UNIX (LF).

Issue #3067, opened 2024-10-11
Analysizing data from Aston Martin 433MHz Smartire TPMS


MatthewRuddy:
I have been trying to reverse decode the TPMS sensors for a 2008 Aston Martin, which uses a Smartire system with 433MHz TPMS sensors. I have no doubt they're similar to something already decoded with rtl_433, but can't quite figure out which sensors they are most like. I'm also pretty new to this!

I've been able to capture some data when running `rtl_433 -S unknown`, which I've attached. The sensors I'm using isn't inside a tyre, so it will be reading 0/1kpa, FYI. Temperature should work. I have an Autel TS408 TPMS programming device, which can activate the sensor remotely, which I'm doing to capture the data here with a Nooelec NESDR Smart V5 and an antenna.

Can anyone help point me in the right direction? I'm sure this sensor is similar to something already figured out here.

Thanks.

[g012_433.92M_250k.cu8.zip](https://github.com/user-attachments/files/17340489/g012_433.92M_250k.cu8.zip)
[g013_433.92M_250k.cu8.zip](https://github.com/user-attachments/files/17340491/g013_433.92M_250k.cu8.zip)

zuckschwerdt:
Try the sample file on https://triq.org/pdv/ -- it's quiet slow (for a TPMS) at 167 µs a half-bit, 6 kBps. And it uses OOK -- we normally see FSK at 50 µs.
Use a decoding of MC with 167 / 167. Note that there is a 3 half-bit desync at the start.

Try a flex decoder like this and record some changes in temperature or pressure to spot those data fields.
`rtl_433 -R 0 -X 'n=name,m=OOK_MC_ZEROBIT,s=167,l=167,r=600'`
The data from e.g. your first sample is `{56}518c001e187f4b`.

MatthewRuddy:
Thanks for your reply @zuckschwerdt. I'm afraid I'm out of my depth here so I will have to figure out what most of that means 😅  I would make sense for the system to be slow - it was developed in the early 2000's and is known to be a pretty mediocre.

I ran `rtl_433 -R 0 -X 'n=name,m=OOK_MC_ZEROBIT,s=167,l=167,r=600'` and recorded two scenarios. One with the sensor just sitting on a table, and another with me spinning it around in a circular motion (to simulate a wheel). When spinning it, upon reading with the TPMS programming device it showed 40kpa pressure. It seems moving the sensor around triggers it into providing a KPa value other than 0.

On table (no movement):
[...]

PR #3072, opened 2024-10-19
Add support for Aston Martin Smartire TPMS sensor, Vantage DB9 protocol

ProfBoc75:
Related to issue #3067 about Aston Martin Smartire TPMS sensor.

Few flags not yet discovered but most of the key metrics (pressure & temp) are properly guessed.


src/devices/tpms_aston_martin.c
```diff
@@ -0,0 +1,151 @@
...
+        decoder_logf(decoder, 1, __func__, "Unknown message type %x", msg_type);
+        return DECODE_ABORT_EARLY;
+    }
+
+    int inflate = (b[4] & 0x80) >> 7;
+    int flags   = b[4] & 0x7f;
+
+    /* clang-format off */
+    data_t *data = data_make(
+            "model",         "",              DATA_STRING, "Aston-Martin-Smartire",
```
zuckschwerdt, line 116: Only a single dash, and should be shorter. Since SmarTire is the manufacturer maybe `SmarTire-AM`?
And decoder file could be tpms_smartire.c perhaps?

ProfBoc75: you're right, renamed everywhere.

merbanan:
Supported in master branch.

Issue #3074, opened 2024-10-21
Device support for Technoline WS 9767


Sola85:
I would like to add device support for my [Technoline WS 9767](https://www.technoline-berlin.de/product/de/WS_9767) temperature sensor. I have recorded several signals and understand them to the point where I can extract the Sensor ID, Battery state, button state, temperature and I can reproduce the checksum. 

Upon replacing the battery in my sensor, I noticed that only then did my sensor get recognised as an Acurite-606TX (and the data provided by rtl_433 matches the data that I had already reconstructed using a flex-decoder). 

I emphasize, the sensor is only recognised by rtl_433 if the battery state bit is set to 1. Unfortunately the sensor spends 99% of its battery life in the self-reported low-battery state, so for 99% of the time, the sensor is not recognized.

Would you prefer if I attempt to make the existing Acurite-606TX decoder work with both Acurite and my Technoline sensor, or would you prefer a completely new decoder, despite the data and encoding being seemingly almost identical?

zuckschwerdt:
Did you find out what prevents the sensor from being decoded when the battery bit is unset?

Best case we'd fix or enhance the existing decoder. But if the battery bit is inverted then we would need a new decoder to override the Acurite-606TX -- and it would need to be default-disabled (only enabled by users who know that they have the sensor and want the battery state the right way around).

Sola85:
I'm currently guessing that the issue is [this check](https://github.com/merbanan/rtl_433/blob/master/src/devices/acurite.c#L1634) in the Acurite decoder. 
Both Technoline and Acurite sensors seem to send 33 bits, which get rounded up to 4 bytes. For Acurite, the last bit (=last byte) seems to be always zero, at least that is what the line above in the Acurite decoder sanity-checks. But for my Technoline sensor, the last bit seems to be the negated battery bit (i.e. the battery status is encoded in 2 bit positions, for whatever reason), hence the final-byte sanity check fails, depending on the battery state.

Here are two example datasets, obtained using `-X name=Technoline,modulation=OOK_PPM,s=1948,l=3900,g=4000,r=9000,priority=n,bits=32`

```
time      : 2024-10-22 17:16:01
model     : Technoline   count     : 1             num_rows  : 4             rows      :
len       : 33           data      : eb80d75e0,
len       : 33           data      : eb80d75e0,
len       : 33           data      : eb80d75e0,
len       : 32           data      : eb80d75e
[...]

PR #3078, opened 2024-10-26
Improve Acurite-606TX decoder to also decode the Technoline TX960 sensor

Sola85:
Fixes https://github.com/merbanan/rtl_433/issues/3074, see that issue for background. 

In short: The Technoline TX960 is almost identical to the existing Acurite 606TX temperature sensor. The only difference seems to be the final bit of each message, which is zero for Acurite but is the negated battery bit for the Technoline sensor.

By removing a sanity check on this last bit, both sensors get decoded reliably.


Sola85:
> Very nice, thanks! Can you run `./maintainer_update.py` to also update conf man man, just add another commit.

Done. (This changed a few lines unrelated to this PR, which I didn't include in the commit. Do you want me to add all changes caused by this?)

Issue #3082, opened 2024-11-01
Add support for ThermoPro TP828B 


rbray89:
This device is similar enough to the ThermoPro TP829B that I was able to key off of the support implemented for it, but different enough that the existing encoder isn't working.

I have the following output with the first probe inserted, and the second unused, ending at 86F:

Temps:
Probe 1: Target=170, Actual=70's - 86
Probe 2: Target=165, Actual=None

```
rtl_433 -d rtl_tcp://radio-rtl-2.home:1234 -f 915M -F json -X "n=tp828b,m=FSK_PCM,s=102,l=102,r=2000,preamble=d2552dd4,bits>=100" | jq --unbuffered -r '.codes[0]'
rtl_433 version 23.11 (2023-11-28) inputs file rtl_tcp RTL-SDR SoapySDR

New defaults active, use "-Y classic -s 250k" if you need the old defaults

Use "-F log" if you want any messages, warnings, and errors in the console.
{154}772c2eceaa4f3eddeaa4d7b2d2d2d2d2d200000
{164}772c2eceaa4f3eddeaa4d7b2d2d2d2d2d20000000
{154}772c5d9d549e3eddeaa4d7b2d2d2d2d2d200000
{154}772c2f8eaa4f3eddeaa4d7e8d2d2d2d2d200000
{182}772c369d549e7dbbd549af61a5a5a5a5a4000000000000
{166}772c33ceaa4f3eddeaa4d703d2d2d2d2d200000000
{170}772c341eaa4f3eddf5526bcfe969696969000000000
{154}772c335eaa4f3eddeaa4d7ccd2d2d2d2d200000
{156}772c331eaa4f3eddeaa4d70dd2d2d2d2d200000
{169}772c35fd549e7dbbd549ae39a5a5a5a5a4000000000
{154}7716194eaa4f3eddeaa4d759d2d2d2d2d200000
{154}772c31deaa4f3eddeaa4d778d2d2d2d2d200000
```

Let me know if you need anything else.

ProfBoc75:
Hi @rbray89 

The data layout is not exactly the same,, here more data / more information (temp,  target, unknown ) than tp829b which provides only temp from the 4 probes. Can you share more samples to confirm, please.

rbray89:
@ProfBoc75 

Here you are:
```
Temperatures in C:

<DATA>                                                                       <TARGET1/PROBE1> <TARGET2/PROBE2>
[...]

PR #3085, opened 2024-11-06
Add support for Thermopro TP828B Meat Thermometers 2 probes

ProfBoc75:
Relate to issue #3082 

- This model TP828B is close to the other model TP829B.
- Previous decoder have been renamed and updated accordingly.


zuckschwerdt:
Most things in the decoder exist twice, once for the TP829B and once for the TP828B.
I'm not sure, it might be easier to use two decoders? Keep one `thermopro_tp82xb.c` but make it two `r_device` and two decoder functions, each much shorter and without the branching.
What do you think?

ProfBoc75:
> Most things in the decoder exist twice, once for the TP829B and once for the TP828B. I'm not sure, it might be easier to use two decoders? Keep one `thermopro_tp82xb.c` but make it two `r_device` and two decoder functions, each much shorter and without the branching. What do you think?

Yes, I will try, my last version of the decoder can be simplified, as the first data part until temp probe 1 is the same, after we have a gap.

I can also consolidate into a single `data_t *data = data_make ...` and put some DATA_CON to show the good model & good values too.

- keep the model guess part
- extract the common data (id, flag, display unit, temp probe 1)
- Depends on the model , extract the data accordingly.
- Output all data into a single data_make list with DATA_CON for model, probes, lo and hi target temp ... like it's already the case in fact (except model) because probe(s) can be removed in that case 0xedd is set by default so I exclude this value.

I'm busy, so will try later, your proposition too...

zuckschwerdt:
Thanks! To clarify: the idea is to have two very simple decoders, not because it's less code, but because it's then a very simple flow without any branching or DATA_COND. ("clever" decoders are hard to maintain later on…)

Generally I think: If we want a different "model" (because the protocol is somewhat different) then we should try to have multiple decoders. And if we want the same "model" (because only some data bits are different) then we should try to have only one decoders.

zuckschwerdt:
This looks much nicer and easier to follow, I think. Thanks!
The file info block should only contain licence info. If you like lines 10-15 to stay then put them in just a `/* */` block (i.e. not `/**`).
The first doc-comment line (21 and 137) should end with a dot (`.`) :)

PR #3092, opened 2024-11-14
Add support for Bresser/Explore Scientific ST1005H

DonKult:
I am very new to this and my SDR hardware is old DVB-T equipment (RTL2832U+FC0012+20cm rod antenna), so probably not quality cu8 files (they are at least rather large for something so trivial)… attaching one anyhow as it shows that what is send by this variant of the Bresser Thermo-/Hygro-Sensor 3CH is misdetected by the "Oregon Scientific SL109H Remote Thermal Hygro Sensor" decoder if the bits align. I am happy to collect some for the test repository though if that helps and they turn out good enough for you. [bresser-misdetect-as-oregon_433.92M_250k.cu8.gz](https://github.com/user-attachments/files/17754278/bresser-misdetect-as-oregon_433.92M_250k.cu8.gz)

The timings were thankfully provided by `-A` and the protocol is so simple that I think I managed to figure the meanings of the individual bits out except for the last 6bits that are probably a checksum, but I lack the experience to make a proper guess as to what type I am afraid.

As such, this remains a draft for the time being. Also as I don't know much about the sensor… I got two in a sale a while ago for far cheaper than the current asking price (from a vendor who sells them for even more than Bresser now?!) under the impression/hope that this might be the ones already supported by rtl_433… they have the same name and feature set after all. Of course, they were not. Could hence be an outdated model mainly sold as (pricey) replacements (covering up the "EXPLORE SCIENTIFIC" branding that seems to be used on other Bresser sensors, too) and so maybe not the most pressing/useful to support.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -275,6 +275,7 @@
     DECL(rosstech_dcu706) \
     DECL(risco_agility) \
     DECL(thermopro_tp828b) \
+    DECL(bresser_st1005h) \
 
     /* Add new decoders here. */
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -72,6 +72,7 @@ add_library(r_433 STATIC
     devices/bresser_7in1.c
     devices/bresser_leakage.c
     devices/bresser_lightning.c
+    devices/bresser_st1005h.c
     devices/bt_rain.c
     devices/burnhardbbq.c
     devices/calibeur.c
--- a/src/devices/bresser_3ch.c
+++ b/src/devices/bresser_3ch.c
@@ -14,6 +14,8 @@ Bresser sensor protocol.
 
 The protocol is for the wireless Temperature/Humidity sensor
 - Bresser Thermo-/Hygro-Sensor 3CH
+  Another sensor sold under the same generic name is handled by
+  bresser_st1005h.c.
 - also works for Renkforce DM-7511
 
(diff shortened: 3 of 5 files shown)


zuckschwerdt:
It seems like the first "bit" of the 38 needs to be ignored (have you ever seen the first bit set?), then the fields mostly align to bytes and checksum is field-wide addition, e.g. something like `1x ID:8h FLAGS:4h TEMP:12h HUM:7d CHK:6h` in BitBench, with checksum maybe ID+FLAGS+TEMP+HUM? But maybe HUM counts 2x, i.e. padded to 8-bit value, not sure.

I think the bresser shouldn't be changed, this looks like a very different protocol. Not sure if ST1005H is the original name then?

DonKult:
Thanks for the quick and spot-on hints! The first bit is indeed always 0 & humidity is doubled in the checksum. All that was left was figuring out that the individual nibbles of the values are added together to form the checksum.

I moved the implementation to a new `bresser_st1005h` as it indeed shares nothing but the generic name with `bresser_3ch`. I did keep `bresser` in the name through instead of `explore_scientific` as the later does not seem to sell them anymore nor have they any info on their website about it. Also, `bresser_6in1` is the only other sensor making a mention of that brand name with bresser prefixed so as long name I picked `Bresser Thermo-/Hygro-Sensor Explore Scientific ST1005H`. I think that would have given me a reasonable chance to find it.
As usual, naming things is hard… so more than happy to change that further.

Issue #2267, opened 2022-12-04
Support for Quinetic 433Mhz Button


pazzernick:
Hi,

Firstly, thanks for the fantastic rtl_433 tooling. Amazing work!

I'm trying to add support for a kinetic-powered switch, which is typically 'paired' via RF to one or more receivers such as Smart Plugs, Lighting Relay, etc. Model: [Quinetic QUGDMK](https://www.tlc-direct.co.uk/Products/QUGDMK.html)

I've demodulated the samples in **Universal Radio Hacker** and think I've found the data...
* Device Model (48-bits from bit 85)
* Unique ID (32-bits from bit 133)

Signal Characteristics:
* Signal Type: FSK PCM
* Signal captured at 2048k sample-rate
* Button is designed to broadcast 3 times (mentioned in product docs) which is why 3 (sometimes 4) pulses occur.
* I've tried to capture signals at 250k and 1024k sample rate, no success with URH or rtl_433.

Command used to capture samples ( antenna attached to R820T )
```
rtl_433 -s 2048000 -R 0 -Y autolevel -Y squelch -g 20 -S unknown
```

Analysis output:
```
# rtl_433 -s 2048000 -R 0 -A button_11_g003_433.92M_2048k.cu8

rtl_433 version 22.11-9-g6e1f1202 branch master at 202211291952 inputs file rtl_tcp RTL-SDR
Use -h for usage help and see https://triq.org/ for documentation.
Trying conf file at "rtl_433.conf"...
Trying conf file at "/usr/local/etc/rtl_433/rtl_433.conf"...
Trying conf file at "/etc/rtl_433/rtl_433.conf"...
Disabling all device decoders.
Registered 0 out of 223 device decoding protocols [ ]
Test mode active. Reading samples from file: button_11_g003_433.92M_2048k.cu8
Detected OOK package	@0.738655s
Analyzing pulses...
Total count:    3,  width: 8.11 ms		(16604 S)
Pulse width distribution:
 [ 0] count:    3,  width: 1313 us [1308;1321]	(2690 S)
Gap width distribution:
 [ 0] count:    2,  width: 2083 us [2083;2083]	(4266 S)
Pulse period distribution:
 [ 0] count:    2,  width: 3399 us [3395;3404]	(6962 S)
Pulse timing distribution:
 [ 0] count:    3,  width: 1313 us [1308;1321]	(2690 S)
 [ 1] count:    2,  width: 2083 us [2083;2083]	(4266 S)
[...]

PR #3098, opened 2024-11-21
Add Quinetic Switches and Sensors

pazzernick:
Relates to #2267

Checksumming added: ID and Action codes are now consistent

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [265]  Rosstech Digital Control Unit DCU-706/Sundance/Jacuzzi
     [266]  Risco 2 Way Agility protocol, Risco PIR/PET Sensor RWX95P
     [267]  ThermoPro Meat Thermometers, TP828B 2 probes with Temp, BBQ Target LO and HI
+    [268]  Quinetic Switches and Sensors
 
 * Disabled by default, use -R n or a conf file to enable
 
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
   [-d <RTL-SDR USB device index>] (default: 0)
   [-d :<RTL-SDR USB device serial (can be set with rtl_eeprom -s)>]
 	To set gain for RTL-SDR use -g <gain> to set an overall gain in dB.
-	SoapySDR device driver is not available.
+	SoapySDR device driver is available.
   [-d ""] Open default SoapySDR device
   [-d driver=rtlsdr] Open e.g. specific SoapySDR device
 	To set gain for SoapySDR use -g ELEM=val,ELEM=val,... e.g. -g LNA=20,TIA=8,PGA=2 (for LimeSDR).
--- a/README.md
+++ b/README.md
@@ -353,7 +353,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [265]  Rosstech Digital Control Unit DCU-706/Sundance/Jacuzzi
     [266]  Risco 2 Way Agility protocol, Risco PIR/PET Sensor RWX95P
     [267]  ThermoPro Meat Thermometers, TP828B 2 probes with Temp, BBQ Target LO and HI
-    [268]  Quinetic Switches and Sensors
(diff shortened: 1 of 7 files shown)


zuckschwerdt:
Thanks! Can you run `./maintainer_update.py`, the protocol in rtl_433.example.conf should be commented out?

Please format the quinetic_switch_decode() doc-comment as valid markdown. First line (and first line in file) should be `Quinetic Switches and Sensors.` with a dot at the end. Use `## Frame Layout` not divider lines, use dashes for lists, indent (only) `...PPPP SS IISCC`, see other files.

Generally just check a checksum and do not print it, it's not data.

Use `int id = (payload[0] << 8) | payload[1];` and then `"id", "ID", DATA_FORMAT, "%04x, DATA_INT, id,`

Use `int action = payload[2];` and then refer to only that var.

Use `b` as name instead of `payload`, like we do in almost all decoders.

Use `char *action_str` and static assignment instead of copy.

The names `action` and `action_str` should something with `button`, see other decoders.

Don't indent inside parens like this `( strcmp(` but use `(strcmp(`

zuckschwerdt:
Looking good.

Parens go around the bytes, i.e. it needs to be `int id = (b[0] << 8) | (b[1]);` (parens around b[1] optional).

Is `if (button_code < 192) { button_action = "press"; }` correct? The comment says 128 is release?
I think you want `if ((button_code & 0x80) == 0) {` -- this tests if the top bit (128) is unset.

Checksum might just work, depending on byte order, try:
```
    int crc = crc16(b, 5, 0x1021, 0x1D0F);
    if (crc != 0) {
        return DECODE_FAIL_MIC;
    }
```

Otherwise checksum can be compared as int, i.e.
```
    int crc = crc16(b, 3, 0x1021, 0x1D0F);
    int chk = (b[3] << 8) | b[4];
    if (crc != chk) {
        return DECODE_FAIL_MIC;
    }

```

pazzernick:
Thanks for feedback.

* rows representing "press" have button_code=[01,02,03] representing button number on switch
* rows representing "release" have button_code=192 ( but no way to indentify button number on switch )
* use of value 128 was in error

I'm thinking that it's best to have a field named **button** to help with MQTT + HASS, with the **button_code** int value.

Equally, it might be best to discard the "release" codes, and just emit the "press" codes which seem to identify button number.

Would it make sense to use a "channel" K/V to identify the button within switch? e.g. 1, 2, 3

zuckschwerdt:
We usually use "button" as DATA_INT with 1=pressed, 0=not-pressed.
Button decoders are more likely to use `unit` then `channel` and `state` (proove). or  `command` (newkaku) for on/off.
Generaly `event` would be the raw event code (what you have as `button_code`).
So perhaps button_code -> `event` and then also on press button_code -> `unit`, `command` = 1, and on release just `command` = 0. Or maybe `button` instead of `command`. What do you think?

Can combined pressed be transmitted, e.g. 1+2? I know that EnOcean kinetic switches support this (hold one button down and press another). We would need an output to support that then.

Someone would need to go over all our button-like decoders, capture the current practice and recommend a universal schema (with hopefully little changes needed to get all decoders in line).

pazzernick:
Thanks for your clarification and suggestions.

* I can't see any way to identify the unique button "release" so just discarding those packets for a predicatable outcome.
* Need to order some multi-button switches to test further.

Short of getting into "schema-like" thoughts, I've chosen to employ the following fields for Quinetic Switch:

* id
* channel (button integer)

One notable benefit of doing that is that **rtl_433_mqtt_hass.py** will add HASS entities for each button within a switch.

That means no other tweaks or adaptations, and the **channel** represents the known button number.

zuckschwerdt:
A reasonable and minimal approach. We can refine the output later, if we find a common schema for buttons.
Btw. the support for channel (and button) as automation triggers in HASS was added with #2102 about 2 years ago. We didn't get around to review button outputs since then (https://github.com/merbanan/rtl_433/pull/2102#issuecomment-1195581778) so this refinement might not happen too soon ;)

zuckschwerdt:
Looks good for merge. Lets wait a week for others to get a word in. Closes #2267

pazzernick:
If I rebase and put a new PR in, are we ok to merge?

zuckschwerdt:
Sorry, I'll merge now.

PR #3100, opened 2024-11-24
examples: Improve: Round battery level to display friendly format

danoh:
Current code, after fix from #1887, generates sometimes weird results. For example, for battery level 0.99 is the result  90.10000000000001, due to float conversions. 

I propose a fix that rounds the result to one decimal place, which should be fine for battery levels expressed in percentages. Rounding to whole numbers might even work well in this case.


danoh:
agreed and changed the pull request

PR #3116, opened 2024-12-15
CCITT whitening test sample

mr-sven:


merbanan:
The actual code is wrong as the buffers are not increased properly to hold the data. Right now there is a zero vector, I assume it works properly this I dont see the need for this change. Reopen explain and fix the code if that is not the case.

Issue #3117, opened 2024-12-17
Invalid attribute


OgreTransporter:
https://github.com/merbanan/rtl_433/blob/11d2c7a0a35e527948fd83edebcd58742dd7c35e/src/devices/deltadore_x3d.c#L177

The MSVC does not recognise `__attribute__((packed))`. Microsoft uses `#pragma pack` for this. (https://learn.microsoft.com/en-us/cpp/preprocessor/pack?view=msvc-170).With the attribute I get an error message when compiling:
```
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(177,32): error C2061: Syntax error: Identifier ‘deltadore_x3d_message_payload’
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(177,32): error C2059: Syntax error: ’;’
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(177,62): error C2449: ‘{’ found at file level - is the function header missing?
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(186,1): error C2059: Syntax error: ‘}’
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(219,34): warning C4013: ‘deltadore_x3d_read_le_u24’ undefined; assumption: external with return type int
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(241,5): error C2037: An undefined struct/union ‘deltadore_x3d_message_payload’ is determined on the left-hand side of ‘retry’
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(242,5): error C2037: An undefined struct/union ‘deltadore_x3d_message_payload’ is determined on the left-hand side of ‘transfer’
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(243,5): error C2037: An undefined struct/union ‘deltadore_x3d_message_payload’ is determined on the left-hand side of ‘transfer_ack’
1>E:\rtl_433\src\src\devices\deltadore_x3d.c(244,5): error C2037: An undefined struct/union ‘deltadore_x3d_message_payload’ is determined on the left-hand side of ‘target’
```

zuckschwerdt:
Thanks! We generally don't want value extraction in decoders by using structs, and it seems we are not actually doing that here, the attribute can be dropped.

PR #3118, opened 2024-12-17
Fix invalid attribute for MSVC #3117

OgreTransporter:
Replace `__attribute__((packed))` with `#pragma pack` for MSVC to fix compiler errors. Solves issue #3117.

### DIFF
--- a/src/devices/deltadore_x3d.c
+++ b/src/devices/deltadore_x3d.c
@@ -174,7 +174,12 @@ struct deltadore_x3d_message_header {
     int16_t header_check;
 };
 
+#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
I think just pragma should do. I.e. 
```
#pragma pack(push, 1)
struct deltadore_x3d_message_payload {
...
#pragma pack(pop)
```

zuckschwerdt:
I just noticed that no packing is actually needed. Change the PR to just remove the attribute please :)

OgreTransporter:
> I think just pragma should do. I.e.

Unfortunately, I don't know if other compilers can handle `#pragma pack`, so I added the distinction.

PR #3125, opened 2024-12-27
Add support for Revolt ZX-7717 power meter

zuckschwerdt:
S.a. discussion #3122

Draft:
- needs to handle other message lengths,
- could also output `apparentpower_VA`,
- key `pf` should likely be `powerfacor`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,6 +359,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [271]  Landis & Gyr Gridstream Power Meters 9.6k
     [272]  Landis & Gyr Gridstream Power Meters 19.2k
     [273]  Landis & Gyr Gridstream Power Meters 38.4k
+    [274]  Revolt ZX-7717 power meter
 
 * Disabled by default, use -R n or a conf file to enable
 
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -500,6 +500,7 @@ convert si
   protocol 271 # Landis & Gyr Gridstream Power Meters 9.6k
   protocol 272 # Landis & Gyr Gridstream Power Meters 19.2k
   protocol 273 # Landis & Gyr Gridstream Power Meters 38.4k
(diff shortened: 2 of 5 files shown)


AkaBoing:
it works,
thx Christian.
Revolt new code

AkaBoing:
output apparentpower_VA is in work,
same for pf > powerfacor.

zuckschwerdt:
We only need to decide wether `apparentpower_VA` would be good to have. It's not in the data and not clear if `I * U` here is actually real VA.
Same for `pf` or `powerfactor`, nice to have and shown on the main unit, but it's not in the data -- we need to decide common keys and something of a policy for power meter decoders.

A more interesting thing would be to collect messages of different lengths and handle those somehow. At least checksum and log the payload.

AkaBoing:
a DIEHL meter shows the same PF, both for inductive ( a 185W compressor) and a 25W halogen lamp.
V*A is not always VA, if a switching Power Unit or Inductive Unit is used,  but thats another story.
apparentpower_VA is not shown on Revolt Monitor zx7716 !
So drop both ?
Give me a hint please.
And for the other messages, i have no clue what to do with them.

zuckschwerdt:
I'm leaning towards not outputting VA or PF. We don't know if V and A are true RMS derived over some time or just some instantaneous values. A user wouldn't know that those outputs are simple (`VA=V*A` and `PF=P/VA`) guesswork.

I'll add checksumming and dumping of other messages soon.

AkaBoing:
its all about clipping, the zx-7717 HF output is far too  strong..
With a bad 866Mhz !! antenna i get a better signal- nearly no clipping.
And (im logging for 2 hours now), no further message beside 120 (112)

i dropped PF and VA

Revolt 28-01

AkaBoing:
Sorry  for IBIS callback, im testing on another PC with -R 80 , but with your code.

AkaBoing (approved):

AkaBoing (approved):

PR #3127, opened 2024-12-29
Added support for Honeywell ActivLink Door / Window Sensors

gedger:
This adds support for ActivLink Honeywell Door / Windows Sensors. This includes open/close together with tamper for when the cover is removed.


merbanan:
This looks harmless enough.

PR #3137, opened 2025-01-05
Change data_make to type-safe calls

zuckschwerdt:
A preview for the (scripted) change from the `data_make()` varargs convention to the new type-safe and compact data api.

This can later be shortened from e.g.
```
data_t *data = NULL;
data = data_str(data, "model", ...);
data = data_int(data, "id", ...);
```
to
```
data_t *data = data_new();
data_str(data, "model", ...);
data_int(data, "id", ...);
```
if we use introduce a fixed header for data_t lists.

### DIFF
--- a/src/devices/abmt.c
+++ b/src/devices/abmt.c
@@ -70,11 +70,10 @@ static int abmt_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     temp_c = (float)temp;
 
     /* clang-format off */
-    data_t *data = data_make(
-             "model",         "",            DATA_STRING, "Basics-Meat",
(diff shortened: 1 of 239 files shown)


merbanan:
@zuckschwerdt Any update on this ?

zuckschwerdt:
It still applies and works well. I want to implement this change asap. I'm undecided on the possible `data_new();` though.

A list initialization (so we don't need to account for the `data` pointer changing) needs an empty head element in the list. In most cases it will be replaced with the current time anyway, so no real problem.

Still, should we change to start any `data_t` list with an empty head element? There are real benefits like ergonomics and type safety and easy expansibility -- should I just go for it and try it out?

PR #3139, opened 2025-01-07
output_mqtt: Use $HOSTNAME if set

3v1n0:
Use the environment variable for setting the hostname in mqtt reporing if set

### DIFF
--- a/src/output_mqtt.c
+++ b/src/output_mqtt.c
@@ -511,7 +511,12 @@ struct data_output *data_output_mqtt_create(struct mg_mgr *mgr, char *param, cha
     if (!mqtt)
         FATAL_CALLOC("data_output_mqtt_create()");
 
-    gethostname(mqtt->hostname, sizeof(mqtt->hostname) - 1);
+    const char *hostname = getenv("HOSTNAME");
+    if (hostname == NULL || *hostname == '\0')
+        gethostname(mqtt->hostname, sizeof(mqtt->hostname) - 1);
+    else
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
`HOSTNAME` is a Bash variable but it's not a Bash environment variable, i. e. it's not passed in environment to child processes.

I don't understand the intention.

gdt:
Sure, that might help.    Given that the topic is configurable, I'd say no to adding env var for HOSTNAME, and suggest a comment in the code that does gethostname that it's only for when topic has not been configured.

Sorry if I'm sounding cranky but my take is that the fewest mechanisms to let people do what they need is best, and complexity will hurt many over time, even if it's mostly invisible.

zuckschwerdt:
E.g. add this to the help text:
```
If MQTT is enabled the defaults are
  base=rtl_433/<HOSTNAME>
  devices=<base>/devices[/type][/model][/subtype][/channel][/id]
  events=<base>/events
  states=<base>/states
```
But there is still the confusion that `base` is only applied to the default topics and can't be used by the user. Maybe we should add `[base]` as a valid token, maybe even `[hostname]` -- but they must not appear in decoder outputs then.

And TL;DR @3v1n0 you can use `-F "mqtt://yourbroker,base=rtl_433/$HOSTNAME"` if you really need it.

gdt:
I see.   It's just a matter of explaining.  Something like:

base can be set; the default value is `rtl_433/${HOSTNAME}`, where `${HOSTNAME}` is obtained by asking the system for the hostname.

zuckschwerdt:
Fixed with d9249e2

PR #3142, opened 2025-01-11
Add support for universal fan controller

marcelrv:
Adding support for universal (reversable) fan controller 24V incl the non standard CRC

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -282,6 +282,7 @@
     DECL(gridstream192) \
     DECL(gridstream384) \
     DECL(revolt_zx7717) \
+    DECL(universalfanctrl) \
 
     /* Add new decoders here. */
 
--- a/src/devices/universalfanctrl.c
+++ b/src/devices/universalfanctrl.c
@@ -39,19 +39,6 @@ Example:
 
 static int universalfan_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 {
-    for (uint16_t i = 0; i < bitbuffer->num_rows; i++) {
-        uint16_t l = bitbuffer->bits_per_row[i];
-        decoder_logf(decoder, 1, __func__, "bits_per_row in row [%d] (total rows: %d %d) = %d", i, bitbuffer->num_rows, bitbuffer->num_rows, l);
-        if (l >= 33) {
-            uint8_t *b    = bitbuffer->bb[i];
-            uint8_t cksum = 0xA;
-            for (int j = 0; j < 28; j += 4) {
-                cksum ^= (b[j / 8] >> (4 * (1 - (j % 8) / 4))) & 0x0F;
-            }
-            int crc_msg = (b[3] & 0x0F);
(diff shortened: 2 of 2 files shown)


zuckschwerdt:
Thanks. Btw, the "CRC" is not CRC but actually a nibble sum without carry (xor), right?

marcelrv:
> Thanks. Btw, the "CRC" is not CRC but actually a nibble sum without carry (xor), right?

:-) i directly believe you....   The formula works I know.. sum of each 4 bits xor'ed

It is not documented anywhere, I figured it out from looking to the occurring value...
As the rotating counter is only 3 bits I tested out all possible values

If it is not a true crc...Should I rename it to checksum?

zuckschwerdt:
Preferably rename `crc` to `chk`.
The formular `crc ^= (b[j / 8] >> (4 * (1 - (j % 8) / 4))) & 0x0F;` seems strange, e.g. the first 8 counts `b[0]` will be shifted, the first 4 times by 4 then 4 times by 0, right? That's a no-op with xor as each two xor's cancel out?

marcelrv:
> Preferably rename `crc` to `chk`. The formular `crc ^= (b[j / 8] >> (4 * (1 - (j % 8) / 4))) & 0x0F;` seems strange, e.g. the first 8 counts `b[0]` will be shifted, the first 4 times by 4 then 4 times by 0, right? That's a no-op with xor as each two xor's cancel out?

no, not really, the idea is that you xor eacht time 4 bits. 
As the array is in bytes, b[j / 8]  makes that we 2x select byte 0, 2x byte 1 , etc.  The (4 * (1 - (j % 8) / 4))) piece makes we either shift by 4 or we don't shift..  the final & 0F makes we always only take the 4 bits. for the xor' ing

zuckschwerdt:
Yes! I Missed the `j += 4` :)
We usually write this with a byte-wise stepping and fold the nibbles, e.g `chk ^= (b[j] >> 4) ^ (b[j] & 0xf)`, the last incomplete byte then needs a last `chk ^= (b[3] >> 4)`.
There is a utility function for this, e.g. `int sum = add_nibbles(b, 3) + (b[3] >> 4);`

zuckschwerdt:
You need to run `./maintainer_update.py` to get the build system changes. Commit those changes too.

marcelrv:
int sum = add_nibbles(b, 3) + (b[3] >> 4); did not work well, as it does not apply the xor.
I added a new function to do that, indeed in a more readable way based on the add_nibbles.

zuckschwerdt:
> int sum = add_nibbles(b, 3) + (b[3] >> 4); did not work well, as it does not apply the xor.

Oops. Confused myself there. I should have given a different example. Also `sum == chk` for xor is the same as full xor and then ` sum == 0` ;)
```
int sum = xor_bytes(b, 4); // xor message bytes, last byte also has the checksum
sum = (sum >> 4) ^ (sum & 0xf); // fold nibbles
if (sum != 0xa) …
```

The xor_nibbles is a nice addition. Remove the init, sum full bytes then fold after the loop. Again sorry for the add-with-carry mixup.

zuckschwerdt:
Merged with code style updates pushed on top.

Issue #2344, opened 2023-01-24
Implement Nice Flor-s decryption (see PR #2238)


uski:
Hi!

Feature request time!

The protocol to encrypt and decrypt Nice Flor-s encryption has been published here: https://github.com/DarkFlippers/unleashed-firmware/blob/dev/lib/subghz/protocols/nice_flor_s.c

It can be implemented in rtl_433. Currently rtl_433 does support Flor-s, but it does not decrypt the remote ID, which means that every time the remote is pressed, the ID seems random. This greatly reduces the usefulness of the decoder. Decoding the remote ID would allow using Nice Flor-s remotes completely.

The only issue is that for some reason, they have decided to obfuscate the encryption key. It is stored as an encrypted asset, decrypted through the secure enclave of the Flipper Zero. It should be possible to decrypt it by building an alternative firmware to Flipper Zero and simply doing a printf() of the decrypted bitstream, or by doing a bit more exploration of the Flipper Zero to discover the encryption key. No idea what they were thinking when doing this.

uski:
The AES is just used to obfuscate the master key within the Flipper Zero. AES is not used for the actual Flor-S algorithm.

Once the master key is deobfuscated it can be stored in plain text in rtl_433 and no encryption is required besides basic stuff like xor.

uski:
If you look at the encrypted master key, it is very short: https://github.com/DarkFlippers/unleashed-firmware/blob/dev/assets/resources/subghz/assets/nice_flor_s

The IV is obfuscated through some short assembly code visible in the firmware of Flipper Zero

I think the course of action is to find anyone with a Flipper Zero and just printf() the decrypted key. If I had a Flipper Zero I'd have done it already, but I don't unfortunately

Then it is just a matter of porting the decryption function

zuckschwerdt:
There is a working decryption in that PR I take it? Not sure why the 16-bit lookup is needed if the decryption is an xor stream though.

merbanan:
I have 2 Flipper Zeros. Post some instructions and I'll eventually get to it.

jurek4321:
@merbanan 
To read this decryption tables from flipper zero you will need to modify source to display decrypted keys.
You can see here https://github.com/DarkFlippers/unleashed-firmware/blob/dev/lib/subghz/subghz_keystore.c 
[...]

PR #2238, opened 2022-11-17
Fix Nice Flor-s decoding Serial and Code

zulea1:
Is using some kind of reverse-lookup tables (not the best way). 
If there is any person with a Flipper Zero device in hand, maybe can help me to make this code in a short way (no more using this 128kb reverse table).

PR #3159, opened 2025-01-24
Decodes the Nice FloR code and serial values

chirrindul:
It is interesting to see the unencrypted values for "code" and "serial". In this way you can see how you have cloned a new remote control and you can see how the "code" is increased on every new use of the remote control.

Credits to 
* https://github.com/TimelessNL/NiceFlor-ESPHome/tree/main/niceflors
and 
* https://github.com/Jev1337/NiceFlor-Encoder

This is my first pull request, so forgive errors.


zuckschwerdt:
S.a. #2238 and #2344

This looks good, but we are unlikely to add cryptographic routines to rtl_433. Using a script or binary to post-process outputs might be better suited.

chirrindul:
I understand your position. Thanks for your work. If you think I must delete my fork, tell me and I'll do it


________________________________

zuckschwerdt:
No harm in keeping this. And you can have your own fork and distribute it if you like.
The principle here is that rtl_433 will, best as possible, pull data bits from rf and output them as-is. There is some level of decoding but we won't e.g. derive weather data (rainfall).
This can be added as, e.g. Python, script to the [examples](https://github.com/merbanan/rtl_433/tree/master/examples).

merbanan:
Well calling this thing crypto was generous. I think we can accept the code (with some cleanup). The ikea sparsnäs code is similar. There we unscramble the data out of the box.

merbanan:
Hmm, but what is this large table needed for?

The following code has no large table like that:
https://github.com/DarkFlippers/unleashed-firmware/blob/dev/lib/subghz/protocols/nice_flor_s.c#L433

chirrindul:
I suppose using tables is the brute force way when people didn't know the formula behind.
It's cool to know the code that makes the tables. Thanks.
As you see I'm a bad programmer.

Enviado desde Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

chirrindul:
Oh, I didn't see the code. After doing it I remember it, it's part of FlipperZero.
There were also tables that are loaded from the filesystem and probably are encrypted.

Enviado desde Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

PR #3160, opened 2025-01-25
Add two new arguments for a TLS certificate and TLS key to allow TLS mutual authentication

madpilot:
examples: Improve TLS mutual authentication support in rtl_433_mqtt_hass.py by adding --cert and --key arguments


zuckschwerdt:
Thanks!

PR #2287, opened 2022-12-23
Add support for Interlogix Smoke Detector and Pet Immune SAW PIR Motion Sensor

yannicklarrivee:

Issue #1856, opened 2021-11-10
"Unknown" subtype for Interlogix smoke detector


atkulp:
I have the GE Interlogix Wireless Combo Heat/Smoke Detector (TX-6010-01-1) but it shows up as unknown for device subtype.  I'm not sure what I can provide, but here are some payloads with no alarm and cover closed:

`64012c`
`e401bc`
`0400a4`
`840034`

I'd be happy to add the device type ID to the code, but I'm not completely sure how to get it.  I'm also not completely sure of which switch to check for a detection event.  I've been watching switch1, but it didn't change when we burned something the other day.

zuckschwerdt:
The device subtype is the first nibble (character) of the ID, not the payload (raw message).
But there is no known mapping for most subtypes, we are considering to just remove subtype.
Do none of the messages show a flip in one of the five switches?

atkulp:
Please don't remove it!  It works great other than smoke.  My device ID is `21db11`, so `0x2` seems to be smoke subtype.

zuckschwerdt:
Do all your smoke detectors start with 0x2? @Sunoo reported a smoke detector in #1708, also 0x2.
I distincly remember a discussion that the ID can't be trusted though?

atkulp:
I only have one smoke detector so I can't be sure.  Across all my other devices the id/subtypes are 100% consistent though.

As for the payload, it's always OPEN|CLOSED|CLOSED|CLOSED|CLOSED.  I'll try a few more tests.  I only had that one event so I'll try to force another so I can see if switch1 really never closes.  I assumed tamper would show opening it too.  I'll report back on that after a bit more testing.

Sunoo:
For what it’s worth, I never tracked down the other 0x2 device, and I only see it occasionally. Perhaps a neighbor has an Interlogix smoke detector just on the edge of radio range or something.

zuckschwerdt:
There is test spray to trigger smoke detectors. E.g. "LOGILINK RP0011" worked for me.

atkulp:
Now you tell me!  😭
![image](https://user-images.githubusercontent.com/3195410/141371529-2f72596b-a799-4538-8a56-cfd3006dc8f2.png)

zuckschwerdt:
Where is the "This Is Fine" dog with his coffee? ;)

gdt:
Is this still an issue with git master?  Is anyone working on this?

merbanan:
This issue should already be solved in the master branch.

Issue #1774, opened 2021-08-05
problems with GE/interlogix PIR motion sensor decode


Arol450:
GE/interlogix motion sensors (PIR) send out a heartbeat once an hour. This is independent from actually having or not having detected any motion.
RTL_433's interlogix 'module' decodes this to:
{  switch1     => 0,
  switch2     => 1,
  switch3     => 0,
  switch4     => 1,
  switch5     => 1,}

When I actually have a true motion I get the very same:
{  switch1     => 0,
  switch2     => 1,
  switch3     => 0,
  switch4     => 1,
  switch5     => 1,}

Obviously this is a problem. I am expecting one of the 'switches' to be different.

I have analyzed the raw messages:
no motion = heartbeat:
"0114a4", 0000 0001 0001 0100 1010 0100
"c1152c", 1100 0001 0001 0101 0010 1100
"e115bc", 1110 0001 0001 0101 1011 1100
"61152c", 0110 0001 0001 0101 0010 1100
"a114ac", 1010 0001 0001 0100 1010 1100
"4115b4", 0100 0001 0001 0101 1011 0100
"a114ac", 1010 0001 0001 0100 1010 1100
  811434, 1000 0001 0001 0100 0011 0100
  611520, 0110 0001 0001 0101 0010 0000
  211434, 0010 0001 0001 0100 0011 0100

confirmed motion:
"e91520", 1110 1001 0001 0101 0010 0000
  091434, 0000 1001 0001 0100 0011 0100
"8914ac", 1000 1001 0001 0100 1010 1100
"49152c", 0100 1001 0001 0101 0010 1100
"c915bc", 1100 1001 0001 0101 1011 1100
"a9143c", 1010 1001 0001 0100 0011 1100

The bit we are interested here is #5 (from the left).
I am using rtl_433-21.05 (2020-05-09)
And I am using a whole bunch of different aged PIR sensors - all with the same problematic.
I DO NOT have any problems with all my other sensors, door contacts, smoke detectors, etc, so I am assuming this is specific to the motion detectors.

If somebody else encounters this problem as well and needs a quick workaround I recommend checking for bit 5 in the raw message.

zuckschwerdt:
"Bit 5" in raw_message is b[3] & 0x10 which is low_battery. Is that maybe a low battery notice from the motion sensor?
https://github.com/merbanan/rtl_433/blob/master/src/devices/interlogix.c#L181-L192

Arol450:
[...]

PR #3163, opened 2025-01-26
Recognize Interlogix smoke alarm device type

NasaGeek:
Observed on TX-6010-01-1 smoke alarm unit with ID 0x2bd130

### DIFF
--- a/src/devices/interlogix.c
+++ b/src/devices/interlogix.c
@@ -168,6 +168,7 @@ static int interlogix_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     case 0x4: device_type = "motion"; break;
     case 0x6: device_type = "heat"; break;
     case 0x9: device_type = "glass"; break; // switch1 changes from open to closed on trigger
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Can you also check https://github.com/merbanan/rtl_433/issues?q=Interlogix esp. #1708 #1774 #1856 #2287
There was some confusion about that smoke alarm type and we did not come to a solid conclusion.

NasaGeek:
Thanks for pointing me to those. I'm confident that the 0x2 ID maps to smoke (ID is printed on a label attached to the device; #1708 seems likely to be picking up some other home's smoke alarm). I have some motion detectors that seem to exhibit the same behavior as #1774, so I can try to gather some more info about that. I'll do some testing on the smoke alarm to try and understand the switch states better.

zuckschwerdt:
Let's add this as there is little harm. Owners of Interlogix please comment if there are false reports of non-smoke detectors.

PR #3174, opened 2025-01-31
send proper Manufacturer and Model strings to  hassio mqtt device discovery

fallenpegasus:
Currently setting hassio mqtt device discovery model to "Mfgr-Model" from rtl_433 data and device discovery manufacturer to "rtl_433". 

Instead actually set correct model and manufacturer strings.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -954,8 +954,13 @@ def publish_config(mqttc, topic, model, object_id, mapping, key=None):
         config["state_topic"] = topic
         config["unique_id"] = object_name
         config["name"] = readable_name
-    config["device"] = { "identifiers": [object_id], "name": object_id, "model": model, "manufacturer": "rtl_433" }
 
+    try:
+        hassio_device_mfgr, hassio_device_modl = model.split('-', 1)
+        config["device"] = { "identifiers": [object_id], "name": object_id, "model": hassio_device_modl, "manufacturer": hassio_device_mfgr }
+    except ValueError:
+        config["device"] = { "identifiers": [object_id], "name": object_id, "model": model, "manufacturer": "rtl_433" }
+  
     if args.force_update:
         config["force_update"] = "true"
 


zuckschwerdt:
The "model" key is just a short and unique identifier, it is not a good source of information. Also being too precise here will lead to confusion as we can only decode "protocols" and those are shared by sometimes very different devices and manufactureres.

fallenpegasus:
While that is true, having hassio display "manufacturer" "Acurite" and "model" "Tower" is still better than having hassio display "manufacturer" "rtl_433" and "model" "Acurite-Tower".  Its more correct and no more confusing when there is a weird decode.

zuckschwerdt:
It might somewhat work for your use-case and with the example you have but not in the general case. E.g. "BurnhardBBQ", "Elantra2012" can't be split and are not plain model names, "WS2032" can't be split and isn't some manufacturer name. "FT-004B" isn't anything recognizable when split. "Lightwave-RF", "Oil-SonicStd", "Secplus-v2" are fixed terms pertaining a protocol and confusing when split.

The "model" key is a machine-readable short unique identifier for a decoder protocol and should not be used for advanced display purposes. There is generally no structured info to be extracted.

fallenpegasus:
Lightwave-RF when presented to [examples/rtl_433_mqtt_hass.py] is "LightwaveRF-Lightwave-RF" and will split properly.

Oil-SonicStd when presented to [examples/rtl_433_mqtt_hass.py] is "Oil Ultrasonic STANDARD FSK-Oil-SonicStd"  and will split properly.

I can check the others, but I strongly suspect that the answer for all the others is also "and will split properly".

If it can't be split anyway, this patch falls back to the original behavior.

The "model" key is already being "used for advanced display purposes" by rtl_433_mqtt_hass.py to create human readable device names in hassio.  That ship has sailed.

This is entirely about [examples/rtl_433_mqtt_hass.py] which is a convenience script used *only* by hassio users. This change has no impact on any other rtl433 users, or any other rtl433code, data flows, or decoders, and makes devices look better and be more reasonable inside the hassio ui.

merbanan:
Well lets see what the hassio users think of this.

PR #3178, opened 2025-02-01
minor: Add rapid_change to PMV-107J from status flag

dfries:
This adds more documentation to the status flags.

For my car PMV-107J sensors the rapid pressure change is set for both increasing and decreasing pressure.

This fills in the status bits in the documentation.  With this change there's only one unknown status flag left.  Looking at my logs I've seen it set once, but it wasn't my tires, and there was only one reading from that id, so I have no way to track it down further.


zuckschwerdt:
Very good, thanks!

PR #3180, opened 2025-02-03
Deltadore X3D - Fixed wrong struct size

mr-sven:
Raw message data was missing 2 bytes because of missing packed attribute of struct.

### DIFF
--- a/src/devices/deltadore_x3d.c
+++ b/src/devices/deltadore_x3d.c
@@ -174,7 +174,7 @@ struct deltadore_x3d_message_header {
     int16_t header_check;
 };
 
-struct __attribute__((packed)) deltadore_x3d_message_payload {
+struct deltadore_x3d_message_payload {
     uint8_t retry;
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
I don't see `deltadore_x3d_message_payload` used as cast. And `raw_msg` uses the frame bytes, right? What did I miss?
Packing is something we really try to avoid as there are pitfalls and it's not fully portable.

mr-sven:
ahm, I used it for size calculation in `deltadore_x3d_parse_message_payload` but I can change it to return a fixed number.
An other question, I thought about declare the `deltadore_x3d_read_le_u16` functions as inline, is that ok for you?

zuckschwerdt:
Inline is okay, but not needed. It's only a hint to the compiler and with static functions compiler optimizations should inline already. You can check with e.g. `objdump -t` on the `.o`.

mr-sven:
Changed to return a fixed value

PR #3185, opened 2025-02-03
Update gridstream.c

MidwestTechnologies:
Removed commas from "location" because it will denote a new column in log.csv.  This was causing data to be shifted into the next column and was mislabeled.


zuckschwerdt:
@krvmk can you take a look? Breaking, but I guess this would be good to change.

krvmk:
Makes sense to change. I'm surprised there's no note about using commas in the output fields.

zuckschwerdt:
Thanks. Hasn't come up yet as custom text strings is not the usual output.

PR #3194, opened 2025-02-10
Add support for HG9901 moisture sensor

zuckschwerdt:
See #3189 #3190

Note that this protocol has no checksum and is default disabled, use `-R 275`

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [272]  Landis & Gyr Gridstream Power Meters 19.2k
     [273]  Landis & Gyr Gridstream Power Meters 38.4k
     [274]  Revolt ZX-7717 power meter
+    [275]* HG9901 Moisture sensor
 
 * Disabled by default, use -R n or a conf file to enable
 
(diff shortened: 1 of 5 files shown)


Tuxyso:
I would be very interested to get support for the HG9901 into the main branch. I also owned such sensor which is also sold in Germany under the brand "dr meter", see https://www.amazon.de/Dr-meter-Feuchtigkeitsmesser-Feuchtigkeitsmessger%C3%A4t-Bodenwassermonitor-Gartenarbeit/dp/B0CQL27NZL 

The sensors are interesting because they have a very strong signal and are very inexpensive.

What currently avoids to add HG9901 to the main branch?
If there are still technical issued I can provide recorded signals, but as I understood https://github.com/merbanan/rtl_433/issues/3190 correctly everything seems to be fine?!

onlymejosh:
I’ve been trying to get this branch to work whilst comparing it to the code in #3190 there are some slight variations, by the looks of it (my limited understanding) it’s just tidying. However the last in the comments works somewhat, but I can’t get this branch to log out any of the correct values.

obzerving:
I've tried out this branch on a Raspberry Pi 3B and I've gotten results for moisture and temperature that correlate to the display that comes with the sensors. Obviously, I can't correlate the light level, but the light_lux readings have been rational.

pi@rtlsdr:~ $ rtl_433 -F json -M utc -R 275
rtl_433 version 24.10-59-g1a6c32f7 branch feat-hg9901 at 202502102227 inputs fil
Use "-F log" if you want any messages, warnings, and errors in the console.
Detached kernel driver
[R82XX] PLL not locked!
{"time" : "2025-05-16 14:00:18", "model" : "HG9901", "id" : 16615, "battery_ok" : 1.000, "temperature_C" : 20, "moisture" : 61, "light_lux" : 5100}

zuckschwerdt:
Duplicated and continued in #3299.

PR #3199, opened 2025-02-14
add weather station  LaCrosse WS6262

sebmorio:
Adding a new protocol for the LaCrosse WS6262 weather station

This protocol is based on an existing model, emax.


merbanan:
Support added to master

PR #3202, opened 2025-02-16
Fix: fix a memory leak bug in function sdr_open_soapy.

mugitya03:
`dev->dev_info` points to a memory object allocated at line 927.

When `r!=0`, the field `dev_info` is not freed before freeing pointer `dev`. Thus we add a free operation.


zuckschwerdt:
Thanks! The `dev_info` should indeed be free'd on any error.

Issue #3200, opened 2025-02-15
Steelmate TPMS decoder issues


e100:
I have some new after market TPMS that sometimes are detected by the Steelmate decoder -R 59 and I'm not sure if I should tweak Steelmate decoder or make a new one because they seem so similar yet different.

What I found is that sometimes my sensors send 73 bits, sometimes only 72 bits where the decoder only decodes 72 bits.
So I modified the code to decode 72 or 73 bits:
```
--- a/src/devices/steelmate.c
+++ b/src/devices/steelmate.c
@@ -42,13 +42,23 @@ static int steelmate_callback(r_device *decoder, bitbuffer_t *bitbuffer)
         uint8_t *b = bitbuffer->bb[row];
 
         //Length must be 72 bits to be considered a valid packet
-        if (bitbuffer->bits_per_row[row] != 72)
+        if ((bitbuffer->bits_per_row[row] != 72) && (bitbuffer->bits_per_row[row] != 73))
             continue; // DECODE_ABORT_LENGTH
 
```

Ignoring the 73rd bit, if it exists, it now decodes the signals from my TPMS, the checksum passes too.  However, the values returned from rtl_433 do not match what is displayed on the display panel included with my TPMS sensors. The values are close but not the same and I would expect them to be the same if rtl_433 was decoding the signal in the same manner as the display does. While assumptions are bad I would expect the display bundled with the TPMS sensors would properly decode the data so making rtl_433 decode to match the display seems to make sense to me.

The attached zip has a recording where the sensor sent the same data, one is 73 bits, the other is 72 bits. 

```
Display   rtl_433
30 PSI    34 PSI
12C/70F   71F
27 PSI    30.5 PSI
24 PSI    27 PSI
18 PSI    20 PSI
10 PSI    11.5 PSI
7 PSI     8 PSI
```

What I noticed is the higher the pressures the more rtl_433 diverged from the display.
My assumption was that the value was not transmitted in PSI but rather kPa. I found that multiplying the value by 3.15 to get Kpa, then divide by 6.895 to get PSI the values closely matched the TPMS display between 0 and 30 psi. I do not have a way to test beyond 30 PSI yet but that needs to be done to reach any final conclusions.

If we plot the original conversion to PSI to calculating it as kPA then converting to PSI we see the same thing as my real data above, the higher in pressure we go the more the two calculations diverge.

![Image](https://github.com/user-attachments/assets/c22d6653-ef61-4b66-8542-bade7ae4490a)

[...]

PR #3209, opened 2025-02-22
Fix Steelmate pressure formula

ProfBoc75:
As per #3200 a new formula for the tire pressure is guessed and applied in this PR.
The previous formula got diverged values with TPMS display.


zuckschwerdt:
Very nice! I think pressure_psi should be removed and replaced by pressure_kpa in the output.

ProfBoc75:
> Very nice! I think pressure_psi should be removed and replaced by pressure_kpa in the output.

In that case,  it's a breaking change, right ? 
Or to provide both , but, I'm not sure that is properly managed by smart home automation clients, like HA ?

PR #3271, opened 2025-05-07
Steelmate tpms improvements

buxtronix:
Improvements to this protocol per discussions in #3200. I'm adding this to the existing pending branch.

I did some fairly thorough analysis on the bench involving a bike pump, power supply, fridge, 3d printer heated bed and a transmitter. The following changes are made:

- Temperature is degrees C + 50, it's much more accurate than degrees F. I left the unit as F though for backwards compatibility. My display hasn't got temperature, I just heated it and read the values. @e100 also figured this out.
- Based on analysis from @e100 and my own PSU tests, i updated the voltage formula to something that is very close to the real value. Incremented voltages on a PSU by 0.1v and read the transmitted value, and compared with two graph lines. This new formula almost exactly matches the line.
- When the voltage byte is 0xFF or 0xFE, the display shows alerts. Added these.
- Both short and long packets are sent. Same data, different preamble size. Now accounts for both including demodulation artifacts.
- Better preamble detection.
- Tweaked signal timings for more reliable detection.

I also have ensured rtl_433_tests are updated/passing and will send a PR for this.

### DIFF
--- a/src/devices/steelmate.c
+++ b/src/devices/steelmate.c
@@ -57,10 +57,9 @@ static int steelmate_callback(r_device *decoder, bitbuffer_t *bitbuffer)
 {
     uint8_t const preamble_pattern[] = {0x00, 0x00, 0x7f}; // inverted, raw value is 0x5a
     //Loop through each row of data
-    for (int row = 0; row < bitbuffer->num_rows; row++)
-    {
+    for (int row = 0; row < bitbuffer->num_rows; row++) {
         //Payload is inverted Manchester encoded, and reversed MSB/LSB order
-        uint8_t *b = bitbuffer->bb[row];
+        uint8_t *b       = bitbuffer->bb[row];
         unsigned row_len = bitbuffer->bits_per_row[row];
 
         //Length must be 72, 73, 208 or 209 bits to be considered a valid packet
@@ -69,9 +68,9 @@ static int steelmate_callback(r_device *decoder, bitbuffer_t *bitbuffer)
 
         //Valid preamble? (Note, the data is still wrong order at this point. Correct pre-amble: 0x00 0x00 0x01)
         unsigned bitpos = bitbuffer_search(bitbuffer, row, 0, preamble_pattern, 24);
-        if (bitpos >= row_len-64)
+        if (bitpos >= row_len - 64)
             continue; // DECODE_ABORT_EARLY
-        b = &b[bitpos/8];
+        b = &b[bitpos / 8];
 
         //Preamble
         uint8_t preamble = ~reverse8(b[2]);
@@ -84,21 +83,21 @@ static int steelmate_callback(r_device *decoder, bitbuffer_t *bitbuffer)
         uint8_t p1 = ~reverse8(b[5]);
 
         //Temperature is sent as degrees Celcius + 50.
-        uint8_t v1 = ~reverse8(b[6]);
+        uint8_t v1          = ~reverse8(b[6]);
         uint8_t tempCelcius = v1 - 50;
 
         //Battery voltage is stored as 100*(3.9v-<volt>).
-        uint8_t b1 = ~reverse8(b[7]);
-        int battery_mV = (int)3900-((double)b1*10.0f);
+        uint8_t b1     = ~reverse8(b[7]);
+        int battery_mV = (int)3900 - ((double)b1 * 10.0f);
 
         //Checksum is a sum of all the other values
-        uint8_t payload_checksum = ~reverse8(b[8]);
+        uint8_t payload_checksum    = ~reverse8(b[8]);
         uint8_t calculated_checksum = preamble + id1 + id2 + p1 + v1 + b1;
         if (payload_checksum != calculated_checksum)
             continue; // DECODE_FAIL_MIC
 
         int sensor_id      = (id1 << 8) | id2;
-        float pressure_kpa = p1 * 3.125f; // as guessed in #3200
+        float pressure_kpa = p1 * 3.125f;                         // as guessed in #3200
         float pressure_psi = pressure_kpa * (1.0f / 6.89475729f); // Keep the PSI value to not Break the decoder after new formula, #3200.
 
         char sensor_idhex[7];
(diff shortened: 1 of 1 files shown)


src/devices/steelmate.c
```diff
@@ -52,19 +55,22 @@ Bytes 2 to 9 are inverted Manchester with swapped MSB/LSB:
...
+        //Length must be 72, 73, 208 or 209 bits to be considered a valid packet
+        if (row_len != 72 && row_len != 73 && row_len != 209 && row_len != 208)
             continue; // DECODE_ABORT_LENGTH
 
         //Valid preamble? (Note, the data is still wrong order at this point. Correct pre-amble: 0x00 0x00 0x01)
-        if (b[0] != 0x00 || b[1] != 0x00 || b[2] != 0x7f)
+        unsigned bitpos = bitbuffer_search(bitbuffer, row, 0, preamble_pattern, 24);
+        if (bitpos >= row_len - 64)
             continue; // DECODE_ABORT_EARLY
+        b = &b[bitpos / 8];
```
zuckschwerdt, line 73: We should either `bitbuffer_extract_bytes()` or make sure `bitpos % 8 == 0`

src/devices/steelmate.c
```diff
@@ -76,22 +82,23 @@ static int steelmate_callback(r_device *decoder, bitbuffer_t *bitbuffer)
         //Pressure is stored as 32 * the Bar
         uint8_t p1 = ~reverse8(b[5]);
 
-        //Temperature is stored in Fahrenheit. Note that the datasheet claims operational to -40'C, but can only express values from -17.8'C
-        uint8_t tempFahrenheit = ~reverse8(b[6]);
+        //Temperature is sent as degrees Celcius + 50.
+        uint8_t v1          = ~reverse8(b[6]);
```
zuckschwerdt, line 86: These can simply be `int`s.

zuckschwerdt:
Thanks! Yes, let's keep output units for now, but after this is merged please prepare a PR with the pressure_PSI to pressure_kPa and temperature_F to temperature_C change.

src/devices/steelmate.c
buxtronix: Updated to use better bit wrangling.

src/devices/steelmate.c
buxtronix: Done

src/devices/steelmate.c
```diff
@@ -52,46 +55,47 @@ Bytes 2 to 9 are inverted Manchester with swapped MSB/LSB:
...
+        //Length must be 72, 73, 208 or 209 bits to be considered a valid packet
+        if (row_len != 72 && row_len != 73 && row_len != 209 && row_len != 208)
             continue; // DECODE_ABORT_LENGTH
 
         //Valid preamble? (Note, the data is still wrong order at this point. Correct pre-amble: 0x00 0x00 0x01)
-        if (b[0] != 0x00 || b[1] != 0x00 || b[2] != 0x7f)
+        unsigned bitpos = bitbuffer_search(bitbuffer, row, 0, preamble_pattern, 24);
+        if (bitpos > row_len - 72)
             continue; // DECODE_ABORT_EARLY
+        bitbuffer_extract_bytes(bitbuffer, row, bitpos, b, 72);
```
zuckschwerdt, line 73: Remove the `uint8_t *b = bitbuffer->bb[row];` above and add directly before extract:
`uint8_t b[9];`

src/devices/steelmate.c
```diff
@@ -52,46 +55,47 @@ Bytes 2 to 9 are inverted Manchester with swapped MSB/LSB:
...
+        uint8_t id2 = b[4];
 
         //Pressure is stored as 32 * the Bar
-        uint8_t p1 = ~reverse8(b[5]);
+        uint8_t p1 = b[5];
 
-        //Temperature is stored in Fahrenheit. Note that the datasheet claims operational to -40'C, but can only express values from -17.8'C
-        uint8_t tempFahrenheit = ~reverse8(b[6]);
+        //Temperature is sent as degrees Celcius + 50.
+        int tempCelcius = b[6] - 50;
```
zuckschwerdt, line 90: should be all snake_case, i.e. `temp_c`

src/devices/steelmate.c
```diff
@@ -52,46 +55,47 @@ Bytes 2 to 9 are inverted Manchester with swapped MSB/LSB:
...
-        uint8_t tmpbattery_mV = ~reverse8(b[7]);
-
-        //Checksum is a sum of all the other values
-        uint8_t payload_checksum = ~reverse8(b[8]);
-        uint8_t calculated_checksum = preamble + id1 + id2 + p1 + tempFahrenheit + tmpbattery_mV;
-        if (payload_checksum != calculated_checksum)
-            continue; // DECODE_FAIL_MIC
+        //Battery voltage is stored as 100*(3.9v-<volt>).
+        uint8_t b1     = b[7];
+        int battery_mV = (int)3900 - ((double)b1 * 10.0f);
```
zuckschwerdt, line 94: preferably all lower case `battery_mv`

src/devices/steelmate.c
buxtronix: done

src/devices/steelmate.c
buxtronix: done

src/devices/steelmate.c
```diff
@@ -52,46 +55,47 @@ Bytes 2 to 9 are inverted Manchester with swapped MSB/LSB:
...
-        uint8_t tmpbattery_mV = ~reverse8(b[7]);
-
-        //Checksum is a sum of all the other values
-        uint8_t payload_checksum = ~reverse8(b[8]);
-        uint8_t calculated_checksum = preamble + id1 + id2 + p1 + tempFahrenheit + tmpbattery_mV;
-        if (payload_checksum != calculated_checksum)
-            continue; // DECODE_FAIL_MIC
+        //Battery voltage is stored as 100*(3.9v-<volt>).
+        uint8_t b1     = b[7];
+        int battery_mv = (int)3900 - ((double)b1 * 10.0f);
```
zuckschwerdt, line 94: All good for merge, though I'd like to mention that this is perhaps not doing what is intended:
`(int)3900` is already int but will be promoted to double because of the right side.
`10.0f` will be promted to double because of the casted b1. 
I guess it would suffice to calc: `int battery_mv = 3900 - b1 * 10;`, right?

PR #3274, opened 2025-05-08
Change Steelmate TPMS to native units BREAKING CHANGE

buxtronix:
The device protocol uses metric units natively, so reflect these in the decoding.


zuckschwerdt:
Let this sit for ~2 weeks to allow comments. If we forget this please ping :)

buxtronix:
Hopefully this looks good now?

PR #3280, opened 2025-05-21
Add UCS vendor to LandisGyr-GS

nicolerenee:
Adding CRC value 0xba1f for United Cooperative Services in Texas


zuckschwerdt:
Thanks!

PR #3282, opened 2025-05-22
Add output file reopen on sighup

### LINKED ISSUE #3281: Improve file output resilience
I'm now successfully using `rtl_433` to log the output of my weather sensor(s) to JSONL but I would like to periodically take that file away and compress it so I have a (sorted) backup. Unfortunately, if I take away the file when the program is running, it seems like the program gets a bit confused and no longer writes any entries. I hadn't (yet) configured it as a systemd service (I probably should, to make restarting it easier) but I'd like to propose adopting one of the following approaches:

1. Provide filename templating with an ISO 8601 date placeholder, e.g. for daily rollover. That way any files older than "today" can be safely removed.
2. Create a new output file when the existing one has been removed.

zuckschwerdt:
This addresses #3281

On receiving the SIGHUP signal rtl_433 will reopen all configured outputs. This way you can rotate log files.

Unfinished because:
- [ ] the CSV output needs to write a new column header line, but only if the file is truely new
- [ ] the output reopen function should be using a DRY helper

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -186,6 +186,7 @@ typedef struct data_output {
     void (R_API_CALLCONV *print_int)(struct data_output *output, int data, char const *format);
     void (R_API_CALLCONV *output_start)(struct data_output *output, char const *const *fields, int num_fields);
     void (R_API_CALLCONV *output_print)(struct data_output *output, data_t *data);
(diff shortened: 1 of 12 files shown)


zuckschwerdt:
Let's merge to get the ability to rotate log files. We can care about CSV headers when that pops up in practice (I hope it does not).

PR #3283, opened 2025-05-23
Change dB estimate display to true RSSI

zuckschwerdt:
This change removes the upper bound of 0 dB from the RSSI reading to show when signals are clipping.
The SNR is still limited to 0 dB, i.e. `snr = min(rssi, 0 dB) - noise`.

The OOK pulse detect is slightly changed as the high estimate tracks a little higher now but that should have no impact.

We still need a dedicated clipping warning for FSK signals as the RSSI is only shown on successful decodes. (OOK can handle a clipping signal very well, FSK can't.)

### DIFF
--- a/src/pulse_detect.c
+++ b/src/pulse_detect.c
@@ -206,7 +206,7 @@ int pulse_detect_package(pulse_detect_t *pulse_detect, int16_t const *envelope_d
             int att = pulse_detect->use_mag_est ? mag_to_att(am_n) : amp_to_att(am_n);
             att_hist[att] += 1;
         }
-        int16_t ook_threshold = (s->ook_low_estimate + s->ook_high_estimate) / 2;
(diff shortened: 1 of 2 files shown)


zuckschwerdt:
I use the term "clipping" similar to audio jargon: the signal is distorting. It's a parameter of the input, we can't do anything about that.

The pulse detection had a (sensible) upper bound of 0 dB for the OOK "high level" (OOK on). That bound is still there but the output will show the true reading. It would only be an issue if noise ("low level"; OOK off) was close to 0 dB.

Regarding the SNR, I don't feel that >0 dB is useful headroom when quantifying signal quality. I.e. to me a +1 dB signal with 10 dB noise is the same (or possibly worse) 10 dB quality as a 0 dB signal. It's not 11 dB as the upper 1 dB is something akin to noise.

gdt:
Normally, it isn't really SNR but (signal + noise + distortion) / noise, or SINAD which is (signal + noise + distortion) / (noise + distortion).

PR #3286, opened 2025-05-30
Rename steelmate.c to tpms_steelmate.c

buxtronix:
This is for consistency with other TPMS modules.


zuckschwerdt:
There is no agreed scheme for filenames. Older decoders might have less uniform names. That's not a problem though and shuffling files around might be a nuisance when e.g. looking for the full file history.

PR #3305, opened 2025-06-20
Add [honeywell.c]  support for 2GIG branded Tilt Sensor

afallows:
Add support for decoding a 2GIG-branded Tilt Sensor that uses the Honeywell protocol. 
Changes are isolated to honeywell.c
No regressions in decoding other Honeywell devices
Signal verified with real-world sensor activity
Confirmed message decoding on multiple tilt and restore events


zuckschwerdt:
Looks good, thank you!
Can you change the comment to `// 2GIG brand, also Type 0xC for Tilt Sensor` and add a commen at line 20, something like:
Also: 2Gig Tilt sensors (type 0xC).
Can you also record a sample file and add here as zip? https://triq.org/rtl_433/ANALYZE.html

PR #3309, opened 2025-06-25
Add pulse analyzer as filter to sample grabber

zuckschwerdt:
This is a preliminary experiment to use the pulse analyzer to filter out uninteresting files before saving with `-S undecoded`.

### DIFF
--- a/include/pulse_analyzer.h
+++ b/include/pulse_analyzer.h
@@ -16,6 +16,9 @@
 
 struct r_device;
 
+/// Check the statistics of a pulse data structure and return a summary.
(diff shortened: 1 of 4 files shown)


zuckschwerdt:
Notes: this currently just checks if there are any pulses but should be enhanced to calculate a score accounting for number of different symbols and number of possible data bits. Possibly allowing an argument to demand a minimum "quality" of signal.

Also the code should be DRYed to create an analyzer struct that can be queried or printed.

merbanan:
LGTM

Issue #3306, opened 2025-06-21
Support for ThermoPro TX-7B sensor


akaSurreal:
Hello!

I recently bought 3 of these ThermoPro TX-7B outdoor sensors in use with a TP260B. I would love to get these onto my Home Assistant.

Attached the capture samples for a 35 second period, so it should have caught all 3 sensors. 

The temps were around this at time of capture:
Ch1: 88.9F 10% RH
Ch2: 91.2F 10% RH
Ch3: 88.2F 12% RH

I used this command to capture:
rtl_433 -f 915M -s 1000k -S unknown -T 35

Version: 25.02 branch  at 202502191252
Using device 0: Nooelec, NESDR SMArt v5, SN: XXXXX, "Generic RTL2832U OEM"

[tx7b_915M_1000k_samples.zip](https://github.com/user-attachments/files/20849555/tx7b_915M_1000k_samples.zip)

Thanks for any help!

ProfBoc75:
@akaSurreal ,

Only the g004 contains useful data.

To get sample, please remove antenna, put the sensor around 50 cm from RTL device, and check the cu8 into pdv https://triq.org/pdv/ , keep only signals with some green/yellow information, with orange/white it's too strong, clipping, with blue not enough, too low.


From g004 file, we have classic thermopro data layout and preamble/sync word (very similar to tp82xb)

    rtl_433 -X "n=tx7b,m=FSK_PCM,s=104,l=104,r=1000" -Y minmax -Y filter=100 g004_915M_1000k.cu8 2>&1 | grep codes
    codes     : {212}eaaaaaaaaaaaaaad2552dd425202ca00daa55aabbd2d2d2d2d200

After preamble/sync word 0xd2552dd4 we have:

    rtl_433 -X "n=tx7b,m=FSK_PCM,s=104,l=104,r=1000,preamble=d2552dd4"
    {120}25202ca00daa55aabbd2d2d2d2d200

Possible Data Layout

    Byte Position    0  1  2  3  4  5  6  7  8  9 10 11 12 13 14
    Sample          25 20 2c a0 0d aa 55 aa bb d2 d2 d2 d2 d2 00
                    II UF ?? ?? ?? ?? ?? ?? CC TT TT TT TT TT TT

Where II = ID
U = Display Unit, 0x2 = F, 0x0 = C
F = Flags ?
?? to guess with Temp and Humidity  --> Need more sample using the flex command. 
CC = Checksum/CRC ?
[...]

PR #3311, opened 2025-06-28
Add support for ThermoPro TX-7B Outdoor Thermometer Hygrometer

ProfBoc75:
Related to issue #3306 Support for ThermoPro TX-7B

This decoder add support for this temp/hum outdoor sensor.

### DIFF
--- a/src/devices/thermopro_tx7b.c
+++ b/src/devices/thermopro_tx7b.c
@@ -1,0 +1,136 @@
+/** @file
+    ThermoPro TX-7B Outdoor Thermometer Hygrometer.
+
+    Copyright (C) 2025 Bruno OCTAU (\@ProfBoc75)
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "decoder.h"
+
+/**
+ThermoPro TX-7B Outdoor Thermometer Hygrometer.
+
+- Outdor Sensor with Temperature and Hymidity
+- Compatible with ThermoPro TP260B/TP280B stations.
+- Issue #3306
+- Product web page : https://buythermopro.com/product/tx7b/
+- Very similar protocol and data layout with ThermoPro TP829b
+
+Flex decoder:
+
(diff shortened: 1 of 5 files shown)


src/devices/thermopro_tx7b.c
```diff
@@ -0,0 +1,136 @@
...
+
+    if ((msg_len - offset ) < 96 ) {
+        decoder_logf(decoder, 1, __func__, "Packet too short: %d bits", msg_len);
+        return DECODE_ABORT_LENGTH;
+    }
+
+    offset += sizeof(preamble_pattern) * 8;
+    bitbuffer_extract_bytes(bitbuffer, 0, offset, b, 9 * 8);
+
+    // checksum is a Galois bit reflect and byte reflect, gen 0x98, key 0x25, final XOR 0x00
```
zuckschwerdt, line 84: Does `lfsr_digest8_reflect()` work here instead?

ProfBoc75: No, I tried already here and into other ThermoPro decoder.

I guess there is some confusion and wording issue here, between reflect and reverse at bit level and at byte level.
But may be I'm wrong into the implementation, or I missed something into bit_util and revdgst tool.

Here the bytes in the array need to be in a reversed order, then the lsfr_digest8 is working, the lsfr_digest8_reflect is not touching the order of the byte within the array b[], but the order of the bits for each byte, and like that it doesn't work.

zuckschwerdt: You are right, `lfsr_digest8_reflect()` is the Reverse-Galois (shl) LFSR with bit and byte reflect, which is symmetrical with the Forward-Galois (shr) without any reflection. I need to clean that up someday.
We are missing the fourth option, either the Forward-Galois with byte reflection or the Reverse-Galois with bit reflection. Btw. the Reverse-Galois for this gen 98 would be gen 31 (roll left one bit).

zuckschwerdt: We *do* have the variant of `lfsr_digest8()` with the bytes reversed: `lfsr_digest8_reverse()`, right?
Can you add your set of good codes for the digest to #3306, so I can possibly clean up the algorithm names later?

ProfBoc75: @zuckschwerdt : it works. I'm updating the code and add the last findings with TX pressed and Low battery too. Will commit soon the updates.

PR #3320, opened 2025-07-04
examples: De-support python 2.7

### LINKED ISSUE #3019: Formally deprecate python2.7, enabling gcing support
Right now, changes to examples/*.py are tested under relatively recent python 3.x and we are ignoring 2.7.  At least I am.
And, we say (I said):
```
Generally, python scripts should work with relatively recent Python 3.
(While they might work with 2.7, no one is paying attention to that.)
```

I hereby run up the flagpole that under examples/ we don't care about 2.7 to the point that we should
  - document that example scripts are for python3 only even harder than the current text.
  - greenlight removal of 2.7 accomodations

At this point it is hard to conceive of anyone not being able to have python3, and these are examples anyway.   (People running LTS that doesn't have python3 and not wanting to build it and prereqs themselves is in my view something we should not cater to, 4 years after python2.7 extended EOL, and "LTS considered troublesome" anyway.)

This would just be an edit to examples/README.md.

This might or might not actually be a breaking change because we don't know if any scripts work with 2.7, because... nobody cares if they do.

gdt:
Fixes #3019


zuckschwerdt:
Thanks! I'm all for this deprecation. Merge if no one speaks up. Make the Commit message e.g. `examples: Remove Python 2.7 support (#3320)` (the issue number is added by default already).

gdt:
Kind of nuts that you have to push, open PR, get the PR number, and then amend the commit and force push!
I did it though....

zuckschwerdt:
Oh sorry, if you "Squash and merge" you get to adjust the commit message and the issue number will be added there by default.

You now would need to remove one instance of the issue number ;)

PR #3340, opened 2025-07-30
Fix visonic decode even if good package is not the first in buffer

kollix:
I often receive data where the first package is corrupted/too short, but the following 2 to 5 are correct. Therefore make the code work when the first good packet is not the very first in the buffer. Also allow to decode when only 2 repeated packets are received. Since there is a checksum in the packet, it might also be enough to use the first good packet, but let's stick to 2 for now.


zuckschwerdt:
Thanks! We usually want the majority of rows to match. Otherwise a majority of possibly good rows might be ignored for a few bad but matching rows.
Let's try and see if this causes any problems for users.

Issue #3350, opened 2025-08-25
Flex decoder specs induce memory leaks


dandwyer:
If I run the following on commit fd5f714223e3bc56ccf163fce7d1db27b86765b3:

```
cmake -GNinja -B build \
    -DCMAKE_BUILD_TYPE=Debug \
    -DFORCE_COLORED_BUILD:BOOL=ON

cmake --build build -j "$(nproc)"

touch empty_sample_file.cu8

build/src/rtl_433 \
    -X n=name,m=FSK_PCM,s=10,l=10,r=10240 \
    -r empty_sample_file.cu8
```

I get 4 reported memory leaks:
```
=================================================================
==4681==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 152 byte(s) in 1 object(s) allocated from:
    #0 0x7f6c4bafd340 in calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x563245170beb in decoder_create /home/runner/work/rtl_433/rtl_433/src/decoder_util.c:21
    #2 0x5632450a8ba6 in flex_create_device /home/runner/work/rtl_433/rtl_433/src/devices/flex.c:627
    #3 0x563244f1a904 in parse_conf_option /home/runner/work/rtl_433/rtl_433/src/rtl_433.c:1210
    #4 0x563244f1568c in parse_conf_args /home/runner/work/rtl_433/rtl_433/src/rtl_433.c:905
    #5 0x563244f20255 in main /home/runner/work/rtl_433/rtl_433/src/rtl_433.c:1644
    #6 0x7f6c4a82a1c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 282c2c16e7b6600b0b22ea0c99010d2795752b5f)
    #7 0x7f6c4a82a28a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 282c2c16e7b6600b0b22ea0c99010d2795752b5f)
    #8 0x563244f06c14 in _start (/home/runner/work/rtl_433/rtl_433/build/src/rtl_433+0x3abc14) (BuildId: a88cdcfd23440e897473930dd8a7d75ce559705e)

Direct leak of 35 byte(s) in 1 object(s) allocated from:
    #0 0x7f6c4baf74e8 in strdup ../../../../src/libsanitizer/asan/asan_interceptors.cpp:578
[...]

PR #3351, opened 2025-08-25
minor: Fix flex device memory leak during destruction (#3350)

dandwyer:
This PR is meant to address: https://github.com/merbanan/rtl_433/issues/3350

1. `build: Add memory leak detection in "Build Analyze Check" GitHub Action` highlights the issue as can be seen by the workflow in my forked repo: https://github.com/dandwyer/rtl_433/actions/runs/17200738585/job/48790690802#step:9:387
2. `minor: Fix flex device memory leak during destruction` addresses the issue as can be seen here: https://github.com/dandwyer/rtl_433/actions/runs/17200928948

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,3 +1,5 @@
+# https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax
+# https://docs.github.com/en/actions/reference/runners/github-hosted-runners
 name: Build Analyze Check
 on: [push, pull_request]
 jobs:
@@ -16,25 +18,47 @@ jobs:
         run: cmake -B build
       - name: Build
         run: cmake --build build
+      - name: Test
+        run: |
+          cmake --build build --target test || \
+          { cat build/Testing/Temporary/LastTest.log && false; }
 
   build_check_job:
     runs-on: ubuntu-latest
-    name: Build with CMake
+    strategy:
+      fail-fast: false
+      matrix:
+        cmake-generator: [Unix Makefiles, Ninja]
+        cmake-build-type: [Debug, Release]
+    name: Build ${{ matrix.cmake-build-type }} ${{ matrix.cmake-generator }}
     steps:
       - uses: actions/checkout@v5
-      - name: Setup
+      - name: Apt Update
+        run: sudo apt-get update -q -y --no-install-recommends
+      - name: Install Build Tools
+        run: sudo apt-get install -q -y --no-install-recommends cmake ninja-build
+      - name: Install Libraries
+        run: sudo apt-get install -q -y --no-install-recommends libsoapysdr-dev librtlsdr-dev
+      - name: Versions
         run: |
-          sudo apt-get update -q -y
-          sudo apt-get install -q -y --no-install-recommends cmake ninja-build
-          sudo apt-get install -q -y libsoapysdr-dev librtlsdr-dev
-      - name: Configure CMake+Ninja
-        run: cmake -GNinja -B ${{ runner.workspace }}/b/ninja -DENABLE_RTLSDR=OFF -DENABLE_SOAPYSDR=OFF
-      - name: Build CMake+Ninja
-        run: cmake --build ${{ runner.workspace }}/b/ninja
-      - name: Configure CMake+UnixMakefiles
-        run: cmake -G"Unix Makefiles" -B ${{ runner.workspace }}/b/unixmakefiles -DENABLE_RTLSDR=OFF -DENABLE_SOAPYSDR=OFF
-      - name: Build CMake+UnixMakefiles
-        run: cmake --build ${{ runner.workspace }}/b/unixmakefiles
+          cmake --version
+          lsb_release -a
+          make --version
+          echo Ninja Version: $(ninja --version)
+          uname -a
+      - name: Configure
+        run: |
+          cmake \
+            -B build \
+            -DCMAKE_BUILD_TYPE=${{ matrix.cmake-build-type }} \
+            -DENABLE_RTLSDR=OFF -DENABLE_SOAPYSDR=OFF \
+            -G "${{ matrix.cmake-generator }}"
+      - name: Build
+        run: cmake --build build
+      - name: Test
+        run: |
+          cmake --build build --target test || \
+          { cat build/Testing/Temporary/LastTest.log && false; }
 
   doc_check_job:
     runs-on: ubuntu-latest
--- a/include/r_device.h
+++ b/include/r_device.h
@@ -68,7 +68,7 @@ typedef struct r_device {
     float sync_width;
     float tolerance;
     int (*decode_fn)(struct r_device *decoder, struct bitbuffer *bitbuffer);
-    struct r_device *(*create_fn)(char *args);
+    struct r_device *(*create_fn)(char const *args);
     unsigned priority; ///< Run later and only if no previous events were produced
     unsigned disabled; ///< 0: default enabled, 1: default disabled, 2: disabled, 3: disabled and hidden
     char const *const *fields; ///< List of fields this decoder produces; required for CSV output. NULL-terminated.
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -331,7 +331,8 @@
     DECL(tpms_schrader_motorcycle) \
     /* Add new decoders here. */
 
-#define DECL(name) extern r_device name;
+#define DECL(name) extern r_device const name;
+DECL(flex_decoder)
 DEVICES
 #undef DECL
 
--- a/src/devices/flex.c
+++ b/src/devices/flex.c
@@ -12,6 +12,7 @@
 #include "decoder.h"
 #include "optparse.h"
 #include "fatal.h"
+#include "rtl_433_devices.h"
 #include <stdlib.h>
 
 enum uart_mode {
@@ -657,10 +658,7 @@ static unsigned parse_uart_mode(char const *str)
     return 0;
 }
 
-// NOTE: this is declared in rtl_433.c also.
-r_device *flex_create_device(char *spec);
-
-r_device *flex_create_device(char *spec)
+static r_device *flex_create_device(char const *spec)
 {
     if (!spec || !*spec || *spec == '?' || !strncasecmp(spec, "help", strlen(spec))) {
         help();
@@ -673,15 +671,18 @@ r_device *flex_create_device(char *spec)
     struct flex_params *params = decoder_user_data(dev);
     int get_count = 0;
 
-    spec = strdup(spec);
-    if (!spec)
+    char * mutable_spec = strdup(spec); // spec will be mutated by getkwargs()
+    if (!mutable_spec)
         FATAL_STRDUP("flex_create_device()");
+    spec = mutable_spec;
 
+    // Copy const fields from static global struct
+    dev->create_fn = flex_create_device;
     dev->decode_fn = flex_callback;
     dev->fields = output_fields;
 
     char *key, *val;
-    while (getkwargs(&spec, &key, &val)) {
+    while (getkwargs(&mutable_spec, &key, &val)) {
         key = remove_ws(key);
         val = trim_ws(val);
 
@@ -858,6 +859,13 @@ r_device *flex_create_device(char *spec)
                 params->min_rows, params->min_bits, params->min_repeats, params->invert, params->reflect, params->match_len, params->preamble_len);
     */
 
-    free(spec);
+    free((void*)spec);
     return dev;
 }
+
+r_device const flex_decoder = {
+        .name        = "General purpose decoder",
+        .create_fn   = &flex_create_device,
+        .decode_fn   = &flex_callback,
+        .fields      = output_fields,
+};
--- a/src/devices/flex.c
+++ b/src/devices/flex.c
@@ -65,9 +65,12 @@ static unsigned long extract_number(uint8_t *data, unsigned bit_offset, unsigned
     return val;
 }
 
+#define FLEX_GET_STR_LEN 32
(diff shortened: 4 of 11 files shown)


include/r_device.h
```diff
@@ -68,7 +68,8 @@ typedef struct r_device {
     float sync_width;
     float tolerance;
     int (*decode_fn)(struct r_device *decoder, struct bitbuffer *bitbuffer);
-    struct r_device *(*create_fn)(char *args);
+    struct r_device *(*create_fn)(char const *args);
```
dandwyer, line 71: Made param `const` since only 3 decoders use this method and `flex.c` was confusingly mutating this param.

include/rtl_433_devices.h
```diff
@@ -292,7 +292,8 @@
 
     /* Add new decoders here. */
 
-#define DECL(name) extern r_device name;
+#define DECL(name) extern r_device const name;
```
dandwyer, line 334: Made `const` to avoid mutation.

include/rtl_433_devices.h
```diff
@@ -292,7 +292,8 @@
...
+DECL(flex_decoder)
```
dandwyer, line 335: I had made a `flex.h` header in previous revision, but ultimately felt like it was best to be consistent w/ other decoders to extent possible.

src/devices/flex.c
```diff
@@ -631,15 +638,19 @@ r_device *flex_create_device(char *spec)
...
 
+    // Copy const fields from static global struct
+    dev->create_fn = flex_create_device;
+    dev->destroy_fn = flex_destroy_device;
     dev->decode_fn = flex_callback;
     dev->fields = output_fields;
 
     char *key, *val;
-    while (getkwargs(&spec, &key, &val)) {
+    while (getkwargs(&mutable_spec, &key, &val)) {
```
dandwyer, line 687: Every iteration of `getkwargs` advances the `mutable_spec` pointer to the next key/value pair, which was causing the call to `free(spec);` to not be freeing the `strdup` allocation in the code on the left.

src/devices/flex.c
```diff
@@ -814,6 +825,14 @@ r_device *flex_create_device(char *spec)
...
     */
 
-    free(spec);
+    free((void*)spec);
     return dev;
 }
+
+r_device const flex_decoder = {
+        .name        = "General purpose decoder",
+        .create_fn   = &flex_create_device,
```
dandwyer, line 869: In theory, could likely have struct only have `create_fn` and let the create method populate everything else.

src/r_api.c
```diff
@@ -306,9 +309,13 @@ void register_protocol(r_cfg_t *cfg, r_device *r_dev, char *arg)
 
 void free_protocol(r_device *r_dev)
 {
-    // free(r_dev->name);
-    free(r_dev->decode_ctx);
-    free(r_dev);
+    if (r_dev->destroy_fn) {
+        r_dev->destroy_fn(r_dev);
```
dandwyer: `flex.c` is only decoder w/ `destroy_fn`, but this seemed more extensible than giving special treatment for protocol `0`.

src/r_api.c
```diff
@@ -279,6 +279,9 @@ void register_protocol(r_cfg_t *cfg, r_device *r_dev, char *arg)
     r_device *p;
     if (r_dev->create_fn) {
         p = r_dev->create_fn(arg);
+        // create_fn() doesn't know or apply the main() assigned
+        // protocol number
+        p->protocol_num = r_dev->protocol_num;
```
dandwyer: I had a different iteration where `free_protocol` was using `protocol_num` to give special treatment to `0`. Although this isn't strictly necessary for this revision, it was a point of confusion for me why protocols w/ `create_fn` were having zeroed `protocol_num` before.

tests/CMakeLists.txt
```diff
@@ -32,6 +32,12 @@ endforeach(testSrc)
 # Define integration tests
 ########################################################################
 add_test(rtl_433_help ../src/rtl_433 -h)
+add_test(flex_decoder_help ../src/rtl_433 -X)
+add_test(flex_decoder bash -c
+    "touch empty_file.cu8 && \
+    ../src/rtl_433 \
+        -X n=name,m=FSK_PCM,s=10,l=10,r=10240 \
+        -r empty_file.cu8")
```
dandwyer, line 41: This induces the memory error: https://github.com/dandwyer/rtl_433/actions/runs/17200738585/job/48790690802#step:9:387

dandwyer (commented):
Annotated for reviewers.

merbanan:
@zuckschwerdt , parts of this looks good. Maybe all of it. Comments?

zuckschwerdt:
Yes, adding `const` where possible is great and the little changes make it future-proof.
I'll post a few comments.

src/r_api.c
zuckschwerdt: We intentionally didn't assign a number to the dynamically created protocols as no two of them are truely alike. Generally a protocol number names a fixed decoder and is unique. But we could `create_fn()` the same decoder template multiple times with different args. This needs a better solution, maybe along the lines of adding a "protocol-instance" number.

src/r_api.c
zuckschwerdt: This change makes it nicely symmetric and extensible. But it feels like establishing a route we don't really want to go. We need something cleaner than random decoders creating and keeping state.

merbanan: As I posted in another issue. I think we should keep state outside of the decoders. A decoder should be able to return a state struct that has a life time of the reset limit duration. The decoder can then build upon the state and eventually release a message.

But for a real implementation there would probably other issues that I missed.

src/r_api.c
merbanan: Do you mean one number that means FLEX decoder?

dandwyer: My comment applied to all functions that use a `create_fn`, so `blueline.c`, `fineoffset.c`, and `flex.c`. My addition was meant to give parity between the revised `if` and the historical `else`, such that `protocol_num` is defined in both:
```
    if (r_dev->create_fn) {
        p = r_dev->create_fn(arg);
        // create_fn() doesn't know or apply the main() assigned
        // protocol number
        p->protocol_num = r_dev->protocol_num;
    }
    else {
        ... 
        *p = *r_dev; // copy (implicitly assigns p->protocol_num)
    }
```

I can remove this if it is offensive, recalling above note that this isn't depended on by any other change in the PR, but was a point of confusion for me when investigating substance of PR.

dandwyer:
Noting that there was conflicts in flex.c and check.yml, that had accrued, I just performed a rebase to HEAD of master w/o any other changes to resolve the conflicts.

src/r_api.c
dandwyer: Crux of issue is that `flex_create_device` dynamically allocates `params->name` and `dev->name`, which are not deallocated by the `free_protocol` function. If a destructor is offensive, and there is no other specific solution proposal, the only thing I can think to do is to:
1. put a  [flexible array member](https://en.wikipedia.org/wiki/Flexible_array_member) at end of `flex_params` to accommodate the two dynamic names.
2. Alternatively, I could do 1, statically constrain the name length for simplicity
 
Let me know if either of these proposals or anything else is preferred relative to what is already provided.

dandwyer:
Last force push [fixed flex decoder `getter` related memory leaks](https://github.com/merbanan/rtl_433/compare/b64f7cff459bb6142a39b064f496a5efb62568b6..e41016f7982fa4dfee10710d1c25f001e7aa53a3) that I found scrubbing for `strdup` and `malloc` within `flex.c`.

src/r_api.c
dandwyer: Given latest findings detailed [here](https://github.com/merbanan/rtl_433/pull/3351#issuecomment-4293068106), I don't think option 1 is viable anymore. Option 2 is still viable, although I can imagine upper bounds on static length of names, getter formats, and getter mapped values could likely end up as source of contention without guidance.

zuckschwerdt: A quick look through the user contributed flex conf files shows the longest model name to be 38 chars.
I think that is excessive and can be truncated. The longer names seem to be around 24 chars.
I'd say a 32 byte fixed buffer (for 31 char names) should do (32 chars is the length we already document as maximum).
```
Jasco/GE Choice Alert Security Devices
Xmas-Tree-Remote-2APJZ-CW002
Heatmiser_PRT-W_Thermostat
Heatilator-gas-log-remote
Tesla charge port opener
Continental-Remote-code
Continental-Remote-time
ATC Technology LMT-430
GE-Smartremote-RF108
Chungear_BCF-0019x2
```
The longest getter names are 16 chars. We can cap it there or maybe provide 24 byte fixed buffers.
```
Leaking_detected
SYSTEM_ADDRESS
controller_id
TOOL_ADDRESS
rolling_code
button_code
AckLeaking
battery_ok
```
Will those 3 fixed buffers (dev->name is params->name +27) counter the need for a destructor?

dandwyer: 3rd commit of newest PR push gets rid of destructor in favor of [statically allocated device members](https://github.com/merbanan/rtl_433/pull/3351/commits/6f5b9b043de75faec28429d7d494cd56467a9ade). Preceding 2nd commit registers the flex decoder so that `free_protocol` will be called for device cleanup. 1st commit adds test cases. All are rebased to latest HEAD to remove merge conflicts.

src/r_api.c
```diff
@@ -232,7 +232,7 @@ void r_free_cfg(r_cfg_t *cfg)
 
 /* device decoder protocols */
 
-void register_protocol(r_cfg_t *cfg, r_device *r_dev, char *arg)
+void register_protocol(struct r_cfg *cfg, struct r_device const *r_dev, char *arg)
```
zuckschwerdt: Is there a reason for using the `struct` type? We do have the alias available here.

.github/workflows/check.yml
```diff
@@ -15,26 +17,48 @@ jobs:
       - name: Configure
         run: cmake -B build
       - name: Build
-        run: cmake --build build
+        run: cmake --build build -- -j $(nproc)
```
zuckschwerdt: Does this work? macOS usually doesn't have `nproc` and seems to need `sysctl -n hw.logicalcpu` or `getconf _NPROCESSORS_ONLN` -- or Ninja, which defaults to a sane cpu value.

zuckschwerdt:
I'm unsure about the use of `nproc` (on macOS) otherwise very well written and LGTM.

If you are up to renaming the commits, please change `fix: …` to `Fix …` (s.a. https://triq.org/rtl_433/CONTRIBUTING.html#commit-messages) and add ` (#3351)` to the 3rd commit so we have a reference when rebase-merging (squash merge would add the issue automatically but is not good here).

src/r_api.c
dandwyer: I was just making it consistent with the header definition [here](https://github.com/merbanan/rtl_433/pull/3351/changes#diff-ee9b5e9a39097834ad5c2b594f34844f4687e63dc58261df332b1e30f581cf81R34), but could remove the struct from both if that is preferable.

.github/workflows/check.yml
dandwyer: Good catch as there is a warning (that does not block compilation) thrown for macos-14 and 15 via GitHub action [here](https://github.com/dandwyer/rtl_433/actions/runs/28557120460/job/84666817429). 
> `/Users/runner/work/_temp/599a5776-2448-46f2-8dc0-3a1e2f2cf254.sh: line 1: nproc: command not found`

I'll work on reverting this portion.

src/r_api.c
zuckschwerdt: It's needed in r_api.h because the struct is only forward declared at that point (we don't want to pull internals into those headers). Usually we use the aliases.

dandwyer:
> I'm unsure about the use of `nproc` (on macOS) otherwise very well written and LGTM.
> 
> If you are up to renaming the commits, please change `fix: …` to `Fix …` (s.a. https://triq.org/rtl_433/CONTRIBUTING.html#commit-messages) and add ` (#3351)` to the 3rd commit so we have a reference when rebase-merging (squash merge would add the issue automatically but is not good here).

Will address commit verbage tomorrow at same time I address `nproc` issue. For reference, I missed above subtlety and was following [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) guidance.

zuckschwerdt:
> was following [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) guidance.

Good practice. Things like git-cliff are really nice. At least we have a clean commit graph and some kind of standard. We might switch someday :)

src/r_api.c
dandwyer: Understood. Will revert this line (excepting the `const` declaration for `r_dev`).

dandwyer: Resolved in latest push.

.github/workflows/check.yml
dandwyer: Reverted this line in latest push.

dandwyer:
> If you are up to renaming the commits, please change `fix: …` to `Fix …` (s.a. https://triq.org/rtl_433/CONTRIBUTING.html#commit-messages) and add ` (#3351)` to the 3rd commit so we have a reference when rebase-merging (squash merge would add the issue automatically but is not good here).

Changed commit language in latest push, noting that I put "(closes #3350)" on 3rd commit given there was an associated issue. I can change this to "(#3351)" and/or annotate all 3 commits if that is favorable.

I have made an attempt to address all feedback herein, so this push is a candidate for merge.

src/r_api.c
dandwyer: Not probable it is ever needed, but original solution w/ destructor is on a branch [here](https://github.com/dandwyer/rtl_433/tree/refs/heads/issue_3350_retain_dynamic) if static length strings ever prove non-extensible.

PR #3360, opened 2025-09-20
Add Baldr Thermo-Hygrometer protocol

smaeul:
Corresponding tests PR: https://github.com/merbanan/rtl_433_tests/pull/483

There's no checksum, but the data is repeated 8 times. I don't know if that is considered sufficient.

The first and last byte are always constant for a given device (in my limited sample set, at least), so I concatenated them to form the device ID.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -289,6 +289,7 @@
     DECL(nexus_sauna) \
     DECL(homelead_hg9901) \
     DECL(maverick_xr50) \
+    DECL(baldr_therm) \
 
     /* Add new decoders here. */
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -64,6 +64,7 @@ add_library(r_433 STATIC
     devices/auriol_hg02832.c
     devices/badger_water.c
     devices/baldr_rain.c
+    devices/baldr_therm.c
     devices/blueline.c
     devices/blyss.c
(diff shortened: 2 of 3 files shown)


zuckschwerdt:
Thanks, this looks great! We just need to figure out if there might be regular collisions since there is no checksum.

merbanan:
I have been pondering if we should allow repeats as a MIC class. In this bitstream we need 4 bits of ones and 16 bits of 0 (and some more). If you then would require 6 of 8 rows to be the same and the pattern to match we can be pretty sure this message is valid. We have CRC-4 as MIC in one case. 6 repeats should be worth more then that.

gdt:
8 transmission and look for 6 that match is a valid but primitve form of integrity, and I think we should alllow it if somone(tm) writes the code.    But if it is a form of MIC, then the decoder should output exactly one frame, and not output any without enough of a match.   Unless there's some no-MIC argument, which also ouputs frames with bad CRCs.

zuckschwerdt:
I think the important part here is not the repeats but the "need 4 bits of ones and 16 bits of 0" in fixed places. There might very well be other repeated PPM 1000/2000 µs messages that fit the length requirement.

merbanan:
Yes, repeats of a unique pattern of alternating bits.

zuckschwerdt:
Let's merge and see if it produces false positives for users. We might want to reduce the required rows to 6 or so later on.

PR #3361, opened 2025-09-20
Add Interlogix sensor types

TechnicalLee:
Added sensor descriptors for glassguard, smoke, freeze, and panic types.


TechnicalLee:
Also see #3163, add smoke type has been requested several times now without being merged.

zuckschwerdt:
Let's add this as there is little harm. We don't want to change the existing "glass" break type though. The newer Shatterpro also maps to "glass" now. I hope that's ok.

Owners of Interlogix please comment if there are false reports of detector types.

Issue #3372, opened 2025-10-04
Add decoder for Technoline Cost Control RC / Voltcraft Energy Count 3000 (ec3k) energy meter


nospam2000:
I already implemented and slightly tested the decoding, you can find it here: https://github.com/nospam2000/rtl_433/tree/add-ec3-decoder
When everthing is finished, I will raise a merge request.

I thought I would need access to `pulse_data_t *pulses` to manually imlement a custom slicing, that why I added this as a parameter to the `decode` callback of all decoders and places where this callback signature is used. I will remove this in a final version.

There is also some test code and commented code that needs to be removed.

If anyone wants to test the current version or review the file `src/devices/ec3k.c` that would be highly welcome.
I can collect some testdata to test the decoding without the device.

Here the output from the command `rtl_433 -f 868000k -s 1000k -d :00000002 -v -F json:rtl868.log -M time:iso:utc:usec -M protocol -M level -M stats -M bits`:

```jsonl
{"time" : "2025-10-04T16:30:48.015280", "protocol" : 282, "model" : "Voltcraft Energy Count 3000", "id" : 48027, "power" : 51.100, "energy" : 758.149, "energy2" : 758.149, "mic" : "CRC", "time_total" : 64942080, "time_on" : 57786240, "power_max" : 186.500, "reset_counter" : 4, "flags" : 8, "mod" : "FSK", "freq1" : 868.183, "freq2" : 868.263, "rssi" : 1.264, "snr" : 25.068, "noise" : -25.068}
{"time" : "2025-10-04T16:30:50.887781", "protocol" : 282, "model" : "Voltcraft Energy Count 3000", "id" : 47858, "power" : 65.300, "energy" : 1022.611, "energy2" : 1022.611, "mic" : "CRC", "time_total" : 64942080, "time_on" : 64942080, "power_max" : 205.800, "reset_counter" : 4, "flags" : 8, "mod" : "FSK", "freq1" : 868.199, "freq2" : 868.279, "rssi" : 0.620, "snr" : 24.362, "noise" : -24.363}
[...]

PR #3373, opened 2025-10-04
Feat ec3k

nospam2000:
add new decoder for Voltcraft Energy Count 3000 (ec3k, sold by Conrad) / Technoline Cost Control RT-110

see also issue #3372

### DIFF
--- a/src/devices/ec3k.c
+++ b/src/devices/ec3k.c
@@ -1,0 +1,368 @@
+/** @file
+    Decoder for Voltcraft Energy Count 3000 (ec3k, sold by Conrad) / Cost Control RT-110 , tested with:
+      - Technoline Cost Control RT-110 (https://www.technotrade.de/produkt/technoline-cost-control-rt-110/), EAN 4029665006208
+
+    Should also work with these devices from other manufacturers that use the same protocol or are even identical:
+      - Voltcraft ENERGYCOUNT 3000 ENERGY LOGGER (Item No. 12 53 53, https://conrad-rus.ru/images/stories/virtuemart/media/125353-an-01-ml-TCRAFT_ENERGYC_3000_ENER_MESSG_de_en_nl.pdf)
+        (don't mix that up with similar products from the same company like "Energy Check 3000", "Energy Monitor 3000" and "Energy Logger 4000")
+      - Velleman (type NETBESEM4)
+      - La Crosse Techology Remote Cost Control Monitor” (type RS3620)
+
+    Copyright (C) 2025 Michael Dreher <michael(a)5dot1.de> @nospam2000
+
+    decoding info taken from these projects:
+      - https://github.com/EmbedME/ec3k_decoder (using rtl_fm)
+      - https://github.com/avian2/ec3k (using python and gnuradio)
+      
+    Some more info can be found here:
+      - https://www.sevenwatt.com/main/rfm69-energy-count-3000-elv-cost-control/
+      - https://batilanblog.wordpress.com/2015/01/11/getting-data-from-voltcraft-energy-count-3000-on-your-computer/
+      - https://web.archive.org/web/20121019130917/http://forum.jeelabs.net:80/comment/4020
+
+    The device transmits every 5 seconds (if there is a change in power consumption) or every 30 minutes (if there is no change).
+    It uses BFSK modulation with two frequencies between 30 and 80 kHz apart (e.g. 868.297 and 868.336 MHz).
+    The bit rate is 20 kbit/s, so bit time is 50 us.
+
+    The used chip is probably a AX5042 from On Semiconductor (formerly from Axsem), datasheet: https://www.onsemi.com/download/data-sheet/pdf/ax5042-d.pdf
+    HDLC mode follows High−Level Data Link Control (HDLC, ISO 13239) protocol. HDLC Mode is the main framing mode of the AX5042.
+    HDLC packets are delimited with flag sequences of content 0x7E. In AX5042 the meaning of address and control is user defined.
+    The Frame Check Sequence (FCS) can be programmed to be CRC−CCITT, CRC−16 or CRC−32.
+    The CRC is appended to the received data. There could be an optional flag byte after the CRC.
+
+    The packet is NRZI encoded, with bit stuffing (a 0 is inserted after 5 consecutive 1 bits).
+    The packet is framed by 0x7E (01111110) bytes at start and end.
+    The packet length is 41 bytes (328 bits) excluding the two framing bytes.
+
+    The following fields are decoded:
+        id -- 16-bit ID of the device
+        time_total -- time in seconds since last reset
+        time_on -- time in seconds since last reset with non-zero device power
+        energy -- total energy in Ws (watt-seconds)
+        power_current -- current device power in watts
+        power_max -- maximum device power in watts (reset at unknown intervals)
+        reset_counter -- total number of transmitter resets
+        device_on_flag -- true if device is currently drawing non-zero power
+        crc
+
+    The CRC is calculated over the packet excluding the leading and trailing framing byte 0x7E and the crc-value itself.
+    The CRC bytes in the packet are in little-endian order (low byte first). I didn't find the parameters
+    for a standard implementation, so I took the implementation from the python code at https://github.com/avian2/ec3k.
+
+    Decoding works good with this params:
+        rtl_433 -f 868300k -s 250k
+        rtl_433 -f 868000k -s 1000k
+        rtl_433 -f 868200k -s 1000k
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+ */
+ 
+#include "decoder.h"
+#include "r_device.h"
+#include "rtl_433.h"
+#include "bitbuffer.h"
+#include <stdint.h>
+
+// --- Configuration ---
+#define DECODED_PAKET_LEN_BYTES (41)
+static const uint32_t BITTIME_US = 50;
+
+// values for 200 kHz sample rate, need to be adapted to actual sample rate
+#define PAKET_MIN_BITS (90)
+#define PAKET_MAX_BITS (PAKET_MIN_BITS * 5 / 2) // NRZ encoding, stuffing and noise
+
+static int ec3_decode_row(r_device *const decoder, const bitrow_t row, const uint16_t row_bits, const pulse_data_t *pulses);
+static int ec3k_decode(r_device *decoder, bitbuffer_t *bitbuffer, const pulse_data_t *pulses);
+static uint16_t calc_ec3k_crc(uint8_t *buffer, size_t len);
+static uint16_t update_ec3k_crc(uint16_t crc, uint8_t ch);
+
+static inline uint8_t bit_at(const uint8_t *bytes, int32_t bit)
+{
+    return (uint8_t)(bytes[bit >> 3] >> (7 - (bit & 7)) & 1);
+}
+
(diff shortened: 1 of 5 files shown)


nospam2000:
I should probably change the model name to the more common Technoline model name.
The units of time datafields are unknown, I need to figure them out.

zuckschwerdt:
Some early comments: Split the doc-comment, the `@file` only contains a copyright, the `@fn` has decoder info.
You should not need any other include but `#include "decoder.h"`

nospam2000:
@zuckschwerdt : thanks for your feedback! I already fixed some of the issues but I didn't understand what the error "Error: STICKY-PAREN error" in the workflow approval pipeline means

zuckschwerdt:
> "Error: STICKY-PAREN error" in the workflow approval pipeline means

that's e.g. `if (...){` vs. `if (...) {` -- notice the space before `{`

The doc-comments need the `@` escaped, e.g. `\@nospam2000`

nospam2000:
I cannot find any `STICKY-PAREN error` like the one you mentioned. Line numbers or the statements themselve in the error message would really help to find them.

 I use a space after parentheses intentionally for such statements:

- `(unpack_nibbles(packetbuffer, 80, 2) << 8))`
- type casts like `(uint16_t)ch`

There was one formatting error which I fixed. I also used clang-formatter now to reformat.

zuckschwerdt:
When you go to the "Files changed" or "Commits" tab you'll see the lines annotated, e.g. https://github.com/merbanan/rtl_433/pull/3373/commits/5992bb5c91bcab9af09374a290049d175e84fb90#diff-44ab634b6da1a54a76b396e7607ededb1ad9b0c5bfbc2a6c8b008a125e57274eL176
...and I was wrong, it's sticking to the left, i.e. `if(` vs `if (`

zuckschwerdt:
The "symbol error" is because the nesting of functions is too much for the symbol extractor to handle.
Best to unwrap the `data_make()` call to be in the `decode_fn`, i.e. `ec3k_decode()`. And return directly if possible, no need to wait for `rc` to make it to the bottom.

The "CRC" isn't one, is it? It looks more like a LFSR?

E.g.
```c
static int ec3k_decode(r_device *decoder, bitbuffer_t *bitbuffer)
{
    if (bitbuffer->num_rows != 1 || bitbuffer->bits_per_row[0] < PAKET_MIN_BITS) {
        decoder_logf(decoder, 3, __func__, "bit_per_row %u out of range", bitbuffer->bits_per_row[0]);
        return DECODE_ABORT_LENGTH; // Unrecognized data
    }

   // body of ec3_decode_row()...
}
```

nospam2000:
I merged the decoding functions like you suggested.

What is this symbol extractor, what is it good for?

Thanks for you help, without you it would be impossible for me to get though that stuff.

My biggest issue when starting was to get the units of the parameters of the r_device, e.g. µs for long/short/tolerance. I thought tolerance would be a percentage and the length would be in bits not in µs. Is there any documentation for that? Some comments there would be really helpful especially when porting code which uses different units.

zuckschwerdt:
> What is this symbol extractor, what is it good for?

A Python script to gather decoder info, run checks and supply e.g. https://triq.org/explorer/

> My biggest issue when starting was to get the units of the parameters of the r_device,

We have https://github.com/merbanan/rtl_433/blob/master/src/devices/new_template.c -- but there could be more info on the units. PR additions if you like!

nospam2000:
I checked symbolizer.py. It wants `r_device const ec3k` but I had `const r_device ec3k` which is my preferred style because when dealing with pointers it would become `const r_device *ec3k` which is more clear for me to read to keep the same style.

I changed it to the wanted form and now it works without error.

zuckschwerdt:
> but I had `const r_device ec3k` which is my preferred style because when dealing with pointers

In C you can't consistently write types which read left to right, our style uses the fact that reading types right to left can be consistent. E.g. compare `char const * const p` and `const char * const p`, thus we use `char const *p` :)

zuckschwerdt:
> The "CRC" isn't one, is it? It looks more like a LFSR?

I had another look. Quite convoluted but expanding the xor in the loop we basically get this?
```
uint8_t ch = buffer[i];
chk = (chk & 0x0f)
    ^ (chk >> 8)
    ^ (chk << 8)
    ^ (chk << 12)
    ^ ((chk & 0xff) >> 4)
    ^ ((chk & 0xff) << 3)
    ^ ((chk & 0x0f) << 7)
    ^ (ch & 0x0f)
    ^ (ch << 3)
    ^ (ch << 8)
    ^ (ch << 12)
    ^ (ch >> 4)
    ^ ((ch & 0x0f) << 7);
```
or with always the mask:
```
chk = 0
    ^ ((chk & 0x000f) << 0)
    ^ ((chk & 0xff00) >> 8)
    ^ ((chk & 0x00ff) << 8)
    ^ ((chk & 0x000f) << 12)
    ^ ((chk & 0x00f0) >> 4)
    ^ ((chk & 0x00ff) << 3)
    ^ ((chk & 0x000f) << 7)
    ^ ((ch & 0x0f) << 0)
    ^ ((ch & 0xff) << 3)
    ^ ((ch & 0xff) << 8)
    ^ ((ch & 0x0f) << 12)
    ^ ((ch & 0xf0) >> 4)
    ^ ((ch & 0x0f) << 7);
```

Perhaps that can still be written in a more obvious way, but this at least allows us to "follow" the bits -- separately for mixing the check and inserting data.
It's something akin to a LFSR (sans the "L"), not a CRC for sure.

src/devices/ec3k.c
```diff
@@ -0,0 +1,317 @@
+/** @file
+    Decoder for Voltcraft EnergyCount 3000 (ec3k, sold by Conrad) / Technoline Cost Control RT-110 , tested with:
+      - Technoline Cost Control RT-110 (https://www.technotrade.de/produkt/technoline-cost-control-rt-110/), EAN 4029665006208
```
zuckschwerdt: Can you move all the non-copyright info to the decoder doc-comment?

src/devices/ec3k.c
```diff
@@ -0,0 +1,317 @@
...
+{
+    uint32_t val = 0;
+    for (int32_t i = 0; i < num_nibbles; i++) {
+        val = (val << 4) | ((buf[(start_nibble + i) / 2] >> ((1 - ((start_nibble + i) & 1)) * 4)) & 0x0F);
+    }
+    return val;
+}
+
+/**
+    decoding info taken from these projects:
```
zuckschwerdt: This should start with a repeat of lione 2 (Decoder for Voltcraft EnergyCount 3000) and the indent here is 0, start a the left column.

src/devices/ec3k.c
```diff
@@ -0,0 +1,317 @@
...
+
+    Check here for some example captures: https://github.com/merbanan/rtl_433_tests/tree/master/tests/ec3k/01
+ */
+static int ec3k_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    int rc = DECODE_ABORT_EARLY;
+
+    if (bitbuffer->num_rows != 1 || bitbuffer->bits_per_row[0] < PAKET_MIN_BITS) {
+        decoder_logf(decoder, 3, __func__, "bit_per_row %u out of range", bitbuffer->bits_per_row[0]);
+        rc = DECODE_ABORT_LENGTH; // Unrecognized data
```
zuckschwerdt: Can you flatten this / early return and remove the "else"?

src/devices/ec3k.c
```diff
@@ -0,0 +1,317 @@
...
+    uint16_t calculated_crc  = calc_ec3k_crc(packetbuffer, DECODED_PAKET_LEN_BYTES - 2);
+
+    // convert to common units
+    uint64_t energy_Ws       = energy_high | energy_low;
+    const double energy_kWh  = energy_Ws / (1000.0 * 3600.0); // Ws to kWh
+    const double energy2_kWh = energy2 / (1000.0 * 3600.0);   // Ws to kWh
+    uint32_t time_total      = (uint32_t)time_total_low | ((uint32_t)time_total_high << 16);
+    uint32_t time_on         = (uint32_t)time_on_low | ((uint32_t)time_on_high << 16);
+
+    if (pad_1 == 0 && pad_2 == 0 && pad_3 == 0 && pad_4 == 0) {
```
zuckschwerdt: Can you flatten this / early return? Pull up the else, e.g. `if (pad != 0 || ...) { decoder_log(); return DECODE_FAIL_SANITY;}`

src/devices/ec3k.c
```diff
@@ -0,0 +1,317 @@
...
+        if (calculated_crc == received_crc) {
+            /* clang-format off */
+            data_t *data = data_make(
+                "model",            "",              DATA_STRING, "Voltcraft-EnergyCount3000",
+                "id",               "",              DATA_FORMAT, "%04x", DATA_INT, id,
+                "power",            "Power",         DATA_DOUBLE, power_current,
+                "energy",           "Energy",        DATA_DOUBLE, energy_kWh,
+                "energy2",          "Energy 2",      DATA_DOUBLE, energy2_kWh,
+                "mic",              "Integrity",     DATA_STRING, "CRC",
```
zuckschwerdt: "mic" needs to be the last field

src/devices/ec3k.c
```diff
@@ -0,0 +1,317 @@
...
+        "reset_counter",
+        "flags",
+        "mic",
+        NULL,
+};
+
+r_device const ec3k = {
+        .name        = "Voltcraft-EnergyCount3000",
+        .modulation  = FSK_PULSE_PCM,
+        .short_width = BITTIME_US,
```
zuckschwerdt: Use 50 here and remove the BITTIME_US constant.

nospam2000: > Use 50 here and remove the BITTIME_US constant.

that really hurts me ;-) We tell our developers never to use magic constants. Hopefully none of my developers ever sees that I have done that.

I understand that source code now more like a macro language for symbolizer.py which can also be feed into a C compiler.

I need to check your feedback later, I don't know if I will have time the next days.

zuckschwerdt: > I understand that source code now more like a macro language for symbolizer.py which can also be feed into a C compiler.

Yes, with nearly 300 decoders now we need a strict and often dumb style so we have a base for information extraction and future refactoring.

nospam2000:
> > The "CRC" isn't one, is it? It looks more like a LFSR?

It is a real crc:
`calc_ec3k_crc((uint8_t *)"123456789", 9) == crc16lsb((uint8_t *)"123456789", 9, 0x8408, 0xffff) == 0x6F91`

You can find the parameters when searching for `0x6F91` here:
https://reveng.sourceforge.io/crc-catalogue/legend.htm

The polynom 0x1021 needs to be reverted to 0x8408, then it matches.

`0x6F91 ^ 0xFFFF == 0x906E` which is the value for ([SDLC](https://reveng.sourceforge.io/crc-catalogue/16.htm#crc.cat.crc-16-ibm-sdlc)) XOR 0xffff. That perfectly matches the documentation of the AX5052:
"The Frame Check Sequence (FCS) can be programmed to be CRC−CCITT, CRC−16 or CRC−32"

zuckschwerdt:
> It is a real crc: crc16lsb 0x8408, 0xffff

A CRC without a conditional for the data bits or a LUT? I learned something new, thanks! Do you know of a way to construct these?

zuckschwerdt:
Can you update to use `crc16lsb()`?
The `model` should be `Voltcraft-EC3k,` but the `.name` can be "Voltcraft EnergyCount 3000 (ec3k)".
The `data_make()` continued lines lack one indent level.

nospam2000:
From my side everything should by done now.
Tests can be done using [these example captures](https://github.com/merbanan/rtl_433_tests/tree/master/tests/ec3k/01)

Issue #3083, opened 2024-11-01
Silent Gliss SG11490 blinds remote decode help


hamchapman:
I've got an electric blinds remote that I'm trying to decode and I'm struggling to work out how to make progress.

Some links about the remote:

* https://www.sg-s.co.uk/spares-parts/electrical-curtain-track-system-parts/5600-system-parts/11490?srsltid=AfmBOopMpzI4QPV2duq-pkS7fTJx6W-xJeDVfEFnO7E0nQz_4kt-Z3mq&dsort=sort_order
* https://www.discountelectriccurtains.co.uk/accessories/5-channel-wireless-wall-switch-11490/?srsltid=AfmBOor6ifFjxo5H4jX31uemS4POdMaD2S4ITrwGdznX0UmyBXGqiDyZ

A PDF with some info: 

![11490_5_channel_wall_switch_868-915](https://github.com/user-attachments/assets/9f181819-ed6b-4af9-afd2-f1b3c8ed9f70)

I've captured some samples of pressing some different buttons and attached them as `sg-samples.zip`:

[sg-samples.zip](https://github.com/user-attachments/files/17603010/sg-samples.zip)

I captured them using the following command:

```
rtl_433 -f 868M -S unknown
```

Hopefully the filenames explain what each sample is but the summary is that there are 5 different types of sample, with 3 instances of each. The different "types" are:

* Channel 1, pressing up button
* Channel 1, pressing down button
* Channel 2, pressing up button
* Channel 2, pressing down button
* Channel 1, pressing stop button

Please let me know if you can suggest some ways for me to progress with getting to a decoder 🙏

zuckschwerdt:
It seems there is much noise and little of a sine waveform to clearly see the FSK.
(try https://triq.org/spectrogram-next/ to load a cu8 and mouse-scroll zoom in to see no real sine waves.)

Try a higher sample rate (`-s 2048k` ?) and different placement or antenna and such.
The first 20 or so bits in the decoding on https://triq.org/pdv/ should be a regular pattern like 101010101…

hamchapman:
Thanks for the suggestions. I've tried using a higher sample rate but things look pretty much the same, same with different placement. 

I'm using an RTL-SDR v4 with [this antenna](https://www.amazon.co.uk/dp/B08WJ9LRMB). Should that work as a combination to be able to get a good read on the signals? Any suggestions for other things I could try if not? Thanks!

zuckschwerdt:
[...]

PR #3374, opened 2025-10-05
Add support for Orion ME ENC Badger water meter GIF2014W-OSE

ProfBoc75:
This pr to add support for the Orion ME ENC endpoint water meter from Badger Meter inc.

Related to issue #2995

### DIFF
--- a/src/devices/badger_orion_me_enc.c
+++ b/src/devices/badger_orion_me_enc.c
@@ -1,0 +1,149 @@
+/** @file
+    Orion Me Enc Water Meter.
+
+    Copyright (C) 2025 Bruno OCTAU (\@ProfBoc75)
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+*/
+
+#include "decoder.h"
+
+/**
+Orion Me Enc Water Meter.
+
+Manufacturer: Badger Meter Inc
+
+Model / FCCID : GIF2014W-OSE
+
+- Water meter endpoint
+- Issue #2995
+- Message is encoded using IBM Whitening Algorithm
+- Other models look compatible, to be confirmed.
+
+Flex decoder:
+
+    rtl_433 -X "n=orion_me_enc,m=FSK_PCM,s=10,l=10,r=1000,preamble=aaaaec62ec62" 2>&1 | grep codes
+
+    codes: {185}eb e1 1d 9a ed 6d 4a 4d e8 71 93 3a 78 23 57 0a ae ce 2d d8 7d 3f 4e 0
+    codes: {184}eb e1 1d 9a ed 5f 99 7c e8 71 92 31 42 62 14 0a b3 95 6e d8 7d 59 7e
+
+Data layout:
+
+    Byte Position   0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25
+    Sample         eb e1 1d 9a ed 6d 4a 4d e8 71 93 3a 78 23 57 0a ae ce 2d d8 7d 3f 4e 0
+    unwhiten       14 00 00 00 00 e8 79 69 02 0b 41 03 08 b4 00 00 fa b3 00 00 10 32 f4 4
+                   LL 11 11 11 11 SS SS SS SS xx |x xx RR RR RR RR DD DD DD DD 22 CC CC TT TT TT 
+                                                 |
+                                              +--+---+
+                                              | xxLx |
+                                              +------+
+
+
+- LL: {8} Message length except CRC, mostly 0x14 = 20 bytes, to be confirmed.
+- II:{32} Fixed value, 0x00000000, could be reverse flow water counter ?
+- SS: {32} Serial Number, little-endian value
+- xx: Unknown, values look fixed and depends on the model, could be flags also, battery level too, to be guessed
+- L:{1} Leak
+- xx: other unknown values, flags, model, unit, battery low ? to be guessed.
+- RR: {32} Reading value, scale 10 gallon, little-endian value
+- DD: {32} Daily Reading Value, scale 10 gallon, little-endian value
+- FF:{8} Fixed value, always 0x10
+- CC:{16} CRC-16, poly 0x8005, init 0xFFFF, final XOR 0x0000, from previous 21 bytes.
+- TT: Trailing bits
+
+*/
+static int orion_me_enc_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    uint8_t const preamble_pattern[] = { 0xaa, 0xaa, 0xec, 0x62, 0xec, 0x62};
+
+    uint8_t b[23];
+
+    if (bitbuffer->num_rows > 1) {
+        decoder_logf(decoder, 1, __func__, "Too many rows: %d", bitbuffer->num_rows);
+        return DECODE_FAIL_SANITY;
+    }
+    int msg_len = bitbuffer->bits_per_row[0];
+
+    if (msg_len > 290) {
+        decoder_logf(decoder, 1, __func__, "Packet too long: %d bits", msg_len);
+        return DECODE_ABORT_LENGTH;
+    }
+
+    int offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, sizeof(preamble_pattern) * 8);
+
+    if (offset >= msg_len) {
+        decoder_log(decoder, 1, __func__, "Sync word not found");
+        return DECODE_ABORT_EARLY;
+    }
+
+    if ((msg_len - offset ) < 184 ) {
+        decoder_logf(decoder, 1, __func__, "Packet too short: %d bits", msg_len);
+        return DECODE_ABORT_LENGTH;
+    }
+
+    offset += sizeof(preamble_pattern) * 8;
+    bitbuffer_extract_bytes(bitbuffer, 0, offset, b, 23 * 8);
+
+    // Unwhiten the data coded with IBM Whitening Algorithm LFSR.
+    ibm_whitening(b,23);
+
+    decoder_log_bitrow(decoder, 2, __func__, b, 23 * 8, "Unwhiten MSG");
+
+    if (crc16(b, 23, 0x8005, 0xffff)) {
+        decoder_log(decoder, 1, __func__, "CRC 16 do not match");
+        return DECODE_FAIL_MIC;
+    }
+
+    decoder_log_bitrow(decoder, 2, __func__, b, 23 * 8, "MSG");
+
+    // int msg_len   = b[0]; not used
+    int id            = b[8] << 24 | b[7] << 16 | b[6] << 8 | b[5];
+    int flags_1       = b[9] << 16 | b[10] << 8 | b[11];
+    int leaking       = (b[10] & 0x20) >> 5;
+    int reading_raw   = b[15] << 24 | b[14] << 16 | b[13] << 8 | b[12];
+    int daily_raw     = b[19] << 24 | b[18] << 16 | b[17] << 8 | b[18];
+    int flags_2       = b[20];
+    float volume_gal  = reading_raw * 0.1f; // scale or decimal could defer
+    float dly_vol_gal = daily_raw * 0.1f; // scale or decimal could defer
+
+    /* clang-format off */
+    data_t *data = data_make(
+            "model",             "",                    DATA_STRING, "Orion-MEENC",
+            "id",                "",                    DATA_INT,    id,
+            "leaking",           "Leaking",             DATA_INT,    leaking,
+            "volume_gal",        "Volume-Gallon",       DATA_FORMAT, "%.1f gal",  DATA_DOUBLE, volume_gal,
+            "daily_volume_gal",  "Daily Volume-Gallon", DATA_FORMAT, "%.1f gal",  DATA_DOUBLE, dly_vol_gal,
+            "flags_1",           "Flags-1",             DATA_FORMAT, "%06x", DATA_INT, flags_1,
+            "flags_2",           "Flags-2",             DATA_FORMAT, "%02x", DATA_INT, flags_2,
+            "mic",               "Integrity",           DATA_STRING, "CRC",
+            NULL);
+    /* clang-format on */
+
+    decoder_output_data(decoder, data);
+    return 1;
+}
+
+static char const *const output_fields[] = {
+        "model",
+        "id",
+        "volume_gal",
+        "daily_volume_gal",
+        "leaking",
+        "flags_1",
+        "flags_2",
+        "mic",
+        NULL,
+};
+
+r_device const orion_me_enc = {
+        .name        = "Orion ME ENC from Badger Meter, GIF2014W-OSE, water meter, 100kbps (-f 916.7M -s 1600k)",
+        .modulation  = FSK_PULSE_PCM,
+        .short_width = 10,
+        .long_width  = 10,
+        .reset_limit = 1000,
+        .decode_fn   = &orion_me_enc_decode,
+        .fields      = output_fields,
+};
--- a/include/bit_util.h
+++ b/include/bit_util.h
@@ -191,6 +191,18 @@ uint16_t lfsr_digest16(uint8_t const message[], unsigned bytes, uint16_t gen, ui
 /// @param buffer_size number of bytes to process
 void ccitt_whitening(uint8_t *buffer, unsigned buffer_size);
 
+/// Apply IBM data whitening to a buffer.
+///
+/// The IBM data whitening process is built around a 9-bit Linear Feedback Shift Register (LFSR).
+/// CCITT data whitening processes data packets byte-per-byte, whereas IBM data
+/// whitening processes the data packet bit-per-bit
+/// Same, the initial value of the data whitening key is set to all ones, 0x1FF.
+/// s.a. https://www.nxp.com/docs/en/application-note/AN5070.pdf s.5.1
+///
+/// @param buffer bytes of message data
+/// @param buffer_size number of bytes to process
+void ibm_whitening(uint8_t *buffer, unsigned buffer_size);
+
 /// Compute bit parity of a single byte (8 bits).
 ///
 /// @param byte single byte to check
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -289,6 +289,7 @@
     DECL(nexus_sauna) \
     DECL(homelead_hg9901) \
     DECL(maverick_xr50) \
+    DECL(orion_me_enc) \
 
     /* Add new decoders here. */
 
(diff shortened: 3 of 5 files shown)


zuckschwerdt:
Maybe add a test like the ccitt_whitening test at the bottom of bit_util.c

ProfBoc75:
> Maybe add a test like the ccitt_whitening test at the bottom of bit_util.c

Good point, and I discovered a little correction where the line showing the results passed and failed just before the ccitt test , must be moved after and just before the `return failed;`

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+    if ((msg_len - offset ) < 184 ) {
+        decoder_logf(decoder, 1, __func__, "Packet too short: %d bits", msg_len);
+        return DECODE_ABORT_LENGTH;
+    }
+
+    offset += sizeof(preamble_pattern) * 8;
+    bitbuffer_extract_bytes(bitbuffer, 0, offset, b, 23 * 8);
+
+    // Unwhiten the data coded with IBM Whitening Algorithm LFSR.
+    ibm_whitening(b,23);
```
klyubin: Nit: Consider hard-coding the 23 bytes from the IBM whitening stream to avoid wasting CPU cycles for every broadcast to do all the bitshift logic to compute the keystream. This step here would then simply XOR these 23 bytes from the IBM data whitening stream with the 23 bytes from the broadcast.

The keystream is: `ffe11d9aed853324ea7ad2397097570a547d2dd86d0dba`.

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    int offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, sizeof(preamble_pattern) * 8);
+
+    if (offset >= msg_len) {
+        decoder_log(decoder, 1, __func__, "Sync word not found");
+        return DECODE_ABORT_EARLY;
+    }
+
+    if ((msg_len - offset ) < 184 ) {
```
klyubin: This is not accounting for the length of the matched preamble. I think this should be:

```suggestion
    if ((msg_len - offset - sizeof(preamble_pattern) * 8) < 184 ) {
```

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+        decoder_logf(decoder, 1, __func__, "Packet too short: %d bits", msg_len);
```
klyubin: Suggestion: Consider being more informative by reporting the length after preamble + sync instead of the full length of the packet. E.g., if the packet starts with 0xaaaaaaaaaaaaaaaaaaaaa, then then full length of the packet might not tell the full story why this error,

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+
+    offset += sizeof(preamble_pattern) * 8;
+    bitbuffer_extract_bytes(bitbuffer, 0, offset, b, 23 * 8);
+
+    // Unwhiten the data coded with IBM Whitening Algorithm LFSR.
+    ibm_whitening(b,23);
+
+    decoder_log_bitrow(decoder, 2, __func__, b, 23 * 8, "Unwhiten MSG");
+
+    if (crc16(b, 23, 0x8005, 0xffff)) {
```
klyubin: - [x]  I think there are two issues here:

* The `crc16` function is supposed to take the first 21 as input, not 23.
* The `crc16` function returns the computed CRC value (over those input bytes) and yet this code doesn't check the returned value against the CRC from the packet (at byte index 21 and 22). I'm surprised this works, because this if statement currently rejects any CRC which is not 0x0000. See https://github.com/merbanan/rtl_433/blob/7693dcc51931438a6e250dce8a04006dac7b0a04/include/bit_util.h#L138-L145.

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    decoder_log_bitrow(decoder, 2, __func__, b, 23 * 8, "MSG");
+
+    // int msg_len   = b[0]; not used
+    int id            = b[8] << 24 | b[7] << 16 | b[6] << 8 | b[5];
+    int flags_1       = b[9] << 16 | b[10] << 8 | b[11];
+    int leaking       = (b[10] & 0x20) >> 5;
+    int reading_raw   = b[15] << 24 | b[14] << 16 | b[13] << 8 | b[12];
+    int daily_raw     = b[19] << 24 | b[18] << 16 | b[17] << 8 | b[18];
```
klyubin: - [ ] `b[16]` instead of `b[18]` at the very end:

```suggestion
    int daily_raw     = b[19] << 24 | b[18] << 16 | b[17] << 8 | b[16];
```

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+    int flags_2       = b[20];
+    float volume_gal  = reading_raw * 0.1f; // scale or decimal could defer
```
klyubin: I'd suggest reporting these readings as raw values (as in, not adjusted in any way), since we don't yet know where in the packet the unit of measurement is stored. In theory, the very same ORION ME ENC could be connected to a reader which reports in full gallons, CCF, or even imperial gallons or liters.

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+    float dly_vol_gal = daily_raw * 0.1f; // scale or decimal could defer
+
+    /* clang-format off */
+    data_t *data = data_make(
+            "model",             "",                    DATA_STRING, "Orion-MEENC",
```
klyubin: Nit: Technically this works for various other ORION models. We could hard-code the mapping from serial number ranges to model names. According to https://badgermeter.widen.net/content/vodetxkyxh/original?download=false&x.app=api:
* 30000000…59999999: ME or SE
* 60000000…69999999: Mobile M
* 70000000…89999999: Classic
* 110000000…119999999: LTE
* 120000000…129999999: LTE-M or LTE-MS
* 130000000…139999999: C or CS
* 140000000…148999999: HLA
* 149000000…149999999: HLC
* 150000000…159999999: HLB
* 160000000…169999999: HLD
* 170000000…179999999: HLFX
* 180000000…189999999: HLG

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+            "id",                "",                    DATA_INT,    id,
```
klyubin: I'd suggest calling this same as in Badger's own docs: serial number or serial.

zuckschwerdt:
> the line showing the results passed and failed just before the ccitt test, must be moved

Oops. Yes!

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+            "leaking",           "Leaking",             DATA_INT,    leaking,
+            "volume_gal",        "Volume-Gallon",       DATA_FORMAT, "%.1f gal",  DATA_DOUBLE, volume_gal,
```
klyubin: Consider calling this "reading" or "current raw meter
reading" -- they way it's referred to in Badger's docs.

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+            "daily_volume_gal",  "Daily Volume-Gallon", DATA_FORMAT, "%.1f gal",  DATA_DOUBLE, dly_vol_gal,
```
klyubin: Consider calling this "daily reading"

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+        "daily_volume_gal",
+        "leaking",
+        "flags_1",
+        "flags_2",
+        "mic",
+        NULL,
+};
+
+r_device const orion_me_enc = {
+        .name        = "Orion ME ENC from Badger Meter, GIF2014W-OSE, water meter, 100kbps (-f 916.7M -s 1600k)",
```
klyubin: Should we specify all the frequencies we've seen instead of just one?

zuckschwerdt: We might need a way to capture more information and a way to query that from a decoder. Perhaps something linked online where a discussion, hints and reports can be added.

src/devices/badger_orion_me_enc.c
zuckschwerdt: A CRC checksum is the remainder after division. Adding that remainder in then gives a remainder of zero. I.e. summing a CRC and including the check bytes (depending on endianness) should results in zero :)

klyubin: Oh, wow! TIL! Thank you!

src/devices/badger_orion_me_enc.c
ProfBoc75: indeed this must be updated as discussed into the issue, the range is much more than that, I need also to add comment into the decoder about the hopping frequencies, as it's key for this sensor.

src/devices/badger_orion_me_enc.c
ProfBoc75: At rtl_433 level, we are always using id for serial number.

I saw one device with ` "id",           "Serial number",   DATA_INT, sn' all others are like I did.

src/devices/badger_orion_me_enc.c
ProfBoc75: Yes, I can, it's more simple like that. Main purpose of rtl_433 is to show rf message with less modification as possible.
Then, proceed outside for any conversion with python, HA or so on.
When the unit is clearly identified, we provide it. Here looks like several possibilities, depends on the model and/or how it's setup by the water company.

src/devices/badger_orion_me_enc.c
ProfBoc75: May be keep model Orion-MEENC or Orion-Endpoint and add the sub-model based on the serial number.

src/devices/badger_orion_me_enc.c
ProfBoc75: yes a typo

src/devices/badger_orion_me_enc.c
klyubin: Sounds good. I'm new to rtl_433 and have no idea about its code base and coding style yet. So, please feel free to disregard my suggestions.

src/devices/badger_orion_me_enc.c
ProfBoc75: in fact line 87 should be done just before 82.

src/devices/badger_orion_me_enc.c
ProfBoc75: In fact, another sensor (#3083) may requires such function, with a different key length, the reason why I took the opportunity to add the function. But agree that rtl_433 may works on low performance system, and such optimization could help.

src/devices/badger_orion_me_enc.c
ProfBoc75: We are already after the preamble/sync word is successfully found, error message could be "Packet too short, expected 184 bits, but received only : %d bits"

src/devices/badger_orion_me_enc.c
ProfBoc75: No conversion, happy with that, see below same remark.

src/devices/badger_orion_me_enc.c
merbanan: Add the mapping as it has inherent value, and add the url inside the doxygen comment section.

src/devices/badger_orion_me_enc.c
```diff
@@ -0,0 +1,149 @@
...
+        return DECODE_FAIL_MIC;
+    }
+
+    decoder_log_bitrow(decoder, 2, __func__, b, 23 * 8, "MSG");
+
+    // int msg_len   = b[0]; not used
+    int id            = b[8] << 24 | b[7] << 16 | b[6] << 8 | b[5];
+    int flags_1       = b[9] << 16 | b[10] << 8 | b[11];
+    int leaking       = (b[10] & 0x20) >> 5;
+    int reading_raw   = b[15] << 24 | b[14] << 16 | b[13] << 8 | b[12];
```
klyubin: I think this is unsigned, so, `uint32_t`. (same for `daily_raw` below).

klyubin (approved):
Looks great! Thank you!!!

src/devices/badger_orion_endpoint.c
```diff
@@ -0,0 +1,228 @@
...
+
+    uint8_t b[23];
+
+    if (bitbuffer->num_rows > 1) {
+        decoder_logf(decoder, 1, __func__, "Too many rows: %d", bitbuffer->num_rows);
+        return DECODE_FAIL_SANITY;
+    }
+    int msg_len = bitbuffer->bits_per_row[0];
+
+    if (msg_len > 290) {
```
klyubin: Optimization:  also reject messages which are too short: shorter than `sizeof(preamble) * 8 + 184`. I've seen a ton of broadcasts captured which are shorter than this, so, this will help save CPU cycles on those.

src/devices/badger_orion_endpoint.c
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    offset += sizeof(preamble_pattern) * 8;
+
+    if ((msg_len - offset ) < 184 ) {
+        decoder_logf(decoder, 1, __func__, "Expected 184 bits, Packet too short: %d bits", msg_len - offset);
+        return DECODE_ABORT_LENGTH;
+    }
+
+    bitbuffer_extract_bytes(bitbuffer, 0, offset, b, 23 * 8);
```
klyubin: Memory safety suggestion: consider using `sizeof(b) * 8` everywhere instead of 23. Or even declaring a constant with a value 23 which is then used for declaring `b`, `ibm_whiten_key`, and everywhere where 23 or 184 is used. This way it'll be harder to accidentally use the wrong length / access past the expected array boundaries.

src/devices/badger_orion_endpoint.c
ProfBoc75: Yes , valid only from 232 to 290 bits.

src/devices/badger_orion_endpoint.c
ProfBoc75: Yes. I can declare constant values.

ProfBoc75:
> This is working great on my ORION:
> 
> ```shell
> $ rtl_433 -f 912.1M -E hop -s 1600k -M level -R 282
> rtl_433 version 25.02-51-g04f3ad95 branch feat-orion_me_enc at 202510071947 inputs file rtl_tcp RTL-SDR with TLS
> 
> New defaults active, use "-Y classic -s 250k" if you need the old defaults
> 
> Found Rafael Micro R828D tuner
> RTL-SDR Blog V4 Detected
> [SDR] Using device 0: RTLSDRBlog, Blog V4, SN: 00000001, "Generic RTL2832U OEM"
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> time      : 2025-10-07 17:00:34
> model     : Orion-Endpoint                         id        : 39365338
> Endpoint Model: ME or SE Leaking   : 0             Reading   : 4464087       Daily Reading: 4462445    Flags-1   : 0b4008        Flags-2   : 10            Integrity : CRC
> Modulation: FSK          Freq1     : 912.4 MHz     Freq2     : 912.6 MHz
> RSSI      : -12.1 dB     SNR       : 16.8 dB       Noise     : -28.9 dB
> 
> $ rtl_433 -f 912.1M -E hop -s 1600k -M level -R 282 -F csv
> rtl_433 version 25.02-51-g04f3ad95 branch feat-orion_me_enc at 202510071947 inputs file rtl_tcp RTL-SDR with TLS
> 
> New defaults active, use "-Y classic -s 250k" if you need the old defaults
> 
> Use "-F log" if you want any messages, warnings, and errors in the console.
> time,msg,codes,mod,freq,freq1,freq2,rssi,snr,noise,model,id,reading,daily_reading,leaking,flags_1,flags_2,mic
> Found Rafael Micro R828D tuner
> RTL-SDR Blog V4 Detected
> 2025-10-07 17:03:00,,,FSK,,912.308,912.565,-3.096,28.634,-31.730,Orion-Endpoint,39365338,4464087,4462445,0,737288,16,CRC
> 2025-10-07 17:03:17,,,FSK,,912.351,912.538,0.791,31.730,-31.730,Orion-Endpoint,39365338,4464087,4462445,0,737288,16,CRC
> 2025-10-07 17:03:32,,,FSK,,912.394,912.561,0.675,31.730,-31.730,Orion-Endpoint,39365338,4464087,4462445,0,737288,16,CRC
> 2025-10-07 17:03:49,,,FSK,,912.368,912.562,-2.813,26.109,-28.922,Orion-Endpoint,39365338,4464087,4462445,0,737288,16,CRC
> ```

Oops, missing the endpoint model in json or csv output, because missing endpoint_model declaration in output field.

src/devices/badger_orion_endpoint.c
```diff
@@ -82,54 +91,54 @@ Data layout:
...
 
     if (bitbuffer->num_rows > 1) {
         decoder_logf(decoder, 1, __func__, "Too many rows: %d", bitbuffer->num_rows);
         return DECODE_FAIL_SANITY;
     }
     int msg_len = bitbuffer->bits_per_row[0];
 
-    if (msg_len > 290) {
-        decoder_logf(decoder, 1, __func__, "Expected less 290 bits, Packet too long: %d bits", msg_len);
+    if (msg_len < MSG_MIN_BITLEN && msg_len > MSG_MAX_BITLEN) {
```
klyubin: - [ ] This needs to be an `||` otherwise it's always `false`.
```suggestion
    if (msg_len < MSG_MIN_BITLEN || msg_len > MSG_MAX_BITLEN) {
```

src/devices/badger_orion_endpoint.c
```diff
@@ -11,6 +11,15 @@
 
 #include "decoder.h"
 
+#define PREAMBLE_BYTELEN   6
+#define DATA_BYTELEN      23
+
+#define PREAMBLE_BITLEN   48  //  6*8
+#define DATA_BITLEN      184  // 23*8
```
klyubin, line 18: Nit: Can be `(PREAMBLE_BYTELEN * 8)` (similar stuff for DATA_BITLEN) so that that they match and you then don't have to have a comment explaining why the number.

```suggestion
#define PREAMBLE_BITLEN   (PREAMBLE_BYTELEN * 8)
#define DATA_BITLEN       (DATA_BYTELEN * 8)
```

src/devices/badger_orion_endpoint.c
ProfBoc75: You're right,

src/devices/badger_orion_endpoint.c
ProfBoc75: Yes I hesitated , I checked other decoders and we avoid calculation at define level, so I put directly the values like others.

ProfBoc75:
Hi @klyubin , @zuckschwerdt : I guess we can squash and merge this version ?

PR #3381, opened 2025-10-16
New device Raddy WF-100C Lite

ChadMatsalla:
This is the Raddy WF-100C Lite Weather Station 13-in-1 Wireless Indoor Outdoor with Temperature, Barometric, Humidity, Wind Gauge, Rain Gauge, Weather Forecast, Moon Phrase, Alarm Clock.
[Raddy's Site](https://iraddy.com/products/wf-100c-lite)

Raddy WF-100C

It currently appears as `Cotech-367959` so I'll be basing my work on that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [280]  Homelead HG9901 (Geevon, Dr.Meter, Royal Gardineer) soil moisture/temp/light level sensor
     [281]  Maverick XR-50 BBQ Sensor
     [282]  Orion Endpoint from Badger Meter, GIF2014W-OSE, water meter, hopping from 904.4 Mhz to 924.6Mhz (-s 1600k)
+    [283]  Raddy WF-100C Lite
 
 * Disabled by default, use -R n or a conf file to enable
 
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -521,6 +521,7 @@ convert si
   protocol 280 # Homelead HG9901 (Geevon, Dr.Meter, Royal Gardineer) soil moisture/temp/light level sensor
   protocol 281 # Maverick XR-50 BBQ Sensor
   protocol 282 # Orion Endpoint from Badger Meter, GIF2014W-OSE, water meter, hopping from 904.4 Mhz to 924.6Mhz (-s 1600k)
+  protocol 283 # Raddy WF-100C Lite
 
 ## Flex devices (command line option "-X")
 
--- a/src/devices/raddy_wf100c.c
+++ b/src/devices/raddy_wf100c.c
@@ -41,216 +41,122 @@ Data layout:
 */
 static int raddy_wf100c_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 {
-    /*
-     * Early debugging aid to see demodulated bits in buffer and
-     * to determine if your limit settings are matched and firing
-     * this decode callback.
-     *
-     * 1. Enable with -vvv (debug decoders)
-     * 2. Delete this block when your decoder is working
-     */
-    //    decoder_log_bitbuffer(decoder, 2, __func__, bitbuffer, "");
+    uint8_t const preamble[] = {0x01, 0x40}; // 12 bits
 
-    /*
-     * If you expect the bits flipped with respect to the demod
-     * invert the whole bit buffer.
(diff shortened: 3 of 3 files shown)


zuckschwerdt:
It's great to see that you are interested in contributing to rtl_433 and any contribution is welcome. It seems you are new to this, so here are some notes:

See https://triq.org/rtl_433/CONTRIBUTING.html#adding-a-new-decoder on how to add a decoder.

People are notified of every change you push to a PR and will ignore you after so many notifications.
Try to get something working on your fork before opening a PR and before  commiting a change.

ChadMatsalla:
Thank you for the advice. `maintainer_update.py` wanted a clean tree, unfortunately, so to get it to show there I had to commit and commit. I apologize.

ChadMatsalla:
I'm afraid I've reached them limit of what I can do - I'm not a decoder.

I'm trying to find the baromatric pressure at the end of this:
`cad000001702eb844b2ffffbfb4f`. It should be 1008 hPa or around 29.7 inhg. It's after 10 but I just can't find it. I can create a test and upload a sample and some metadata for what should be in the sample if that helps.

zuckschwerdt:
> `maintainer_update.py` wanted a clean tree, [...]so I had to commit

You only need to stage, no commit needed ;)

> I'm trying to find the baromatric pressure at the end of this

Get codes with very little variation in the values then you should be able to spot the low bits changing.

But atmospheric pressure isn't a value that differs inside and outside, so it usually a sensors on the inside display unit.

ChadMatsalla:
Argh. I didn't know that. It seems that this machine is sending 16 bytes of data and that's why I assumed it was there. It seems this might have been for nothing given that I can get all of the rest of the data from this machine using `Cotech-367959`, unless you think otherwise. It's base station is sending this to wunderground and I don't see any other sensor-based information `/weatherstation/updateweatherstation.php?ID=XXXX&PASSWORD=XXXX&indoortempf=73.6&indoorhumidity=36&tempf=69.8&humidity=46&dewptf=48.1&windchillf=69.8&absbaromin=28.07&baromin=29.75&windspeedmph=0.0&windgustmph=0.0&winddir=160&windspdmph_avg2m=0.0&winddir_avg2m=162&windgustmph_10m=0.0&windgustdir_10m=104&rainin=0.0&dailyrainin=0.0&weeklyrainin=0.21&monthlyrainin=0.21&dateutc=2025-10-16%2017:41:24&action=updateraw&realtime=1&rtfreq=5&`

zuckschwerdt:
The 16 bytes is where the Cotech has 14 bytes? Where are those two located? At the end? Are they included in the checksum? If so did the checksum then move (e.g. from byte 14 to byte 16)?

Did you notice any difference in the fields between Raddy WF-100C Lite Weather Station and Cotech 36-7959 Weatherstation? If it's only minor then maybe add that to the original decoder. If they are the same then just document Raddy WF-100C Lite Weather Station as also supported in the original decoder.

merbanan:
@ChadMatsalla do you have sample signals for this device?

ChadMatsalla:
> @ChadMatsalla do you have sample signals for this device?

No, but I could probably get some. I was persuaded by the comment above that the barometric pressure is likely sampled at the bast station so there was no need for this work at all given that I can get the data from `Cotech-367959`. What's your interest?

ProfBoc75:
> > @ChadMatsalla do you have sample signals for this device?
> 
> No, but I could probably get some. I was persuaded by the comment above that the barometric pressure is likely sampled at the bast station so there was no need for this work at all given that I can get the data from `Cotech-367959`. What's your interest?

@ChadMatsalla , @merbanan I reviewed the code here and confirm that it's not good, few errors and mismatchs:
- The comments about data layout don't match the code
- The original code from Cotech was copy/paste and few modifications were done.
- The baro value is extract from data which contains the crc byte, not good at all.

And the point here, it's possible that you have a model with the baro sensor on the outside sensor, to confirm that we need samples, raw codes or few cu8 files. Then we can help and correct the code or confirm our assumptions and cancel this pr.

merbanan:
Ok, now I have read through the history a bit more.

If there are a new parameters to be extracted then the current cotech decoder should be extended with those. But I think it is only the barometric pressure that is missing and that is handled internally by the receiver most likely. With that conclusion this PR can be closed.

Is that the case @ChadMatsalla ?

ChadMatsalla:
Yes, this is the case.

PR #3392, opened 2025-10-31
Added device for GIF2020OCECNA

SRGDamia1:
The only difference from the current device for the Badger Orion (the 2014 version) is the bit width. See #2995.

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -299,7 +299,7 @@
     DECL(ec3k) \
     DECL(orion_endpoint_2020)
 
-/* Add new decoders here. */
+    /* Add new decoders here. */
 
 #define DECL(name) extern r_device name;
 DEVICES


mdjnewman:
I'm not getting anything using this, but I'm wondering if my Orion Cellular C just isn't broadcasting over RF at all? From reading manuals I think it's configuration dependent, and RF can be set as a backup only:

> Dynamic RF Messaging: Specific configurations transmit a backup RF message to facilitate Quick Reads or help troubleshoot in the field in the rare event of a network communication outage. The endpoint begins to transmit this additional RF message when it has been unsuccessful at communicating with BEACON via the cellular network for a defined period of time.

(from https://badgermeter.widen.net/content/0dupukb5ii/original?download=false&x.app=api)

I could definitely just be making some other newbie mistake though!

mdjnewman:
@SRGDamia1 I think the entry in `include/rtl_433_devices.h` is missing from this PR - I added it locally for testing.

mdjnewman:
Do you see `13xxxxxx` serial numbers with these changes? IIUC all GIF2020OCECNA devices should be in that range.

SRGDamia1:
Sorry, I just fixed up the PR to put the device in the include and update the ReadMe

zuckschwerdt:
Thanks. Can you indent the `/* Add new decoders here. */` line in rtl_433_devices.h so it is unchanged?

SRGDamia1:
Oops. Done.

PR #3396, opened 2025-11-03
Add support for NetAtmo weather station outdoor temp/hum and wind sensors

wxkpr:
This decoder handles the transmissions from the NetAtmo weather station outdoor modules in the 868.9 Mhz band. The base station initiates these transmissions every 6 seconds. Base station decoding is not included.

Normal operation of the NetAtmo WS is to collect the raw data from outdoor sensors and update the measurements in the cloud at 5 min intervals. I could confirm the data from the decoder by comparing with the average/min/max or on-demand measurements which are displayed in the cloud.

Also implemented an offset correction for the raw wind measurements (an ultrasonic anemometer), since my wind sensor showed an ageing effect with high offset.

Hope this is interesting for other NetAtmo owners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [287]  Fine Offset Electronics WS85 weather station
     [288]  Oria WA150KM freezer and fridge thermometer
     [289]  Voltcraft EnergyCount 3000 (ec3k)
+    [290]  NetAtmo temp/hum and wind sensors
 
 * Disabled by default, use -R n or a conf file to enable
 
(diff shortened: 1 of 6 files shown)


zuckschwerdt:
Thanks, seems like a good addition to the decoders.
The doc-comment (from line 13) needs to attach to the decoder function by being directly preceeding or using `@fn`.
Also the comment needs to be formatted as markdown -- try to use a markdown editor or preview to refine the current text, or build with `-DBUILD_DOCUMENTATION=ON` and view the resulting html.

It's not clear if we want to include the adjustments. rtl_433 should only output what's received, with the exception of straight conversion to known units if available.

wxkpr:
I have addressed all your comments, removed the adjustment code, and made some cleanup, mainly consolidation of parameter names that are shown in the json and csv output.  
Also tested a reduced sample rate of 300 K and it works well.

wxkpr:
after correction of markdown syntax, now ready for further review

merbanan:
Code merged.

PR #3404, opened 2025-11-14
Add support for La Crosse View LTV-WSDR1 Cyclone Wind and Rain Sensor

corhere:
The signal is just different enough to not be picked up by the existing LTV-WR1 decoder. (Though tragically lacrosse_r1_decode incorrectly matches on the LTV-WSDR1 signals, with garbage output.)

The La Crosse View V41CA weather station display shows precipitation on a bar graph with axis labels at halfway and top. At the smallest scale, the labels are 1.27 mm and 2.54 mm for measurements from an LTV-WSDR1. One tip of the bucket causes the raw_rain1 value to increment by one, and the displayed precipitation to increment by ~0.25 mm. (The raw_rain2 value takes on the previous sequence number's raw_rain1 value, presumably to afford measuring the instantaneous precipitation rate from a single packet.) The LTV-WSDR1 therefore most likely measures rainfall in units of 0.254 mm, or 0.01 inches. And it is a pretty safe bet that the LTV-WR1 also measures 0.01 inch per bucket tip since it has identical specifications to the LTV-WSDR1 and transmits in the same format, and other rain La Crosse rain gauges in the same product family (LTV-R1, LTV-R3) are also known to measure ~0.01 inch per bucket tip. Change the rain outputs for the LTV-WR1 and LTV-WSDR1 decoder to be scaled to physical units.

Make the flags output field conditional so it is only output if any of the flags are set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [287]  Fine Offset Electronics WS85 weather station
     [288]  Oria WA150KM freezer and fridge thermometer
     [289]  Voltcraft EnergyCount 3000 (ec3k)
+    [290]  LaCrosse Technology View LTV-WSDR1 Cyclone Wind and Rain Sensor
 
 * Disabled by default, use -R n or a conf file to enable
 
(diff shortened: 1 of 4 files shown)


merbanan:
Do you have samples for this?

merbanan:
No answers or samples. Closing.

PR #3405, opened 2025-11-14
Change FSK metadata to log freqs at kHz precision

corhere:
The FSK tones in a signal may be only a handful of kilohertz apart. Format the freq1 and freq2 to three decimal places so the user can see the difference between them.


zuckschwerdt:
Duplicate of  #1631
It's just a very rough estimation and there is not enough accuracy behind it, sorry.
We might add more digits when we add FFTW as option.

PR #3406, opened 2025-11-14
Add support for La Crosse View LTV-TH2i Thermo/Hygro Sensor

corhere:
Much like its predecessor the LTV-TH2, the CRC on some of the packets transmitted by the LTV-TH2i will be calculated with a fixed nonzero initial value. Unlike the LTV-TH2, the TH2i uses init=0xb2. The lacrosse_th3 decoder is still able to decode the TH2i packets with init=0x00 CRC values, but it discards the majority of the reports as corrupt. Add a check for init=0x2b CRC to the decoder to improve its support for decoding packets from LTV-TH2i sensors.

### Change LaCrosse LTV-TH decoder to output battery_ok, retransmit

Experimentation with a La Crosse View LTV-TH2i sensor has revealed the semantics of two of the flags. One signals a low battery, and the other signals that the packet is a repeat broadcast triggered by a button press. (The La Crosse V41CA display audibly beeps upon reception of a packet with this flag set.) Break these flags out into their own output fields and squelch the display of the remaining unknown flags unless one of them is set.

### DIFF
--- a/src/devices/lacrosse_th3.c
+++ b/src/devices/lacrosse_th3.c
@@ -60,12 +60,15 @@ Sequence# 0,1,3,4,5 & 7
     CHK is CRC-8 poly 0x31 init 0xac over 7 bytes following SYN
 
 LTV-TH2i
-    SYNC:32h ID:24h ?:4b SEQ:3b ?:1b TEMP:12d HUM:12d CHK:8h TRAILER:96h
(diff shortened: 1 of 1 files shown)


merbanan:
@corhere do you have any sample files for this change?

merbanan:
Code merged in current master.

PR #3410, opened 2025-11-22
Add support for Emos GS558 Smoke detector

pselak:


zuckschwerdt:
Emos is just a reseller not a brand and we already have the GS558 protocol. Maybe add your findings to that decoder?

pselak:
Hmm, I thought that Emos is a brand. Here is a link to product, that I have https://amzn.eu/d/fZCD0jI
I guess it has a gs558 chip in. 
I tried to decode with existing gs558 smoke detector (86) but it has different bit length and some other differeces. And its disabled by default.
The current decoder expects 24 bits per row, but my device is sending 20 bits + 7 bits (likely a stop/sync pattern).

Besides this, I don't have device at my desk for current 86 gs558 smoke detector to test my modifications.
I was affraid to break something at existing gs558 devices, that's why I decided to create a new decoder.

If you insist in implementation logic for Emos smoke detectors in existing decoder, I will try to include it.

Thanks for your time

zuckschwerdt:
Can you upload a .cu8 sample file of your smoke detector? The OOK should look like a EV1527 encoder I think.

pselak:
Here's a link -> https://we.tl/t-nETtSA6nHx
Concise I just noticed you have a separate tests project: https://github.com/merbanan/rtl_433_tests/tree/master. I can upload my files there, but before I do, could you confirm whether this should be added as a new sensor or as a modification to an existing one? 
Also, I'm new to this—thanks for your patience.

P.S. This are flex decoder parameters for .cu8 file
rtl_433.exe -s 250000 -r C:\work\rtl_433-win-x64-23.11\EMOS GS558_433.92M_250k.cu8 -R 0 -X "n=GS558,m=OOK_PWM,s=460,l=1220,g=1284,r=11648,t=304,invert" -Y autolevel -vv -M bits -F kv

zuckschwerdt:
The "trailing 3 bits" you see are not actually data but a kind of wakeup pulse. It's just 24 data bits. The symbols are ~1200 µs pulse + ~500 µs gap / ~500 µs pulse + ~1200 µs gap. The wakeup is ~500 µs + ~500 µs.

It's the same protocol/chip and best handled by the same decoder. Add documentation and needed changes to https://github.com/merbanan/rtl_433/blob/master/src/devices/smoke_gs558.c if you like.

You can PR a test file into a `02` folder to https://github.com/merbanan/rtl_433_tests/tree/master/tests/smoke_gs558 if you like.

pselak:
OK, thanks for info. 
Great, so I have existing recordings to test signals. I will put my recordings to `02` folder
I'll modify existing decoder to handle both signals.
What do you suggest to do with this PR? Should I close this and open another "clean" one?

zuckschwerdt:
If you want to archive your work then close this PR and open a new one. Otherwise you can reuse this PR as there is some good info already.

pselak:
Hello @zuckschwerdt, thanks for all your advices. I looked again an established debugger in my environment, and found out, that only one condition was problematic. All my devices has id set to 0. So I changed this condition of checking min/max value.
I was trusting too much to Copilot's advices :(.

zuckschwerdt:
My devices came with a random ID. An all-zero ID isn't great, can you run the setup procedure to get a random ID assigned?
(There is a pairing procedure, IIRC just start it and let it timeout.)

pselak:
Hello, well it looks you were right. There is no need to change code. I did  a reset to all smoke sensors and after reset I got an ID, that is not zero.
I'll close this pull request.

zuckschwerdt:
Great to hear that it worked out!

I just noticed that the GS558 manufacturer is indeed noted on your device: Siterwell :)
And about EMOS, they are a the Czech seller https://www.emos.cz/ that has expanded into some other EU countries https://www.emos.eu/ somewhat recently.

Issue #2306, opened 2023-01-10
Support for Watchman Sonic Advanced/Plus


garethpotter:
Yet another Watchman oil tank monitor, with an apparently different format.

A few samples are at https://github.com/merbanan/rtl_433_tests/pull/440.

I have managed to decode these using URH following the process outlined in https://github.com/merbanan/rtl_433/issues/1333. The signal is clear enough in URH's demodulated view:

<img width="1093" alt="image" src="https://user-images.githubusercontent.com/108409039/211536243-4e43ef35-f33a-40c3-bc6d-00a541b5f6ed.png">

...but I am not sure how to translate that process into a rtl_433 flex decoder. Running the captures through `rtl_433 -A -a 4` doesn't yield much of use, at least from what I can see.

zuckschwerdt:
That is a pretty neat signal capture. You can even just drop the cu8 onto https://triq.org/pdv/ and see the same curve.
It's GFSK, which rtl_433 isn't optimized for. But the `minmax` demod can pick that up somewhat, e.g.
`rtl_433 -Y minmax -A g100_433.92M_1024k.cu8` -- the timings are off though, actually it's 500 µs PCM like you manually found. E.g. should work, but does not: `rtl_433 -Y minmax -X 'n=name,m=FSK_PCM,s=500,l=500,r=9000'`

We really need someone to build a GFSK-optimized demod someday.

garethpotter:
> We really need someone to build a GFSK-optimized demod someday.

I suspect this is well beyond my capabilities but I will have a dig around.

zuckschwerdt:
The code is here: https://github.com/merbanan/rtl_433/blob/master/src/pulse_detect_fsk.c#L158
I suspect that setting FSK_EST_SLOW to up to 10x the value might help: https://github.com/merbanan/rtl_433/blob/master/src/pulse_detect_fsk.c#L23

rolandshiner:
I just got a Kingspan Watchman Sonic Advanced which I was hoping would work with a rtl-433, but like others here I find sadly it doesn't.

I'm pretty new to rtl-sdr etc but I've been sniffing the airways and have found some useful data. I think I have an ID and both temperature and height but there is other data here I haven't understood yet. The below very long link is my test data on bitbench. *I'm not sure how best to link to that.

The heights shown are always within 1-2 cm of the bottom of the device and the floor. I'm pretty confidant the height data is correct. 

[...]

PR #3411, opened 2025-11-23
minor: watchman advanced depth is 12 bits

markferry:
GH-2306

Observed with an 0x0106 device with 8-bit depth overflowing from 0xf4 to
0x(1)07:

    2dd40e 0106 00011a 80 4b a6 30 f4 01170000c9020
    2dd40e 0106 00011a 80 51 a6 41 07 01170000bb240

### DIFF
--- a/src/devices/oil_watchman_advanced.c
+++ b/src/devices/oil_watchman_advanced.c
@@ -38,8 +38,9 @@ Data Layout:
   - 0x80 - the first one or two transmissions after the sync period when the device seems to be calibrating itself
   - 0x98 - normal, live value that you'll see on every transmission when the device is up and running
 - 1 byte temperature, in intervals of 0.5 degrees, offset by 0x48
-- 2 byte - varying bytes which could be the raw sensor reading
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks! What maximum depth is specified for your sensor? with 9 bits we'd already be over 5 meters? Is it really 12 bit?

markferry:
> Thanks! What maximum depth is specified for your sensor? with 9 bits we'd already be over 5 meters? Is it really 12 bit?

See my recent comment on #2306 .

My 0x0106 model sensors report in *mm*.
I haven't been able to establish if the precision is in the payload or hardcoded at both ends. I'll create a separate PR to support that.

These sensors were originally backed by a cloud service so it seems reasonable that the supplier would get the highest resolution they can and massage the data on their side.

markferry:
Data from my sensor showing the overflow:  https://snapshots.raintank.io/dashboard/snapshot/jYV2To2xkOrbQOnlA4Zb2Ykdj8Mup3WU

PR #3427, opened 2025-12-17
Update README.md - split and update the Building and Installation part

serialrf433:
Many things have changed in over 5 years. Lets update the readme. Every supported Debian release these days have the package available. apt-get is now just apt. CPU architectures like RISC-V are more common now. ESP32 devices with build in LoRa chipsets exist that can be used for FSK and OOK decoding, and much more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,20 +10,22 @@ Actively tested and supported are Realtek RTL2832 based DVB dongles (using RTL-S
 
 ![rtl_433 screenshot](./docs/screenshot.png)
 
-## Building / Installation
+## Building
 
-rtl_433 is written in portable C (C99 standard) and known to compile on Linux (also embedded), MacOS, and Windows systems.
+rtl_433 is written in portable C (C99 standard) and known to compile on Linux (also embedded), FreeBSD, MacOS, and Windows systems.
 Older compilers and toolchains are supported as a key-goal.
-Low resource consumption and very few dependencies allow rtl_433 to run on embedded hardware like (repurposed) routers.
-Systems with 32-bit i686 and 64-bit x86-64 as well as (embedded) ARM, like the Raspberry Pi and PlutoSDR are well supported.
+Low resource consumption and very few dependencies allow rtl_433 to run on embedded hardware like routers running [OpenWrt](https://openwrt.org) or [ESP32](https://github.com/NorthernMan54/rtl_433_ESP).
+Most common 32-bit and 64-bit CPU architectures like amd64, arm64, armhf, i386, ppc64el, riscv64, s390x and others are well supported.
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Work on our docs is welcome. Thanks!

About the contents, we don't know of FreeBSD users (of NetBSD we know), and I think we also dont't know about or test i386, ppc64el, riscv64, s390x.

Not sure if we want to feature OpenWrt specifically. An ESP32 note should be bolder, like "for ESP32 devices there is the rtl_433_ESP spin off."

I hope Debian and Ubuntu will be familiar to users of derived distros. Naming Trisquel, PureOS, Kali seems random.

serialrf433:
FreeBSD was named before in the doc. I left it there. I did the check if its still available and it is. The maintainer of the rtl-433 package for FreeBSD is rodrigo@freebsd.org . The availability of rtl-433 in FreeBSD is listed in https://www.freshports.org/comms/rtl-433, in https://repology.org/project/rtl-433/versions and in https://pkgs.org/download/rtl-433 .

i386 - its the most basic 32bit architecture on x86. The listed i686 before was like if you name x86_64_v3 instead of the basic x86_64. The i386 package is stable in debian: https://packages.debian.org/search?keywords=rtl-433&searchon=names&suite=stable&section=all

Now you also know about the source of the CPU architecture rtl_433 is listed to be working and stable. Debian is conservative and is testing all the things before they list them as stable.

ppc64el and riscv64 are also more common. Especially riscv64 (64bit RISC-V) is even available on some modern laptops. 
I would just use the current debian stable cpu architecture list to show off how many different cpu architectures are supported and put at the end the word others to show even further flexibility. ESP32, MIPS and others like that are proven to be working in OpenWrt.

I added OpenWrt because routers as a target was listed in there before. Also the rtl-433 package is just available to install with opkg: https://repology.org/project/rtl-433/versions
It is highly practical to run additional things on your already 24/7 running device you use for internet access purpose.

ESP32 is sort of a hardware architecture. I named it in general and linked to a project rtl_433 is most commonly used to run on ESP32. Also OpenWrt is just the most common tool to run rtl_433 on your router instead of building your own Linux OS and putting it on there.

I took the other Debian-like distro from here https://repology.org/project/rtl-433/versions (ctrl+f for debian and look at the maintainer). The idea behind that was to list more of them to show the wide availability of rtl_433 also in not as common used distro and on the other side show to the people some other options they could use. Just naming ubuntu as the only debian based distro is not fair and correct these days.

I had spend time into thinking in detail how to change the documentation at the page. I hope explaining it here more in detail helps to merge it.

merbanan:
Applied with changes.

serialrf433:
Thanks for taking up the ideas and merging them the way you think they fit the best.

PR #3434, opened 2025-12-29
build: Add support for running on OpenBSD - rtlsdr_read_sync

n0xa:
rtl_433 failed to start on OpenBSD:
```
[Input] Reading samples in async mode...
[SDR] rtlsdr_set_center_freq 433920000 = 0
[SDR] Tuned to 433.920MHz.
[acquire_thread] acquire_thread enter...
[Input] Input device start failed, exiting!
[sdr_stop] EXITING...
[sdr_stop] JOINING...
[rtlsdr_read_loop] LIBUSB_ERROR_NOT_SUPPORTED: Operation not supported or unimplemented on this platform! Check your RTL-SDR dongle, USB cables, and power supply.
[rtlsdr_read_loop] rtlsdr_read_async done
[acquire_thread] acquire_thread async stop...
[SDR] async read failed (-12).
[acquire_thread] acquire_thread done...
[sdr_stop] EXITED.
```
switching to rtlsdr_read_sync on this platform made it work. Not shown: Accurately decoding weather stations and TPMS nearby.

```
[SDR] Found 1 device(s)
[SDR] trying device 0: Realtek, RTL2838UHIDIR, SN: 00000004
[SDR] Using device 0: Realtek, RTL2838UHIDIR, SN: 00000004, "Generic RTL2832U OEM"
[R82XX] PLL not locked!
[SDR] Sample rate set to 250000 S/s.
[Input] Bit detection level set to 0.0 (Auto).
[SDR] Tuner gain set to Auto.
[Input] Reading samples in async mode...
[SDR] rtlsdr_set_center_freq 433920000 = 0
[SDR] Tuned to 433.920MHz.
[acquire_thread] acquire_thread enter...
[rtlsdr_read_loop] Using synchronous read mode for OpenBSD
```

### DIFF
--- a/src/sdr.c
+++ b/src/sdr.c
@@ -549,6 +549,32 @@ static int rtlsdr_read_loop(sdr_dev_t *dev, sdr_event_cb_t cb, void *ctx, uint32
 
     dev->running = 1;
 
+#ifdef __OpenBSD__
+    // OpenBSD workaround: use synchronous reads instead of async
+    print_log(LOG_DEBUG, __func__, "Using synchronous read mode for OpenBSD");
+
+    unsigned char *sync_buf = malloc(buf_len);
+    if (!sync_buf) {
+        WARN_MALLOC("rtlsdr_read_loop() sync_buf");
+        return -1;
+    }
+
+    while (dev->running) {
+        int n_read = 0;
+        r = rtlsdr_read_sync(dev->rtlsdr_dev, sync_buf, buf_len, &n_read);
+        if (r < 0 || !dev->running) {
+            print_logf(LOG_ERROR, __func__, "rtlsdr_read_sync failed: %d", r);
+            break;
+        }
+        if (n_read > 0) {
+            rtlsdr_read_cb(sync_buf, n_read, dev);
+        }
+    }
+
+    free(sync_buf);
+    print_log(LOG_DEBUG, __func__, "rtlsdr_read_sync done");
+#else
+    // Standard async read for other platforms
         r = rtlsdr_read_async(dev->rtlsdr_dev, rtlsdr_read_cb, dev, buf_num, buf_len);
         // rtlsdr_read_async() returns possible error codes from:
         //     if (!dev) return -1;
@@ -570,6 +596,7 @@ static int rtlsdr_read_loop(sdr_dev_t *dev, sdr_event_cb_t cb, void *ctx, uint32
             dev->running = 0;
         }
     print_log(LOG_DEBUG, __func__, "rtlsdr_read_async done");
+#endif
 
     return r;
 }


zuckschwerdt:
Thanks!
The async function (https://gitea.osmocom.org/sdr/rtl-sdr/src/branch/master/src/librtlsdr.c#L1867) queues multiple transfers compared to sync (https://gitea.osmocom.org/sdr/rtl-sdr/src/branch/master/src/librtlsdr.c#L1708) which is only running a simple usb transfer. We might need to do more there so we don't lose input data.

gdt:
Thanks for posting this.  I will see how it works on NetBSD.

Have you run rtl_433 for a long time, e.g. 48h, and verified that it keeps working and doesn't drop data?   High-speed reading of data from USB is tricky in terms of getting the new read request quickly after completion to not lose samples.    This is why my group implemented read-ahead in NetBSD, which is sort of like async, but in the USB layer instead of the read/write layer.  The concept is that you turn it on and the device just reads and then your read calls take out of the buffer, so you can keep streaming going.  This was in 2005 for a USRP, but I don't think the fundamentals have changed.

n0xa:
This may impact other BSDs, so testing with and without this patch is welcome, but may require tweaking the ifdef. I am not a highly technical developer, but from what I can tell, OpenBSD's USB stack (/sys/dev/usb/) doesn't implement a few things that libusb async requires, so this is an architecture problem.  
I'm kind of hackathonning SDR stuff on my favorite OS over my winter break, and this theme keeps popping up where rtlsdr_read_async breaks stuff. I let rtl_433 run overnight (about 10 hours) and it found dozens of TPMS sensors, a handful of simplisafe, and my two neighbors' weather stations. Output volume and device list looks similar to overnight runs on other platforms I've done in the past.  I have restarted it on my desktop PC and will leave it running for a few days.

gdt:
Sure, happy to help with extending the ifdef, and yes, openbsd/netbsd usb is similar and doesn't have the async stuff yet. Adding async to open/net is really the right approach, but harder, so if this works we should not hold it up because of theoretical better fix later.

zuckschwerdt:
We somewhat recently changed SDR data acquisition to run in a thread (#1978).
This might now help with the sync loop which previously would have lost data (due to demod taking quite long).
Maybe extend the comment with a line of `// This depends on the read callback finishing quickly`

n0xa:
I have had this running for about 2 and a half days. It's still holding steady (last decode less than 30 seconds ago):

   1   "model": "Hyundai-VDO",
   1   "model": "Renault-0435R",
   3   "model": "Abarth-124Spider",
   3   "model": "Watts-WFHTRF",
   4   "model": "Citroen",
   6   "model": "Renault",
  10   "model": "Schrader-EG53MA4",
  38   "model": "DSC-Security",
 101   "model": "Schrader",
 171   "model": "SimpliSafe-Sensor",
21242   "model": "Acurite-3n1",

I have a log running it `./rtl_433 -vv -F json -F log 2>&1` and there's a lot of log spam, but it's mostly preamble and decode debug output. About 13MB of text all in all. I could upload it somewhere or if there are any specific things I ought to be looking for in it, I'd be happy to assist in that.  For my purposes, it's working exactly as I need it to. I'd love to hear from others testing it. I'm happy to add the clarifying comment above, but might as well do one last commit with anything else that's needed if this seems like a decent candidate for merging.

Cheers, and happy new year!

gdt:
Thanks for the testing and report.

What happens if you call rtl_sdr_read_async on OpenBSD?   Is it defined? Does it return an error?    I would really like to see the ifdef converted into something that doesn't use an OS name so it will work on other similar systems without a list, as well as use async when openbsd/netbsd get async.

I still intend to try this; digging out from testing a totally unrelated program as well as things other than computers.

n0xa:
The initial comment the PR was created with shows exactly what happens with rtlsdr_read_async. Maybe testing for this condition in the configure stage would be better? I could try to figure that out.

merbanan:
This workaround should be done either in librtlsdr, libusb or fixed in openbsd not in rtl_433. Until that happens this patch should live in the ports package manager in openbsd. And it seems to already exists: https://openports.pl/path/comms/rtl_433

So something is done there to get rtl_433 working.

PR #3441, opened 2026-01-05
Tweek times for ThermoPro TX2-C

SRGDamia1:


zuckschwerdt:
Do you have samples with timings of an actual device? What packet distance (gap width) do you see?

merbanan:
No answer for a while. Reopen if there is still an issue.

PR #3448, opened 2026-01-24
Add support for JCHENG SECURITY Contact and PIR sensors

hertzg:
## Summary

Add decoders for JCHENG SECURITY sensors:
- **Contact Sensor** (`Jcheng-Contact`): Door/window contact sensor with 16-bit ID, reports closed state, battery status, and event flag
- **PassiveIR Sensor** (`Jcheng-PIR`): PIR motion sensor with 12-bit ID, reports motion, tamper, and battery status

Both sensors use OOK PWM modulation at 433.92MHz with 25-bit messages.

## Protocol Details

**Contact Sensor (25 bits):**
```
PPPP IIII IIII IIII IIII OBSX XX
```
- P: 4-bit preamble (0xF)
- I: 16-bit sensor ID
- O: is_on flag
- B: battery_ok
- S: state (1=closed, 0=open)

**PIR Sensor (25 bits):**
```
PPPP PPPP IIII IIII IIII TMBX XX
```
- P: 8-bit preamble (0xAA)
- I: 12-bit sensor ID
- T: tamper flag
- M: motion detected
- B: battery_low

## Notes

- Decoders are **disabled by default** as there is no checksum/CRC in the protocol
- Users can enable with `-R 292` (contact) or `-R 293` (PIR)

## Test data

I can provide test recordings for the **Contact Sensor** only. Unfortunately, I no longer have the PIR sensor device.

**@zuckschwerdt**: Would you prefer I remove the PIR decoder from this PR and submit it as Contact Sensor only? I can add the PIR decoder later if/when I have access to the device again.

## Test plan

- [ ] Add test recordings for Contact Sensor to rtl_433_tests
- [ ] Verify decoding with actual sensor recordings

### DIFF
--- a/include/rtl_433_devices.h
+++ b/include/rtl_433_devices.h
@@ -299,6 +299,8 @@
     DECL(ec3k) \
     DECL(orion_endpoint_2020) \
     DECL(geevon_tx19) \
+    DECL(jc_contact) \
+    DECL(jc_pir) \
 
     /* Add new decoders here. */
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -164,6 +164,7 @@ add_library(r_433 STATIC
     devices/interlogix.c
     devices/intertechno.c
     devices/jasco.c
+    devices/jc_sensor.c
     devices/kedsum.c
     devices/kerui.c
     devices/klimalogg.c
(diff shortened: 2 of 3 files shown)


zuckschwerdt:
Thanks! The code looks great. If there is no checksum and no advanced decoding I'd always recommend to use a flex decoder. Given that there is the 0xa/0xaa check it might warrant a full decoder though.

merbanan:
merged as flex decoders

Issue #3445, opened 2026-01-14
Sainlogic SA8 / Gevanti SA8 signal analysis


lrbison:
Looking for any signal analysis experts to help determine the modulation used by this weather station I recently purchased.

I've captured a snippet on my RTL-SDR v4 on windows using the 25.12 release.  It transmits every minute at 433.9MHz, but both me and rtl_433 are confused by the modulation scheme.

Here is the [triq link](https://triq.org/pdv/#AAB02C0801FFFF00D0019C0268037C048044B4073C8292919192A391939191919392929191A2B1A19191A2A29293A455+AAB0130801FFFF00D0019C0268037C048044B4073C9455+AAB0180801FFFF00D0019C0268037C048044B4073C91B191C1A1A455+AAB0210801FFFF00D0019C0268037C048044B4073CB19191B19192B291A3C192B291A3C455+AAB0190801FFFF00D0019C0268037C048044B4073C91C3939191A2D755+AAB0160801FFFF00D0019C0268037C048044B4073C91C1A19755+AAB0160801FFFF00D0019C0268037C048044B4073C9191C29755+AAB0130804FFFF00D0019C0268037C048044B4073C9755+AAB0160801FFFF00D0019C0268037C048044B4073C9391A1A755+AAB0150801FFFF00D0019C0268037C048044B4073C92D19755+AAB0140801FFFF00D0019C0268037C048044B4073C919455+AAB0140801FFFF00D0019C0268037C048044B4073C919755+AAB0170801FFFF00D0019C0268037C048044B4073C91A291919755+AAB0130802FFFF00D0019C0268037C048044B4073C9755+AAB0150801FFFF00D0019C0268037C048044B4073C9193C455+AAB01A0801FFFF00D0019C0268037C048044B4073C9391B1939191B3E055).

and the .cu8 file can be downloaded here: https://paste.c-net.org/PrevailHardon

and here is a screenshot:

<img width="1082" height="278" alt="Image" src="https://github.com/user-attachments/assets/aab5c06c-1a58-41e4-9091-ebf5415566c8" />

The timing analysis from rtl_433.exe: 

```
Analyzing pulses...
Total count:   96,  width: 185.79 ms            (46448 S)
Pulse width distribution:
 [ 0] count:    1,  width: 81392 us [81392;81392]       (20348 S)
 [ 1] count:   62,  width:  216 us [216;224]    (  54 S)
 [ 2] count:   15,  width:  424 us [424;432]    ( 106 S)
[...]

PR #3450, opened 2026-01-25
Add support for Sainlogic SA8, Gevanti SA8 Weather Station

ProfBoc75:
This is related to the issue #3445 for Saintlogic SA8 Weather Station analysis.

Temp, Hum, Wind Gust/Avg/Direction/Rain gauge are guessed, but need to be confirmed, and very few data not yet guessed, like CRC/Checksum. The protocol is therefore disabled by default.

Please add `-R 292` option to force this protocol.

With `-R 292:vv` you can pickup the UART decoded MSG for any deep analysis.

### DIFF
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -453,7 +453,7 @@ convert si
   protocol 212 # Renault 0435R TPMS
   protocol 213 # Fine Offset Electronics WS80 weather station
   protocol 214 # EMOS E6016 weatherstation with DCF77
-  protocol 215 # Emax W6, rebrand Altronics x7063/4, Optex 990040/50/51, Orium 13093/13123, Infactory FWS-1200, Newentor Q9, Otio 810025, Protmex PT3390A, Jula Marquant 014331/32, TechniSat IMETEO X6 76-4924-00, Weather Station or temperature/humidity sensor
+  protocol 215 # Emax W6, rebrand Altronics x7063/4/x7064A, Optex 990040/50/51, Orium 13093/13123, Infactory FWS-1200, Newentor Q9, Otio 810025, Protmex PT3390A, Jula Marquant 014331/32, TechniSat IMETEO X6 76-4924-00, Weather Station or temperature/humidity sensor
 # protocol 216 # ANT and ANT+ devices
   protocol 217 # EMOS E6016 rain gauge
   protocol 218 # Microchip HCS200/HCS300 KeeLoq Hopping Encoder based remotes (FSK)
@@ -530,7 +530,8 @@ convert si
   protocol 289 # Voltcraft EnergyCount 3000 (ec3k)
   protocol 290 # Orion Endpoint from Badger Meter, GIF2020OCECNA, water meter, hopping from 904.4 Mhz to 924.6Mhz (-s 1600k)
   protocol 291 # Geevon TX19-1 outdoor sensor
-# protocol 292 # Sainlogic SA8, Gevanti SA8 Weather Station
+  protocol 292 # WallarGe CLTX001 Outdoor Temperature Sensor
+  protocol 293 # Sainlogic SA8, Gevanti SA8 Weather Station
 
 ## Flex devices (command line option "-X")
 
--- a/README.md
+++ b/README.md
@@ -300,7 +300,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [212]  Renault 0435R TPMS
     [213]  Fine Offset Electronics WS80 weather station
     [214]  EMOS E6016 weatherstation with DCF77
-    [215]  Emax W6, rebrand Altronics x7063/4, Optex 990040/50/51, Orium 13093/13123, Infactory FWS-1200, Newentor Q9, Otio 810025, Protmex PT3390A, Jula Marquant 014331/32, TechniSat IMETEO X6 76-4924-00, Weather Station or temperature/humidity sensor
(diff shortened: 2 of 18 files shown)


src/devices/saintlogic_sa8.c
```diff
@@ -0,0 +1,150 @@
...
+    //int unknown_chk   = b[39] << 8  | b[40]; // big endian until guessed
+
+    float temp_c      = temp_raw * 0.1f;
+    float gust_km_h   = gust_raw * 0.036f; // orignal value is m/s but for Customary conversion, km/h is better
+    float wind_km_h   = wind_raw * 0.036f; // orignal value is m/s but for Customary conversion, km/h is better
+    float rain_mm     = rain_raw * 0.15f;
+
+    /* clang-format off */
+    data_t *data = data_make(
+            "model",           "",                 DATA_STRING, "Saintlogic_SA8",
```
zuckschwerdt: a typo: `Saintlogic-SA8`

conf/rtl_433.example.conf
```diff
@@ -530,6 +530,7 @@ convert si
   protocol 289 # Voltcraft EnergyCount 3000 (ec3k)
   protocol 290 # Orion Endpoint from Badger Meter, GIF2020OCECNA, water meter, hopping from 904.4 Mhz to 924.6Mhz (-s 1600k)
   protocol 291 # Geevon TX19-1 outdoor sensor
+# protocol 292 # Saintlogic SA8 Weather Station
```
lrbison: Maybe safer to label this Gevanti SA8.  I suspect it's the same as Sainlogic, but I'm not fully confident.  Or at least remove the "t" if you keep it sainlogic.

src/devices/saintlogic_sa8.c
```diff
@@ -0,0 +1,158 @@
...
+- ??: {16} fixed value 0xf300
+- TT: {16} little endian LSB/MSB, signed value, Temp °C, scale 10
+- HH:  {8} Humidity %
+- 00: {40} Fixed value to 0
+- GG: {16} little endian LSB/MSB, Wind Gust in m/s, scale 100
+- WW: {16} little endian LSB/MSB, Wind Average in m/s, scale 100
+- DD: {16} little endian LSB/MSB, Wind Direction in °, 0 = North, 180 = South
+- RR: {16} little endian LSB/MSB, Rain Gauge in mm scale 0.42893617f
+- ??: {16} little endian LSB/MSB, another unknown counter
+- LL:  {8} To be confirmed, probably battery level information here.
```
lrbison: My Guess is this field (LL: {8}) is related to solar panel and or the internal battery.

src/devices/saintlogic_sa8.c
```diff
@@ -0,0 +1,158 @@
...
+
+    if (bitbuffer->num_rows != 1) {
+        decoder_logf(decoder, 1, __func__, "Too many rows: %d", bitbuffer->num_rows);
+        return DECODE_ABORT_EARLY;
+    }
+
+    unsigned offset = bitbuffer_search(bitbuffer, 0, 0, preamble_pattern, 16) + 16;
+
+    if (offset >= bitbuffer->bits_per_row[0]) {
+        decoder_log(decoder, 1, __func__, "Sync word not found");
```
lrbison: When I run with a single :v, I get very many of these debug logs, so the point I grep -v to ignore them.  Maybe it's too verbose?  I leave it to you to decide.

ProfBoc75: I increased the verbose for this value, now you'll need  -vv to get more details instead of -v

src/devices/saintlogic_sa8.c
ProfBoc75: In fact, I have another approach.
Looks like all the metrics are little-endian 16b, same for the last, probably the battery level, in mV.

Looking at all the values you shared, from the first older to the more recent (look at the "flags" value):
(because you have 3 batteries into the weather station the decimal value is divided by 3 to get the estimation)

| Raw value (Flags) | little-endian | Decimal | Comment | AA Battery estimation |
--- | --- | --- | --- | --- 
F110 | 10F1 |  4337 | 4337 mV = 4.337 V => 3 x AA battery, so around 1.4456 V / battery, which is normal value for recent batteries, expected around 1.5 V or little more | 1.4456 V |
C610 | 10C6 | 4294 |  4.294 V |  1.431 V 
BE10 | 10BE | 4286  | 4.286 V | 1.428 V
8B10 | 108B | 4235 | 4.235 V | 1.412 V
xx01 | 01xx |  256 to 511 | Battery removed, 0.256 V to 0.511 V | 

The value slowly decreased, like normal battery behavior, and the solar panel helps to reduce the battery wear/discharge.


If we keep the current Battery OK bit flag, 0x10, this means all values from 4096 to 8191 are Battery OK if less Battery LOW. So less than 1.365 V / Battery, may be still working ?
When you did the battery test, did you noticed a battery low icon on the display ? I can't found this information from the SA8_User_Manual.

conf/rtl_433.example.conf
ProfBoc75: corrected

src/devices/saintlogic_sa8.c
lrbison: >When you did the battery test, did you notice a battery low icon on the display ? 

Yes.

I am also considering finding some nearly-dead batteries and seeing what they indicate.

ProfBoc75:
The CRC formula is guessed, so the protocol is no more disabled.
Notice that the protocol number is now 293 instead of 292, since another device was added in the meantime.

PR #3453, opened 2026-01-26
Add support for WS69 pressure sensor

zuckschwerdt:
Adds support for the extra pressure sensor found on newer WS69.

See also #3059 and #3426


merbanan:
LGTM @zuckschwerdt

Issue #3452, opened 2026-01-26
Support for Apator Metra E-ITN 30.2


Pavlajs:
Hi,
I am trying decode data from Apator Metra E-ITN 30.2 electronic heat cost allocator.
The manufacturer uses its own protocol and he refused to sell me a device to read the values.
I am trying to read it with rtl_433 -f 868.95M -s 1024k -g 20 -X "n=metra,m=FSK_PCM,s=25,l=25,r=1000"
A am receiving this data. Btw I added the ID, value and value last year. That should be correct.
Thank you for the advice.

```
[ID:31975929 Value:517 Value last year:2605]
5555 5555 5554 d335 dd84 bdb7 1c00 7a2b 0995 be6c f1f2 6183 ef7b 8dd8
5555 5555 5554 d335 dd84 bdb7 1c00 7a2b 0995 7e6c f1f2 6015 ff7b f04a

[ID: 31975930 Value:0 Value last year:1]
5555 5555 5554 d335 dd9e bdb7 1d12 67c8 a5a0 3e72 e1f2 6195 6d7b d88c
5555 5555 5554 d335 dd9e bdb7 1d12 67c8 a5a1 be72 e1f2 6195 c77b 21d7

[ID:31975868 Value:298 Value last year:1351]
5555 5555 5554 d335 dcb6 bdb7 1de8 7961 3ba4 bf6e f9f2 60fa c17b c03c
5555 5555 5554 d335 dcb6 bdb7 1de8 7961 3ba5 bf6e f9f2 60db 797a 54c4

[ID:31975867 Value:415 Value last year:1966]
5555 5555 5554 d335 dcb4 bdb7 1cc2 65f6 e603 6216 f9f2 6015 fd7a 0c72
5555 5555 5554 d335 dcb4 bdb7 1cc2 65f6 e602 2216 f9f2 6005 e17a 42ca
```

zuckschwerdt:
For the values we expect to see lots of zeros, esp. for the second sensor (Value:0 Value last year:1), but we don't. The data might be (XOR) encrypted, maybe a simple Whitening scheme.

The ID is usually at the front. We expect at least 32 bits with most bits the same as your IDs are so close together. I guess it's the `dd84 bdb7` part but with different endianess or LSB/MSB bit-order (and/or whitening).

How far apart in time are the two readings per sensor? Can you grab a sequence to hopefully spot a single value tick upwards?

The last 16 bit look like a checksum, maybe a digest.

[...]

PR #3458, opened 2026-02-01
Add support for Apator Metra E-ITN 30 and Improve E-RM 30 water meter

ProfBoc75:
This is related to the issue #3452 to add support for the Apator Metra E-ITN 30 Heat cost allocator meter.

- This PR to add the support for this sensor, but also to improve the overall decoding part for the E-RM 30 using the same protocol.
- The previous version of the decoder did not unwhitened the data using the IBM coding, using the more complex s-boxes/key-schedule method.

@carpalex : Can you please review my code, in order to check if there is any changes that could BREAK the previous version.
I noticed that the ID is not decoded the same, your code using the s-boxes/key-schedule was always giving the good ID.
With my code, using the IBM unwhiten and 1 s-box, I need to do a XOR to get the good ID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,7 +300,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [212]  Renault 0435R TPMS
     [213]  Fine Offset Electronics WS80 weather station
     [214]  EMOS E6016 weatherstation with DCF77
-    [215]  Emax W6, rebrand Altronics x7063/4, Optex 990040/50/51, Orium 13093/13123, Infactory FWS-1200, Newentor Q9, Otio 810025, Protmex PT3390A, Jula Marquant 014331/32, TechniSat IMETEO X6 76-4924-00, Weather Station or temperature/humidity sensor
+    [215]  Emax W6, rebrand Altronics x7063/4/x7064A, Optex 990040/50/51, Orium 13093/13123, Infactory FWS-1200, Newentor Q9, Otio 810025, Protmex PT3390A, Jula Marquant 014331/32, TechniSat IMETEO X6 76-4924-00, Weather Station or temperature/humidity sensor
     [216]* ANT and ANT+ devices
     [217]  EMOS E6016 rain gauge
     [218]  Microchip HCS200/HCS300 KeeLoq Hopping Encoder based remotes (FSK)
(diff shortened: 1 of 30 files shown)


carpalex:
Hi,
So I have ~5 million raw bit strings captured from around 100 water meters that I have in range, during several months. To double check everything, I am running those bit strings through `rtl_433 -y <code>`, on both implementations (the old one with s-boxes/key-schedule and the new one with IBM whitening).

Running through all of them takes some time, but so far I've processed around 500k, and both implementations yield **identical** results. I think this is a solid indicator that the new implementation is also correct for the water meter and does not break anything. 
I'll also let you know when I have everything processed.

Looking at the code itself, the new implementation is far superior to the old one and more likely to match what is actually happening at the hardware level. Having a LSFR and a single substitution table makes a lot sense.

I do have one question / suggestion: would it make sense to split the combined detector into 2 separate ones? This way, one would be able to decode messages coming from only one type of device, if needed.

ProfBoc75:
> Hi, So I have ~5 million raw bit strings captured from around 100 water meters that I have in range, during several months. To double check everything, I am running those bit strings through `rtl_433 -y <code>`, on both implementations (the old one with s-boxes/key-schedule and the new one with IBM whitening).
> 
> Running through all of them takes some time, but so far I've processed around 500k, and both implementations yield **identical** results. I think this is a solid indicator that the new implementation is also correct for the water meter and does not break anything. I'll also let you know when I have everything processed.
> 
> Looking at the code itself, the new implementation is far superior to the old one and more likely to match what is actually happening at the hardware level. Having a LSFR and a single substitution table makes a lot sense.
> 
> I do have one question / suggestion: would it make sense to split the combined detector into 2 separate ones? This way, one would be able to decode messages coming from only one type of device, if needed.

@carpalex , Thanks a lot for your feedback ! Everything is possible, and if we split into 2 decoders, the discrepancy is at the message length level, this way, each decoder will be more simple and just checking if 0x11 or 0x13, and no more variable length to manage, but fixed value. And, later, if one of the decoder needs to be corrected / improved, may be more simple too.

carpalex:
I processed all the ~5 million samples that I had and, as expected, the two implementation gave identical results.
Which can also be proved mathematically: I you take the inverses of the 16 S-boxes from the old implementation, they can be derived from one another by XOR-ing sbox_0_0 with 0x1, 0x2, ... 0xf, respectively. Which is the equivalent of whitening. And instead of the key_schedule we now use the IBM whitening key.

BTW, the code with 2 decoders looks great! I think it's more readable and extensible this way.

carpalex:
Hi, are we waiting for more input, or can this PR be merged?

carpalex:
@zuckschwerdt Do you think this PR is ready to be merged?

Issue #3215, opened 2025-03-02
Correcting Build instructions for Windows in `BUILDING.md`


eabase:
I'd like to see some improvement in the documentation for Building on Windows.

In particular, the section for `Visual Studio 2017` need to be updated. 
As it is presented today, it is useless since it is not reproducible.
* https://github.com/merbanan/rtl_433/blob/master/docs/BUILDING.md

Instead I can provide detailed build/install instructions for using `Visual Studio 2022` + `SoapySDR` + `RTL-SDR`, via **CMake**.

---

### A Quick Build Summary for Windows-11 

Assumptions:

* Have installed native Windows versions of:
  - `LIBUSB` - also included below
  - `RTL-SDR` - also included below
  - `SoapySDR` - optional, but [build instructions here](https://github.com/pothosware/SoapyRTLSDR/issues/79)
  - `Doxygen` - optional
  - `Visual Studio 2022`
  - `CMake`
  - `Windows-11`
* You are able to start a **"VS Dev Shell"** in Admin mode, in `Windows Terminal`.
* You know how to set and order the `User` and `Machine` PATH in Windows with:  
  `SystemPropertiesAdvanced`

Here are the steps. 
<sub>Please ensure to replace the `\path\to\` part with your own PATH.</sub>

```powershell
#----------------------------------------------------------------------
# Quick Build - Summary 
#----------------------------------------------------------------------

#------------------------------------------------
# 1. Install Requirements
#------------------------------------------------

# (a) For OpenSSL
winget install -e --id 'FireDaemon.OpenSSL'

# (b) For vcpkg [pkgconfig, pthreads]
# Which is needed to install pkgconfig and pthreads...
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg; .\bootstrap-vcpkg.bat -disableMetrics

# (b.1) Configure the VCPKG_ROOT environment variable.
# Make these permanent, once confirmed to work.
$env:VCPKG_ROOT = "C:\path\to\vcpkg"
$env:PATH = "$env:VCPKG_ROOT;$env:PATH"
$env:VCPKG_DISABLE_METRICS=1

# (b.2) Add dependencies and project files
vcpkg version
vcpkg new --application
vcpkg add port fmt

vcpkg list
vcpkg search pkgconf
vcpkg search pthread

# (b.3) Install 2 most suitable packages
vcpkg install pkgconf:x64-windows
vcpkg install pthreads:x64-windows-static

# (c) libusb
[...]

PR #3464, opened 2026-02-09
Change OOK input file parser to handle `;received` timestamp lines

kevinmehall:
When writing an OOK file, `rtl_433` includes `;received` lines with the timestamp of the packet. When reading OOK, these timestamps were ignored and all packets were reported received at the start time of the process.

Correctly reporting these timestamps might be helpful to look at packet timings when replaying an OOK file, but my main motivation is to have a long-running `rtl_433` with piped input from a [process generating OOK](https://codeberg.org/kevinmehall/gpio_433) and have correct timestamps passed through to the output.

### DIFF
--- a/src/r_util.c
+++ b/src/r_util.c
@@ -288,3 +288,40 @@ char const *nice_freq (double freq)
      snprintf (buf, sizeof(buf), "%f", freq);
   return (buf);
 }
+
+#ifdef _TEST
+#define TEST_PARSE_TIME_STR(str, epoch, millis) \
+    do { \
+        struct timeval tv; \
+        if (parse_time_str(str, &tv) != NULL) { \
+            if (tv.tv_sec == (epoch) && tv.tv_usec == (millis)) { \
+                ++passed; \
+            } \
+            else { \
+                ++failed; \
+                fprintf(stderr, "FAIL: parse_time_str \"%s\" = %ld.%06ld, expected %ld.%06d\n", str, tv.tv_sec, tv.tv_usec, (long)(epoch), (millis)); \
+            } \
+        } \
+        else { \
+            ++failed; \
+            fprintf(stderr, "FAIL: parse_time_str failed to parse \"%s\"\n", str); \
+        } \
+    } while (0)
+
+int main(void)
+{
+    unsigned passed = 0;
+    unsigned failed = 0;
+
+    TEST_PARSE_TIME_STR("2026-02-11T12:34:56.123456Z", 1770813296, 123456);
+    TEST_PARSE_TIME_STR("2026-02-11 12:34:56.123456Z", 1770813296, 123456);
+    TEST_PARSE_TIME_STR("2026-02-11 12:34:56.123Z", 1770813296, 123000);
+    TEST_PARSE_TIME_STR("2026-02-11 12:34:56.111111111Z", 1770813296, 111111);
+    TEST_PARSE_TIME_STR("2026-02-11 12:34:56Z", 1770813296, 0);
+    TEST_PARSE_TIME_STR("2026-02-11 12:34:56.123456-0700", 1770838496, 123456);
+    TEST_PARSE_TIME_STR("2026-02-11 12:34:56.123456+0845", 1770781796, 123456);
+
+    fprintf(stderr, "r_util test (%u/%u) passed, (%u) failed.\n", passed, passed + failed, failed);
+    return failed;
+}
+#endif /* _TEST */
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -20,7 +20,7 @@ endif()
 ########################################################################
 # target_compile_definitions was only added in CMake 2.8.11
 add_definitions(-D_TEST)
-foreach(testSrc bitbuffer.c fileformat.c optparse.c bit_util.c)
+foreach(testSrc bitbuffer.c fileformat.c optparse.c bit_util.c r_util.c)
     get_filename_component(testName ${testSrc} NAME_WE)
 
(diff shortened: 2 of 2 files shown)


zuckschwerdt:
Very well written and a good addition, thanks! Good for merge.

kevinmehall:
The test failure is because `strptime` is non-standard and requires `#define _XOPEN_SOURCE` in clang. However, when looking at the compatibility of that, I noticed that Windows doesn't have `strptime` at all.

I'll replace `strptime` with `sscanf` when I get a chance.

kevinmehall:
Updated with sscanf and tests

DennisKehrig:
This change doesn't work as is on Windows:
```
error C1083: Cannot open include file: 'sys/time.h': No such file or directory
```

kevinmehall:
Try https://github.com/kevinmehall/rtl_433/commit/6ae1b807737b097cb9035a65e04e278da855eff9

DennisKehrig:
> Try [kevinmehall@6ae1b80](https://github.com/kevinmehall/rtl_433/commit/6ae1b807737b097cb9035a65e04e278da855eff9)

```
r_util.obj : error LNK2019: unresolved external symbol gettimeofday referenced in function get_time_now
r_util.obj : error LNK2019: unresolved external symbol timegm referenced in function parse_time_str
```

DennisKehrig:
I'm roughly using the approach outlined here to compile with VS 2022: https://github.com/merbanan/rtl_433/issues/3215
When I compile inside of VS 2017, I only get a complaint about `timegm` missing, not `gettimeofday`. I see that the latter is part of `compat_time.c`, so I'm not sure yet why the VS 2022 approach complains about that, too. But feel free to ignore that one for now, that might be specific to my setup.

kevinmehall:
I'm confused by the `gettimeofday` one since that was already there, and as you noted, has a compatibility shim. Were you able to build a previous commit, or is this your first time building rtl_433?

For `timegm`, it looks like Windows has an equivalent [`_mkgmtime`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/mkgmtime-mkgmtime32-mkgmtime64?view=msvc-170), so the fix might be to add
```c
#ifdef _WIN32
#define timegm _mkgmtime
#endif
```

in either `r_util.c` or `compat_time.h`.

DennisKehrig:
I looked into it more closely, my VS 2022 build also compiles the tests, so I'm getting both errors for test_r_util.vcxproj, and only the timegm one for rtl_433.vcxproj, while VS 2017 doesn't seem to compile the tests the first time I click on CMake >  Build Only > All Targets, but when I click it again I see the same errors as with VS 2022.

With less omitted context:
```
r_util.obj : error LNK2019: unresolved external symbol gettimeofday referenced in function get_time_now [F:\Repositories\rtl_433\build\tests\test_r_util.vcxproj]
r_util.obj : error LNK2019: unresolved external symbol timegm referenced in function parse_time_str [F:\Repositories\rtl_433\build\tests\test_r_util.vcxproj]
F:\Repositories\rtl_433\build\tests\Release\test_r_util.exe : fatal error LNK1120: 2 unresolved externals [F:\Repositories\rtl_433\build\tests\test_r_util.vcxproj]

r_433.lib(r_util.obj) : error LNK2001: unresolved external symbol timegm [F:\Repositories\rtl_433\build\src\rtl_433.vcxproj]
F:\Repositories\rtl_433\build\src\Release\rtl_433.exe : fatal error LNK1120: 1 unresolved externals [F:\Repositories\rtl_433\build\src\rtl_433.vcxproj]
```

So maybe compat_time.c needs to be pulled in now for test_r_util?

I'm working on adding support for another temperature sensor I recently got, and was able to compile again after rebasing onto 1628d4a3a165c7496ad609dc32c083b4222dcd44 (the commit prior to merging this PR).

DennisKehrig:
> For `timegm`, it looks like Windows has an equivalent [`_mkgmtime`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/mkgmtime-mkgmtime32-mkgmtime64?view=msvc-170), so the fix might be to add
> 
> ```c
> #ifdef _WIN32
> #define timegm _mkgmtime
> #endif
> ```
> 
> in either `r_util.c` or `compat_time.h`.

Adding it to `compat_time.h` helps, now I just get the gettimeofday one for test_r_util:
```
r_util.obj : error LNK2019: unresolved external symbol gettimeofday referenced in function get_time_now [F:\Repositories\rtl_433\build\tests\test_r_util.vcxproj]
F:\Repositories\rtl_433\build\tests\Release\test_r_util.exe : fatal error LNK1120: 1 unresolved externals [F:\Repositories\rtl_433\build\tests\test_r_util.vcxproj]
```

DennisKehrig:
This change to `tests/CMakeLists.txt` seems to do the trick:
```
foreach(testSrc bitbuffer.c fileformat.c optparse.c bit_util.c r_util.c)
    get_filename_component(testName ${testSrc} NAME_WE)

    add_executable(test_${testName} ../src/${testSrc} ../src/compat_time.c)

    add_test(${testName}_test test_${testName})
endforeach(testSrc)
```
(Added compat_time.c to add_executable)
There's probably a more targeted approach.

EDIT: Added r_util.c back in

DennisKehrig:
@kevinmehall Thank you for your help! I'm glad it's a reasonably straightforward fix.

zuckschwerdt:
I've commited these fixes with c0036e0

DennisKehrig:
@zuckschwerdt I rebased my changes and it's compiling just fine now. Thank you!

PR #3467, opened 2026-02-13
Add support for Ecowitt WN20

rromanchuk:
Related tests: https://github.com/merbanan/rtl_433_tests/pull/487

### DIFF
--- a/src/devices/ambientweather_wh31e.c
+++ b/src/devices/ambientweather_wh31e.c
@@ -1,11 +1,12 @@
 /** @file
-    Ambient Weather WH31E, EcoWitt WH40 protocol.
+    Ambient Weather WH31E, EcoWitt WH40, EcoWitt WN20 protocol.
 
     Copyright (C) 2018 Christian W. Zuckschwerdt <zany@triq.net>
     based on protocol analysis by James Cuff and Michele Clamp,
     EcoWitt WH40 analysis by Helmut Bachmann,
     Ecowitt WS68 analysis by Tolip Wen improved by Bruno Octau,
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
Thanks, looks great!
- You missed one indent space in the `int id` line.
- What is that `extra` data, isn't the code just 10 bytes?
- The `battery_ok` should be 0/1 (not sure where to cut off if there is no battery low flag?),
- And add `"battery_pct", "Battery level", DATA_INT, battery_lvl, // Note: this might change with #3103`

rromanchuk:
I will take another pass shortly. i struggled with battery in general and trying to align it with convention

src/devices/ambientweather_wh31e.c
```diff
@@ -317,6 +343,47 @@ static int ambientweather_whx_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+                decoder_log(decoder, 1, __func__, "WN20 bad SUM");
+                continue; // DECODE_FAIL_MIC
+            }
+
+            int id         = (b[2] << 8) | b[3];
+            int battery_raw  = b[4];
+            float battery_v  = battery_raw * 0.02f; // 2xAAA, nominal ~3.0V
+            int battery_lvl  = battery_raw <= 90 ? 0 : 100 * (battery_raw - 90) / 60; // 1.8V-3.0V is 0-100
+            int rain_raw     = (b[5] << 8) | b[6];
+            char extra[11];
```
merbanan, line 364: This is to large and it needs to be initialized to 0.

merbanan:
Code merged in master.

PR #3474, opened 2026-02-16
Add support for Restaurant Pager (EV1527-variant, 25-bit)

jflaflamme:
## Summary

Decoder for cheap restaurant pager systems (the coaster buzzers you
get at fast food places). Reverse-engineered over many espressos with
an RTL-SDR, capturing signals from a JianTao JT-913 at 315 MHz.
These are sold under dozens of no-name brands and also show up at
433.92 MHz.

Protocol is an EV1527 variant -- 25-bit OOK PWM with a 16-bit
system ID, 4-bit pager address, 4-bit function nibble, and a stop bit.
The base station blasts ~40-50 repeated frames per button press.

Pulse timings from signal analysis: short ~204 us, long ~636 us,
gap ~880 us, reset ~7312 us.

## Test data

.cu8 captures attached with 3 different pager codes (0, 4, B).

[restaurant_pager_captures.zip](https://github.com/user-attachments/files/25331629/restaurant_pager_captures.zip)

## Output

    time      : 2026-02-16 10:33:08
    model     : Restaurant-Pager    System ID : FCFF
    Pager Code: 4                   Function  : Buzz    Raw Command: 4D


zuckschwerdt:
All EV1527-style codes without checksum and without complex decoding are best implemented as flex decoder.
Transmission parameters like timing and coding details might vary and flex decoders make those easy adaptable.

jflaflamme:
> All EV1527-style codes without checksum and without complex decoding are best implemented as flex decoder. Transmission parameters like timing and coding details might vary and flex decoders make those easy adaptable.

Make sense, thanks for this @zuckschwerdt , I will check the Flex decoder approach instead.

zuckschwerdt:
Thanks!

PR #3477, opened 2026-02-19
﻿Add support for Shenzhen Wale WL-TH6R (was: Tuya WSD024-W-433)

DennisKehrig:


DennisKehrig:
Companion PR: merbanan/rtl_433_tests/pull/489

ProfBoc75:
@DennisKehrig

Why not using the bitbuffer_find_repeated_row() function ? like that you can analyse MIC only the repeated row and it's much more simple, no ?

ProfBoc75:
Few advice to simplify your code, lines 83 to 137 can be replaced by these few lines:

bitpos is probably = 0, this is the first bit position. You can directly replace bitpos by 0.


```
    int row = bitbuffer_find_repeated_row(bitbuffer, 4, BITS_PER_ROW); // return the row number if the row is found 4 times into the frame, with a BITS_PER_ROW length (ie 72)

    if (row < 0) {  // repeated rows not satisfied.
        return DECODE_ABORT_EARLY;
    }

    if (bitbuffer->bits_per_row[row] > BITS_PER_ROW) { // row too long
        return DECODE_ABORT_LENGTH;
        }

    bitbuffer_invert(bitbuffer);
    bitbuffer_extract_bytes(bitbuffer, row, bitpos, b, BITS_PER_ROW); // you may replace bitpos by 0, to be confirmed/tested.

```
Then proceed with the dewhitening code.

DennisKehrig:
> Few advice to simplify your code, lines 83 to 137 can be replaced by these few lines:
> 
> bitpos is probably = 0, this is the first bit position. You can directly replace bitpos by 0.
> 
> ```
>     int row = bitbuffer_find_repeated_row(bitbuffer, 4, BITS_PER_ROW); // return the row number if the row is found 4 times into the frame, with a BITS_PER_ROW length (ie 72)
> 
>     if (row < 0) {  // repeated rows not satisfied.
>         return DECODE_ABORT_EARLY;
>     }
> 
>     if (bitbuffer->bits_per_row[row] > BITS_PER_ROW) { // row too long
>         return DECODE_ABORT_LENGTH;
>         }
> 
>     bitbuffer_invert(bitbuffer);
>     bitbuffer_extract_bytes(bitbuffer, row, bitpos, b, BITS_PER_ROW); // you may replace bitpos by 0, to be confirmed/tested.
> ```
> 
> Then proceed with the dewhitening code.

I disagree. This fails to meet the stated goal of reporting every unique valid row with plausible data. Your version rules out such rows if they occur less than four times, mine does not. Your version would also ignore valid data from another sensor contained in the same bitbuffer, mine does not. I have some recordings with five valid rows from one sensor and five valid rows from a second sensor, all in the same recording. And many where they occur less than five times, in several cases only one valid row per sensor made it into the recording.

Example scenario with overlapping transmissions from two sensors:
```
Sensor 1: 1 2 3 4 5
Sensor 2:   1 2 3 4 5
```
Only the first row from sensor 1 and the last row from sensor 2 make it through clearly, the rest will likely cause many shorter rows.

Less severe scenario:
```
Sensor 1: 1 2 3 4 5
Sensor 2:       1 2 3 4 5
```
I'd want to report the data from sensor 1 with count 3 and the data from sensor 2 with count 3, in the same decoder invocation.

If you feel like I can make this more clear in my comments, please point out how.

src/devices/tuya_wsd024w433.c
```diff
@@ -0,0 +1,263 @@
+/** @file
+    Tuya WSD024-W-433 Temperature & Humidity Sensor
```
zuckschwerdt: Needs to be closed with a dot `.`

src/devices/tuya_wsd024w433.c
```diff
@@ -0,0 +1,263 @@
...
+
+#include "decoder.h"
+
+#define BITS_PER_ROW       72
+#define BYTES_PER_ROW      9
+#define DATA_BYTES_PER_ROW 7
+#define MAX_CANDIDATES     4
+
+/**
+# Tuya WSD024-W-433 Temperature & Humidity Sensor.
```
zuckschwerdt: Needs to be a plain first sentence without a H1 heading.

src/devices/tuya_wsd024w433.c
```diff
@@ -0,0 +1,263 @@
...
+
+        decoder_logf(decoder, 2, __func__, "#%u is valid", candidate_index + 1);
+        successes++;
+
+        /* clang-format off */
+        data_t *data = data_make(
+                "model",            "Model",            DATA_STRING,    "Tuya-WSD024W433",
+                "id",               "Sensor ID",        DATA_FORMAT,    "%06X", DATA_INT, sensor_id,
+                "battery_ok",       "Battery",          DATA_COND,      battery_pct < 20, DATA_INT, 0, // The mobile app sends a low battery notification below 20%
+                "battery_pct",      "Battery level",    DATA_FORMAT,    "%d %%", DATA_INT, battery_pct,
```
zuckschwerdt: A `// Note: this might change with #3103` reminder comment might be good until we finalize this

zuckschwerdt:
@ProfBoc75's advice is how we generally handle robustness in decoders. We expect that transmission collisions are rare and not fully recoverable.
If a majority of packets are successfully demodulated, or the MIC is strong we have an easy win without anything extra to do for partially received transmissions.
We consider a bitbuffer with packets from different transmissions to be extremely rare and not recoverable at all.

It's is good that you thoroughly explained the special circumstances (high transmission rate, many senders) in the comments. Generally your code comments are on point and very nice to have!

It sounds like you've done the testing with good results and improved robustness. (Your work on the checksum is especially impressive, AI or not.) We just need to emphasize that the usual decoder should look at lot simpler ;) We need to maintain ~300 after all!

The `bitbuffer_find_repeated_row()` is really not sophisticated enough. Could your scheme be abstracted into some improved variant of that? Would be great to have this as tool where needed or maybe even to benefit all decoders? Ideal case with low complexity in decoders. E.g. some ranking of rows instead of just the one most common row.

I'm not sure about the `count` output. We did ponder something like a "reliability" or "quality" output, e.g. basic case `actual_good_rows / nominal_expected__rows`. We should use this opportunity to find a good standard for this case of reporting or just apply a best guess to report or drop the decoding -- how frequent do you see and how do you rank/trust counts of 1 to 5?

zuckschwerdt:
Do you have search keywords to find these sets? I could not locate any "Tuya WSD024 433", are they mostly the same as the ubiquitous Tuya Zigbee Thermometer Hygrometer sets?

DennisKehrig:
> Do you have search keywords to find these sets? I could not locate any "Tuya WSD024 433", are they mostly the same as the ubiquitous Tuya Zigbee Thermometer Hygrometer sets?

You can find more info in the README.md of the companion PR: https://github.com/merbanan/rtl_433_tests/pull/489
I'm curious whether you feel like most of that info should be in this repo instead, or whether more of it should be duplicated across both. Since I can add images in the rtl433_tests repo I'm more inclined to more details there, but many regular users probably will never even see that repo.
"WSD024-W-433" is what it says on a sticker on the box for the extra sensors when you order a kit more with more than 2, while "WSD023-WIF-433-W12" is the base station (based on a sticker on its box and an image in the description on [AliExpress](https://www.aliexpress.us/item/3256808493034721.html) saying "WSD023").
While the base station has a Tuya CBU module and uses the Tuya cloud infrastructure, the sensors themselves might not actually be made by Tuya, based on what a Tuya support person suggested. He provided details about their BT sensors and when I clarified I'm asking about these 433 MHz ones, he suggested their module doesn't have 433 MHz support and I should contact "the manufacturer". Hmmm. I suspect the actual manufacturer is that SMATRUL seller on AliExpress since most of the other products they sell has their branding, but not this one.
When you search for "WSD024-W-433" you should find this: arendst/Tasmota#23291, but that's just someone else making the same assumption as I did.

DennisKehrig:
> @ProfBoc75's advice is how we generally handle robustness in decoders. We expect that transmission collisions are rare and not fully recoverable. If a majority of packets are successfully demodulated, or the MIC is strong we have an easy win without anything extra to do for partially received transmissions. We consider a bitbuffer with packets from different transmissions to be extremely rare and not recoverable at all.
>
> It's is good that you thoroughly explained the special circumstances (high transmission rate, many senders) in the comments. Generally your code comments are on point and very nice to have!
>
> It sounds like you've done the testing with good results and improved robustness. (Your work on the checksum is especially impressive, AI or not.) We just need to emphasize that the usual decoder should look at lot simpler ;) We need to maintain ~300 after all!

I appreciate that and agree in general. I consider this one a special case (albeit based on limited knowledge of the market) because someone might easily have ten of these sensors, and because they don't transmit regularly unless they see a change in the measurements, so missing a transmission is more consequential than with my WallarGe sensors, which transmit every 31-35 seconds no matter what. And especially if we require a large number of identical rows, chances are you'll ignore valid transmissions from two sensors at the same time. I'd like to empower users to make that choice for themselves, but I recognize that this is a design decision that could be useful more generally and might be odd to have such isolated support for.
Personally, I'd be happy to trust everything that has a valid checksum and occurs at least twice, but I know that means throwing away some likely valid transmissions, too.

During my testing, with the MIC algorithm still being a mystery, I only considered cases where I've seen five identical rows, and frequently I saw the sensor's LED blink, and nothing showed up, which was frustrating. Now as a user I'm glad I can simply display everything with a valid MIC, and highlight when I saw a row only once.
Of course a regular user might not really care at all, depending on the use case. I want to show something like a heatmap of the temperature in my apartment over time after opening a window, so I'd like the data to be as complete as possible.

> The `bitbuffer_find_repeated_row()` is really not sophisticated enough. Could your scheme be abstracted into some improved variant of that? Would be great to have this as tool where needed or maybe even to benefit all decoders? Ideal case with low complexity in decoders. E.g. some ranking of rows instead of just the one most common row.

I agree, that would be nice. Especially if it enabled flex decoders with a config like "bits=72, repeats>=2, unique" to produce data for all unique 72-bit rows that occur at least twice, not just the first one.

The way I approach it here is already fairly general, turning a list of rows into a list of unique rows and how often they occur. I'm limiting it to four candidates which works well in this case where the worst case I've seen is valid data from two sensors and one row having a bad checksum (thus producing a unique third candidate), but that could be adjustable.

The nice thing about bitbuffer_find_repeated_row is that it doesn't need to store a lot of data and can stop early, while my approach has to consider every single row. In specific cases there could be optimizations, like "if the last row is the same as the first, chances are low there'll be valid data from a different sensor inbetween those rows", but a generic solution can't make those assumptions.

One way to make it easier to reuse is with a struct to hold the row index and the number of times the row occurs, then the caller could allocate an array with however many candidates they expect to need at most and pass a pointer to that array to the function. A second variant of the function could omit the count information and just provide indexes of unique rows that are repeated enough times (closer to bitbuffer_find_repeated_row). If the number of candidates is limited, we could choose to replace the oldest existing candidate that has only been seen once so far since duplicates tend to be consecutive, at the risk of missing that the just replaced candidate occurs a second time later, now looking like a new candidate (I hope that makes sense).

One question to ponder is how often anyone will see transmissions from more than two devices in the same batch of rows. If only two are expected, we could do the regular bitbuffer_find_repeated_row to find candidate A, then move backwards from the last row until it we either see A again or find a different row the required number of times.

So in a case like this:

```
Sensor 1: 1 2 3 4 5
Sensor 2:         1 2 3 4 5
```
We could find 1+2 from sensor 1, then 5+4 from sensor 2 and stop, skipping all the rows in the middle.
The user couldn't do their own filtering, but that's not a new situation, for example with the Kerui decoder requiring at least 9 identical rows, with no option for the user to change that limit.

> I'm not sure about the `count` output. We did ponder something like a "reliability" or "quality" output, e.g. basic case `actual_good_rows / nominal_expected__rows`. We should use this opportunity to find a good standard for this case of reporting or just apply a best guess to report or drop the decoding -- how frequent do you see and how do you rank/trust counts of 1 to 5?

Yeah, there's a lot of options for that part. I considered calling it "repeats", but don't like the ambiguity that technically something that is sent once is not repeated at all and therefore should have "repeats: 0", but people tend to think of something transmitted five times as "repeats: 5", not "repeats: 4".
Or similar to battery_ok not showing when it's 1, it could be something like "confident: 0" or "redundant: 0" that's not present when a minimum redundancy threshold has been crossed.
The more standardized it is, rtl_433 could also filter it out itself based on a user configurable setting insted of requiring external post-processing, i.e. users could choose whether they want low confidence values reported.

src/devices/tuya_wsd024w433.c
DennisKehrig: Should I submit the same change for the WallarGe CLTX001 as separate PR or sneak it into this one? I also want to move the comment block there to directly above the decoder function so it doesn't show as documentation for the first `#define` statement, I missed that in the last one.

src/devices/tuya_wsd024w433.c
DennisKehrig: Will add

zuckschwerdt:
> You can find more info in the README.md of the companion PR

My bad. That's actually the preferred way to present the info. General protocol details here and specific device details with the tests.

There'll be some integration of the different types of info soon.

src/devices/tuya_wsd024w433.c
zuckschwerdt: Thanks, I'll fix the other decoder with a direct commit now.

avg-I:
I've recently got the same kind of a sensor.
It looks very similar and uses exactly the same protocol.
I can confirm that the code from this PR works well with the sensor:
```
time      : 2026-03-15 18:11:58
Model     : Tuya-WSD024W433                        Sensor ID : A51C16
Battery level: 109 %     Temperature: 34.2 C       Humidity  : 25 %          Cycle     : 36            Count     : 5x            Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2026-03-15 18:13:39
Model     : Tuya-WSD024W433                        Sensor ID : A51C16
Battery level: 109 %     Temperature: 33.3 C       Humidity  : 25 %          Cycle     : 46            Count     : 5x            Integrity : CHECKSUM
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2026-03-15 18:14:29
Model     : Tuya-WSD024W433                        Sensor ID : A51C16
Battery level: 109 %     Temperature: 33.0 C       Humidity  : 26 %          Cycle     : 51            Count     : 5x            Integrity : CHECKSUM
```
It was trivial to resolve merge conflicts with master, they are only because protocols / devices 298 and 299 have already been merged.

Product link: https://www.aliexpress.com/item/1005010587281687.html
It mentions Onenuo as a brand name and TH15 as a model number.
Via a different product I also found model number TH6R.
Tuya is also mentioned, but I guess it's more about the integration with the WiFi bridge.

There is also this documentation link for a similar set of devices:
https://manuals.plus/ae/1005008114787930

The sensor has a slightly different build internally.
For example, the LED is visible only through a tiny hole behind the back cover.
The front "window" leads to U6 chip on the PCB (looks like it's _the_ sensor).
That chip looks exactly the same as U6 chip in image_1_sensor.jpg from the companion PR.

DennisKehrig:
@avg-I Thank you for your comment, interesting findings! The manual you link to mentions TH16 as the model name for the base station and TH6R as the model name for the sensors, which led to me finding a "SHENZHEN WALE GROUP CO., LTD" that has FCC filings for a [(WL)TH16R](https://fcc.report/FCC-ID/2A2X7-WLTH16R/) (the base station) and a [(WL-)TH6R](https://fcc.report/FCC-ID/2A2X7-WL-TH6R/) (the sensor). Not sure if WL is supposed to indicate wireless or is short for (Shenzhen) Wale. Certainly a bunch of good data points to add to the documentation to make this easier to find.

I'm wondering whether I should rename the model accordingly. Just "TH6R"? "WL-TH6R"? "Shenzhen Wale WL-TH6R"?

As a user, what would you prefer?
1. Missing valid transmissions that only one row was received of
2. Seeing transmissions that have a valid checksum, but may have some corruption regardless

avg-I:
@DennisKehrig, I'd probably choose 1. The sensor transmits frequently enough (comparing to its measurement inertia) that some missing transmissions are not a problem for me. But invalid data may be more of an an issue.

And thank you very much for this PR and all your work on it!

DennisKehrig:
@ProfBoc75 @zuckschwerdt The code now only considers the first row with at least 72 bits that occurs at least twice and ignores everything else

zuckschwerdt:
Thanks for the detective work on the true manufacturer! I'd say "ShenzhenWale-WL-TH6R" is too long for a model id. Maybe "Wale-TH6R" or "WL-TH6R" or "Wale-WLTH6R"?

DennisKehrig:
@zuckschwerdt Makes sense, I went with "WL-TH6R"

src/devices/tuya_wsd024w433.c
```diff
@@ -0,0 +1,197 @@
...
+    // Ignore rows with an invalid MIC value
+    if (b[8] != mic) {
+        decoder_logf(decoder, 2, __func__, "Row index %d has MIC %u, expected %u", row, b[8], mic);
+        return DECODE_FAIL_MIC;
+    }
+
+    // The temperature is an int16_t starting in b[3], use a cast to sign-extend
+    int temp_raw = (int16_t)((b[3] << 8) | b[4]);
+    // The temperature is stored as multiples of 0.1°C
+    double temp_c = temp_raw * 0.1f;
```
merbanan: Use float.

src/devices/shenzhen_wale_wl_th6r.c
```diff
@@ -0,0 +1,197 @@
...
+        return DECODE_ABORT_LENGTH;
+    }
+
+    // Pointer through which we will access the bytes in the selected row
+    uint8_t *b = bitbuffer->bb[row];
+
+    // Invert just the bits in this row
+    for (int i = 0; i < BYTES_PER_ROW; i++) {
+        b[i] = ~b[i];
+    }
```
merbanan, line 96: bitbuffer_invert(bitbuffer); no good ?

src/devices/shenzhen_wale_wl_th6r.c
```diff
@@ -0,0 +1,197 @@
...
+
+    // Dewhiten the data
+    for (int i = 0; i < 7; i++) {
+        b[i] ^= b[7];
+    }
+
+    // Calculate the MIC value (made possible by a >7h run of Claude Opus 4.6 to identify the algorithm)
```
merbanan, line 103: Ahh, VC funds burning :), but seriously Claude Opus 4.6 has proven to be very capable.

src/devices/tuya_wsd024w433.c
```diff
@@ -0,0 +1,197 @@
...
+    // Extract the pairing mode bit
+    int pairing = b[7] >> 7;
+    // Extract the cycle counter. Every ~10s the sensor takes a measurement and increments the counter, but only
+    // transmits when there's enough of a change (e.g. +/- 0.3°C or +/- 3% RH) or more than five minutes have passed.
+    // A good way to test the counter behavior is to press the pairing button, which causes the sensor to send
+    // 13 transmissions in a row, incrementing the counter each time.
+    // Normally the counter goes from 0-64 (0x40) and then starts over, but there's some odd behavior for the
+    // first transmission after inserting batteries where the bits for the counter might be uninitialized,
+    // seemingly reusing memory used for the battery level. When inserting sufficiently fresh batteries this could
+    // lead to counter values > 64, so let's cap the value to something inside the expected range just in case.
```
merbanan: Use /**
        *
        */
style block comments here.

merbanan:
LGTM and regarding temperature sensors they change value slowly so prioritise reporting only good temperatures.

Regarding mic, xors and sums are usually sensitive to all 00 or all ff data just make sure that those vectors does not pass the check. And I think 2*72 bits + mic is a strong enough check to be sure that you have the intended transmission.

src/devices/shenzhen_wale_wl_th6r.c
DennisKehrig: I find it wasteful to invert more than just the one row we actually extract data from. Maybe premature optimization.

Long term it would be great if instead of short/long durations we could specify one/zero durations, and the slicer could use whichever one is lower as short (to ignore shorter pulses), and right away set 0s and 1s accordingly, avoiding the need for inversion altogether.

src/devices/shenzhen_wale_wl_th6r.c
DennisKehrig: Indeed. Various other models and even Codex 5.3 failed to crack this.

src/devices/tuya_wsd024w433.c
DennisKehrig: Really `/**` or just `/*`?

merbanan: /*

DennisKehrig:
> LGTM and regarding temperature sensors they change value slowly so prioritise reporting only good temperatures.

Thank you for the review!

> Regarding mic, xors and sums are usually sensitive to all 00 or all ff data just make sure that those vectors does not pass the check. And I think 2*72 bits + mic is a strong enough check to be sure that you have the intended transmission.

This particular checksum often doesn't change for consecutive values (e.g. only increasing the battery level by 1 often doesn't affect the checksum), sometimes even in groups of 4, and it oscillates, i.e. even larger differences in a single byte often produce the same checksum.
I did find actual corrupted data with valid MICs in practice (easy to detect when it affects the sensor ID since that has a low cardinality). Would be fun to see if other algorithms would have fared better with the same corrupted data I've seen, maybe it's not actually worse than others.

DennisKehrig:
@merbanan Thank you for the review, all done from my end!

PR #3489, opened 2026-03-14
Change [Arad / Master Meter Dialog3G] preamble alignment, overrides, and documentation

avicarmeli:
Improve robustness of the Arad / Master Meter Dialog3G decoder.

Preamble detection now uses a configurable nibble window inside the observed 48-bit preamble while payload extraction always aligns to the start of the full preamble. This keeps decoding stable even if the match window changes.

Add a diagnostic field "unmatched_preamble" which outputs the preamble nibbles outside the match window in inverted nibble form:

    <before>_..._<after>

This helps analyze variants of the Dialog3G transmission.

Add decoder options:
- mandatory serial input
- gear override
- unit override

These allow forcing decoding parameters when meter configuration or protocol variants differ from the automatically detected values.

Update protocol documentation to reflect observed messages:

* the trailing FF byte is not constant
* the serial suffix byte sometimes matches the printed meter letter
* the suffix byte together with the following byte likely indicates meter unit and gear ratio

Add BitBench alignment reference:

    PREAMBLE_ALIGN = c196f51385

(invert of 3e690aec7a)

Format string used for analysis:

    UID:16h SERIAL:<24d 8h 8h
    COUNTER:<32d 8h8h 8h8h 8h8h
    SUFFIX:hh

Example frame used during analysis:

3e690aec7ac84b9aa50900404a7a02000c42bb5e8cc0f8

### DIFF
--- a/src/devices/arad_ms_meter.c
+++ b/src/devices/arad_ms_meter.c
@@ -1,6 +1,6 @@
 /**
  * @file
- * Arad/Master Meter Dialog3G water utility meter (protocol 260).
+ * Arad / Master Meter Dialog3G water utility meter (protocol 260).
  *
  * Copyright (C) 2022 avicarmeli
  *
@@ -113,8 +113,8 @@
  * Several fields previously assumed constant have been observed to vary in
  * newer captures. This documentation reflects current understanding based on
  * empirical analysis and may be updated as additional frames are decoded.
-
-  * ---------------------------------------------------------------------------
+ *
+ * ---------------------------------------------------------------------------
  * BitBench notation
  * ---------------------------------------------------------------------------
  *
@@ -135,10 +135,10 @@
  * ---------------------------------------------------------------------------
  * This decoder accepts arguments through:
  *
- *     rtl_433 -R <id>:<args>
+ *     rtl_433 -R ID:ARGS
  *
  * Mandatory:
- * - serial=<list> / serials=<list>
+ * - serial=LIST / serials=LIST
  *     At least one serial must be provided, otherwise the decoder is silent
  *     and prints a warning once.
  *
@@ -150,27 +150,27 @@
  *         SERIAL-SUFFIX
  *
  *     Examples:
- *         -R <id>:serial=9444602
- *         -R <id>:serials=9444602;1234567;0xfa1c90
- *         -R <id>:serials=09444602-73;01234567-53
+ *         -R ID:serial=9444602
+ *         -R ID:serials=9444602;1234567;0xfa1c90
+ *         -R ID:serials=09444602-73;01234567-53
  *
  *     NOTE:
  *     If other options are combined with serials, commas separate key=value
  *     pairs. In that case prefer semicolons inside serials:
  *
- *         -R <id>:serials=9444602;1234567,gear=0.1,units=m3
+ *         -R ID:serials=9444602;1234567,gear=0.1,units=m3
  *
  * Optional:
- * - gear=<value>
+ * - gear=VALUE
  *     Flow multiplier / resolution.
  *     Allowed values: 0.01, 0.1, 1, 10, 100
  *     Default: 0.1
  *
- * - units=<value>
+ * - units=VALUE
  *     Overrides the native unit interpretation of the meter.
  *     Allowed values (case-insensitive): m3, Liters, CF, USG
  *
- * - convert=<value>
+ * - convert=VALUE
  *     Converts the decoded numeric volume to the requested output unit.
  *     Allowed values (case-insensitive): m3, Liters, CF, USG
  *
@@ -178,7 +178,7 @@
  * Outputs
  * ---------------------------------------------------------------------------
  * - model               : decoder model name
- * - id                  : "SERIAL-SUFFIX"
+ * - id                  : SERIAL-SUFFIX
  *                         serial as 8-digit decimal + '-' + suffix as hex
  * - volume              : decoded volume in selected output units
  * - unit                : output unit string (m3 / Liters / CF / USG)
@@ -188,10 +188,10 @@
  * - unmatched_preamble  : inverted preamble nibbles outside the matched
  *                         nibble window, formatted as:
  *
- *                             <before>_..._<after>
+ *                             BEFORE_..._AFTER
  *
- *                         where <before> is the unmatched part before the
- *                         matched preamble window and <after> is the unmatched
+ *                         where BEFORE is the unmatched part before the
+ *                         matched preamble window and AFTER is the unmatched
  *                         part after it
  *
  * ---------------------------------------------------------------------------
@@ -203,11 +203,11 @@
  *
  * The decoder may search only a configurable nibble window inside this full
  * preamble. The unmatched nibbles are not validated and are reported in the
- * output as "unmatched_preamble" after nibble-wise inversion.
+ * output as unmatched_preamble after nibble-wise inversion.
  *
  * Important:
  * - The payload extraction offset is always derived from the start of the
- *   FULL 48-bit preamble, even if only a sub-range of nibbles is matched.
+ *   full 48-bit preamble, even if only a sub-range of nibbles is matched.
  * - This ensures serial / suffix / counter extraction stays aligned when the
  *   preamble match window is changed.
  *
@@ -249,9 +249,9 @@
  *     auto -> gear override -> units override -> convert
  *
  * Meaning:
- * - gear=<...>   overrides the detected native gear
- * - units=<...>  overrides the detected native unit
- * - convert=<...> converts the numeric output volume to the requested unit
+ * - gear=...    overrides the detected native gear
+ * - units=...   overrides the detected native unit
+ * - convert=... converts the numeric output volume to the requested unit
  *
  * Notes:
  * - units= changes the native interpretation
@@ -1011,12 +1011,11 @@ static r_device *arad_ms_meter_create(char *args)
         if (!val)
             val = (char *)"";
 
-        if (arad_ieq(key, "serial") || arad_ieq(key, "serials")) {
+ if (arad_ieq(key, "serial") || arad_ieq(key, "serials")) {
             collecting_serials = 1;
             serial_buf[0]      = '\0';
             if (*val) {
-                strncat(serial_buf, val, sizeof(serial_buf) - 1);
-                serial_buf[sizeof(serial_buf) - 1] = '\0';
+                snprintf(serial_buf, sizeof(serial_buf), "%s", val);
             }
         }
         else if (arad_ieq(key, "gear")) {
--- a/src/devices/arad_ms_meter.c
+++ b/src/devices/arad_ms_meter.c
@@ -1,271 +1,13 @@
-/**
- * @file
- * Arad / Master Meter Dialog3G water utility meter (protocol 260).
- *
- * Copyright (C) 2022 avicarmeli
- *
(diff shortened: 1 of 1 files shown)


avicarmeli:
Tested locally on Windows/MSVC and in GitHub CI.

Example frames used during analysis:
3e690aec7ac84b9aa50900404a7a02000c42bb5e8cc0f8
3e690aec7ac84b9aa50900404c7a02000c671a4f1a10f8

zuckschwerdt:
The file should start with a line `/**@file` and the first doc-comment block should only contain the copyright.
Can you move the documentation to the decode function doc-comment? Note that not `*` indentation is wanted in the decode function doc-comment.

avicarmeli:
Done.

I updated the file header to the requested /** @file format and moved the full protocol documentation to the decode function. The decode function doc comment was also reformatted to use the no-leading-* style as requested.

ProfBoc75:
@avicarmeli : 
I'm working on this decoder to take into account our last findings with the XOR checksum.
This is impacting the data layout and the sync word pattern value, as the last byte 0x4b is part of the data and not part of the UID, could be 0x6b in case of water leak and because this byte is taken into account into the checksum.
And about this sync word, we can use only the last 2 byte 0x7ac8 to increase the success.
I will update later as I need sometime update the code accordingly.

avicarmeli:
@ProfBoc75 Great, I was about to do that. Go a head. If you look at the last code I pushed there are parameters to select the size of the preamble at lines 795 and 796.  For now it is set to nibble 1 to 8 so from full preamble of 3e690aec7ac84b the check is to match for e690aec7. Proceed with caution when you change the **_`payload_start`_**  variable. Let me know if you need any assistance.

ProfBoc75:
@avicarmeli , I was testing your (impressive) code before adding the checksum feature.

Is there any reason why the serial number is mandatory here ? This is breaking the current decoder.

Notice that rtl_433 is working as a gateway between rf frames and applications without filter, as much simple as possible: transmitting validated data with message integrity check (mic) based on checksum/crc/parity ... if no mic formula the decoder is disabled by default, like it's now the case for this Arad water meter and it's what we want to add.

At rtl_433 level, if we know the unit, then we can properly convert the raw data to the good units, and any other data analysis or filters should be done externally, using python, smart home solution like HA or so on.

This is the case here, you are able to pickup the units, the gear, and you provide accurate counter value, this part can be an improvement for the current decoder, this requires few corrections (see below) but added value.

- We can provide the 2 volume values, the native and the convert, except if both have the same units, only one is needed.
- This means that we must put data conditions to trigger outputs, classic approach already done on other decoders.
- We can improve your result, the unit can be appended to the value with DATA_FORMAT

For example, extract of the 2 improvement we can do into your code: 

```c

    // Output selection
    double out_volume    = volume_m3;  // give volume m3 by default
    arad_unit_t out_unit = ARAD_UNIT_M3; // give volume m3 by default

.../...

            //"volume",       "Volume",        DATA_DOUBLE, out_volume,
            //"unit",         "Unit",          DATA_STRING, arad_unit_str(out_unit),
            //"volume_m3",    "Volume (m3)",   DATA_DOUBLE, volume_m3,
            "volume_m3",    "Volume (M3)",     DATA_COND, native_unit == ARAD_UNIT_M3,                             DATA_FORMAT, "%.3f m3",    DATA_DOUBLE, volume_native,
            "volume_l",     "Volume (Liters)", DATA_COND, native_unit == ARAD_UNIT_LITERS,                         DATA_FORMAT, "%.3f l",     DATA_DOUBLE, volume_native,
            "volume_cu_ft", "Volume (CF)",     DATA_COND, native_unit == ARAD_UNIT_CF,                             DATA_FORMAT, "%.3f cu ft", DATA_DOUBLE, volume_native,
            "volume_gal",   "Volume (USG)",    DATA_COND, native_unit == ARAD_UNIT_USG,                            DATA_FORMAT, "%.3f gal",   DATA_DOUBLE, volume_native,
            "volume_m3",    "Volume (M3)",     DATA_COND, out_unit != native_unit && out_unit == ARAD_UNIT_M3,     DATA_FORMAT, "%.3f m3",    DATA_DOUBLE, out_volume,
            "volume_l",     "Volume (Liters)", DATA_COND, out_unit != native_unit && out_unit == ARAD_UNIT_LITERS, DATA_FORMAT, "%.3f l",     DATA_DOUBLE, out_volume,
            "volume_cu_ft", "Volume (CF)",     DATA_COND, out_unit != native_unit && out_unit == ARAD_UNIT_CF,     DATA_FORMAT, "%.3f cu ft", DATA_DOUBLE, out_volume,
            "volume_gal",   "Volume (USG)",    DATA_COND, out_unit != native_unit && out_unit == ARAD_UNIT_USG,    DATA_FORMAT, "%.3f gal",   DATA_DOUBLE, out_volume,
            "gear",         "Gear",            DATA_DOUBLE, native_gear,
            //"native_unit",  "Native Unit",     DATA_STRING, arad_unit_str(native_unit),

```

Like that, we have always the native value, the first, and the second value is the converted one, or m3 if the native is other units. If native and convert are the same, only one value is reported. Units are appended to the value, more simple use for external processing. I used the official units gal, l, m3, cu ft

What could be done at rtl_433 is to add the volume conversion into the r_api and r_util codes, I had already a discussion about that, as now, the pressure, the temp, the speed and few others are part of the conversion, but not the volumes.
rtl_433 conversion is reclaimed with the `-C native | si | customary` Convert units in decoded output.
Like that, you don't need to add options to the Arad decoder, only provide official units

avicarmeli:
@ProfBoc75 Thanks for the detailed feedback — much appreciated.

Regarding the mandatory serial:
The main reason for enforcing it is not technical but [practical/legal](https://github.com/merbanan/rtl_433/discussions/3414#discussioncomment-15863832). In some regions, receiving and decoding water meter data from neighboring devices may be considered sensitive or even prohibited. This was raised by several users in the community, and the serial filter was added as a safeguard to ensure the decoder only processes explicitly authorized meters.

That said, I understand rtl_433’s role as a generic RF gateway and agree that mandatory filtering is not ideal at this level. I’m open to making the serial filter optional (disabled by default), while still allowing users to enable it when needed.

Regarding units and gearing:
At the moment, I’m not able to reliably derive the exact gearing and units from the transmitted data in all cases. While some patterns exist, they are not fully consistent across meters. For that reason, I exposed optional overrides so users can correct the interpretation when needed.

I agree that the decoder should provide the best possible native interpretation and follow rtl_433 conventions. I’ll review how to better align this with the DATA_COND / DATA_FORMAT approach and minimize custom options where possible.

Thanks again — I’ll incorporate your suggestions and adjust the implementation accordingly.

merbanan:
> In some regions, receiving and decoding water meter data from neighboring devices may be considered sensitive or even prohibited.

This is true for alot of things but we can not comply with all restrictions around the world. IMO it is the user that has to take responsibility for their actions.

src/devices/arad_ms_meter.c
```diff
@@ -7,100 +7,899 @@
...
+} arad_unit_t;
+
+typedef struct {
+    // serial filter (MANDATORY)
+    unsigned count;
+    arad_mm_serial_filter_t items[ARAD_MM_MAX_SERIAL_FILTER];
+
+    // user overrides
+    int user_gear_set;
+    double user_gear;
```
merbanan, line 42: Use float. Here and in all other places.

avicarmeli:
@merbanan @ProfBoc75 
If I sum up your remarks I think I should change the code as follows:
1. Make Serial non-mandatory
2. Remove user unit conversion (but leave the option to change auto- selected  Gear and Units, as we are not sure of the auto-select)
3. Add MIC and make decoder on by default
4. Change the var type as per @merbanan remark.

Please advise if you think I should address differently the raised issues.

ProfBoc75:
@avicarmeli , correct.

I drafted this "simple" decoder below, based on the current "official" one but including MIC, some other sources from Boing and you.
- The UID could be truncated or wrong if the first bytes are shifted or altered, this part can't be validated as outside the checksum.
- Thanks to the checksum which validating the frame, I reduced the sync word to 2 bytes to increase the success (pro) but UID could be wrong (cons).
- I let you test it and copy the checksum part for your final version.
- I tested it on thousand samples you shared into excel files and it works fine.
- I didn't put in place 3 bit corrections with the checksum information.
- I did the water volume conversion with gear value based on your codes.


[arad_ms_meter.c](https://github.com/user-attachments/files/26759160/arad_ms_meter.c)

avicarmeli:
Implemented the requested changes:

* Decoder enabled by default
* Serial filter is now optional
* Added MIC/checksum validation and enabled the decoder by default
* Added checksum-based correction (up to 3 bit errors)
* Removed user-side conversion output logic
* Kept optional gear/unit overrides only, since auto-detection is still not fully reliable
* Changed numeric types to `float` as requested

I kept 4-byte sync matching for robustness and lower false positives. With checksum correction enabled, reception remains good even for weaker signals.

Tested locally on real captures with repeated valid meter receptions and checksum validation.

ProfBoc75:
@avicarmeli , Thx. I will try it this week and hope to confirm by the week-end.

avicarmeli:
@ProfBoc75 Any findings?

ProfBoc75:
> @ProfBoc75 Any findings?

Yes, sorry, I was busy and on vacation

Except the remark at the beginning of the code (include part), I did not review all the code nor all options but I checked the result from few samples.

So, little remark about the integrity message with CHECKSUM-CORR when you did corrections, up to 3.

As I'm not sure how this CHECKSUM-CORR can be handle by external program,
may I propose to keep CHECKSUM and add `corrections` data, to get corrections = 0, 1, 2 or 3

```diff
-             "mic",         "Integrity",     DATA_STRING, corrections > 0 ? "CHECKSUM-CORR" : "CHECKSUM",
+             "corrections", "Corrections",   DATA_INT,    corrections,
+             "mic",         "Integrity",     DATA_STRING, "CHECKSUM",
```

About the volume and the unit:
Like corrections, is it also possible to show the gear conversion, whatever default or set by user. Like that, we are aware about the conversion factor (gear).

For the rest I trust you.

merbanan:
Merged in master

zuckschwerdt:
This needs to be looked at again when someone has the time. Many structs, helpers and this
https://github.com/merbanan/rtl_433/blob/a9cd06aade5eba2640df9fcd4c81597e6ef43337/src/devices/arad_ms_meter.c#L14-L18
will bite use later on for sure.

avicarmeli:
@zuckschwerdt If you can explain what change is needed I will do it.

zuckschwerdt:
Thanks! Review every include and document why you need it. Try to do without the includes.

E.g. `ctype.h` is used for `isspace()`, but that can also be a check for spaces, tabs and newline characters. And are tabs and newline characters even valid here? Should this perhaps just be a check for space?

We want our decoders to be a minimal subset of the C language and GLIBC calls to improve portability and maintainability.

avicarmeli:
@zuckschwerdt 
Thanks, understood.

I will review each added include and remove any that are not strictly needed. I will also simplify the argument parsing so it only accepts the whitespace that is actually valid for decoder arguments, rather than using the broader isspace() behavior.

I will start with removing ctype.h and replacing the generic whitespace handling with explicit checks, then verify whether the remaining standard headers are already provided through the rtl_433 headers or can be avoided by simplifying the helpers.

Just rise tomb up if I am in the right direction.

PR #3493, opened 2026-03-14
Add support for Govee H5059 Water Leak Detector

neffrw:
This adds a new decoder for the Govee H5059 leak sensor.

Background: I bought these without realizing they differ from the H5054 leak detector. As [structure77 pointed out in this reddit comment](https://www.reddit.com/r/Esphome/comments/1jnktrf/comment/nbcv2q9/), the payloads are encrypted. Fortunately, I was able to dump the H5059 firmware on the ARM Cortex-M0+ over JTAG (no readout protection set), which allowed me to reverse engineer the protocol details and encryption key.

Note: I used a mix of tools during investigation and implementation, including JTAG/OpenOCD, Ghidra, rpitx, and several AI assistants to speed up the reverse engineering process and code generation, analysis, and cleanup. Once I had a working Python prototype, I ported the logic into rtl_433 and validated behavior against real RF captures from two different H5059 devices.

I do not have a Govee gateway, so some flag meanings are still best-effort. Field naming is intentionally conservative where meaning is not fully confirmed.
I’d really appreciate feedback from anyone with a gateway who can compare RF events against the state in the govee app, especially around subtype/state transitions and battery-related fields.

I eventually plan to publish an rpi_tx helper repo that can emulate gateway-side packets so users without a gateway can still pair sensors / assign IDs, as this took the most time to get figured out on my end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [297]  ThermoPro TP211B Thermometer
     [298]  TRW TPMS OOK OEM and Clone models
     [299]  TRW TPMS FSK OEM and Clone models
+    [300]  Govee Water Leak Detector H5059
 
 * Disabled by default, use -R n or a conf file to enable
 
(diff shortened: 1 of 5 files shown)


zuckschwerdt:
Thanks! Well written and commented. Kudos for the reversing effort! Ready for merge as-is.

PR #3498, opened 2026-03-27
Add support for EN2058 temperature sensor

srwalter:
Support the EN2058 four probe temperature sensor


merbanan:
LGTM with some small nits.

merbanan:
Decoder merged in master.

PR #3524, opened 2026-04-27
TFA Marbella: add battery status decoding

kickouille:
- Extract battery low flag from msg[6] bit 7
- Add standardized battery_ok field
- Fix counter extraction using correct 3-bit mask (bits 1-3)

Tested with:
- battery low sensor state
- battery replaced sensor state

Both temperature and counter decoding remain unchanged.


zuckschwerdt:
Thanks! Well written PR.

Issue #3525, opened 2026-05-05
Watchman Sonic Advanced - CRC check


steve-burke:
After experimenting with different gains I have finally managed (in a test environment) to get regular half-hour outputs from my Watchman Sonic Advanced.  However, although the message is supposed to be 192 bits long, I always get more than 192 bits (usually 194, occasionally 193).  If any additional bit is a 1 (which it always is when I get 194 bits) the CRC check fails.  193 bit messages generally have a 0 bit at the end, and pass the CRC check. (I use -R234:vvv to see these results.)

Although most messages fail the CRC check, the actual contents seem fine (depth and temperature don't vary by more than expected and other values stay consistent).

Does anyone know: shouldn't the messages be truncated to 192 bits, and is it possible the additional bits are messing up the CRC checks?

Here's a typical output for one reading:

oil_watchman_advanced_decode: failed CRC check
pulse_slicer_pcm: Watchman Sonic Advanced / Plus, Tekelek codes [{194}aaaaaaaaaa2dd40e040153b5cb9066a3906a010703008c954] bits [1010 1010 1010 1010 1010 1010 1010 1010 1010 1010 0010 1101 1101 0100 0000 1110 0000 0100 0000 0001 0101 0011 1011 0101 1100 1011 1001 0000 0110 0110 1010 0011 1001 0000 0110 1010 0000 0001 0000 0111 0000 0011 0000 0000 1000 1100 1001 0101 01]

Incidentally, the status code I am getting is 0x90, and the 'version no.' is 1.7.3.0.

ProfBoc75:
@steve-burke if you get more bits than expected is probably due to wrong timing / wrong pulse, too short , here must be around 500 us. If you are able to increase by 10 us  the value (at the end of decoder, you will need to rebuild) and test ? Could be also a battery issue, no more able to respect the exact timing/ bit rate.

Can you share a sample file to let us analyze this timing with `-S known` option , can be confirm by `-A` option to see pulse values.

Another advice from the decoder documentation : use `-Y minmax` option could help as the modulation is GFSK, not well decoded by rtl_433.

I got more success with smaller preamble (ThermoPro sensor using also GFSK modulation), remove the 3 first byte 0xaa, keep only the 3 last byte, change PREAMBLE_SYNC_LENGTH_BITS to 16 (instead of 40). 

[...]

PR #3526, opened 2026-05-08
minor: Update Oil Watchman Advanced with better CRC validation and Add sensor version

ProfBoc75:
Related to issue #3525.

Because of the GFSK modulation, the frame could contain an extra 1 bit.
Therefore the last part of the frame is shifted by 1 bit but this part contains the CRC 2 byte, in result, lot of CRC failed.

Here the commit improvement by adding a second CRC validation with the last 16 bit of the frame, taking into account the shift.

Data layout:

         PREAMBLE   SYNC LEN MESSAGE                      SHIFTED CRC --> First check failed
    {194}aaaaaaaaaa 2dd4 0e  040153b5cb906389806701070300 3b308

If you shift by 1 bit left 0x3b308 << 1 = 0x7661 = expected crc 16 value ( poly 0x8005, init 0x0000, final XOR 0x0000)

Same message second check with unshifted CRC:

         PREAMBLE   SYNC LEN MESSAGE                      UNSHIFTED CRC --> Second check, CRC is validated
    {194}aaaaaaaaaa 2dd4 0e  040153b5cb906389806701070300 7661

This approach/method may be applied to other sensors, like ThermoPro using also GFSK modulation with lot of failing validations.

### DIFF
--- a/src/devices/oil_watchman_advanced.c
+++ b/src/devices/oil_watchman_advanced.c
@@ -28,21 +28,26 @@ The format might be most easily summarised in a BitBench string:
 
 Data Layout:
 
-- 40 bits of preamble, i.e. 10101010 etc.
-- 2 byte of 0x2dd4 - 'standard' sync word
-- 1 byte - message length, fixed 0x0e (14)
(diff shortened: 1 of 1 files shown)


merbanan:
LGTM in @ProfBoc75 we trust.

PR #3527, opened 2026-05-11
devices/fineoffset.c: Rototill battery voltage to SOC translation

gdt:
This PR is based on my experience with (Amazon Basics, which seem quite good) alkaline cells in the Ecowitt flavor of the device.   The device-read voltage had been 1300 mV for a long time; that's far above the 700 mV in the code for 0%.   Then the voltage dropped to 1200 mV in the middle of the night, and a few hours later I started getting bad values and the device went offline.

### DIFF
--- a/src/devices/fineoffset.c
+++ b/src/devices/fineoffset.c
@@ -664,10 +664,51 @@ static int fineoffset_WH51_callback(r_device *decoder, bitbuffer_t *bitbuffer)
     char id[7];
     snprintf(id, sizeof(id), "%02x%02x%02x", b[1], b[2], b[3]);
     int boost           = (b[4] & 0xe0) >> 5;
-    int battery_mv      = (b[4] & 0x1f) * 100;
-    float battery_level = (battery_mv - 700) / 900.0f; // assume 1.6V (100%) to 0.7V (0%) range
+    int battery_mv_bits = b[4] & 0x1f;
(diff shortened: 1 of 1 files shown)


gdt:
Seems to work.  I have 3 sensors, with battery 1.3, 1.4, 1.6 V, reporting, and sort/uniq-c historram of battery_ok is
```
  27 0.1
  27 0.5
  27 1
```

Therefore requesting review.

zuckschwerdt:
Sorry for the late feedback. Looking at discharge curves there should be some power left at 1200 mV. But it's better to signal battery low early. The change looks reasonable to me.
Note: `battery_mv_bits` should be `battery_raw`
I'd prefer one if else if... for all 5 cases, i.e. roll the 4 cases into the default. Use `{ }` blocks even for single statements.

gdt:
Thanks for the review - adjusted.

I know what you mean that 1200 should have some energy, but the WH51 failed to operate correctly within hours of the the transition from 1300 to 1200.   Ideally, battery% should decline linearly from 100% to 0% from fresh battery to failure of the device -- where remaining power is not relevant.   My intent was to report a % that reflects the most-discharged battery that reports a given level, because battery% is really about "should the human replace the battery now, or when".   I think the values I picked are good enough to give a person looking at the battery% the right impression.    The big point is that once you see 1300, you should change the battery, and that you do have a week or so to do it.

PR #3529, opened 2026-05-12
build: Change version info to generated header file

zuckschwerdt:
This changes the mechanism to define version information for the builds from compiler arguments to a header file.

Allows for incremental builds even across version info changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,9 +35,6 @@ if(GIT_COMMIT) # is a git repo
     git_timestamp(GIT_TIMESTAMP)
     message(STATUS "Found Git version: ${GIT_REFSPEC} commit ${GIT_COMMIT} from ${GIT_TIMESTAMP_ISO}")
     message(STATUS "Using Git version tag: ${GIT_VERSION} on ${GIT_BRANCH} at ${GIT_TIMESTAMP}")
-    ADD_DEFINITIONS(-DGIT_VERSION=${GIT_VERSION})
-    ADD_DEFINITIONS(-DGIT_BRANCH=${GIT_BRANCH})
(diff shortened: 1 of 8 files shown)


zuckschwerdt:
@gdt can you test if this works for you?

One remaining problem though: we would need a `version.h` in the Git tree to allow analyzers and checkers to run. But we don't want the generated `version.h` to dirty the work tree.

zuckschwerdt:
- [x] refactor `version_string()` from r_api to it's own `version.c`

gdt:
Thanks!  I should be able to test later today.    Also, github seems messed up right now.

Wouldn't the analyzers run a build to get other things, and that would write a version.h in the build dir?  Surely everybody else has the same issue.

zuckschwerdt:
E.g. when you start up an editor with Language Server support or just run a static analyzer you won't have the source configured. But that is solved, we actually can include headers depending on wether they exist.

gdt:
Thanks!  I should be able to test later today.    Also, github seems messed up right now.

Wouldn't the analyzers run a build to get other things, and that would write a version.h in the build dir?  Surely everybody else has the same issue.

gdt:
Thanks!  I should be able to test later today.    Also, github seems messed up an hour ago.

Wouldn't the analyzers run a build to get other things, and that would write a version.h in the build dir?  Surely everybody else has the same issue.

gdt:
I checked out your branch, and ran make.   It rebuilt some things, but that's ok.   Then I cherry-picked my pending PR to change fineoffset.c (and change the git commit).   I reran make, and there were 2 recompiles, 1 hit 1 miss, and that seems great.

Reading the diff, it looks fine to me.

Thanks for digging into this!

PR #3540, opened 2026-05-26
fix(http): web UI sends corrupt responses on /

deviantintegral:
This fixes an off-by-one error in calculating the Content-Length header. It also fixes sending two responses to a single request.

You can replicate this by running `-D manual -F http`, and then `curl http://localhost:8433`. You'll get an error about compressed data. If you try to load in a browser, you'll get a variety of different errors depending on the browser used.

This also fixes sending multiple responses - it looks like the `handle_redirect` should have been removed in c76d5e69.

I don't see any tests around the http server, so none added to cover this since that would involve setting up some sort of harness.


zuckschwerdt:
Thanks! Five years and nobody noticed :/

PR #3544, opened 2026-05-30
Add Nedis WIFIWEST500WT 6-in-1 weather station decoder

elcodedocle:
Adds support for Nedis WIFIWEST500WT 6-in-1 weather station

### DIFF
--- a/src/devices/nedis_wifiwest500wt.c
+++ b/src/devices/nedis_wifiwest500wt.c
@@ -1,0 +1,222 @@
+/** @file
+    Nedis WIFIWEST500WT weather station sensor.
+
+    Copyright (C) 2026 elcodedocle
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    @sa https://github.com/make-all/tuya-local/discussions/4136
+
+    Manufacturer : Fujian Youtong Industries Co., Ltd. ??
+    Sold as      : Nedis WIFIWEST500WT (Tuya-compatible)
+
+    --- Radio Layer ---
+
+    Frequency   : 868.4 MHz FSK
+    Modulation  : FSK-PWM over FSK carrier
+                  Each bit is one 7-chip symbol at 122 us/chip (854 us/symbol):
+                    Bit 1 = F1 long (610 us) + F2 short (244 us)
+                    Bit 0 = F1 short (244 us) + F2 long (610 us)
+                  Preamble: ~66 equal-duty symbols (~850 us each) before each packet.
+    Bit rate    : ~1170 baud
+    Packet      : 17 bytes, repeated 8x per burst.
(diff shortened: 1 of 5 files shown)


src/devices/nedis_wifiwest500wt.c
```diff
@@ -0,0 +1,222 @@
+/** @file
+    Nedis WIFIWEST500WT weather station sensor.
+
+    Copyright (C) 2026 elcodedocle
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
```
zuckschwerdt: Please split this comment like other files do. Use `{}` with every if/for/...

elcodedocle:
Be aware that using the lower nibble of wind dir bytes to extend the range of the speed measurements is an unverified guess. I am basing it on user manual claims of 0-50m/s range. The reason I can't verify this easily is that we don't often see wind gusts above 25.5m/s around here. I thought about plugin the anemometer's rotor to a drill or something to reach that speed, but I only have this one sensor and I would prefer not to destroy it. I am trying to set up a radio to send mocks and verify it with the vendor's receiver; Otherwise suggestions are welcomed. I'll write a comment here if I get any relevant validation results.

ProfBoc75:
> Be aware that using the lower nibble of wind dir bytes to extend the range of the speed measurements is an unverified guess. I am basing it on user manual claims of 0-50m/s range. The reason I can't verify this easily is that we don't often see wind gusts above 25.5m/s around here. I thought about plugin the anemometer's rotor to a drill or something to reach that speed, but I only have this one sensor and I would prefer not to destroy it. I am trying to set up a radio to send mocks and verify it with the vendor's receiver; Otherwise suggestions are welcomed. I'll write a comment here if I get any relevant validation results.

Try with a garden sprayer with only air to simulate wind ?
If the weather station is reachable, try with a hair dryer but on the cool setting to avoid damage.

elcodedocle:
> > Be aware that using the lower nibble of wind dir bytes to extend the range of the speed measurements is an unverified guess. I am basing it on user manual claims of 0-50m/s range. The reason I can't verify this easily is that we don't often see wind gusts above 25.5m/s around here. I thought about plugin the anemometer's rotor to a drill or something to reach that speed, but I only have this one sensor and I would prefer not to destroy it. I am trying to set up a radio to send mocks and verify it with the vendor's receiver; Otherwise suggestions are welcomed. I'll write a comment here if I get any relevant validation results.
> 
> Try with a garden sprayer with only air to simulate wind ? If the weather station is reachable, try with a hair dryer but on the cool setting to avoid damage.

although I am definitely going to try some of that and have fun with it I doubt any garden hose or hair dryer can reach 25m/s hurricane wind speed blows (which the device does not seem robust enough to endure, regardless of the spec anyways), but it is a cool suggestion, thank you!

merbanan:
@elcodedocle we need sample cu8 signals so the decoder can be tested. Closing until there are sample signals we can test this code with.

PR #3546, opened 2026-05-31
Change [gridstream] add ci field and surface 0xD2 encrypted frames

swannman:
## Summary

Three improvements to the Landis+Gyr GridStream decoder, validated against a
full Puget Sound Energy deployment capture (CRC init 0x142A):

- **Emit the CI (Control Information) byte** as a new `ci` field. It is the
  first byte the CRC covers (immediately after the length field) and identifies
  the frame class — the field a downstream parser keys on. It is parsed past
  today but never surfaced.
- **Surface 0xD2 / CI=0x52 encrypted frames instead of dropping them.** These
  carry an AES-encrypted payload that ends in a 2-byte authentication tag rather
  than a CRC, so they can never pass `gridstream_checksum` and are silently
  discarded today. They are now detected before the CRC gate and emitted flagged
  `encrypted`.
- **Doc-comment fixes:** label the previously-unmarked CI byte (`N`) in the
  subtype 0x55 / 0xD2 / 0xD5 layouts, and reference a field-level protocol writeup.

Relates to #2531.

## Validation

1. **Live over-the-air capture** (AirSpy R2): the new `ci` field parses
   correctly on live PSE frames — e.g. `subtype=0xD5, ci=0x22` and `ci=0x29`,
   CRC-validated against 0x142A.
2. **Corpus replay:** real 0xD2/CI=0x52 frames from an 84k-frame PSE corpus,
   GFSK-remodulated and fed back through the decoder, now emit
   `{subtype:210, ci:82, encrypted:1}`. The modulator was self-validated by
   round-tripping a known CRC-OK 0xD5 frame (decodes back to networkID 142a,
   CRC-valid). None of the 16 known CRC inits validate the 0xD2/CI=0x52 frames,
   confirming the trailer is an auth tag, not a CRC — stock drops all 3,296 of
   them in the corpus.

Full dataset, methodology, and protocol reference:
https://github.com/swannman/gridstream-protocol

## Test plan

- [x] Builds clean
- [x] Existing 0xD5 / 0x55 frames still decode and CRC-validate (round-trip)
- [x] New `ci` field appears on decoded frames
- [x] 0xD2 / CI=0x52 frames emit `encrypted=1` instead of being dropped

### DIFF
--- a/src/devices/gridstream.c
+++ b/src/devices/gridstream.c
@@ -22,21 +22,25 @@ Landis & Gyr Gridstream Power Meters.
 - Syncword v5: 0b0000000001 0b11111111111
 
 This decoder is based on the information from: https://wiki.recessim.com/view/Landis%2BGyr_GridStream_Protocol
+The CI (Control Information) byte and the 0xD2/CI=0x52 encrypted frame handling
+below are validated against a full Puget Sound Energy deployment capture corpus:
+https://github.com/swannman/gridstream-protocol
 Datastream is variable length and bitrate depending on type fields
 Preamble
 Bytes after preamble are encoded with standard uart settings with start bit, 8 data bits and stop bit.
 Data layouts:
     Subtype 55:
-        AAAAAA SSSS TT YY LLLL KK BBBBBBBBBB WWWWWWWWWW II MMMMMMMM KKKK EEEEEEEE KKKK KKKKKK CCCC KKKK XXXX KK
+        AAAAAA SSSS TT YY LLLL NN BBBBBBBBBB WWWWWWWWWW II MMMMMMMM KKKK EEEEEEEE KKKK KKKKKK CCCC KKKK XXXX KK
     Subtype D2:
-        AAAAAA SSSS TT YY LL K----------K XXXX
+        AAAAAA SSSS TT YY LL NN K--------K XXXX
     Subtype D5:
-        AAAAAA SSSS TT YY LLLL KK DDDDDDDD EEEEEEEE II K----------K CCCC KKKK XXXX
+        AAAAAA SSSS TT YY LLLL NN DDDDDDDD EEEEEEEE II K----------K CCCC KKKK XXXX
 - A - Preamble
 - S - Syncword
 - T - Type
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
If the encrypted frames are not valiated by a checksum then false positives could be a problem.

merbanan:
Preamble + sync + 2 fixed bytes. I think we passed the bar to have it enabled by default. @swannman can we get the 12,614 0xD2 messages attached to the ticket?

swannman:
@merbanan the 12,614 0xD2 messages are in the public (anonymized) GridStream corpus: https://github.com/swannman/gridstream-protocol → `capture/corpus.log`, the lines beginning `80FF2AD2` (12,614 of them; 3,296 are the CI=0x52 encrypted type). The LAN addresses are anonymized/scrubbed, but the frame structure, type and CI bytes are untouched, so it's faithful for detection analysis. I can also commit a filtered `corpus_0xD2.log` if a single attached file is easier.

@zuckschwerdt on false positives: the encrypted frames aren't CRC-checked, but detection still requires the full preamble + sync (`80 FF 2A`) plus the fixed type/CI bytes before any payload is accepted, so the trigger surface is narrow. The corpus above is real off-air data if it's useful for evaluating that.

Separately, I've added the first real GridStream IQ to rtl_433_tests as reference captures (cu8 + readme, no expected-json): https://github.com/merbanan/rtl_433_tests/pull/496 — note that these frames only slice with `-Y minmax`; the default detector doesn't pick them up, which is why there's no harness json.

swannman:
Here are the 12,614 0xD2 messages as an attachment: [corpus_0xD2.log](https://github.com/user-attachments/files/28445890/corpus_0xD2.log)

swannman:
Hi folks, was there anything else you’d like to see here?

merbanan:
@swannman merged in https://github.com/merbanan/rtl_433/commit/28f6958c7142ec817194c260a167a4a02e508797

and if you could hook us up with unlimited inference that would be nice :). Thanks for the contribution.

Issue #3541, opened 2026-05-26
Support a simple Bearer token for the http server


deviantintegral:
The http server is great, and I've been working on a [new Home Assistant integration that works over websockets](https://github.com/rtl-433-hass/rtl_433), avoiding the need for an mqtt server in the middle.

However, the http server is not just read, but allows controlling the radio too. If you expose the server on the network, then anyone with access could control it. Even for running on localhost, there's a chance of a malicious app or web page trying to control the server.

What if we added a simple, header-based authentication option? For example, something like `--http-token <value>`, that is then passed in with an `Authorization: Bearer <token>` header. This would be a small amount of code to add compared to a full authentication system, but also allow users to set up something more closely matching a typical mqtt setup (where the traffic is typically cleartext, but users would have to sniff a password to connect).

If there's interest in accepting this, I'm glad to work on this.

zuckschwerdt:
Sound great. There were plans for a read-only mode, maybe this could be combined.
As we don't use long args, it would have to be something like `-F http,token=hunter2` or maybe like `user_auth` (read-only) and `admin_auth`.

deviantintegral:
A read-only mode makes sense too, and I like the idea of supporting two different tokens to split out the functionality. I think it would make sense to write some sort of tests around this too. Do you have any preference in how or what toolset? Personally, I've been finding that adding lots of tests (unit + coverage + mutation + end to end) really helps to keep AI on the rails, but I can skip this if it would get in the way of other work.

zuckschwerdt:
I'm not sure we can integrate and support more in-depth tests than the unit tests we have. And even those are pretty bare bones. Maybe something like a Python script to run against the HTTP interface and assert answers?

deviantintegral:
> Maybe something like a Python script to run against the HTTP interface and assert answers?

Exactly my thinking. I have a proof of concept locally that actually instruments code coverage from test scripts like this. I'll get a PR up soon.

gdt:
It's not clear to me what the right approach is; this is a general authorization problem, and one might want to authorize readers too.    To me, the obvious thing is to have rtl_433 have a set of tuples { username, password, access-set } where access-set is one or more tokens that denote ability to read samples, ability to configure, ?   It's arguably not ok to let randoms on the local net read samples without authorization.

PR #3548, opened 2026-06-01
tests: add HTTP/WS server test coverage

deviantintegral:
A first pass to add more comprehensive tests around the http server. Highlights include:

- Code coverage metrics
- A basic integration test that runs with no radio frames, but just makes sure the http server is functional.
- A mini python script that mimics rtl_tcp so we can inject frames to actually test websocket events.

I've read through this, but leaving as draft for now to think on it more. If anyone is OK with approving workflows, I'd be curious to see if these pass.

Refs: #3541

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -78,6 +78,28 @@ jobs:
         run: |
           ./tests/symbolizer.py check
 
+  coverage_check_job:
+    runs-on: ubuntu-latest
+    name: Run tests with code coverage
+    steps:
+      - uses: actions/checkout@v5
+      - name: Setup
+        run: |
+          sudo apt-get update -q -y
+          sudo apt-get install -q -y --no-install-recommends cmake gcovr curl python3
+      - name: Build with coverage and run tests
+        # Builds with -DENABLE_COVERAGE=ON, runs ctest (incl. the HTTP server
+        # integration/dataflow/ws tests), and writes an HTML report. Exits
+        # non-zero if any test fails.
+        run: ./tests/coverage.sh "$GITHUB_WORKSPACE/build-coverage"
+      - name: Upload coverage report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-report
+          path: build-coverage/coverage-report
+          if-no-files-found: warn
+
   analyzer_check_job:
     # https://github.com/actions/virtual-environments
     # - Ubuntu 24.04 ubuntu-24.04
--- a/tests/http-integration-test.sh
+++ b/tests/http-integration-test.sh
@@ -1,0 +1,178 @@
+#!/bin/sh
+#
+# Black-box integration test for the rtl_433 HTTP/WS API server.
+#
+# Starts rtl_433 in "manual" device mode (-D manual), which runs the HTTP
+# server event loop without any SDR hardware or input file, exercises each
+# endpoint over the loopback interface, then shuts the server down with
+# SIGTERM so it exits cleanly (this is what lets gcov flush .gcda files when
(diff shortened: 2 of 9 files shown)


zuckschwerdt:
Thanks! I'm not sure how we would enable to run PR-added workflows? But it should run on your fork repo, right?

deviantintegral:
I think GitHub requires approval for forked PRs by default in new repos unless you flip the security setting. I assumed that was the case here!

.github/workflows/check.yml
```diff
@@ -78,6 +78,28 @@ jobs:
...
+      - uses: actions/checkout@v5
+      - name: Setup
+        run: |
+          sudo apt-get update -q -y
+          sudo apt-get install -q -y --no-install-recommends cmake gcovr curl python3
+      - name: Build with coverage and run tests
+        # Builds with -DENABLE_COVERAGE=ON, runs ctest (incl. the HTTP server
+        # integration/dataflow/ws tests), and writes an HTML report. Exits
+        # non-zero if any test fails.
+        run: ./tests/coverage.sh "$GITHUB_WORKSPACE/build-coverage"
```
deviantintegral, line 94: Do we want this to fail if test coverage goes down?

Right now, we’re just looking at the http server, but it would also be straightforward to expand this to all files.

.github/workflows/check.yml
```diff
@@ -78,6 +78,28 @@ jobs:
...
+      - name: Upload coverage report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-report
```
deviantintegral, line 99: The HTML reports give line-by-line stats:

Image

tests/http-integration-test.sh
```diff
@@ -0,0 +1,178 @@
...
+HOST=127.0.0.1
+
+if [ ! -x "$RTL_433" ]; then
+    echo "ERROR: rtl_433 binary not found or not executable: $RTL_433" >&2
+    exit 99
+fi
+
+# Pick a free TCP port unless one was given, so parallel ctest runs don't collide.
+if [ -z "${PORT:-}" ]; then
+    PORT=$(python3 -c 'import socket; s=socket.socket(); s.bind(("127.0.0.1",0)); print(s.getsockname()[1]); s.close()' 2>/dev/null)
```
deviantintegral, line 28: This is subject to a race condition, where a port gets used by another process before it’s open, but I also don’t think it’s worth the complexity to make this more robust.

tests/http-integration-test.sh
```diff
@@ -0,0 +1,178 @@
...
+if [ "$ready" != 1 ]; then
+    echo "ERROR: server did not become ready on $BASE" >&2
+    cat "$SERVER_LOG" >&2
+    exit 97
+fi
+
+echo "Server ready on port $PORT (pid $SERVER_PID). Running checks:"
+
+# --- static / UI ---
+expect_status        "GET /  returns 200"          200 "$BASE/"
```
deviantintegral, line 112: We could use https://github.com/bats-core/bats-core for tests and assertions too.

tests/rtl_tcp_serve.py
```diff
@@ -0,0 +1,137 @@
...
+"""
+import argparse
+import math
+import socket
+import struct
+import sys
+import threading
+
+
+def synth_cu8(bits, rate, short_us, long_us, sync_us, gap_us, lead_in_us, reset_us):
```
deviantintegral, line 46: I’m not familiar with this level of decoding to validate this, but I think it’s likely fine given that the PR isn’t touching files in `/src`.

deviantintegral:
This was merged in #3552.

PR #3556, opened 2026-06-03
Add support for Venduro Extractor Hoods

wlcrs:
This PR introduces a new decoder for Venduro extractor hoods (also widely used by Novy as an OEM). It decodes the bidirectional FSK communication between the RF remote control and the extractor hood.

It was tested against my own Venduro extraction hood and [remote](https://www.venduro.be/nl/toebehoren/afstandsbediening) 

#### Protocol Architecture:

* Modulation: FSK_PCM (NRZ) at 868.3 MHz (~64 kbps).
* Hardware: Based on standard CC1101 transceiver formatting.
* Sync Word: 0xD391 (Note: Hardware transmits this as a double sync word D391D391, which the decoder handles).
* Checksum: CC1101 Hardware CRC-16 (Poly 0x8005, Init 0xFFFF) calculated over the explicit length byte and payload.

#### Features Decoded:

* Remote Tx ➔ Hood: Action commands (Light Up/Down/Toggle, Fan Up/Down/Off) and button release states.
* Hood ACK ➔ Remote: Live status updates sent back to the remote, parsing the exact Fan Level (1-4) and the Filter Replacement Alarm bit.
* Extracts Source MAC, Destination MAC, and the rolling Sequence Counter for both transmission directions.

### DIFF
--- a/src/devices/venduro_hood.c
+++ b/src/devices/venduro_hood.c
@@ -89,6 +89,7 @@ static int venduro_hood_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     char src_mac_str[12];
     char dst_mac_str[12];
     char raw_cmd_str[9];
+    char action_buf[64];
     char const *action_str = "Unknown";
     char const *direction  = "Unknown";
     int is_ack             = 0;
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
At a first glance the code style looks good, thanks! Can you add `{ }` also to one-line conditionals?

wlcrs:
I've added the requested braces.

zuckschwerdt:
Missed a few, lines: 39, 46, 60, 71, 74.

wlcrs:
Sorry about that. Should be ok now.

src/devices/venduro_hood.c
```diff
@@ -0,0 +1,215 @@
...
+- Q: 1-byte sequence counter (rolling)
+- C: 4-byte command
+- X: 2-byte CRC-16, poly 0x8005, init 0xFFFF, over length + payload bytes
+
+Remote -> Hood command bytes [11..14]: 00 27 5A CMD
+Hood -> Remote ACK bytes [11..14]:     00 4E 85 STATE
+*/
+static int venduro_hood_decode(r_device *decoder, bitbuffer_t *bitbuffer)
+{
+    if (bitbuffer->num_rows == 0) {
```
zuckschwerdt: Did you encounter this? I don't think we ever call a decoder with no rows.

src/devices/venduro_hood.c
```diff
@@ -0,0 +1,215 @@
...
+
+    // The CC1101 hardware often repeats the sync word. Skip ahead to the true payload.
+    uint8_t tmp[2];
+    while (offset + 32 <= bitbuffer->bits_per_row[row]) {
+        bitbuffer_extract_bytes(bitbuffer, row, offset + 16, tmp, 16);
+        if (tmp[0] != 0xD3 || tmp[1] != 0x91)
+            break;
+        offset += 16;
+    }
+    offset += 16; // Step past the final sync word
```
zuckschwerdt: Put this directly after the first `bitbuffer_search()` at line 45, then we don't need to keep track.

src/devices/venduro_hood.c
```diff
@@ -0,0 +1,215 @@
...
+    uint8_t const sync_word[] = {0xD3, 0x91};
+    unsigned offset = bitbuffer_search(bitbuffer, row, 0, sync_word, 16);
+
+    if (offset >= bitbuffer->bits_per_row[row]) {
+        return DECODE_ABORT_EARLY;
+    }
+
+    // The CC1101 hardware often repeats the sync word. Skip ahead to the true payload.
+    uint8_t tmp[2];
+    while (offset + 32 <= bitbuffer->bits_per_row[row]) {
```
zuckschwerdt: I think two syncwords is the default, one only might be the exception with truncation at the start (it is configurable though).
Maybe something like this to be explicit about the two cases:
```
    uint8_t const sync_word[] = {0xD3, 0x91, 0xD3, 0x91}; // 32 or 16 bit length
    unsigned offset = bitbuffer_search(bitbuffer, row, 0, sync_word, 32);
… else // fallback
    offset = bitbuffer_search(bitbuffer, row, 0, sync_word, 16);
```

src/devices/venduro_hood.c
```diff
@@ -0,0 +1,215 @@
...
+    }
+    offset += 16; // Step past the final sync word
+
+    // Ensure we have at least the length byte
+    if (offset + 8 > bitbuffer->bits_per_row[row]) {
+        return DECODE_ABORT_LENGTH;
+    }
+
+    uint8_t b_len[1];
+    bitbuffer_extract_bytes(bitbuffer, row, offset, b_len, 8);
```
zuckschwerdt: `bitbuffer_extract_bytes(bitbuffer, row, offset, &payload_len, 8);` should also work?

wlcrs:
Thank you for your feedback.

I've simplified the logic and tested it against my device.

wlcrs:
Hi @zuckschwerdt , are there still blocking issues to move this PR forward?

merbanan:
@wlcrs we need cu8 sample signals so we can test the decoder. Closing until samples are available.

PR #3554, opened 2026-06-02
Fix automatic discriminator selection for file input

zuckschwerdt:
The `center_frequency` should be used for discriminator selection in `sdr_callback()`.
With this change frequency information from input files will be used the same way as if it were live input.

This fix adds 13 fails to the tests though. It seems that these samples/protocols need the `classic` discriminator manually set:

- `sharp_spc775/01/sharp-1_917.2M_250k.cu8`
- `sharp_spc775/01/sharp-2_917.2M_250k.cu8`
- `insteon/03/g005_915M_1024k.cu8`
- `thermopro-211b/01/g144_915.1M_1000k_22.9degC.cu8`
- `thermopro-211b/01/g155_915.1M_1000k_23.0degC.cu8`
- `thermopro-211b/01/g097_915.1M_1000k_22.0degC.cu8`
- `thermopro-211b/01/g074_915.1M_1000k_21.2degC.cu8`
- `lacrosse_ltv/BreezePro_LTV-WSDTH01/g001_914.938M_2400k.cu8`
- `lacrosse_ltv/BreezePro_LTV-WSDTH01/g050_914.938M_2400k.cu8`
- `lacrosse_ltv/BreezePro_LTV-WSDTH01/g011_914.938M_2400k.cu8`
- `lacrosse_ltv/BreezePro_LTV-WSDTH01/g012_914.938M_2400k.cu8`
- `lacrosse_ltv/LTV-W1/g007_915M_1000k.cu8`
- `ESIC-EMT7110_power_meter/g016_868.28M_1024k.cu8`

Issue #3549, opened 2026-06-01
cu8 loader produces incomplete readouts


n6ham:
A cu8 loader produces incomplete results depending on a presence of `-f 915M` argument, despite the presence of `915M` in the file name.

In one case, lack of `-f 915M` results in missing one out of two signals
```
 rtl_433 -r g001_915M_1000k.cu8 -F json

{"time" : "@0.139808s", "model" : "ThermoPro-TempSpikeXR", "id" : 138, "color" : "black", "temperature_int_C" : 31.900, "temperature_amb_C" : 32.300, "is_probe" : 1, "probe_batery" : 3, "mic" : "CRC"}

rtl_433 -f 915M -r g001_915M_1000k.cu8 -F json

{"time" : "@0.139808s", "model" : "ThermoPro-TempSpikeXR", "id" : 138, "color" : "black", "temperature_int_C" : 31.900, "temperature_amb_C" : 32.300, "is_probe" : 1, "probe_batery" : 3, "mic" : "CRC"}
{"time" : "@0.161949s", "model" : "ThermoPro-TempSpikeXR", "id" : 138, "color" : "black", "temperature_int_C" : 31.900, "temperature_amb_C" : 32.300, "is_booster" : 1, "booster_battery" : 3, "mic" : "CRC"}
```
In other case, presence of `-f 915M` results in missing one out of three signals
```
rtl_433 -r g004_915M_1000k.cu8 -F json

{"time" : "@0.048643s", "model" : "ThermoPro-TempSpikeXR", "id" : 156, "color" : "white", "is_docked" : 1, "temperature_int_C" : 26.500, "temperature_amb_C" : 32.400, "is_probe" : 1, "probe_batery" : 3, "mic" : "CRC"}
{"time" : "@0.105515s", "model" : "ThermoPro-TempSpikeXR", "id" : 156, "color" : "white", "is_docked" : 1, "temperature_int_C" : 26.500, "temperature_amb_C" : 32.400, "is_probe" : 1, "probe_batery" : 3, "mic" : "CRC"}
{"time" : "@0.162386s", "model" : "ThermoPro-TempSpikeXR", "id" : 156, "color" : "white", "is_docked" : 1, "temperature_int_C" : 26.500, "temperature_amb_C" : 32.400, "is_probe" : 1, "probe_batery" : 3, "mic" : "CRC"}

rtl_433 -f 915M -r g004_915M_1000k.cu8 -F json

[...]

PR #3559, opened 2026-06-06
Fix FSK min/max slicer: add hysteresis to prevent noise chatter

deviantintegral:
I've been having a really difficult time over the past year getting solid reporting from my WH51 soil sensors. I tried different radios, different antenna setups, but would see misses for hours or days. Then I wondered... what if this is a bug in the decoder?

To test, I set up a harness to watch for signals (the battery voltage and ID were good values to look for), and then asked it to see if there was a signal in the noise. Turns out, yes! Here's the change:

> The minmax FSK demodulator sliced bits crossing the midpoint between the tracked min/max levels. On a noisy or weak signal, samples jitter back and forth across that single threshold. This produces spurious Frequency-High <-> Frequency-Low transitions that shatter the bit framing and corrupt the decoded pulse train. This was observed with Fine Offset WH51.
> 
> This fix adds a hysteresis band around the slicing level, sized as 1/8 of the tracked FSK deviation (var_test_max - var_test_min). A transition now requires the signal to move hyst past mid rather than merely touch it. Because the band scales with the measured deviation, it self-adjusts per device and per signal strength.

Unfortunately, I accidentally deleted the sample recording and don't have physical access right now to the radio to plug it in, but I will get that in the next day or two so we can add a test.

### DIFF
--- a/src/pulse_detect_fsk.c
+++ b/src/pulse_detect_fsk.c
@@ -163,15 +163,14 @@ void pulse_detect_fsk_minmax(pulse_detect_fsk_t *s, int16_t fm_n, pulse_data_t *
      * otherwise the min/max trackers won't converge properly
      */
     if (!s->skip_samples) {
+        // Running min/max of the FSK deviation over the package (reset per
+        // package in pulse_detect_fsk_init()). Intentionally tracked without
+        // decay: a fixed per-sample step toward mid is not scaled to the sample
+        // rate and, at high rates, collapses the window faster than one bit
+        // lasts -- which moves mid and shatters the bit framing on weak signals.
         s->var_test_max = MAX(fm_n, s->var_test_max);
         s->var_test_min = MIN(fm_n, s->var_test_min);
         mid = (s->var_test_max + s->var_test_min) / 2;
-        if (fm_n > mid) {
-            s->var_test_max -= 10;
-        }
-        if (fm_n < mid) {
-            s->var_test_min += 10;
-        }
         // Hysteresis band around the slicing level, proportional to the tracked
         // FSK deviation, so a noisy weak signal doesn't produce spurious
         // mid-crossings that shatter the bit framing. Self-scales per device.
--- a/src/pulse_detect_fsk.c
+++ b/src/pulse_detect_fsk.c
@@ -172,6 +172,10 @@ void pulse_detect_fsk_minmax(pulse_detect_fsk_t *s, int16_t fm_n, pulse_data_t *
         if (fm_n < mid) {
             s->var_test_min += 10;
         }
+        // Hysteresis band around the slicing level, proportional to the tracked
+        // FSK deviation, so a noisy weak signal doesn't produce spurious
+        // mid-crossings that shatter the bit framing. Self-scales per device.
+        int16_t const hyst = (s->var_test_max - s->var_test_min) / 8;
 
         s->fsk_pulse_length += 1;
         switch(s->fsk_state) {
@@ -184,7 +188,7 @@ void pulse_detect_fsk_minmax(pulse_detect_fsk_t *s, int16_t fm_n, pulse_data_t *
                 }
                 break;
             case PD_FSK_STATE_FH:
-                if (fm_n < mid) {
+                if (fm_n < mid - hyst) {
                     s->fsk_state = PD_FSK_STATE_FL;
                     fsk_pulses->pulse[fsk_pulses->num_pulses] = s->fsk_pulse_length;
                     s->fsk_pulse_length = 0;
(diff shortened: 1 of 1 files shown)


src/pulse_detect_fsk.c
```diff
@@ -172,6 +172,10 @@ void pulse_detect_fsk_minmax(pulse_detect_fsk_t *s, int16_t fm_n, pulse_data_t *
         if (fm_n < mid) {
             s->var_test_min += 10;
         }
+        // Hysteresis band around the slicing level, proportional to the tracked
+        // FSK deviation, so a noisy weak signal doesn't produce spurious
+        // mid-crossings that shatter the bit framing. Self-scales per device.
+        int16_t const hyst = (s->var_test_max - s->var_test_min) / 8;
```
deviantintegral, line 177: `8` isn't really a scientific number, but at least keeps this as a bit shift instead of requiring real division.

I wonder if with this change if it's run with other samples in the test library if it actually picks up previously undecoded signals.

zuckschwerdt:
Hysteresis sounds like a sensible addition. @merbanan this is your code, is this change sound?

merbanan:
Yepp. But it needs to be tested against a few samples. This will affect the pulse length position. But in theory the previous  transition will be delayed also.

6 juni 2026 20:35:10 Christian W. Zuckschwerdt ***@***.***>:

zuckschwerdt:
> This will affect the pulse length position. But in theory the previous  transition will be delayed also.

Indeed. It might effectively look like an (ideal) low-pass filter or moving average.

deviantintegral:
Here's a sample that doesn't decode on master, but does on this branch.

[wh51_00c50f_915M_1000k.cu8.gz](https://github.com/user-attachments/files/28687961/wh51_00c50f_915M_1000k.cu8.gz)

I went to integrate this into the rtl_433_tests repository, but I realized that as is that repo only is using the classic decoder, regardless of the frequency of the signal. I suppose we could modify it just for this device, but I think a better option would be to:

1. Update the tests to pass in the frequency with `-f` so other settings driven by that are tested appropriately.
2. Assuming the tests don't blow up, we can then run the test suite against this branch to make sure there's nothing unintended.

zuckschwerdt:
> realized that as is that repo only is using the classic decoder, regardless of the frequency of the signal.

We just noticed in #3549 that the frequency is not passed properly, it will be fixed with #3554 but we need to address regressions before. Your change might be a big part of that :)

deviantintegral:
I think I have the regressions fixed at https://github.com/deviantintegral/rtl_433/pull/1. There are two steps I took to get there:

1. Even though this PR’s changes clearly improved decoding for me, over a week it was clear I was still getting some dropouts. I recorded new samples of those and noted how many transmissions were missing.
2. I took the test repo and ran a loop with a goal that my new recordings and all of the test cases identified in #3554 passed.

I’m going to run my branch for a few days to confirm it actually fixed the dropouts. Hopefully that gives me time to fully understand what the proposed changes are 😅 . Otherwise, I wanted to make sure I linked to the draft in case it saves time and the changes are obviously correct.

zuckschwerdt:
While rewritting the FSK pipeline I just noticed that the classic discriminator it is really brittle. In some cases dropping or adding a single sample to the start of an FSK discrimination will break the whole thing. It won't converge on the two frequencies or converge much too late.
E.g. `lacrosse_ltv/LTV-R3/g012_868.3M_1024k.cu8` has a rough start. Skip a few samples at the start and it converges very quickly. Add a single sample (i.e. start the FSK 1 µs earlier) and it gets stuck in FL state. Not great.

merbanan:
Maybe we should drop the classic demod? I'll take a look around the test repo and see what I can find

deviantintegral:
So far, this plus the code at https://github.com/deviantintegral/rtl_433/pull/1 has certainly been an improvement over what's in master, but it's not perfect. One of my WH51's drops out, while two others within a few feet are perfectly solid. Before, all three would drop. But given I'm still getting significant multiple-hour drops, I'm not confident enough to say this code should be opened as a real PR and shipped.

What can I do next? I wonder if we could generate additional samples in the test repository plus finish integration PR tests so we could be building towards something specific.

merbanan:
@deviantintegral computer says that the hysteris divisor should be 3, can you test with that ?

merbanan:
After some testing I am not able to find some optimal parameters that are able to decode all our sample files in the test corpus. The only conclusion to this is that we simply cant have one demodulator that is capable of handling all types of signals. We need more demodulators. In this case we should keep the minmax code as is and tack on another fsk type demod using the same input as the minmax code but implementing the hysteris flow.

I am not sure how far @zuckschwerdt is with the demod rework but that is the only way I see we can take this forward without causing regressions. And we would go from 1 active fsk demod to 3. It should result in more decodes.

Regarding the hysteris demod, when running with /3 it actually started picking up hard fsk signals that currently only was possible to decode with the classic demod. Anyway I think that the hysteris demod should use a divisor !=8.

zuckschwerdt:
Good thing here is that the actual AM/FM *demod* is the cpu heavy part. The *discriminator* (or dichotomization) part (called "detect" in our code) is computational not that expensive. Should not be a problem to run multiple ones.

Re progress, I'm rolling back much of my changes currently as the FSK discriminator nestled inside the OOK discriminator is too brittle to refactor. I can PR a draft is anyone is interested in looking at the problems?
But we can run other, better, discriminators alongside that then.

merbanan:
PR a draft and we'll see where it goes.

PR #3560, opened 2026-06-08
test: run rtl_433_tests in GitHub Actions

deviantintegral:
This PR sets up a test job using the existing rtl_433_tests repository. It's set up to run tests twice:

1. First, with a pinned commit. This one will fail builds, and is pinned so that changes in the test repo can't break all PRs here.
2. Second, tracking `master`, so we can see if newly recorded samples are passing or not.


tests/rtl_433_tests_regression.py
```diff
@@ -0,0 +1,196 @@
...
+import subprocess
+import sys
+
+# Lines emitted by the rtl_433_tests runner that we parse.
+FAIL_RE = re.compile(r"^## Fail with '(.+?)':")
+IGNORE_RE = re.compile(r"^WARNING: Ignoring '(.+?)'")
+MISSING_RE = re.compile(r"^WARNING: Missing '(.+?)'")
+
+
+def enumerate_universe(tests_dir):
```
deviantintegral, line 37: With this name, I expected this to be an infinite loop!

zuckschwerdt:
We had ideas an discussions on how to best run regressions test. Pulling and running the tests repo felt like a heavy workload. One idea was to collect raw bitbuffers and json output from the tests repo and by default run only that. The full demod pipeline would be used only on changes to the demod. Unclear how to detect and switch that though. Also that bitbuffer-to-json input might get stale without care.
Do you see the `rtl_433_tests_baseline.txt` approach as maintainable? Would we update the file on versioned releases (~twice a year) maybe?

deviantintegral:
I wondered about the performance too, but it looks like the jobs complete in about 2 minutes which seems pretty reasonable: https://github.com/merbanan/rtl_433/actions/runs/27145816245/job/80123123330?pr=3560

Do you still think it’s worth optimizing further?

> Do you see the rtl_433_tests_baseline.txt approach as maintainable?

In an ideal world, all samples always pass, and at most there’s a bit of lag where a sample is added to the tests repo, and then a PR bumps the commit hash to pull in a new update (along with rtl_433 improvements that let the new sample pass). I’d expect it to be updated more often than twice a year, likely as new decoders are added or improved. There’s a `--record` option in the python script to update the baseline file. I’ve used processes like this on other projects adding tests and it tends to work pretty well, as long as the goal is to eventually not need it!

deviantintegral:
Back to review. I _thought_ I had discovered an architecture-dependent bug, but that was wrong. Running tests with arm64 proved it was an unrelated issue with my local checkout. We can roll back the matrix if we're worried about the number of jobs, but I do think there's value in proving at least one arm variant works given how common Pi-like machines are.

PR #3564, opened 2026-06-17
Add [Govee-H5310] pool/spa thermometer decoder

Pablo1:
Adds a decoder for the Govee H5310 pool/spa thermometer.

The H5310 uses the same FSK framing (sync word `2c 4c 4a`, 128-byte XOR key,
CRC-16/AUG-CCITT) reverse-engineered by Reece Neff for the H5059 water leak
detector (#3493). This PR adds the four H5310-specific payload layouts:

- **Temperature Update** (LL=0x10): triggered by a forced report or unit-change action
- **Periodic Update** (LL=0x3d): the sensor's autonomous ~10-minute broadcast
- **Status Reply** (LL=0x1f): sent in response to a gateway poll
- **Ping** (LL=0x1c): connectivity heartbeat, no temperature/battery data

Also fixes a bug in the H5059 decoder where top-probe-only water leaks were
not reported (the old condition only matched when both probe flags were set).

Temperature formula: `T_C = (raw - 33168) / 10.0`, confirmed against the
sensor's own display across two physical units (Pool and Spa) at multiple
temperatures. Wire encoding is always Celsius regardless of the display unit
setting.

The H5310 decoder runs at default priority (0); H5059 is set to priority 10
so H5310 claims shared-envelope frames first.

Test captures will be submitted separately to rtl_433_tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,26 +388,27 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [298]  TRW TPMS OOK OEM and Clone models
     [299]  TRW TPMS FSK OEM and Clone models
     [300]  Govee Water Leak Detector H5059
-    [301]  Astrostart 2000 Car Remote (-f 372.4M)
-    [302]  Compustar 1WG3R Car Remote
-    [303]  Chrysler Car Remote (-f 315.1M -s 920k)
-    [304]  Nidec Car Remote (-f 313.8M -s 1024k)
-    [305]  Audiovox PRO-OE3B Car Remote (-f 303.4M)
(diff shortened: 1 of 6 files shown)


zuckschwerdt:
You need to add new protocols at the end. We must not renumber existing protocols.

Pablo1:
Fixed — moved DECL(govee_h5310) to the end of the list and regenerated. H5310 is now at [321].

zuckschwerdt:
@neffrw can you take a look also?

neffrw:
Great find on the top probe flag since I only used the bottom probes when testing the leak detection! Also, thanks for further clarifying the `0x07` subtype.

Happy to see other Govee devices using the same protocol, which will hopefully save us a lot of work for this device family 😁 

It looks good to me!

merbanan:
Decoder merged in master

PR #3568, opened 2026-06-22
Add [Govee-H5112] dual-probe thermometer decoder

Pablo1:
## Summary

Adds a decoder for the Govee H5112 dual-probe thermometer, which transmits over the shared Govee FSK envelope (sync `2c4c4a`, XOR key, CRC-16/AUG-CCITT).

**Sensor encoding**: Both probe temperatures and humidity are packed into one 32-bit little-endian word (`dec[6..9]`) by the device's own firmware. Confirmed by dumping and disassembling the H5112's FR8016HA firmware: it contains a sensor-packing function with this exact symmetric layout — 11 bits each for probe1/probe2 (0.1°C/count, zero-anchored at -40°C) and 10 bits for humidity (0.1%/count). Probe2 is read directly as an absolute value, like probe1 — no per-device state, no wrap-count tracking, no cold-start dependency.

Validated two independent ways: (1) forward-encoding real app-displayed readings reproduces the actual RF bytes exactly across multiple frames/devices/temperature ranges; (2) decoding the in-frame history buffer (no app data involved) traces a smooth, physically continuous temperature curve through real 25.6°C band crossings.

**Frame types decoded:**
- `0x13` periodic sensor report (~10 min interval): full sensor data + 10-entry history buffer at ~1-minute intervals
- `0x71` triggered status response (reply to gateway poll): same sensor fields, no history

Also fixes a latent frame collision between the H5310 and H5112 decoders: H5310 status frames (dec[0]=0x71, dec[9]=0xFF) were being stolen from H5112 0x71 frames at certain probe temperatures. Fixed by adding a `dec[9] != 0xFF` guard in the H5310 status-frame path. Tightens H5059 decoder boundary (DECODE_ABORT_EARLY for unrecognised msg_class) and clarifies POST_ALARM (0x07) semantics.

## Test plan
- [x] Build passes (`make -j4`)
- [x] `ctest` passes (11/11 including style-check)
- [x] Three test captures in companion rtl_433_tests PR: two room-temp devices, one fridge/freezer device with history buffer
- [x] Verified against physical H5112 hardware at room temp, fridge (~2°C), and freezer (~-23°C)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ coverage-report/
 .cache
 .vs
 .idea
+.vscode
 cmake-build-debug/
 
 # Editor files
@@ -31,6 +32,9 @@ cmake-build-debug/
 # Development files
 /_*/
 /rtl_433_tests
+
+# CMake-generated at build time (see include/version.h.in)
+/include/version.h
 
 # File manager files
 .DS_Store
--- a/README.md
+++ b/README.md
@@ -388,26 +388,27 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [298]  TRW TPMS OOK OEM and Clone models
     [299]  TRW TPMS FSK OEM and Clone models
     [300]  Govee Water Leak Detector H5059
-    [301]  Astrostart 2000 Car Remote (-f 372.4M)
-    [302]  Compustar 1WG3R Car Remote
-    [303]  Chrysler Car Remote (-f 315.1M -s 920k)
-    [304]  Nidec Car Remote (-f 313.8M -s 1024k)
-    [305]  Audiovox PRO-OE3B Car Remote (-f 303.4M)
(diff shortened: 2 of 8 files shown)


zuckschwerdt:
The designers couldn't spend 4 (or 6) more bits for absolute temperature? Or send probes alternating? So strange.

Pablo1:
Yeah... it's absolutely insane. I spent more time working on this than I care to admit. Searching through every bit of every message to try to see if those extra bits were somewhere else and we could just decode absolute temperature. Tried a dozen different experiments with two different H5112 units to see if I had missed something. Maybe one day I'll dump the firmware and decompile it and see what they're doing and why.

Pablo1:
Quick update - I couldn't resist. It was a big challenge but managed to bypass the read protection using Claude's help and dump the firmware. Dissassembled it and found the actual encoding. Reimplementing some of the decoding and getting rid of the k counter since there are some extra bits for the probe 2 temperature hiding in some of the bytes for what I thought was just the probe 1 temperature. Update coming shortly.

merbanan:
Code merged in master.

PR #3570, opened 2026-06-29
infactory: accept NC-3982-675 42-bit rows

jtollet:
Some NC-3982-675 variants are sliced as 42-bit rows with two trailing bits after the 40-bit payload. This keeps the existing inFactory decoder and accepts 40/41/42-bit rows; the payload is already aligned, so the existing buffer access is sufficient. It also adds basic channel and humidity sanity checks to avoid false positives from adjacent weather sensor frames. Tested: ctest 11/11, -y {40} 0f305ce761, -y {41} 0f305ce7610, -y {42} 741068f4210, and live RTL-SDR capture at 434.052 MHz with -R 91 only, no -X flex decoder.

### DIFF
--- a/src/devices/infactory.c
+++ b/src/devices/infactory.c
@@ -48,6 +48,9 @@ Payload looks like this:
 
     [00] {40} 0f 30 5c e7 61 : 00001111 00110000 01011100 11100111 01100001
 
+Some NC-3982-675 variants are sliced as 42-bit rows with two trailing bits
+after the 40-bit payload, e.g. {42} 74 10 68 f4 21 0.
+
 (See below for more information about the signal timing.)
 */
 
@@ -66,24 +69,30 @@ static int infactory_crc_check(uint8_t *b)
 
 static int infactory_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 {
-    if (bitbuffer->bits_per_row[0] != 40 && bitbuffer->bits_per_row[0] != 41)
+    if (bitbuffer->bits_per_row[0] != 40 && bitbuffer->bits_per_row[0] != 41 && bitbuffer->bits_per_row[0] != 42) {
         return DECODE_ABORT_LENGTH;
+    }
 
     uint8_t *b = bitbuffer->bb[0];
 
-    // Check that the last 4 bits of message are not 0 (channel number 1 - 3)
-    if (!(b[4] & 0x0F))
+    int channel = b[4] & 0x03;
+    if (!channel) {
         return DECODE_ABORT_EARLY;
+    }
 
-    if (!infactory_crc_check(b))
+    if (!infactory_crc_check(b)) {
         return DECODE_FAIL_MIC;
+    }
 
     int id          = b[0];
     int button      = (b[1] >> 3) & 1;
     int battery_low = (b[1] >> 2) & 1;
     int temp_raw    = (b[2] << 4) | (b[3] >> 4);
     int humidity    = (b[3] & 0x0F) * 10 + (b[4] >> 4); // BCD, 'A0'=100%rH
-    int channel     = b[4] & 0x03;
+
+    if (humidity > 100) {
+        return DECODE_FAIL_SANITY;
+    }
 
     float temp_f    = (temp_raw - 900) * 0.1f;
 


src/devices/infactory.c
```diff
@@ -66,13 +69,14 @@ static int infactory_crc_check(uint8_t *b)
 
 static int infactory_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 {
-    if (bitbuffer->bits_per_row[0] != 40 && bitbuffer->bits_per_row[0] != 41)
+    if (bitbuffer->bits_per_row[0] != 40 && bitbuffer->bits_per_row[0] != 41 && bitbuffer->bits_per_row[0] != 42)
         return DECODE_ABORT_LENGTH;
 
-    uint8_t *b = bitbuffer->bb[0];
+    uint8_t b[5];
+    bitbuffer_extract_bytes(bitbuffer, 0, 0, b, 40);
```
zuckschwerdt: this is not needed as the buffer is already aligned.

zuckschwerdt:
please add `{ }` also to one-line `if`s

zuckschwerdt:
I guess we are reasonably sure that `humidity > 100` never occurs and 0xa0 is the valid error condition.
How about the temp, are we really sure there are not error condition values outside the normal range?

jtollet:
Updated in 50ceb7ee: removed the extra bitbuffer_extract_bytes() call, added braces to the one-line if statements touched by this patch, and dropped the temperature range sanity check. Kept the humidity sanity check only. Re-tested on the target host: ctest 11/11, plus -y checks for the 40-bit example, the observed 42-bit NC-3982-675 row, and the false-positive sample.

zuckschwerdt:
Thanks. Can you post a zip with one 42-bit example .cu8 that previously failed here?

jtollet:
Thanks. I uploaded a zip here:

https://github.com/jtollet/rtl_433/releases/download/pr3570-nc3982-42bit-sample/nc3982_675_42bit_sample.zip

It contains one CU8 sample generated from the observed 42-bit row:

```text
{42} 741068f4210
```

Verification:

```text
parent 3e7bb667: no JSON output with -R 91
current 50ceb7ee: decodes as inFactory-TH id=116 channel=1 temperature_F=77.9 humidity=42
rtl_433 -A: codes : {42}741068f4210
```

SHA256:

```text
29c17a87be94ca2634a431fbac90ff4e7182535b8ac5bc3451b826d793b7944b  nc3982_675_42bit_sample.zip
b0c11a0a5d0baf529d6b2570e12dc7adf934862ea34389d2f1d5072e9a37a324  nc3982_675_42bit_434.052M_250k.cu8
```

zuckschwerdt:
Is it possible to get a cu8 sample from the device? I'd like to see where the demod or discriminator fails. Or maybe the bits are really in the transmission? If you need this private then mail me, I'll delete the sample after inspecting it.

jtollet:
Thanks. I captured a real CU8 sample from the device and uploaded it here:

https://github.com/jtollet/rtl_433/releases/download/pr3570-nc3982-42bit-sample/nc3982_675_real_ch3_434020M_20260630.zip

It contains one real capture from the NC-3982-675 sensor, id 116, channel 3. The receiver was tuned to 434.020 MHz at 250 kS/s; rtl_433 reports the signal around 434.064 MHz.

Verification with current master:

```text
rtl_433 -r g004_434.02M_250k.cu8 -R 91 -M time:iso -M protocol -M level -F json -A

{time: 2026-06-30T13:46:03, protocol: 91, model: inFactory-TH, id: 116, channel: 3, battery_ok: 1, button: 0, temperature_F: 79.100, humidity: 41, mic: CRC, mod: ASK, freq: 434.064, rssi: -1.196, snr: 20.654, noise: -21.850}
```

SHA256:

```text
b90ace06f5b7ce363554dcc5bd4bf77444d8a5085cab80d4a9f065ced947df9a  nc3982_675_real_ch3_434020M_20260630.zip
3ac3c34856815c7e3b9143f0b9601fbff7f72e3e393ee9691cc1a9e862b5627f  g004_434.02M_250k.cu8
```

zuckschwerdt:
Thanks! I think we can spot where the problem is. The end of packet pulse is shortened, a normal pulse here is 500 µs, the last one (after bit 41) is only 50 µs. Then the (warmup, not data) pulse starting the next packet is too soon, there is only a 260 µs gap. rtl_433 is correct in reading this as 42 bit packets :)

g004_434 02M_250k

If we ever set a (PPM) minimum gap length for a data bit then 42 should go to 41 again. And if we demanded a stable (PPM) pulse length then 42 should go to 41 also. If we do both we should get true 40 bit packets.
Those changes might break things on bad reception and generally the workaround of accepting 42 bits is good enough.

jtollet:
Thanks for the detailed analysis. That makes sense, and it is very helpful to understand why rtl_433 legitimately sees those as 42-bit packets from the real signal.

Given the risk of making the generic PPM demodulation stricter and affecting weak receptions, accepting the 42-bit variant in this decoder looks like the right pragmatic fix to me.

PR #3572, opened 2026-07-04
Fix pulse_slicer_dmc missing bitbuffer_clear (closes #3492)

### LINKED ISSUE #3492: several slicers seem to be missing bitbuffer_clear
This report is based on code reading (by a novice to rtl_433) and on seeing output like:
```
codes     : {146}ffff607f34af471f2e79754ecb7fff7b7fffc
codes     : {276}ffff607f34af471f2e79754ecb7fff7b7fffd81fcd2bd1c7cb9e5d53b2dfffdedffff
codes     : {406}ffff607f34af471f2e79754ecb7fff7b7fffd81fcd2bd1c7cb9e5d53b2dfffdedffff607f34af471f2e79754ecb7fff7b7fffc
codes     : {536}ffff607f34af471f2e79754ecb7fff7b7fffd81fcd2bd1c7cb9e5d53b2dfffdedffff607f34af471f2e79754ecb7fff7b7fffd81fcd2bd1c7cb9e5d53b2dfffdedffff
codes     : {666}ffff607f34af471f2e79754ecb7fff7b7fffd81fcd2bd1c7cb9e5d53b2dfffdedffff607f34af471f2e79754ecb7fff7b7fffd81fcd2bd1c7cb9e5d53b2dfffdedffff607f34af471f2e79754ecb7fff7b7fffc
```
The output is produced by using m=OOK_DMC flex decoder and a sample .cu8 file.

I see that several slicers (PWM, PCM, etc) do `bitbuffer_clear` right after `account_event`.
But several others (DMC, etc) do not have `bitbuffer_clear`.

So, it seems that after detecting and reporting a package, such slicers keep appending bits to already reported bits.

I do not see it happening with live reception, only with recorded samples.

munzzyy:
Closes #3492.

`pulse_slicer_dmc()` calls `account_event()` when it hits a reset gap, but unlike every other slicer in this file it never clears the bitbuffer afterward. PCM, PPM, PWM, and manchester_zerobit all do `account_event()` then `bitbuffer_clear(&bits)` right after. DMC just keeps going, so the next message's bits get appended onto the row(s) from the message that was already reported. That matches what the issue reporter saw: the flex decoder output growing longer and repeating itself on every subsequent message instead of resetting.

The fix is one line, added right after the existing `account_event()` call, mirroring the other slicers exactly.

On the maintainer's comment about whether this could be intentional, to avoid splitting DMC into multiple rows on error: that row-splitting (the `bitbuffer_add_row` a few lines up, on a mid-message symbol error) happens before `account_event` is reached. It's unaffected by this change, the two are separate concerns. This clear only fires after a message has already been accounted for, so it just resets state for the next one, same as the sibling slicers.

I went through all the DMC-based device decoders (fordremote, funkbus, hideki, norgo, vaillant_vrt340f, wt450, plus flex.c for generic OOK_DMC use) to check none of them rely on bits persisting across separate messages. None do. They either index row 0 directly expecting one message per call, or iterate rows expecting the fixed repeat structure from a single capture. flex.c just reflects whatever's in the buffer, which is exactly why the growing output was so visible there.

One honest caveat: I don't have an OOK_DMC device on hand, so this is verified by reading the slicer code and the callers, not a live before/after capture. Running the sample .cu8 from the issue through this before merging would confirm it end to end. I'm confident in the read given how consistent the pattern is across the rest of the file, but that capture is the one thing I couldn't do myself.

Also worth flagging since it's visible in the issue thread already: piwm_raw, piwm_dc, and nrzs have the same missing clear. I kept this scoped to DMC since that's what's tracked here, but the same fix would apply to those.

### DIFF
--- a/src/pulse_slicer.c
+++ b/src/pulse_slicer.c
@@ -589,6 +589,7 @@ int pulse_slicer_dmc(pulse_data_t const *pulses, r_device *device)
                 && bits.num_rows > 0) { // Only if data has been accumulated
             //END message ?
             events += account_event(device, &bits, __func__);
+            bitbuffer_clear(&bits);
         }
     }
 
--- a/src/pulse_slicer.c
+++ b/src/pulse_slicer.c
@@ -59,6 +59,9 @@ static int account_event(r_device *device, bitbuffer_t *bits, char const *demod_
         decoder_log_bitbuffer(device, ret > 0 ? 1 : 2, demod_name, bits, device->name);
     }
 
+    // always reset the bitbuffer after accounting, so every slicer starts the next message clean
+    bitbuffer_clear(bits);
+
     return ret;
 }
 
@@ -250,7 +253,6 @@ int pulse_slicer_pcm(pulse_data_t const *pulses, r_device *device)
                 && (bits.bits_per_row[0] > 0 || bits.num_rows > 1)) { // Only if data has been accumulated
 
             events += account_event(device, &bits, __func__);
-            bitbuffer_clear(&bits);
         }
     } // for
     return events;
@@ -329,7 +331,6 @@ int pulse_slicer_ppm(pulse_data_t const *pulses, r_device *device)
                 && (bits.bits_per_row[0] > 0 || bits.num_rows > 1)) { // Only if data has been accumulated
 
             events += account_event(device, &bits, __func__);
-            bitbuffer_clear(&bits);
         }
     } // for pulses
     return events;
@@ -437,7 +438,6 @@ int pulse_slicer_pwm(pulse_data_t const *pulses, r_device *device)
                     || (pulses->gap[n] > s_reset)) // Long silence (OOK)
                 && (bits.num_rows > 0)) {                        // Only if data has been accumulated
             events += account_event(device, &bits, __func__);
-            bitbuffer_clear(&bits);
         }
         else if (s_gap > 0 && pulses->gap[n] > s_gap
                 && bits.num_rows > 0 && bits.bits_per_row[bits.num_rows - 1] > 0) {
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks for the very detailed analysis! This is a great help.
Good that you bring up piwm_raw, piwm_dc. nrzs was added much later, so that is explained through just copying the some other slicer.

But then osv1 just plain returns, what is that about? Could that be a bug too?

zuckschwerdt:
Thanks for the thorough analysis! This is a great help.
Good that you bring up piwm_raw, piwm_dc. nrzs was added much later, so that is explained through copying other slicers.

I clicked through git blame now and, oops, a0b4d09 by accident delete those clear's. Maybe it was a typo, maybe I thought that account_event would reset the bitbuffer, not sure.

If you like, change the PR to restore all clear's or better move it to a last block in account_event.

And then osv1 just aborts remaining pulses with a return? It could well be that remaining pulses can't sensibly be bits but it looks odd.

munzzyy:
Moved it into account_event, per your preference. It's now a single `bitbuffer_clear(bits)` at the end of that function, and I removed the now-redundant explicit clears from pcm, ppm, pwm, manchester_zerobit, and dmc's call sites. piwm_raw, piwm_dc, and nrzs pick up the clear for free without needing any changes of their own.

I checked a0b4d09's actual diff to be sure before changing scope: it deleted the clear from dmc, piwm_raw, and piwm_dc (each had `bitbuffer_clear(&bits);` right after the old inline account_event block, and the refactor dropped it while collapsing the block down to the new helper call). nrzs never had one to begin with since it was added later, so that matches what you said about it just copying an already-broken slicer.

On osv1: I don't think that return is the same kind of bug. Before a0b4d09, osv1 never called bitbuffer_clear either, so there's nothing the refactor deleted there. And structurally it's not a multi-message loop like the others, it parses one preamble + sync + manchester stream per call and returns as soon as it either fails to match the preamble or finishes accounting one message. There's no bits state left over between calls since it's a fresh stack buffer each time, so there's nothing to leak into a next message.

The real question is whether one capture (one reset-gap window from the pulse detector) ever holds two back-to-back OSv1 transmissions. If it did, this function would miss the second one entirely, since it exits on the first account_event. That's a "are we dropping messages" question, separate from the corrupted-bits-carried-forward bug this PR fixes. I'd rather not touch it here since I don't have a real OSv1 capture with two messages in one window to check against, but I can open a separate issue for it if you want it tracked.

zuckschwerdt:
Agreed, best not to change osv1. The slicer was built for one protocol and would have been tested extensively.

PR #3574, opened 2026-07-05
Fix [secplus_v2] doc comment had order/inversion nibbles swapped

### LINKED ISSUE #3557: secplus_v2 note at the top is incorrect
The note at the top of the file is incorrect

https://github.com/merbanan/rtl_433/blob/master/src/devices/secplus_v2.c

the bitmapping is supposed to be order first, then inversion. The code is correct

```
    bitbuffer_extract_bytes(bits, 0, start_pos, buffy, 8);
    start_pos += 8;
    order = buffy[0] >> 4;

    invert = buffy[0] & 0x0f;
```

above code snippet shows order is pulling from the top 4 bits while invert is pulling from the bottom 4 bits. I looked at the raw bits to confirm as well

munzzyy:
Closes #3557.

The header comment for the Security+ 2.0 decoder lists the two 4-bit
indicator fields in the wrong order. It currently reads:

    bits = `AA BB IIII OOOO X*30`
    - IIII = inversion indicator (4 bits)
    - OOOO = Order indicator (4 bits).

but the decode function pulls order from the high nibble and inversion
from the low nibble of that byte:

    order = buffy[0] >> 4;
    invert = buffy[0] & 0x0f;

so in actual transmission order it's order first, then inversion. I
checked this against argilo/secplus (the reference implementation this
decoder is based on, linked at the top of the file) and it extracts the
indicator byte the same way: high nibble picks the reorder permutation,
low nibble picks the inversion pattern. So the code's right, the comment
has the two fields backwards.

Swapped the layout string and field descriptions to match what the code
actually does. No functional change, comment only.

### DIFF
--- a/src/devices/secplus_v2.c
+++ b/src/devices/secplus_v2.c
@@ -28,12 +28,12 @@ Data comes in two bursts/packets.
 
 Layout:
 
-    bits = `AA BB IIII OOOO X*30`
+    bits = `AA BB OOOO IIII X*30`
 
 - AA = payload type  (2 bits 00 or 01)
 - BB = FrameID (2 bits always 00)
-- IIII = inversion indicator (4 bits)
-- OOOO = Order indicator (4 bits).
+- OOOO = Order indicator (4 bits)
+- IIII = inversion indicator (4 bits).
 - XXXX....  = data (30 bits)
 
 ---


merbanan:
Can you fix the commit message so it confirms to how the commit history looks like ?

munzzyy:
Done, amended and force-pushed. Looked through the log for how doc-comment-only fixes are titled (e.g. `minor: Fix Oria-WA150KM doc comment`, `minor: Fix various decoder doc comments`) and matched that: `minor: Fix Security+ 2.0 doc comment`. Tree is unchanged, only the message.

PR #3579, opened 2026-07-06
Add support for Microchip HCS362 KeeLoq Hopping Encoder based remotes

merbanan:
Adapted from Christian Zuckschwerdt's untested sketch in PR #3113 and fixed/verified against real hardware captures in issue #3112 (thanks to yashikada for testing and reporting the corrected bit layout). Supports both of the HCS362's transmission modes: PWM (mode 0) and Manchester (mode 1), sharing a single decode callback that branches on the demod modulation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,8 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [325]  Thermor A6N 132TX temperature sensor
     [326]  Acurite/Chaney 985 Refrigerator / Freezer Thermometer
     [327]  Honda Keyfob KR5V2X/1X (-f 433.6M -s 1024k)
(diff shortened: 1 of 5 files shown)


merbanan:
@zuckschwerdt ok to merge ?

zuckschwerdt:
The `hcs362_crc()` function needs to be moved before the main doc-comment so that doc-comment block directly preceeds `hcs362_decode()`, otherwise a `@fn` annotation is needed.

merbanan:
> The `hcs362_crc()` function needs to be moved before the main doc-comment so that doc-comment block directly preceeds `hcs362_decode()`, otherwise a `@fn` annotation is needed.

Fixed

PR #3585, opened 2026-07-07
Add support for Grill Thermometer (RF-T0912)

merbanan:
Remote Grill Thermometer with probe (FCC ID TXRFPT0912), sold under the RF-T0912 model number. Decodes a 16-bit Fahrenheit temperature with an 8-bit checksum, repeated across multiple rows for reliable detection, and reports an overload condition.


zuckschwerdt:
Note that this is a cleanup of and superseeds #2746

The "temperature_F" needs a format of `"%.0f F"` (and thus, DATA_DOUBLE, temp_f).
The include should move up between the two doc-comments and the decoder doc-comment does not need the `@fn`.

PR #2921, opened 2024-05-06
Add support for TR-502MSV remote controller for RC-710 smart sockets

filipkosecek:
I have written a decoder for TR-502MSV smart socket controller, see https://github.com/merbanan/rtl_433_tests/pull/467.

PR #3586, opened 2026-07-07
Add support for TR-502MSV remote smart socket controller

merbanan:
Decoder for the TR-502MSV remote for RC-710DX smart sockets, based on merbanan/rtl_433#2921 by Filip Kosecek.

Applies the simplifications requested in review (direct mask/shift field extraction instead of bit reformatting helpers, table lookups for socket/command strings) and resolves the previously unknown 2-bit checksum: U1 = C ^ S2 ^ S0, U0 = O ^ S1, derived from the sample table posted in the PR discussion and verified against all 13 captured samples in filipkosecek/rtl_433_tests#467.


zuckschwerdt:
Always `{ }` block braces, even for single (if) statements would be good. Otherwise ready for rebase and merge.

PR #3593, opened 2026-07-08
Feat emax ws6262

merbanan:
WS6262 support and some bug fixes


zuckschwerdt:
Does this supersede #3199?

merbanan:
> Does this supersede #3199?

yes

PR #3596, opened 2026-07-10
Minor changes, primarily affecting output format

hdtodd:
The original 'omni.c' mistakenly defined field values as 'float'; define as doubles so output is correct past 6 digits (e.g, for frequency & barometric pressure).

Original code embedded the CRC init value deep in the program: make it a '#define' at the top so it's easy to change.

Change 'light %' to 'humidity 2'; not likely either value would be used but outdoor humidity is a more likely sensor.


src/devices/omni.c
```diff
@@ -8,13 +8,6 @@
...
     (at your option) any later version.
 */
-
-#include "decoder.h"
-
-#define OMNI_MSGFMT_00 0x00
-#define OMNI_MSGFMT_01 0x01
-
-
 /**
```
merbanan, line 17: No

src/devices/omni.c
```diff
@@ -93,7 +86,7 @@ A format=1 message format is provided as a more complete example.
 It uses the Bosch BME688 environmental sensor as a data source.
 It is an indoor-outdoor temperature/humidity/pressure sensor, and the
 message packet has the following fields:
-    indoor temp, outdoor temp, indoor humidity, light level,
+    indoor temp, outdoor temp, indoor humidity, outdoor humidity,
```
merbanan, line 95: yes

src/devices/omni.c
```diff
@@ -104,14 +97,19 @@ For format=1 messages, the message nibbles are to be read as:
      i: id of device, 0-15
      1: sensor 1 temp reading (e.g, indoor),  °C *10, 12-bit, 2's complement integer
      2: sensor 2 temp reading (e.g, outdoor), °C *10, 12-bit, 2's complement integer
-     h: sensor 1 humidity reading (e.g., indoor), %RH as 8-bit integer
-     g: light level reading, % as 8-bit integer
+     h: sensor 1 humidity reading (e.g., indoor),  %RH as 8-bit integer
+     g: sensor 2 humidity reading (e.g., outdoor), %RH as 8-bit integer
```
merbanan, line 107: yes

src/devices/omni.c
```diff
@@ -104,14 +97,19 @@ For format=1 messages, the message nibbles are to be read as:
...
      v: (VCC-3.00)*100, as 8-bit integer, in volts: 3V00..5V55 volts
      c: CRC8 checksum of bytes 1..9, initial remainder 0xaa,
             divisor polynomial 0x97, no reflections or inversions
 */
 
+#include "decoder.h"
+
+#define initCRC 0xaa
+#define OMNI_MSGFMT_00 0x00
+#define OMNI_MSGFMT_01 0x01
```
merbanan: no

src/devices/omni.c
```diff
@@ -121,13 +119,13 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     if (r < 0 || bitbuffer->bits_per_row[r] > 82) {
         decoder_log(decoder, 1, __func__, "omni: Invalid message");
         return DECODE_ABORT_LENGTH;
-    }
+    };
```
merbanan: no

src/devices/omni.c
```diff
@@ -121,13 +119,13 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
 
     // OK, that's our message buffer for decoding
     uint8_t *b = bitbuffer->bb[r];
 
     // Validate the packet against the CRC8 checksum
-    if (crc8(b, 9, 0x97, 0xAA) != b[9]) {
+    if (crc8(b, 9, 0x97, initCRC) != b[9]) {
```
merbanan: no

src/devices/omni.c
```diff
@@ -146,11 +144,10 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     case OMNI_MSGFMT_00:
         ptr = &hexstring[0];
         // print just the 8 data bytes as payload data
-        for (int ij = 1; ij < 9; ij++) {
+        for (int ij = 1; ij < 9; ij++)
```
merbanan: no

src/devices/omni.c
```diff
@@ -146,11 +144,10 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
             ptr += sprintf(ptr, "%02x", b[ij]);
-        }
-        float itemp_c = (float)(((int16_t)( b[1]<<8  | b[2] )) >> 4) * 0.10f;
-        float volts   = ((float)(b[8])) * 0.01f + 3.00f;
+        double itemp_c = (double)(((int16_t)( b[1]<<8  | b[2] )) >> 4) * 0.10;
+        double volts   = ((double)(b[8])) * 0.01 + 3.00;
```
merbanan: no

src/devices/omni.c
```diff
@@ -166,12 +163,12 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
-        float ihum    = (float)b[4];
-        float light   = (float)b[5];
-        float press   = (float)(((uint16_t)(b[6] << 8)) | b[7]) * 0.10f;
-        volts         = ((float)(b[8])) * 0.01f + 3.00f;
+        itemp_c = (double)(((int16_t)( b[1]<<8  | b[2] )) >> 4) * 0.10;
+        double otemp_c = (double) (((int16_t)(b[2]<<12 | b[3]<<4 )) >> 4) * 0.10;
+        double ihum    = (double)b[4];
+        double light   = (double)b[5];
+        double press   = (double)(((uint16_t)(b[6] << 8)) | b[7]) * 0.10;
+        volts          = ((double)(b[8])) * 0.01 + 3.00;
```
merbanan: no

src/devices/omni.c
```diff
@@ -193,9 +190,8 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
         // Print the whole 10 bytes in hex, with id and fmt explicit
         ptr = &hexstring[0];
         // print just the 8 data bytes as payload data
-        for (int ij = 1; ij < 9; ij++) {
+        for (int ij = 1; ij < 9; ij++)
```
merbanan: no

src/devices/omni.c
```diff
@@ -205,7 +201,11 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             "mic",     "Integrity", DATA_STRING,  "CRC",
             NULL);
         break;
-    }
+
+        /* New decoders follow here */
+        break;
+
+    };
```
merbanan, line 209: Change where there are other cu8 needing this.

src/devices/omni.c
```diff
@@ -215,8 +215,8 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 // List the output fields for various message types
 static char const *const output_fields[] = {
         "model",
+        "fmt",
         "id",
-        "channel",
```
merbanan, line 219: Maybe

merbanan:
The code as committed mostly align with the coding standard. That is why I changed it.

src/devices/omni.c
hdtodd: moved back to original position

src/devices/omni.c
hdtodd: moved back to original position; changed to INITCRC

src/devices/omni.c
hdtodd: fixed

src/devices/omni.c
hdtodd: INITCRC

src/devices/omni.c
hdtodd: fixed

src/devices/omni.c
hdtodd: Don't understand the issue here.

src/devices/omni.c
hdtodd: fixed

src/devices/omni.c
hdtodd: fixed by moving the "others" comment & break to place before default; good catch

src/devices/omni.c
merbanan: Keep the type as float.

hdtodd:
Fixed except for the couple involving "double xxxx" rather than "float xxxx" that I didn't understand.

src/devices/omni.c
```diff
@@ -11,10 +11,10 @@
 
 #include "decoder.h"
 
+#define INITCRC        0xaa
```
merbanan: This define is not needed. Just use the hex when needed.

hdtodd: Ah, I wanted to make it easy for others to set the initial CRC.  Is that really an issue?

It would give others a level of security if they wanted to use a different CRC init.

merbanan: Well ideally rtl_433 will be able to decode all variations. Anyway I now understand the purpose of the omni format.

src/devices/omni.c
hdtodd: I interpreted "DATA_DOUBLE" to mean that the variable should be double rather than float.  

I'll revert.

src/devices/omni.c
hdtodd: I did some error insertion analyses a couple of years ago (https://github.com/hdtodd/ISMErrDetect) and found that the init value didn't seem to affect the reliability of the error detection, except for 0x00 being a bad choice.  So choosing a different CRC init value would give an additional level of security for anyone wanting to keep their communications private (and at least one user has been clear that he does want them private). 

So it just seemed like an easy feature to offer.

src/devices/omni.c
merbanan: That is the fault of the c language/abi.

src/devices/omni.c
merbanan: ok fine

hdtodd:
doubles reverted to floats

{"time":"2026-07-11 17:16:35","protocol":337,"model":"Omni-Multisensor","id":9,"channel":1,"temperature_C":29.6,"temperature_2_C":27.6,"humidity":42.0,"light_pct":50.0,"pressure_hPa":1015.4,"voltage_V":4.57,"mic":"CRC","mod":"ASK","freq":433.93942,"rssi":0.72634,"snr":27.04524,"noise":-27.0458}

merbanan:
output_fields[] now lists "fmt" but data_make() still emits the key "channel"

src/devices/omni.c
```diff
@@ -11,10 +11,10 @@
 
 #include "decoder.h"
 
+#define INITCRC        0xaa  // change this to match your device for security
 #define OMNI_MSGFMT_00 0x00
 #define OMNI_MSGFMT_01 0x01
 
-
```
merbanan, line 17: Cosmetics

src/devices/omni.c
```diff
@@ -104,15 +104,14 @@ For format=1 messages, the message nibbles are to be read as:
...
-     g: light level reading, % as 8-bit integer
+     h: sensor 1 humidity reading (e.g., indoor),  %RH as 8-bit integer
+     g: sensor 2 humidity reading (e.g., outdoor), %RH as 8-bit integer
      p: barometric pressure * 10, in hPa, as 16-bit integer, 0..6553.5 hPa
      v: (VCC-3.00)*100, as 8-bit integer, in volts: 3V00..5V55 volts
      c: CRC8 checksum of bytes 1..9, initial remainder 0xaa,
             divisor polynomial 0x97, no reflections or inversions
 */
 
-
```
merbanan, line 115: cosmetics

src/devices/omni.c
hdtodd: ✔️

src/devices/omni.c
hdtodd: ✔︎

hdtodd:
> output_fields[] now lists "fmt" but data_make() still emits the key "channel"

Yes, that's intentional.

I use the "channel" nibble to indicate the message format, so there are 16 possible message formats from any microdevice (Pico, etc).  My omni.c code has two examples and defaults to just print the hex string if "fmt" isn't 0x00 or 0x01.   Within the omni.ino code that nibble is called "fmt" (https://github.com/hdtodd/omnisensor_433/).  The microdevice can transmit data from any number of different devices using the "fmt" field to tell omni.c how to interpret the remaining bits in the data packet.

But I didn't want to introduce another JSON field name in the output file -- I wanted to stay within the standard existing field names.  So I keep the "channel" label in omni.c and print the value of the "fmt" nibble as the channel number.

In both the omni.ino code and the omni.c code, I have this note for programmers who want to add additional formats to be decoded:

NOTE: the rtl_433 decoder, omni.c, uses the "fmt" or "Format" field                 
here, as transmitted to omni.c, to decode the incoming message.                     
But, through omni.c, rtl_433 reports the packet format-field value                  
as "channel" in its published reporting (JSON, for example),                        
in keeping with the standard nomenclature and order of field-name                   
precedence used within rtl_433 for data fields.

merbanan:
output_fields[] is the superset of all possible json keys. I think channel needs to be there otherwise rtl_433 will complain that the csv isnt matching the keys.

hdtodd:
Ah, didn't realize it would cause a problem for csv.  Added "channel" to output_fields[].

src/devices/omni.c
```diff
@@ -215,8 +216,9 @@ static int omni_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 // List the output fields for various message types
 static char const *const output_fields[] = {
         "model",
-        "id",
+        "fmt",
```
merbanan: I dont see this key used in code.

hdtodd: "fmt" removed.  I'm using "channel" for the format field coming from the device.

Issue #1310, opened 2020-02-09
endless loop in m_bus


Mindavi:
Hey, I've managed to trigger an endless loop in m_bus.
For this checksums/CRCs must be disabled, like I've done in https://github.com/Mindavi/rtl_433/tree/fuzzing-no-checksum.

The input `{163} 2a aa aa aa aa 1e ea a6 68 4a 39 69 64 21 49 bb 89 a0 62 f1 30` triggers the bug in that build.

It seems to take place during record decoding.
https://github.com/merbanan/rtl_433/blob/cbe1fb01979723551463760bf70e991fca066275/src/devices/m_bus.c#L203

the `m_bus_decode_records` function seems to return 0 indefinitely, causing a while loop to keep running.

https://github.com/merbanan/rtl_433/blob/cbe1fb01979723551463760bf70e991fca066275/src/devices/m_bus.c#L273-L319

Probably some extra sanity checking might help avoiding this issue, but I'm not sure what exactly is happening in that code.

This issue is probably reproducible with a valid checksum, when that's set correctly

merbanan:
Not sure how this could have triggered. I changed one parsing point that could have consumed 0 bytes. It's not really correct but we don't support any sensor that uses a dif_coding value of 0.

merbanan:
@Mindavi can you check and see if that fixes it?

Mindavi:
Wasn't sure what you were referring to before, but I see the new commits now.

The patch seems to work well, good stuff 👍 

```
rtl_433-fuzzing-no-checksum/build$ ./src/rtl_433 -y '{163} 2a aa aa aa aa 1e ea a6 68 4a 39 69 64 21 49 bb 89 a0 62 f1 30' -F json
rtl_433 version unknown inputs file rtl_tcp
Use -h for usage help and see https://triq.org/ for documentation.
Trying conf file at "rtl_433.conf"...
Trying conf file at "/home/ricsch/.config/rtl_433/rtl_433.conf"...
Trying conf file at "/usr/local/etc/rtl_433/rtl_433.conf"...
Trying conf file at "/etc/rtl_433/rtl_433.conf"...
Registered 121 out of 148 device decoding protocols [ 1-4 8 11-12 15-17 19-21 23 25-26 29-36 38-60 63 67-71 73-100 102-105 108-116 119 121 124-128 130-148 ]
[...]

PR #3597, opened 2026-07-11
Fix uint8_t offset wraparound in m_bus parse_payload

munzzyy:
`parse_payload()` in `src/devices/m_bus.c` walks the M-Bus data records using
a running offset `off`, declared as `uint8_t`:

```c
uint8_t off = block1->block2.pl_offset;
...
while (off < block1->L) {
    ...
    int consumed = m_bus_decode_records(&data, &b[off], ...);
    ...
    off += consumed;
}
```

`m_bus_decode_records()` can return a `consumed` value up to 192 for the
LVAR (variable-length) DIF coding, taken directly from a byte in the
telegram (`m_bus_decode_val()`: `if (b[0] <= 0xbf) return b[0] + 1;`). After
a couple of records the running `off += consumed` overflows the `uint8_t`
accumulator and wraps modulo 256. The wrapped value can still be smaller
than `block1->L` (also a `uint8_t`, max 255), so the loop condition
`off < block1->L` doesn't catch it - the parser jumps backward into bytes
it already processed and starts reinterpreting them as a new DIF/VIF
record.

I worked through the arithmetic on a crafted example (L=250, off starting
around 198, constant per-record consumption of 192): the offset cycles
through a fixed set of residues mod 256 and only escapes the loop when it
happens to land on one of a handful of values >= L. An attacker who can
place the same DIF/VIF/LVAR-style record at the right repeating offsets in
a telegram with valid per-block CRC (public, computable offline) can make
the offset walk cycle for a long time, or indefinitely, instead of
terminating - a hang against an unattended M-Bus gateway.

`out->data` is a fixed 512-byte buffer and `block1->L` tops out at 255, so
this doesn't corrupt memory - it's a parser desync / DoS, not an
out-of-bounds read or write.

This is a different bug from the one fixed for #1310 (that one was a
`consumed == 0` case, already fixed by making DIF/VIF byte consumption
unconditional). That fix didn't touch `off`'s type, so the large-`consumed`
wraparound is still there.

Fix: widen `off` to `unsigned` so the running total can't wrap, and also
bound the loop against `out->length` (the number of bytes actually copied
into `out->data`), not just `block1->L`. For Format B telegrams
`out->length` is a few bytes smaller than `block1->L`, so this is also a
slightly tighter, more correct bound on its own, not just insurance against
the overflow case. This matches the "don't trust a length taken from the
air without bounding it" approach already used elsewhere in the codebase
(Arexx-ML, GridStream).

Not memory-unsafe, so no crash/ASan repro to attach, just the arithmetic.
I didn't have cmake available to build the full project, but I compiled
`m_bus.c` directly against the project's `-Wall -Wextra -Wsign-compare`
flags with no new warnings, and verified the fix with a small standalone
program reproducing the exact offset walk before and after the type
change.

### DIFF
--- a/src/devices/m_bus.c
+++ b/src/devices/m_bus.c
@@ -762,7 +762,13 @@ static void parse_payload(data_t *data, const m_bus_block1_t *block1, const m_bu
 
     // process standard payload
 
-    uint8_t off = block1->block2.pl_offset;
+    // NOTE: off must not be a uint8_t. m_bus_decode_records() can return a
+    // "consumed" value up to 192 (LVAR variable-length field), and adding
+    // that to off several times can exceed 255. A uint8_t accumulator would
+    // wrap around, and the wrapped (small) value can still satisfy
+    // "off < block1->L", making the parser jump backward and re-process
+    // already consumed bytes instead of terminating.
+    unsigned off = block1->block2.pl_offset;
     const uint8_t *b = out->data;
 
     // Data Record Header DRH, contains Data Information Block DIB and Value Information Block VIB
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks for digging in and the very detailed analysis! We usually want to use `int` (as opposed to smaller types) everywhere as C will integer-promote most calculation to that anyway, somehow this slipped in. And `unsigned` (of same size) is perfect for counter types.

PR #3602, opened 2026-07-12
Add Fine Offset / Ecowitt WH52 soil moisture/temperature/EC decoder

Verstreubulator:
## Verification

Verified against **four physical WH52 units**, recorded live off-air at 915 MHz alongside
the manufacturer gateway/app:

- units span **moisture 25–44 %**, **EC 6–107 µS/cm**, **temperature 18.4–21.0 °C**;
- all frames pass the **CRC-8 (poly 0x31) + sum checksum**, and decoded values match the app;
- temperature matched the app to ~0.1 °F; moisture checked at 0 % (air) and 100 % (submerged);
- EC calibrated against a stepped salt-water series (~340–7400 µS/cm).

The four captures are provided in a companion rtl_433_tests PR and each decodes with this
change via `rtl_433 -r <file>`: **https://github.com/merbanan/rtl_433_tests/pull/508**

## What this adds

A decoder for the Ecowitt **WH52**, a 3-in-1 soil probe (moisture %, temperature, EC µS/cm).
It is a Fine Offset device and shares the WH51's FSK/PCM modulation (~58 µs, preamble
`aa2dd4`), but uses family byte `0xA2` and a 24-byte frame with a CRC-8 (poly 0x31) followed
by a sum checksum. The WH51 decoder rejects it on the family byte, so this doesn't affect
existing support. It was reverse-engineered from the captures above; the full byte layout is
documented in the decoder's header comment.

Fields emitted: `model`, `id`, `temperature_C`, `moisture`, `conductivity` (µS/cm),
`battery_V`, `boost`, `mic` (CRC).

## Honest caveats

This is meant as a working building block others can refine, not a definitive spec:

- the EC scale (÷25.6) is an empirical fit from a single unit and may need adjustment;
- the battery-voltage scale is approximate (units observed only across a narrow 1.5–1.6 V range);
- several bytes are constant per unit and not yet characterised (likely factory
  serial / calibration); they're documented in the header but not interpreted.

I'm not an expert here and would genuinely welcome scrutiny and improvements.

## Note for Home Assistant users

Until this ships in a released build, an interim MQTT bridge that decodes the WH52 into Home
Assistant (via a flex decoder) is available here:
**https://gist.github.com/Verstreubulator/7983285ba6f0eaf7ce1524f70868619a**
It is not part of this PR — just a stopgap for anyone who needs the data in the meantime.

## Thanks

To Benjamin Larsson (@merbanan) and the rtl_433 contributors for the tool, and to Peter
(@pbkhrv) for the "rtl_433 (next)" Home Assistant add-on
(https://github.com/pbkhrv/rtl_433-hass-addons), which made the off-air captures easy.

## Notes on tooling (I contribute from Windows, so a couple of caveats)

- `DECL(fineoffset_wh52)` is registered in `include/rtl_433_devices.h` (appended at the end),
  and the source is added to `src/CMakeLists.txt` so it builds.
- `maintainer_update.py` didn't run to completion in my environment (a Python-3.12 / Windows
  path issue), so I added the `CMakeLists.txt` entry by hand. The README / man page /
  `rtl_433.example.conf` regeneration needs a built binary and wasn't run — I'd be grateful if
  a maintainer could regenerate those, or point me to the right way to do it.
- `clang-format` wasn't available to me; the file follows the 4-space WH51 style it's modeled
  on — happy to reformat anything that's off.
- I couldn't build locally, so the decode path was validated by re-reading the four
  rtl_433_tests captures through `rtl_433 -r` (all decode) and comparing against the app.

### DIFF
--- a/src/devices/fineoffset_wh52.c
+++ b/src/devices/fineoffset_wh52.c
@@ -30,35 +30,27 @@ manufacturer's gateway/app readouts, and by controlled calibration:
 
 Data layout:
 
-    Preamble: aa aa aa 2d d4
+Preamble: aa aa aa 2d d4
 
-    24-byte payload (indices after the preamble):
+24-byte payload (indices after the preamble):
 
-        Byte   0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23
-               A2 II II II Bt tt MM rr Ec cc rr Rr RR RR RR RR RR RR RR RR RR RR YY SS
+    Byte   0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23
+           A2 II II II Bt tt MM rr Ec cc rr Rr RR RR RR RR RR RR RR RR RR RR YY SS
 
-    - Byte 0     : family / type = 0xA2 (WH52 signature)
-    - Byte 1..3  : device ID (24-bit, hex)
-    - Byte 4     : bits 7..5 = transmission boost/retry counter;
-                   bits 4..0 = temperature high 5 bits
-    - Byte 5     : temperature low 8 bits.
-                   temperature_C = (((b4 & 0x1F) << 8) | b5) * 0.1 - 40.0
-    - Byte 6     : soil moisture (%), 0x00..0x64
-    - Byte 7     : moisture raw / gain (not fully characterised)
-    - Byte 8     : bits 7..4 = moisture-raw high nibble (not characterised);
-                   bits 3..0 = EC high nibble (bits 16..19 of the 20-bit EC value)
-    - Byte 9     : EC bits 15..8
-    - Byte 10    : EC bits 7..0
-                   ec_raw = ((b8 & 0x0F) << 16) | (b9 << 8) | b10   (20-bit)
-                   conductivity_uS = ec_raw / 25.6   (empirical; 25.6 = 256/10)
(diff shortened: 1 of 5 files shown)


merbanan:
You need to run maintainer update also. But looks ok otherwise.

src/devices/fineoffset_wh52.c
```diff
@@ -0,0 +1,151 @@
...
+Data layout:
+
+    Preamble: aa aa aa 2d d4
+
+    24-byte payload (indices after the preamble):
+
+        Byte   0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23
+               A2 II II II Bt tt MM rr Ec cc rr Rr RR RR RR RR RR RR RR RR RR RR YY SS
+
+    - Byte 0     : family / type = 0xA2 (WH52 signature)
```
zuckschwerdt: Do not indent lists. Indent is for verbatim/code blocks.

Verstreubulator:
Thank you both for the review — addressed and rebased on master:

- @zuckschwerdt — un-indented the per-byte field list; the byte-layout table is now the only verbatim block.
- @merbanan — ran maintainer_update.py: WH52 is registered as protocol 353 in the README device list and rtl_433.example.conf. The man page has no per-device list, so it needed no change.
- Resolved the rtl_433_devices.h conflict against the Schrader split (kept both DECLs).

Let me know if anything else needs adjusting.

PR #3603, opened 2026-07-13
Refactor Fineoffset WH2 BREAKING change for Agimex Rosenborg 66796

zuckschwerdt:
Enable protocol 356 if you have a Rosenborg-66796.

Removes the create_fn (protocol argument `-R 18:no-wh5`) in favor of a dedicated decoder.


merbanan:
Shouldnt both be enabled by default ?

zuckschwerdt:
The Rosenborg-66796 is likely used by very few people and was not enabled before. Less collision messages this way.

PR #3606, opened 2026-07-14
Add ENABLE_MONGOOSE_SPLIT option to compile Mongoose as split modules

merbanan:
The vendored Mongoose 6.16 ships as one amalgamated src/mongoose.c (~16k lines); compiling it with optimization needs a lot of RAM, which can be a problem on memory-constrained build hosts.

Add an opt-in CMake option, ENABLE_MONGOOSE_SPLIT (default OFF), that re-splits mongoose.c + include/mongoose.h back into their ~30 original modules at configure time via the new tools/unamalgam.py, so each compiles as a small translation unit. The default build is unchanged: it still compiles the amalgamated mongoose.c and needs no Python.

tools/unamalgam.py is a Python 3 port of Cesanta's own un-amalgamation tool (splits on the #ifdef MG_MODULE_LINES / #line markers and restores the /* Amalgamated: #include */ lines), plus the deterministic adaptations rtl_433 needs to build the pieces as separate translation units instead of one:
- an umbrella mongoose.h including the module headers in amalgamation order, defining MG_INTERNAL empty for cross-TU linkage
- #include "mongoose.h" prepended to each module for its feature macros
- mg_uri.c + the mg_http*.c files combined into one TU (mg_http_modules.c) as they share file-static state
- cross-TU prototypes, the feature-header includes before their guards, the SSL backends' mg_internal.h include, and the moved MSG_NOSIGNAL Generated output was verified byte-identical (bar one comment) to a hand-split tree that was itself validated against upstream 6.16.

Generated modules land in the build tree (build/src/mongoose), so the source tree and the default build are untouched. Mongoose 6.16 is end-of-life upstream, so the amalgamated input is frozen.

Verified: both modes build and pass the full ctest suite (15/15, including the HTTP/WS API tests); the split objects export the same globals as the amalgamated object (plus the former MG_INTERNAL statics); and GET / is byte-identical between the two binaries.


gdt:
I have been running with the commit  from `mongoose-split-option-awk` rebased to master for quite some time, and rebuilding, and I think it's ready to merge.

I lean to awk because it's a more basic unixy tool than python, and the resulting implementation feels simpler.

Are you ok with me pushing that commit to master and then cleaning up branches and this PR?     Would you like to do it?  Or is there something problematic with the change that needs discussion?

Issue #2453, opened 2023-03-25
Add Support for TFA 30.3307.02 Windsensor


aZa1905:
Hi there!
I have to preface this with the following:
I have almost none experience when it comes to using Radios / Wireless-stuff. So I'm sorry in advance for not being very helpful.

I recently bought a [TFA 30.3307.02 Windsensor](https://www.tfa-dostmann.de/produkt/solar-funk-windmesser-weatherhub-30-3307/) in the hopes to have a compatibility with rtl_433. It does not, but I thought maybe someone here is so nice and would help me add support or show me how.
I have recorded some samples from that device, I'll attach them below.

To let the device start transmitting, you have to uncover the solar panel and press a small button at the bottom which triggers some kind of initialization sequence. I don't know if it has been received by rtl_433 though. After that it sent some data twice.
I don't know if that is important or not.

![grafik](https://user-images.githubusercontent.com/41471620/227667934-2fa8f52d-e7bf-46d9-a69c-8ddaae734ef1.png)

I'd appreciate any kind of help. And since there is already support for a lot of TFA devices I assume it shouldn't be too hard to add support? But again, I have almost zero experience with this stuff.
If any additional Info is required I will try to supply as much as possible!

[TFA 30330702 windsensor.zip](https://github.com/merbanan/rtl_433/files/11067492/TFA.30330702.windsensor.zip)

ProfBoc75:
Hi @aZa1905 ,

Do you have the WeatherHub and the weather sensor information related each capture like wind speed , wind direction ? 
Without these information it's very difficult or impossible to decode the frame.

Your samples sound to be ook pwm signal , try with rtl_433 -f 868M -X "n=TFA,m=OOK_PWM,s=150,l=650,r=10000", should have something starting with lot fffff then the data related to your weather station.

ProfBoc75:
From the Specifications
Wind speed and gust:
Measuring range: 0.3 … 50 m/s 
Resolution: 0.1 m/s
Measuring range: ± 1m/s < 8m/s | ± 10% > 8m/s (wind speed only)
Wind directions: 16 (resolution 22.5°)
Transmission to gateway: every 7 minutes

Extract from the user guide, this can give some details about the expected data units and multiples.

Try to move manually the weather vane , while your licensing for data, and very slowly , you should have one data that will change from 0 to 15 , 0x0 to 0xf , they are multiples of 22.5 degrees. Usually 0 = North and value clockwise increase.

[...]

PR #3608, opened 2026-07-15
Tfa 30 3307 wind sensor

merbanan:
@zuckschwerdt The code contains a new bit slicer or we can use the pcm mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,6 +451,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [361]* Kidde RF-SM-DC wireless-interconnect smoke alarm
     [362]  Jansite TPMS TY588-EU2
     [363]  Oregon Scientific WMR500 weather station
+    [364]  TFA 30.3307.02 Wind sensor
 
 * Disabled by default, use -R n or a conf file to enable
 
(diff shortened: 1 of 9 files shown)


zuckschwerdt:
The doc-comment needs to exactly preceed the decode function or we need a `@fn` annotation.

The PCM slicer already supports RZ. But this coding "high almost all the time, with brief [...] low dips" reads like Return-to-one or maybe "inverted RZ"? There are returns to one (high) between zero bits, right? We should be able to code this into the PCM slicer. But having it explicit has advantages. Maybe I'll split the RZ out of the PCM too.

Note though that our convention is to use low energy / low frequency / long as zero and high energy / high frequency / short as one. This would not map very well here and should be explained, e.g. "contrary to convention short is the 0-bit and long is the 1-bit in this coding".

zuckschwerdt:
[Return-to-zero inverted (RZI)](https://en.wikipedia.org/wiki/Return-to-zero#Return_to_zero,_inverted) might be a good name, `pulse_slicer_rzi()`.

merbanan:
But which model should I use? New bit slicer or use the fsk pcm way ? I'll rework accordingly.

merbanan:
It is this issue https://github.com/merbanan/rtl_433/issues/2453

zuckschwerdt:
Add the slicer, but name it
`OOK_PULSE_RZI = 13,  ///< OOK Modulation, Return-to-Zero inverted coding, Pulse = 1, No pulse = 0.`

I'll sort through it later. Having it tagged properly as RZI slicer is advantageous.

merbanan:
merged

Issue #2653, opened 2023-10-01
WS90 battery voltage and several others are in mV, while supercap is in V


gdt:
It seems obvious that we should regularize quantities into SI units, or at least all battery-like things should be in the same units.
This issue has two steps:
  1) Confirm or fix docs to state norms (I couldn't find this quickly).
  2) Adjust drivers to meet norms

Split out from #2072 where it is off topic.
@jpochmara @sdalu

gdt:
Data format says `battery_V (battery_mV)`.   I propose to drop the mV form from the docs, so that docs will say V is the right approach, to solve point 1.

sdalu:
The only benefit I see to have `battery_mV` is to avoid dealing with float and rounding errors. Not sure if it is really an issue here.

gdt:
In general, values are floats, and there are not rounding errors that matter due to floating point representation.   That argument would lead to temperatures  in milli-degrees as ints ;-) 

The hard question is where to draw the line between "represent what is in the packet" and "use standard units and in general regularize".  However, we enthusiastically convert numbers that are in units of 0.25psi back to psi, so I think unit conversion is on the "be straightforward" side of the line.

jpochmara:
The point I was trying to make in the other thread was that we should not change the resolution of data when converting between units.  Changing 3.3V to 3300mV makes a statement about the data that we don't know to be true.  The 3.3V could be 3300mV or it could 3399mV or anything in between.  The original data does not have have resolution down to 0.001, only to 0.1.

gdt:
These are simply numbers.  There is absolutely no implied resolution in V, mV, uV, kV or anything else.   So  3300 mV does not make any kind of statement about resolution or accuracy, and is entirely equivalent to 3.3V.   rtl_433 does not have a data type which involves a value and an uncertainty.  If we wanted to do this right, each value would be a tuple containing the value itself and the standard deviation, and people could propagate that.  But this is logging cheap sensors, not a science lab, and the sensors lack the required documentation about their errors -- so I  don't see this as remotely making sense.

I thiink you may be seeing the printed output as honoring what we know about resolution, but I see that as more of a "avoid goofy extra digits that are just rounding noise" and not an attempt to represent precision.

What I have said is that we should steer to only using SI units, and to have everything that is sort of the same thing in a consistent unit.   To me that means V for all voltage measurments, and that doesn't seem to run afoul of what you are worried about, so we don't need to resolve this.

[...]

PR #3613, opened 2026-07-16
docs: standardize battery voltage field on battery_V, deprecate batte…

merbanan:
…ry_mV

Per discussion in #2653, drop battery_mV as an accepted alternate form in the data format docs so battery_V is the single documented norm. Existing decoders that emit battery_mV are not changed here - this is step 1 (confirm/fix the documented norm) before adjusting decoders to match in a follow-up.


merbanan:
@zuckschwerdt yes/no/maybe ?

zuckschwerdt:
Oops, I thought this was merged already;) Sorry, I killed the branch in the clean up just now.

Not a fan of deprecating `battery_mV` as I do see a value in the implied precision of an 4-digit int vs a float.
But I'm not too invested as I don't use those keys. Maybe write "discouraged" instead of "deprecated"?

PR #3619, opened 2026-07-18
Refactor to add proper package flush

zuckschwerdt:
This adds proper on-demand flushing of the demod (discriminator actually).

E.g. on parameter change (freq or rate) and on input file change any currently unfinished packet will be emitted.

This replaces our old workaround of pushing two empty frames through the demod.


merbanan:
LGTM

Issue #1785, opened 2021-08-29
Parameter 'pulse_detect autolevel' could be replaced (and is impractically named)


drws:
The `pulse_detect autolevel` parameter turns on the automatic `minlevel` control and considering other `pulse_detect` parameter names, it has an impractical name which introduces confusion into the group. A more appropriate and nondeceptive name would be `pulse_detect autominlevel` (shorted to `pulse_detect autominlvl` if need be).

Secondly, the `pulse_detect autolevel` is not needed at all since its functionality could be incorporated into `pulse_detect minlevel`, analogous to `pulse_detect level` and other parameters with "auto" settings. The `pulse_detect minlevel=0` could turn on auto-minleveling exactly as `pulse_detect level=0` does for auto-leveling.

zuckschwerdt:
You are right about the autolevel controlling a minimum level. But it's upper bound by and initially value is minlevel.
I.e. `-Y autolevel -Y minlevel=-25` if you known you have a mostly clear band, or `-Y autolevel -Y minlevel=-10` if you known there is much noise.
Using autolevel, squelch, magest is experimental for now but will likely be the default in a few months. In general most of the `-Y` switches are workaround or tests and will be removed or replaced at some time I guess.

drws:
Oh, I can see now why the `-Y minlevel=0` is not a solution. But `-Y autolevel` for auto-min-levelling remains confusing, so I'm still proposing it is renamed to `autominlevel`. Since the `-Y` group is experimental, such change wouldn't strictly require backward-compatible alias.

But the fact that `-Y` is subject to change should be emphasized in documentation, which will at least partially remedy the headaches in the future for both users and devs. The same applies for other work-in-progress parameters, not only pulse-detection-related ones.

gdt:
This is still open 2 years later, and I don't see us changing terms.  If anything, a bug that there is specific documentation missing, or better yet a PR would be useful.   I'm inclined to close this.

drws:
> I don't see us changing terms

Therein lies the problem. Even after 2 years `autolevel` is still a misleading name and changing terms under _experimental_ `-Y` should just be doable.

>  I'm inclined to close this.

Please don't, that's no solution.

gdt:
If you want to make progress, open a PR the changes to new words, and is internally consistent..  I can't guarantee how anyone will react, but a complaint that it should be better is much harder to deal with than an actual change that could perhaps be accepted.

[...]

Issue #1784, opened 2021-08-29
Not all parameters are sufficiently documented in help / manpage / rtl_433.example.conf


drws:
There are two help-related issues regarding parameters' descriptions and their default values. First of all the related documentation is scattered between [help](https://github.com/merbanan/rtl_433/blob/master/src/rtl_433.c), [manpage](https://github.com/merbanan/rtl_433/blob/master/man/man1/rtl_433.1) and/or [`rtl_433.example.conf`](https://github.com/merbanan/rtl_433/blob/master/conf/rtl_433.example.conf). Parameter descriptions are more or less in-sync between these documentation sources, but there are notable differences, some of which are stated below in examples.

The second issue is that some of the `rtl_433` parameters' default settings are not clearly stated in any documentation source. Below I've listed a few examples that stand out the most.

**Parameter `duration`**
Manpage states that the input is in seconds and some other time formats. It fails to mention the default value (run indefinitely) or its numerical value. Example configuration file implies the default value is 0, but it doesn't clearly specify and explain it. It also fails to mention anything regarding input format(s).

**Parameter `hop_interval`**
The fact that this parameter is used only when multiple frequencies are set is emphasized only in the example configuration file, but not in the main help / manpage.

**Parameter `sample_rate`**
The main help fails to mention only a handful of possible values such as in the example configuration file. Default value is exemplary stated in both cases.

**Parameter `gain`**
Manpage fails to mention the numerical value of 0 for auto-setting, while the example configuration file efficiently explains everything with _(default: 0 for auto)_.

**Parameter `convert`**
Help/manual page does not specify the default setting `convert native` (which probably means `convert no`).

**Parameter group `pulse_detect`**
The default values are missing from the manpage and only implied in example configuration files. Default behaviour unambiguitiy is additionaly confusing in cases of `pulse_detect autolevel` and `pulse_detect squelch`. Are they used by default or not?

These are the cases that stand out the most but all in all, there is lot of important information missing in help/man page and this tool practically relies on documenting input parameters in the example configuration file. This can be an issue if these documentation sources are not at least included/autosynced/linked somehow.

zuckschwerdt:
Thanks for taking the time to document this in detail! The man page and example conf should be a copy of the help. So we need to enhance the help, the improve the copy.

drws:
[...]

PR #3623, opened 2026-07-20
Run FSK pulse detectors in parallel

merbanan:


merbanan:
```
• Evaluation Report

  The best result for the difficult Mercedes Sprinter captures from issue #3539 (https://github.com/merbanan/rtl_433/issues/3539) is:

  > Median-of-three filtering followed by non-decaying min/max discrimination, used as the final AUTO fallback.

   Method                           Normal FSK fixtures    Final hard batch      All issue captures
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━
   Original min/max                             566/835                 3/9                  69/133
  ───────────────────────────────  ─────────────────────  ──────────────────  ──────────────────────
   Hysteresis /3                                513/835                 3/9                  60/133
  ───────────────────────────────  ─────────────────────  ──────────────────  ──────────────────────
   Median + decaying min/max                    579/835                 6/9                  76/133
  ───────────────────────────────  ─────────────────────  ──────────────────  ──────────────────────
   Median + non-decaying min/max                574/835                 6/9                  77/133
  ───────────────────────────────  ─────────────────────  ──────────────────  ──────────────────────
   Running centroid                             555/835                 1/9                  72/133
  ───────────────────────────────  ─────────────────────  ──────────────────  ──────────────────────
   CUSUM                                        610/835                 2/9                  72/133
  ───────────────────────────────  ─────────────────────  ──────────────────  ──────────────────────
   Histogram centroid                           560/835                 1/9                  72/133
  ───────────────────────────────  ─────────────────────  ──────────────────  ──────────────────────
   AUTO + median fallback                No regressions                 8/9    Best combined result

  CUSUM has the best standalone result across existing FSK fixtures, but performs poorly on the target hard captures and introduces classification regressions when used broadly. The
  median min/max method is the better production choice because it removes one-sample FM spikes while preserving weak frequency separation.

  With the median method last in AUTO:

  - Full rtl_433_tests: 5511 records, 0 failures
  - CTest: 15/15 passed
  - General corpus: one additional fixture decoded, no losses
  - Final Sprinter package: improved from approximately 3/9 to 8/9

```

merbanan:
@zuckschwerdt this is the results I get with 4 discriminators.

@deviantintegral can you test this patch and see if you are able to receive all your sensors?

@gdt can you test this code and see if it works ok for you.

deviantintegral:
I've switched to the master branch on my install that's for the Ecowitt soil sensors. Those have been by far the most finicky. At the least, all three have shown up since I started it up. It may take a couple days to have any solid feedback.

I could switch my 433MHz weather station and 916.8M doorbell, but probably not until next week.

I was thinking about this further, and one advantage to running detectors like this is that we could start to ship more device specific tunings in other parts of the stack beyond just the decoders. Perhaps that's the escape hatch if there are decoders that are fundamentally incompatible with each other.

deviantintegral:
🤦 I got mixed up between this and another issue where fixes where going directly into master branch.

So, what I can say i on `master`, I'm still seeing significant drops from 2 of my 3 WH51(L) sensors. Note they're all within a few feet of each other, including the "Tomato Planter" one that is solid.

image

In the 30 minutes I've been running this branch, I still see drops (one sensor has been missing for 30 minutes now).

However, I took a look at my SNR readings. 

image

I'm surprised to see so much variance, and to see one sensor SNR declining at the same time the other two rise. This is with auto gain - I'm going to try setting fixed gains and see if that helps.

merbanan:
Can it be the sensor timings overlaps? They are the same models and should have the same timing. What tuner are you using? The FC0012 and PC0013 tuner drivers are buggy.

merbanan:
It could be that rtl_433 just fails to decode the signal. But then we need a huge set of samples to look at.

deviantintegral:
I'm using a Rafael Micro R820T tuner.

The timers are 70 second intervals, and I suppose it's possible they overlap. I'll take a look.

deviantintegral:
After some digging, I discovered `pulse_detect autolevel` looks to improve things. It's only been a few hours, but zero drops so far. I'll watch over the next few days.

But I did a double take - `autolevel`, not `auto`? Surely `autolevel` was a typo or hallucination? `auto` is all that's listed in https://github.com/merbanan/rtl_433/blob/master/conf/rtl_433.example.conf#L88-L90.

Not quite - it's a little further down. Somehow I've never noticed it as an option.

https://github.com/merbanan/rtl_433/blob/master/conf/rtl_433.example.conf#L104-L106

And apparently it's been documented as recommended since at least 2021, along with `squelch`.

https://github.com/merbanan/rtl_433/blob/master/docs/STARTING.md?plain=1#L202-L204

I found https://github.com/merbanan/rtl_433/issues/1784 and https://github.com/merbanan/rtl_433/issues/1785 so I'll continue over there.

merbanan:
Great news! @zuckschwerdt Maybe it is time to set it as the default?

zuckschwerdt:
It think so too. I always need to tell users to please use those and have not gotten any negative feedback so far.

@deviantintegral autolevel will shift the detection threshold (default -12 dB) and help with low signal transmissions, the SNR still needs to be above 9 dB.

PR #3636, opened 2026-07-27
Fineoffset-WH52: report battery_ok + battery_mV, add conductivity to HA autodiscovery

Verstreubulator:
First off — apologies. The original WH52 decoder (#3602) shipped incomplete: it reported battery in a field nothing maps, and it never wired up conductivity for autodiscovery, so the two readings users most want from a WH52 (battery health and EC) don't actually show up in Home Assistant. That was an oversight in the first PR; this follow-up fixes both. Corrections very welcome — the battery scaling is still empirical.

**1. Battery — `battery_V` → `battery_ok` + `battery_mV`.**
The merged decoder emitted the battery reading as `battery_V`. The example HA autodiscovery script (`examples/rtl_433_mqtt_hass.py`) has no mapping for `battery_V`, so no battery entity is ever created — whereas the sibling WH51 decoder emits `battery_ok` + `battery_mV`, which is exactly why WH51 battery entities do appear. This changes WH52 to use the same two fields as WH51, so the battery autodiscovers the same way and the output is consistent across the Fine Offset family.

`battery_mV = b[15] * 20 - 60` (0.02 V/LSB with a −0.06 V offset — the same empirical fit noted in the decoder header, expressed directly in mV). `battery_ok` is graded from that voltage using the same breakpoints as the WH51 decoder (≥1.6 V, 1.5, 1.4, 1.3 V).

**2. Conductivity — add a `conductivity` mapping.**
The decoder already emits `conductivity` (µS/cm), but `MAPPINGS` in `rtl_433_mqtt_hass.py` has no entry for it, so the EC reading — the main reason the WH52 exists over the WH51 — never autodiscovers. Added a `conductivity` entry using Home Assistant's `conductivity` device class and `µS/cm` unit (`UnitOfConductivity.MICROSIEMENS_PER_CM`).

**Verification.**
Values were checked against the two existing WH52 test captures in `rtl_433_tests` and against four physical field units. Examples from the captures: `0058ac` → `battery_mV` 1560 / `battery_ok` 0.9; `005995` → 1620 / 1.0 — matching the WH51 rendering style. The paired expected-output update for those captures is in merbanan/rtl_433_tests#514.

Ran `clang-format` on the decoder (the repo's `.clang-format`). I wasn't able to build/run the binary locally on this machine, so a maintainer sanity-check on the compile is appreciated.

Thanks again to @merbanan and the rtl_433 contributors, and to @pbkhrv for the Home Assistant add-on that makes this usable.

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -284,6 +284,18 @@ mappings = {
         }
     },
 
+    "conductivity": {
+        "device_type": "sensor",
+        "object_suffix": "EC",
+        "config": {
+            "device_class": "conductivity",
+            "name": "Conductivity",
+            "unit_of_measurement": "μS/cm",
+            "value_template": "{{ value|float }}",
+            "state_class": "measurement"
+        }
+    },
+
     "detect_wet": {
         "device_type": "binary_sensor",
         "object_suffix": "moisture",
--- a/src/devices/fineoffset_wh52.c
+++ b/src/devices/fineoffset_wh52.c
@@ -47,7 +47,7 @@ Preamble: aa aa aa 2d d4
 - Byte 9: EC bits 15..8
 - Byte 10: EC bits 7..0; ec_raw = ((b8 & 0x0F) << 16) | (b9 << 8) | b10 (20-bit); conductivity_uS = ec_raw / 25.6 (empirical; 25.6 = 256/10)
 - Byte 11: coarse EC / range indicator (redundant, low nibble fixed 0x6)
-- Byte 15: battery voltage; volts ~= b15 * 0.02 - 0.06 (empirical, fit from 4 field units reading 1.56/1.58/1.58/1.62 V vs bytes 0x51/0x52/0x52/0x54; scale approximate pending a wider range)
+- Byte 15: battery voltage; battery_mV ~= b15 * 20 - 60 (0.02 V/LSB, -0.06 V offset; empirical, fit from 4 field units reading 1.56/1.58/1.58/1.62 V vs bytes 0x51/0x52/0x52/0x54; scale approximate pending a wider range). Reported as battery_ok (low below 1.4 V) + battery_pct + battery_mV.
 - Bytes 12-14, 16-21: per-unit fixed data (factory serial / calibration), constant per unit, differ between units. Not decoded.
 - Byte 22: CRC-8, poly 0x31, init 0x00, over bytes 0..21
 - Byte 23: checksum = sum(bytes 0..22) & 0xFF
@@ -95,13 +95,22 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     char id[7];
     snprintf(id, sizeof(id), "%02x%02x%02x", b[1], b[2], b[3]);
 
-    int boost       = (b[4] & 0xe0) >> 5;
-    int temp_raw    = ((b[4] & 0x1f) << 8) | b[5];
-    float temp_c    = temp_raw * 0.1f - 40.0f;
-    int moisture    = b[6];
-    int ec_raw      = ((b[8] & 0x0f) << 16) | (b[9] << 8) | b[10];
-    float ec_uscm   = ec_raw / 25.6f;      // empirical calibration, see notes above
-    float battery_v = b[15] * 0.02f - 0.06f; // empirical, see notes above
+    int boost      = (b[4] & 0xe0) >> 5;
+    int temp_raw   = ((b[4] & 0x1f) << 8) | b[5];
+    float temp_c   = temp_raw * 0.1f - 40.0f;
+    int moisture   = b[6];
+    int ec_raw     = ((b[8] & 0x0f) << 16) | (b[9] << 8) | b[10];
+    float ec_uscm  = ec_raw / 25.6f;  // empirical calibration, see notes above
+    int battery_mv = b[15] * 20 - 60; // 0.02 V/LSB, -0.06 V offset (empirical, see notes above)
+
+    // Battery reported as an OK flag + a percentage of the usable ~1.3-1.6 V range
+    // (WH51 sibling breakpoints), following the standard battery convention.
+    float batt_lvl = (battery_mv - 1300) * (1.0f / 300.0f);
+    if (batt_lvl < 0.0f)
+        batt_lvl = 0.0f;
+    if (batt_lvl > 1.0f)
+        batt_lvl = 1.0f;
+    int battery_ok = battery_mv >= 1400 ? 1 : 0;
 
     /* clang-format off */
     data_t *data = data_make(
@@ -110,7 +119,9 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
             "temperature_C",    "Temperature",          DATA_FORMAT, "%.1f C", DATA_DOUBLE, temp_c,
             "moisture",         "Moisture",             DATA_FORMAT, "%u %%",  DATA_INT,    moisture,
             "conductivity",     "Conductivity",         DATA_FORMAT, "%.0f uS/cm", DATA_DOUBLE, ec_uscm,
-            "battery_V",        "Battery Voltage",      DATA_FORMAT, "%.2f V",     DATA_DOUBLE, battery_v,
+            "battery_ok",       "Battery",              DATA_INT,    battery_ok,
+            "battery_pct",      "Battery level",        DATA_DOUBLE, 100.0f * batt_lvl,
+            "battery_mV",       "Battery",              DATA_FORMAT, "%d mV",      DATA_INT,    battery_mv,
             "boost",            "Transmission boost",   DATA_INT,    boost,
             "mic",              "Integrity",            DATA_STRING, "CRC",
             NULL);
@@ -126,7 +137,9 @@ static char const *const output_fields_wh52[] = {
         "temperature_C",
         "moisture",
         "conductivity",
-        "battery_V",
+        "battery_ok",
+        "battery_pct",
+        "battery_mV",
         "boost",
         "mic",
         NULL,


merbanan:
Split this into 2 commits and look in the git history how to word the commit text.

Regarding the code do not use double as type.

zuckschwerdt:
`battery_ok` should now be 0/1, add a level as e.g.
`"battery_pct",      "Battery level",    DATA_DOUBLE, 100 * batt_lvl, // Note: this might change with #3103`

Verstreubulator:
Thanks for the review — revised and force-pushed. Split into two commits (decoder, then the hass mapping), dropped the `double`, and `battery_ok` is now binary `0/1` with a separate `battery_pct` (`DATA_DOUBLE, 100 * batt_lvl`) as suggested, matching the `Fineoffset-WS85` convention. Checks are green. The paired expected-output update is in merbanan/rtl_433_tests#514 (checks green there too).

vgabor99:
Hi @Verstreubulator,

Sorry I have no other means to contact you. Thanks for supporting the community! I have recently purchased 3 Ecowitt WH52 sensors, just to find out that they are not yet supported "out of the box" by rtl433. Ended up building your PR and it's working fine for me.

I wanted to reach out to let you know that in case you need it, I am ready and willing to support your efforts with real-life data or testing. Feel free to ping me at: gabor dot varga dot 99 at gmail dot com.

Cheers!

Verstreubulator:
Thanks very much, Vgabor99! I am really pleased to hear it is working for you! It is working great for me too!

Since I had the Ecowitt app going while decoding the WH52, I was able to verify everything right away. I had 4 WH52s at the time and was able to make sure it worked with all of them. Now I have 10 and all work well.

I am new to GitHub and do not fully understand how things work around here. As far as I understand it, the decoder is now properly set up and fully tested. It passed all the tests and the people in charge approved it. It has been merged. It should be included in the next rtl_433 update. Now we just have to wait.

Thanks again! I hope your WH52s work well for you!

vgabor99:
Hi @Verstreubulator! I overlooked this: the battery reading may be off a bit. My 3 field units report around 2870 mV for a fully charged NiMH battery, which is clearly off the valid range. I have noticed this just now, after purchasing a few WH51 sensors and realizing that they report 1400 mV (50%) for the same battery. Let me know if you need field data from my units (some pointers on how to get it are appreciated).

zuckschwerdt:
Manually merged as c21cfdb and 8fa6364

vgabor99:
Hi @Verstreubulator, I have analyzed my voltage measurements as well as yours (vs. WH52 frames). I think that byte 20 carries voltage information, not byte 15. In your measurements, byte 15 was also proportional to the voltage but this might be just a fluke and not an actual relation. For my devices, byte 15 does not change with voltage at all. See the attached plot: devices a-b-c-d are yours, the others are mine. 

Battery voltage vs byte 15 and byte 20 by device

Considering my measurements only, the formula is V = 0.0192 x byte20 + 0.0853
Adding your measurements to the mix, it becomes V = 0.0181 x byte20 + 0.1633 (but the voltage values are not ground truth because they are computed values from the rtl_433_tests.

Do you happen to have the actual voltage measurements taken when you made the recordings for the tests? I yes, please share it with me.

I plan to make more tests with various voltages. Would you mind to record more data with lower voltages (e.g. a NiMH battery)? It would give some clarity.

Thank you!

PS. I have the 868 Mhz (EU) variant of the device but I don't think this is relevant at all in this matter.

zuckschwerdt:
Also `conductivity` should really be named `conductivity_uS_cm` as it is not an arbitrary value.

vgabor99:
> `battery_ok` should now be 0/1, add a level as e.g. `"battery_pct", "Battery level", DATA_DOUBLE, 100 * batt_lvl, // Note: this might change with #3103`

@zuckschwerdt the HomeAssistant autodiscovery script does not support this (yet?). It publishes the `battery_ok` as % value and does not pubish `battery_pct` at all. Are there plans to make the switch? Will this break other sensors?

(Sorry on commenting on a closed PR, I am trying to make my WH52 battery display to work somewhat similarly to WH51.)

zuckschwerdt:
Go ahead and discuss your findings about the battery byte and all. We only closed this because the change is merged. Looks like more changes might be needed.

The new definition of battery_ok and battery_pct should be implemented in the hass script some day, yes.

vgabor99:
> Go ahead and discuss your findings about the battery byte and all. We only closed this because the change is merged. Looks like more changes might be needed.

I'm on it, trying to gather a bit more data.

> The new definition of battery_ok and battery_pct should be implemented in the hass script some day, yes.

I am very new to HomeAssistant, and to this repo as well, so bear with me. Isn't changing the exposed type a breaking change that would cause issues to existing WH51 users? Also, my understanding is that the hass script exposes entities in a "standardized" way, i.e. all supported sensors should be changed to the `battery_ok` (bool) + `battery_pct` (%) type at the same time. Right?

> Also conductivity should really be named conductivity_uS_cm as it is not an arbitrary value

Same question here (except that WH52 is new so the userbase is small and maturity expectations are low). No problem doing a breaking change?

gdt:
There is theory of standardization and where we are, and progress.   More or less I think `we` are reporting battery_ok from 0.0 (dead, device doesn't work) to 1.0 (just replaced).   In theory that should be linear in time.   Then there is reporting the battery voltage.

I think there is a written plan of what should be, and it would be good to look for and read that and align and if not we should fix that.

My own view is that fixing things is more important than back compat, especially for things that are really new, and if they haven't been in an actual release.   No compat with previous master -- only with actual releases!!!!

zuckschwerdt:
> changing the exposed type a breaking change

Yes, and we always try to avoid it or have compatibility for a some time. Here the decoder is new and a few fixes are expected in the first months.

The `battery_ok` is already the result of standardization. But because 0/1 vs 0.0-1.0 output in that key is not easily discoverable we recently moved to split into two keys (if consumers could scan for the `.` they could in theory choose wether to show a binary switch or gauge).
The discussion has not fully concluded but I'd say "settled" towards that move. It will not break what is expected of the battery_ok key (0/1). Guess I'll change the 13 remaining decoders soon.

gdt:
I admit to being fuzzy and not paying attention, but it would be nice if doc/DATA_FORMATS.md said the current plan.

I am a fan of battery_mV (or battery_V, but we should have one rule not varying per decoder).   However, battery_mV is totally a separate concept from a 0.0 to 1.0 gauge of empty to full.

Right now the docs say battery_ok is a double, 0.0 to 1.0, but don't clearly say that if a device will (waving hands, usually) run for X hours on a fresh standard battery, then ideally battery_ok starts at 1.0 and over those X hours declines smoothly to 0.0.   Perhaps more importantly, when battery_ok is based on some low-resolution sensor, the value should represent the fraction of standard lifetime that can be reasonably relied on.

Which is a long way of saying that if the actual values represent 0%, 25%, 50%, 75%, and 100% of capacity, then as soon as the device indicates 0 it should say 0.  if it say 0.25, and X is 1000h, then it should be expected there are 250h left.

But really the most important point is adequate warning.

I tried to do it this way when I redid the WH51 battery indication, which is super coarse.

vgabor99:
> The discussion has not fully concluded but I'd say "settled" towards that move. It will not break what is expected of the battery_ok key (0/1). Guess I'll change the 13 remaining decoders soon.

I see that this discussion has been going for some years and there are two camps: one treats `battery_ok` as a signal to "no change needed / change now", the other one treats it as a countdown to "change now" / "dead now". As a newcomer, I am not taking sides :D

> if consumers could scan for the . they could in theory choose wether to show a binary switch or gauge

The HomeAssistant autodiscovery script can do this too. I think this is not too ugly to add this "migration support" feature for `battery_ok`. I will give it a try. Also adding `battery_pct`.

> I tried to do it this way when I redid the WH51 battery indication, which is super coarse.

That's a kind of 3rd approach, adding a limited number of thresholds to mitigate the nonlinear nature of battery depletion. I was wondering about the difference between WH51 and WH52 battery reporting. Maybe they should be harmonized?

gdt:
> I see that this discussion has been going for some years and there are two camps: one treats `battery_ok` as a signal to "no change needed / change now", the other one treats it as a countdown to "change now" / "dead now". As a newcomer, I am not taking sides 🗡️ I

It is reasonable to have a binary sensor (HA speak) and a battery level sensor both.

> The HomeAssistant autodiscovery script can do this too. I think this is not too ugly to add this "migration support" feature for `battery_ok`. I will give it a try. Also adding `battery_pct`.

I strongly advise against this.    The right thing to do is to map the semantics defined by rtl_433 into the semantics expected by Home Assistant.   Home Assistant, as I understand it this week, wants "battery level" as a number from 0 to 100 which is the % life of the battery remaining.     Resist the temptation to create a long list of things because someone might want one of them; that leads to clutter.   Instead, map rtl_433's value -- a number from 0.0 to 1.0 to Home Assistant's battery_level, 0.0 to 100.0.  The goal, I think, is to produce a device in HA that is faithful to the raw data but expressed in the native HA language.   Within HA, there are all sorts of integrations/etc. to deal with batteries and they all want a 0-100 level entity because that's what everything else creates.

> > I tried to do it this way when I redid the WH51 battery indication, which is super coarse.
> 
> That's a kind of 3rd approach, adding a limited number of thresholds to mitigate the nonlinear nature of battery depletion. I was wondering about the difference between WH51 and WH52 battery reporting. Maybe they should be harmonized?

It's not really a third approach, but an approach to map values 1200, 1300, 1400, 1500 1600 mV to an estimate of fractional/percent remaining life.   What I did for WH51 is based on living with multiple of those sensors over a long time and observing how batteries decay and fail.   My guess is that the newer sensor is similar.    I admit that my "X voltage maps to Y fraction" table is a bit rough, but the real goal is to give the user an impression among "this battery is brand new" "this battery is in great shape" to "this battery is -- waves hands wildly -- sort of half used up", "this battery is getting low but it's not desparate" to "this battery is in bad shape - go change it right now" with percentages that correspond to what a better indication would show, in terms of not leading the user astray.   What's need is to put in a fresh AA alkaline and then record battery_mV over the 1.5 years it takes to run down, and then guess from there what the values should be.

What I replaced was a linear scale from 700 mV to 1600 mV or something like that, and that caused a battery at 60% ish to totally fail within hours (as in bad data and no more transmissions).    Remaining live is not linear in battery, not even close.

I believe that the % for 1400 mV for example, should be the amount remaining the day before it changes to 1300, not the amount remaining the day it transitions from 1500 to 1400.  Always underestimate.     Really it should have measured in 50 mV or 10 mV increments and then this wouldn't matter, but that's not how it is.

vgabor99:
OK, what do you think about this:

**HA autodetection**

- Keep existing legacy exposure of `battery_ok` (0.0...1.0) in order to not having to migrate all sensors now (I can't do it, I don't have them).
- Expose the new types `battery_ok_b` (0/1) and `battery_pct` (0..100), this will work for the new sensors that have already been migrated.

Individual sensors sensors pick the one(s) they want to expose. We already have a bunch of `battery_pct` which is not exposed by hass.

```python
"battery_ok_b": {
        "object_suffix": "BB",
        "config": {
            "device_class": "battery",
            "name": "Battery Status",
            "payload_on": "0",
            "payload_off": "1",
            "value_template": "{{ value | int }}",
            "entity_category": "diagnostic"
        }
    },
"battery_pct": {
        "device_type": "sensor",
        "object_suffix": "BP",
        "config": {
            "device_class": "battery",
            "name": "Battery Level",
            "unit_of_measurement": "%",
            "value_template": "{{ value | int }}",
            "state_class": "measurement",
            "entity_category": "diagnostic"
        }
    },
```

**WH52 decoder**

- I take battery voltage reporting logic from WH51 and use it in WH52, becasue it represents remaining battery life better than the linear scale.
- (Also fix the byte15 -> byte20 issue, that's what I came here for)
- (Also fix the naming `conductivity_uS_cm`)

vgabor99:
See https://github.com/merbanan/rtl_433/pull/3664

gdt:
I think we should make things as they should be in the end, and not worry about back compat with previous master *at all*.  We should only worry about back compat with actual releases.  We are in the midst of a long rationalization, and given the pressure for back compat, I think it's a good time to step back and organize the grand plan and get that written down as a design and then follow it.

For the binary sensor in home assistant, it looks like you have it right, except the name "battery_ok_b" is odd.  Looking at my Home Assistant instance, there are a lot of "Foo Low Battery Level" binary sensors and some "Replace Battery Now" and "battery low" binary sensors.  I lean to the entity being "Foo Battery Low" because that's normal in HA, arguably the main plan, given that the "battery plus" integration produces that.    This doesn't have any suffixes; once it's a binary sensor, the suffix is noise, and we have too much noise in names.

The normal HA name is "Foo Battery Level".   I don't see why there is a BP object suffix; this is in HA-land just a percentage, and the semantics are implied by the device class.

It's entirely possible the WH51 mqtt stuff is not the way I am saying it should be.    I am using MQTT and have hand-written yaml to produce entities.

I'll send a note to the mailinglist, which I hope you're on -- my take is that the community should be centered on the list instead of github.

PR #3638, opened 2026-07-29
Set autolevel to be on by default

### LINKED ISSUE #1785: Parameter 'pulse_detect autolevel' could be replaced (and is impractically named)

deviantintegral:
Following up from discussion in #3623.

This PR changes the pulse detector to be set to autolevel by default. A few notes:

- We support the same types of settings flags as other flags (0 / no / off) at the cost of a bit more code.
- This _doesn’t_ enable squelch by default. I’m open to changing that, but it seems to me that for new users missing signals is worse than high CPU use.
- The “measured on real hardware” is my own setup for WH51 sensors. Totally open to testing other values if if it makes sense.

Test run locally has the same results as master:

|      |records|failures|
|------|-------|--------|
|BEFORE|5545   |71      |
|AFTER |5545   |71      |

82 captures also trigger the tracking, moving the floor from -12 dB to -24.7.

IMO this closes #1785 and partially addresses #1784 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
+- Changed pulse detector `autolevel` to be enabled by default, BREAKING behavioral change to the minimum detection level: it now tracks estimated noise instead of a fixed -12 dB floor; use `-Y autolevel=0` (`no`/`off` also accepted) to restore the previous fixed floor
+
 ## Release 25.12 (2025-12-12)
 
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Changelog
 
-## Unreleased
-
-### Changed
-
-- Changed pulse detector `autolevel` to be enabled by default, BREAKING behavioral change to the minimum detection level: it now tracks estimated noise instead of a fixed -12 dB floor; use `-Y autolevel=0` (`no`/`off` also accepted) to restore the previous fixed floor
-
 ## Release 25.12 (2025-12-12)
 
 ### Breaking Changes
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -103,7 +103,8 @@
 
 # as command line option:
 #   [-Y autolevel] Set minlevel automatically based on average estimated noise.
-# This is now the default; kept enabled here for clarity, use "pulse_detect autolevel=0" to disable.
+# This is now the default; kept enabled here for clarity.
+# Use "pulse_detect autolevel=0" ("no", "off", "false" and "disable" also work) to disable.
 pulse_detect autolevel
 
 # as command line option:
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -94,7 +94,7 @@
 #pulse_detect level=0
 
 # as command line option:
-#   [-Y minlevel=<dB level>] Manual minimum detection level used to determine pulses (-1.0 to -99.0).
+#   [-Y minlevel=<dB level>] Manual minimum detection level used to determine pulses (-1.0 to -99.0) (disables autolevel).
 #pulse_detect minlevel=-12
 
 # as command line option:
@@ -102,13 +102,11 @@
 #pulse_detect minsnr=9
 
 # as command line option:
-#   [-Y autolevel] Set minlevel automatically based on average estimated noise.
-# This is now the default; kept enabled here for clarity.
-# Use "pulse_detect autolevel=0" ("no", "off", "false" and "disable" also work) to disable.
+#   [-Y autolevel] Set minlevel automatically based on average estimated noise (default: on, 0 disables).
 pulse_detect autolevel
 
 # as command line option:
-#   [-Y squelch] Skip frames below estimated noise level to lower cpu load.
+#   [-Y squelch] Skip frames below estimated noise level to reduce cpu load (default: off).
 pulse_detect squelch
 
 # as command line option:
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -103,6 +103,7 @@
 
 # as command line option:
 #   [-Y autolevel] Set minlevel automatically based on average estimated noise.
+# This is now the default; kept enabled here for clarity, use "pulse_detect autolevel=0" to disable.
 pulse_detect autolevel
 
 # as command line option:
--- a/docs/STARTING.md
+++ b/docs/STARTING.md
@@ -199,17 +199,25 @@ for higher frequencies like `868M` the `minmax` pulse detector is used by defaul
 
 Use `-Y classic` or `-Y minmax` to force the use of a FSK pulse detector.
 
-Use `-Y autolevel` to automatically adjust the minimum detection level based on average estimated noise. Recommended.
+### Detection levels
 
-Use `-Y squelch` to skip frames below estimated noise level to reduce cpu load. Recommended.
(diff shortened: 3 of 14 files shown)


deviantintegral:
AI thinks this is the next step to enabling squelch by default. Regardless, this should be a code comment or an issue, not in STARTING so I removed it.

>  Making squelch evaluate noise-only status within a frame instead of across it (sub-frame squelch) is potential future work to make it safe to enable by default; until then it works best where signals stay strong for most of each frame.

src/r_flow.c
```diff
@@ -32,6 +32,12 @@
 #include "logger.h"
 #include "fatal.h"
 
+// Lower bound for the auto-tracked detection floor (min_level_auto).
+// Measured on real hardware: letting the tracked floor fall below -30 dB
+// raises CPU usage sharply (~3.9x at -35 dB vs -30 dB) for negligible
+// additional decode benefit, so auto-level tracking is clamped here.
+#define MIN_LEVEL_AUTO_FLOOR -30.0f
+
```
deviantintegral: This is new but makes sense to me. “Real Hardware” in this case is a Pi Zero 2, so I think that’s about as slow as anyone is running rtl_433 on. If this too narrow a use case we could remove it.

include/auto_level.h
```diff
@@ -1,7 +1,7 @@
 /** @file
     Auto-tracked detection floor for the pulse detector.
 
-    Copyright (C) 2026 Christian W. Zuckschwerdt <zany@triq.net>
+    Copyright (C) 2026 Andrew Berry <andrew@furrypaws.ca>
```
deviantintegral, line 4: hah, I’m not actually sure what the convention is here given this PR introduced this file?

docs/STARTING.md
```diff
@@ -199,17 +199,25 @@ for higher frequencies like `868M` the `minmax` pulse detector is used by defaul
...
+
+`-Y minlevel=<dB level>` still matters with autolevel on: it sets the *initial* detection level, and it is the level the tracked noise must fall at least 3 dB below before autolevel starts lowering the floor further -- it is not a lower bound on how far autolevel can track down. To prevent the tracked floor from drifting arbitrarily low on a quiet band (and pulling in more CPU-costly false candidates), autolevel's own adjustment is clamped to a built-in lower bound of -30 dB, regardless of `-Y minlevel`.
+
+Use `-Y squelch` to skip frames below estimated noise level to reduce cpu load. Squelch decides per received sample frame (the SDR callback buffer, about 1 second at the default 250k sample rate) using that frame's *average* level, so a short weak-signal burst can raise the average too little to change the frame's classification -- the whole frame, burst included, gets discarded.
```
zuckschwerdt: A default frame is 256k, i.e. 128k samples, half a second.

src/r_flow.c
```diff
@@ -177,9 +183,10 @@ int push_sdr_flow(r_cfg_t *cfg, unsigned char *iq_buf, uint32_t len)
         demod->total_frames_squelch += 1;
         demod->noise_level = (demod->noise_level * 7 + avg_db) / 8; // fast fall over 8 frames
         // If auto_level and noise level well below min_level and significant change in noise level
+        float min_level_auto_target = fmaxf(demod->noise_level + 3.0f, MIN_LEVEL_AUTO_FLOOR);
         if (demod->auto_level > 0 && demod->noise_level < demod->min_level - 3.0f
-                && fabsf(demod->min_level_auto - demod->noise_level - 3.0f) > 1.0f) {
-            demod->min_level_auto = demod->noise_level + 3.0f;
+                && fabsf(demod->min_level_auto - min_level_auto_target) > 1.0f) {
```
zuckschwerdt: the sign (IIRC of the hysteresis) changed?

docs/STARTING.md
deviantintegral: Good catch, fixed!

src/rtl_433.c
```diff
@@ -942,10 +942,25 @@ static void parse_conf_option(r_cfg_t *cfg, int opt, char *arg)
         while (p && *p) {
             char const *val = NULL;
             if (kwargs_match(p, "autolevel", &val)) {
-                cfg->demod->auto_level = atoiv(val, 1); // arg_float_default(p + 9, "-Y autolevel: ");
+                // autolevel  autolevel:1  autolevel:on  autolevel:yes
```
zuckschwerdt: why work around atoiv when it could be improved. Do we want no/off? Do we then also want yes/on?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
```
zuckschwerdt: This note should be in the commit message so it gets picked up by git log (and gen_nightly_info.sh)

conf/rtl_433.example.conf
```diff
@@ -103,6 +103,7 @@
 
 # as command line option:
 #   [-Y autolevel] Set minlevel automatically based on average estimated noise.
+# This is now the default; kept enabled here for clarity, use "pulse_detect autolevel=0" to disable.
```
zuckschwerdt: This should be a copy of the help output and preferably not amended

src/r_flow.c
```diff
@@ -177,9 +183,10 @@ int push_sdr_flow(r_cfg_t *cfg, unsigned char *iq_buf, uint32_t len)
...
+            demod->min_level_auto = min_level_auto_target;
```
zuckschwerdt: The flow should try hard to only use demod as state.

include/auto_level.h
zuckschwerdt: I don't care much. The convention I use is to keep trace of who wrote it across refactors, see e.g. analyzer, detect, slicer, util headers.

src/r_flow.c
deviantintegral: I think this didn’t change and is just a rearrangement?

Before:

```
fabsf(demod->min_level_auto - demod->noise_level - 3.0f) > 1.0f
```

After:

```
min_level_auto_target = fmaxf(demod->noise_level + 3.0f, MIN_LEVEL_AUTO_FLOOR);
fabsf(demod->min_level_auto - min_level_auto_target) > 1.0f
```

ie

`A - B - 3 == A - (B + 3)`

src/rtl_433.c
deviantintegral: Agreed, I fixed this in c094622a.

src/r_flow.c
deviantintegral: This has been changed since you reviewed so I think we’re clear here?

conf/rtl_433.example.conf
deviantintegral: Done!

CHANGELOG.md
deviantintegral: Done! Note when you’re building the changelog that this is a breaking change.

deviantintegral:
One commit worth looking at is https://github.com/merbanan/rtl_433/pull/3638/changes/08ce39726a3dd9f08ee6184e5397dd8f63826848

Does that make sense for how to handle specifying minlevel?

PR #3639, opened 2026-07-29
Added support for the Roboguard/IQ-Blue devices

Cale-Torino:
Added device support for the Roboguard/IQ-Blue devices

### DIFF
--- a/src/devices/roboguard.c
+++ b/src/devices/roboguard.c
@@ -1,6 +1,6 @@
 /** @file
     Roboguard (ZA:South Africa) PIR motion detection legacy security devices max 8 zones for 1 HQ(base station).
-    Also works with IQ-Blue Integrated Systems (ZA:South Africa) 433.92 RF line: reed switch, radar sensor, temp alerter etc.
+    Also works with IQ-Blue Integrated Systems (IBIS) (ZA:South Africa) 433.92 RF line: reed switch, radar sensor, temp alerter etc.
 
     Copyright (C) 2026 Cale Torino (ZR5CA)
     https://tutorials.techrad.co.za/2026/07/30/roboguard-protocol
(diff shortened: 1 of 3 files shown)


src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
+/** @file
```
zuckschwerdt, line 1: use two doc-comments, one for copyright, one for the decoder.

src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
...
+    (at your option) any later version.
+
+    ---
+
+    Tested devices:
+    - Roboguard: TXPIR 58B, CPUPIR 58B V5.8
+    - IQ-Blue: Reed switch V1.3
+    - IQ-Blue: Radar detector V1.1
+
+     ______
```
zuckschwerdt: this is markdown. indent verbatim blocks with 4 spaces. look at other decoders.

src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
...
+
+Z2
+Attempting demodulation... short_width: 1220, long_width: 2460, reset_limit: 29760, sync_width: 13224
+Perhaps we can find cyril's phala phala couch cash in the radio wave ethos
+Use a flex decoder with -X 'n=name,m=OOK_PWM,s=1220,l=2460,r=29760,g=0,t=0,y=13224'
+codes : {216}ffffff32bd6832bd6832bd6832bd6832bd6832bd6832bd6832bd68
+*/
+
+r_device const roboguard = {
+        .name        = "RoboGuard",
```
zuckschwerdt: more verbose here please.

src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
...
+        .modulation  = OOK_PULSE_PWM,
+        .short_width = 1230,
+        .long_width  = 2480,
+        .reset_limit = 20000,
+        .disabled  = 0,
```
zuckschwerdt: remove this line

src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
...
+        .decode_fn = &roboguard_decode,
+        .fields    = roboguard_output_fields,
```
zuckschwerdt: just `output_fields`

src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
...
+        "model",
+        "id",
+        "type_dec",
+        "type_str",
+        "bit24",
+        "bit23",
+        NULL,
+};
+
+/*
```
zuckschwerdt: remove this comment

src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
...
+    // ffffff32bd0832bd0832bd0832bd0832bd0832bd0832bd0832bd08
+    // ffffff 32bd08 32bd083 2bd083 2bd083 2bd083 2bd083 2bd083 2bd08
+    // 110110100100 = 1100
+    // getting data from the second repeat b[3]     b[4]     b[5]
+    int32_t id       = ((b[3] & 0x0F) << 16) | (b[4] << 8) | b[5]; // combine 3 bytes to get device_id
+    int32_t type_dec = (b[3] >> 4);                                // signal_type encoded in 1st 4 bits
+    int32_t bit24    = (b[3] << 16) | (b[4] << 8) | b[5];          // gets the 1st full signal (we only need 1)
+    int32_t bit23    = bit24 / 2;                                  // Helps the people using RC-SWITCH to see the signal, you know who you are ;)
+
+    for (uint8_t i = 0; i < SIGNAL_TYPES; i++) // 4
```
zuckschwerdt: use `if ... else if ...` for the four choices, not a lookup table.

src/devices/roboguard.c
```diff
@@ -0,0 +1,198 @@
...
+    preamble = ffffff
+    payload =  32bd08 x 8
+
+*/
+
+#include "decoder.h"
+
+#define SIGNAL_TYPES 4
+
+static int32_t roboguard_decode(r_device *decoder, bitbuffer_t *bitbuffer, int32_t startPulses, uint16_t package_type)
```
zuckschwerdt: this is not the correct prototype.

Cale-Torino (commented):
I used a switch statement and made the necessary changes, should all be good now

zuckschwerdt:
Do you have a sample .cu8 for this signal? (Upload as zip here)
The codes look similar to EV1527, I think.

Cale-Torino:
> Do you have a sample .cu8 for this signal? (Upload as zip here) The codes look similar to EV1527, I think.

it's similar but different, has a preamble instead of just payload.
[g004_433.84M_1024k.zip](https://github.com/user-attachments/files/30546708/g004_433.84M_1024k.zip)

zuckschwerdt:
There are gaps between the packets, try e.g.
`rtl_433 -R 0 -X 'n=name,m=OOK_PWM,s=1200,l=2400,g=3500,r=20000,invert'`

Since there is no decoding and no checksum a simple flex decoder might do:

```
rtl_433 -R 0 -X 'n=name,m=OOK_PWM,s=1200,l=2400,g=3500,r=20000,invert,
get=id:@4:{20}:%05x,
get=cmd:@0:{4}:[2:HEARTBEAT 4:LEARN/TAMPER 5:REMOTE 12:ALERT]'
```

zuckschwerdt:
You can also add `countonly` to summarize the rows. And if you know your remotes then you can use `match=d42f7` to filter.

Cale-Torino (commented):
Added filtering for preamble and combined into 1 row (216 bits) and added longer gap.

Cale-Torino:
> You can also add `countonly` to summarize the rows. And if you know your remotes then you can use `match=d42f7` to filter.

Does work but the gap between preamble and data is a bit too long I wanted it to be 1 row like:
```
{216}ffffffd2bd08d2bd08d2bd08d2bd08d2bd08d2bd08d2bd08d2bd08
```
Instead of 
```
{24}d2bd58, {24}d2bd58, {24}d2bd58, {24}d2bd58, {24}d2bd58, {24}d2bd58, {24}d2bd58, {24}d2bd58
```

Also picks up a lot of interference, that's why I use the preamble in the updated filtering but it does reduce range a bit.

I can false trigger with a 24-bit generic fixed remote but wanted to do some filtering to try get at least a 90% match.

So we know it is indeed the security device and not a generic remote with a very similar encoding.

The joys of congested frequencies 😃

Anyway it's up to you to add it or not.

I didn't know we could create a nice console data table though, thanks for the tips

Maybe I'm overthinking it.

PR #3644, opened 2026-08-01
Fix https truncation on /events

deviantintegral:
This fixes another corrupt JSON issue over the http server. Nearly all the code is adding tests, and the actual production code is just calling a previously introduced function.


deviantintegral:
The first commit references code in the HA integration which doesn't make sense on its own, but IMO this is probably best squashed into one commit so it includes the tests at the same time.

zuckschwerdt:
Thanks!

PR #3646, opened 2026-08-03
Add Ecowitt LDS01 Laser Distance Sensor decoder

StuartUSA:
## Description
This PR adds a native C decoder for the **Ecowitt LDS01 Laser Distance Sensor**.

### Device Specifications
* **Frequency:** 868.35 MHz
* **Modulation:** FSK (PCM)
* **Symbol Rate / Pulse Width:** 100 µs (`short_width = 100`)
* **Payload Size:** 12 bytes (96 bits)

### Telemetry Fields
* `model`: `Ecowitt-LDS01`
* `id`: 16-bit Device ID (Hex)
* `flags`: 16-bit Status Flags (Hex)
* `distance_mm`: Laser distance reading in millimeters
* `distance_m`: Laser distance reading formatted in meters
* `battery_ok`: Battery status flag (1 = OK, 0 = Low)

### Verification
Tested on hardware using RTL-SDR at 868.35 MHz with real sensor transmissions.

### Example JSON Output
```json
{
  "time": "2026-08-02 18:55:27",
  "model": "Ecowitt-LDS01",
  "id": "2A28",
  "flags": "9064",
  "distance_mm": 510,
  "distance_m": "0.510",
  "battery_ok": 1,
  "mic": "CHECKSUM"
}

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -89,10 +89,7 @@ There is a single global set of field mappings to Home Assistant meta data.
 
 """
 
-
-
 # import daemon
-
 
 import os
 import argparse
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by StuartUSA — COMMENTED** [review]
Test sample can be found here: https://github.com/merbanan/rtl_433_tests/pull/516

**Comment by StuartUSA** [comment]
Hope all is OK with this as it is my first one!
:)

**Comment by gdt** [comment]
I don't understand why you have distance in both m and mm.    Data should only be presented once, and my impression is that we tend to Si units, so just _m.

I'm unclear on the details, but I think we've also tended to battery as 0 to 100, even if the only values are 0 to 100.

File: examples/rtl_433_mqtt_hass.py [hunk 1]
```diff
@@ -108,14 +105,34 @@
 # Fields that get ignored when publishing to Home Assistant
 # (reduces noise to help spot missing field mappings)
 SKIP_KEYS = [ "type", "model", "subtype", "channel", "id", "mic", "mod",
-                "freq", "sequence_num", "message_type", "exception", "raw_msg" ]
+                "freq", "sequence_num", "message_type", "exception", "raw_msg",
+                "flags", "protocol", "freq1", "freq2" ]
```
zuckschwerdt: We should only ignore meta-data keys. But ok, flags and sequence_num would be pretty useless in HA.

src/devices/ecowitt_lds01.c
```diff
@@ -0,0 +1,134 @@
+/** @file
+    Ecowitt LDS01 laser distance / level sensor.
+
+    Copyright (C) 2026
```
zuckschwerdt, line 4: If you don't want your name here then use e.g. `Copyright (C) 2026 the rtl_433 authors`

src/devices/ecowitt_lds01.c
```diff
@@ -0,0 +1,134 @@
...
+        if (row_len - pos < sizeof(b) * 8) {
+            decoder_logf(decoder, 2, __func__,
+                    "not enough bits after preamble (row %u, %u left)",
+                    row, row_len - pos);
+            continue;
+        }
+
+        bitbuffer_extract_bytes(bitbuffer, row, pos, b, sizeof(b) * 8);
+
+        uint16_t id          = ((uint16_t)b[0] << 8) | b[1];
```
zuckschwerdt, line 83: no cast needed, and use the plain `int` type.

zuckschwerdt:
Why did you remove the comments from the hass script? Please revert that.

`distance_mm` is the sensor native unit here. I'd drop the `distance_m` conversion.

Run `./maintainer_update.py` to get the needed changes to the CMakeLists.txt

StuartUSA:
I've made the updates and restored the comments, I used an AI to update the code and it removed the comments :(

PR #3654, opened 2026-08-17
examples: fix id parsing in rtl_433_mqtt_hass.py

mw-white:
The existing ID parsing in examples/rtl_433_mqtt_hass.py to limit what devices are created in HA expects only integer IDs from devices.  Some devices seem to send their ID as a hex number - for example, one of my Ecowitt/FineOffset WH51 soil moisture sensors sends "0db71c" as the ID, which isn't allowed by the current arg parsing.

This PR changes the argument parsing for -I/--ids to parse the IDs as strings.  IDs received from mqtt are cast to str when they are checked so that "proper" int IDs are still caught.  Tested and working with the half-dozen devices I use.

Matt


gdt:
This change seems sensible to me.

PR #3655, opened 2026-08-17
Add flex decoder conf for Hampton Bay Ashby Park ceiling fan

coder999:
## Summary

Adds `conf/hamptonbay-ceilingfan.conf`, a flex-decoder config for the Hampton
Bay "Ashby Park" 52" ceiling fan's RF remote protocol (OOK_PWM, 304.25 MHz),
modeled on the existing `honeywell-fan.conf` / `fan-11t.conf` in this directory.

## How this was derived and verified

Reverse-engineered by live RF capture (no datasheet/FCC filing available for
the exact timing) while building a Home Assistant integration for these
fans. The fixture's own RF receiver was confirmed - by direct capture
comparison - to respond identically to its Bond Bridge (Olibra BD-1000)
smart-home controller, so the protocol documented here is the fixture's own
receiver protocol, independent of any specific remote/switch product.

Bit offsets in the `get =` field extraction were verified empirically, not
just derived: captured real RF from known commands (fan power, fan speed at
33/66/100%, light) and confirmed the decoded `id`/`ctr` fields matched
independently-known values in each case. Full detail in the file's own
header comment.

## Note on scope

`docs/CONTRIBUTING.md` documents the full C-decoder contribution path (with
test recordings submitted to `rtl_433_tests`) but doesn't mention `.conf`
files specifically, even though this directory already has ~80 similar
config-only device profiles. Happy to work toward a full C decoder + test
data submission instead if that's preferred for new contributions now -
just let me know.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conf/hamptonbay-ceilingfan.conf
+++ b/conf/hamptonbay-ceilingfan.conf
@@ -1,0 +1,56 @@
+# Hampton Bay "Ashby Park" 52" ceiling fan (and Bond Bridge, which controls
+# the same fixtures over the identical RF protocol).
+#
+# This fan's built-in RF receiver responds to a specific OOK_PWM protocol at
+# 304.25 MHz. Bond Bridge (Olibra BD-1000) was confirmed, by direct RF
+# capture comparison, to transmit bit-identical signals when controlling the
+# same fixture over its own local API - i.e. this is the fixture's own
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks, a .conf flex decoder is exactly what should be used for this.

Issue #3622, opened 2026-07-20
Message integrity levels


merbanan:
The MIC was intended to be used as something you can use to set a level of trust into. That is sort of not true in the current code base.

So how do we want it to work?

I would like to take into account manchester encoding, fixed structure bits, repeats, checksum and crc lengths. It should all add up to some quantifiable levels.

One suggestion of levels are minimal, low, high and very high. Then decoders can be enabled based on the level.

But they questions is if we should do it at all and how many levels and how to quantify the levels.

Anyone have any good ideas?

gdt:
The real issue is the probability of bad decodes.   More or less a preamble, a variable length packet, and a16 bit CRC. has some probability of decoding random energy that's 2^-1*(16+p) from some p that is the effective bit length of the checked preamble.

I think we should use the effective number of bits.  Then people can set their choice as a threshold.  I realize that everybody's random noise is different, but if the values are consistent they are still useful.

A complication is the "X decoded as Y", and we should perhaps look at Y as how many bits of strength does it have in the presence of valid X.  Arguably that's the right metric always: how many bits of strength does a decoder have in the presence of legit signals not from that decoder's devices.

I suspect you can sort this into
  - minimal: <= 5
  - low: 8
  - high: 13
  - very high: 20
 
but I don't see that exercise as useful as it will have us round the results of thinking and then have people have to be quantified.   Just putting in the numbers as we think about a decoder, changeable as we refine our thinking is easier., and will be more useful.

nicolasroche898:
Typically you use False Alarm Rate (FAR) or Frame Error Rate (FER) threshold (same thing), either as a percentage (ex. 0.1% error rate) or as a straight -log2 scale (number of redundancy bits). so, for a 16 bit crc, the FAR is 2^-16, (or just 16), 8 bit checksum has a FAR or 2^-8, a message which contains a 8 bit value repeated twice which is tested to be identical is 2^-8, etc. User can choose the FAR that can be tolerated. preamble  (all 0s or 010101 etc.) should NOT be part of the test, it is expected to vary naturally and contain errors as its purpose is to let the AGC tune to the right level. only test message bits starting from the SYNC word / SYNC bits

most protocols in rtl433 have a FAR of 0,4,8,16 or 24. that would map into: none,verylow,low,high,veryhigh?

gdt:
[...]

PR #3658, opened 2026-08-18
tpms_renault_0435r: report mic as CHECKSUM, not CRC

thenisvan:
`tpms_renault_0435r` reports its integrity check as `CRC`, but computes an 8-bit XOR:

- line 85: `int chk = xor_bytes(b, 9);`
- line 125: `"mic", "", DATA_STRING, "CRC",`

The file's own header says so too — line 2: *"FSK 9 byte Manchester encoded TPMS with xor checksum, 0435R"*, and line 84's comment is `// check checksum (checksum8 xor)`.

Every other TPMS decoder is self-consistent on this: those computing `xor_bytes` report `CHECKSUM` (`tpms_truck`, `tpms_citroen`, `tpms_ford`, `tpms_abarth124`), and those reporting `CRC` compute `crc8` (`tpms_toyota`, `tpms_renault`, `tpms_elantra2012`). As far as I can tell this is the only TPMS decoder where the declared and computed MIC disagree.

Noting it because the `mic` field is load-bearing for the integrity-level work in #3622 — any level derived from it would inherit the error here, and an XOR-8 decoder presenting as CRC is the case most likely to be trusted more than it should be.

No functional change; the `mic` string only.


zuckschwerdt:
Thanks! It might have been just a typo. And sometimes any xor scheme is seen as something CRC, where it really is just addition.

PR #3664, opened 2026-08-22
Improve WH52 battery and conductivity reporting

vgabor99:
WH52 battery voltage calculated from byte 20, not byte 15 (based on empirical observations).
Harmonized battery percent reporting with the WH51 sensor.
Renamed `conductivity` to `conductivity_uS_cm` for consistency and clarity.

The multiplier / scale factor is based on 7 measurements x 2 physical devices in the 1.282...1.617 voltage range, rounded to the nearest plausible values (20mV scale, no offset):

image

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -284,7 +284,7 @@ mappings = {
         }
     },
 
-    "conductivity": {
+    "conductivity_uS_cm": {
         "device_type": "sensor",
         "object_suffix": "EC",
         "config": {
--- a/src/devices/fineoffset_wh52.c
+++ b/src/devices/fineoffset_wh52.c
@@ -47,8 +47,8 @@ Preamble: aa aa aa 2d d4
 - Byte 9: EC bits 15..8
 - Byte 10: EC bits 7..0; ec_raw = ((b8 & 0x0F) << 16) | (b9 << 8) | b10 (20-bit); conductivity_uS = ec_raw / 25.6 (empirical; 25.6 = 256/10)
 - Byte 11: coarse EC / range indicator (redundant, low nibble fixed 0x6)
-- Byte 15: battery voltage; battery_mV ~= b15 * 20 - 60 (0.02 V/LSB, -0.06 V offset; empirical, fit from 4 field units reading 1.56/1.58/1.58/1.62 V vs bytes 0x51/0x52/0x52/0x54; scale approximate pending a wider range). Reported as battery_ok (low below 1.4 V) + battery_pct + battery_mV.
-- Bytes 12-14, 16-21: per-unit fixed data (factory serial / calibration), constant per unit, differ between units. Not decoded.
+- Byte 20: battery voltage; battery_mV ~= b20 * 20 (0.02 V/LSB, no offset; empirical. Reported as battery_ok_b (low below 1.4 V) + battery_pct + battery_mV.
+- Bytes 12-19, 21: per-unit fixed data (factory serial / calibration), constant per unit, differ between units. Not decoded.
 - Byte 22: CRC-8, poly 0x31, init 0x00, over bytes 0..21
 - Byte 23: checksum = sum(bytes 0..22) & 0xFF
 
@@ -101,31 +101,35 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int moisture   = b[6];
     int ec_raw     = ((b[8] & 0x0f) << 16) | (b[9] << 8) | b[10];
     float ec_uscm  = ec_raw / 25.6f;  // empirical calibration, see notes above
-    int battery_mv = b[15] * 20 - 60; // 0.02 V/LSB, -0.06 V offset (empirical, see notes above)
+    int battery_mv = b[20] * 20; // 0.02 V/LSB, no offset (empirical, see notes above)
 
-    // Battery reported as an OK flag + a percentage of the usable ~1.3-1.6 V range
-    // (WH51 sibling breakpoints), following the standard battery convention.
-    float batt_lvl = (battery_mv - 1300) * (1.0f / 300.0f);
-    if (batt_lvl < 0.0f) {
-        batt_lvl = 0.0f;
+    // We have no data yet for the battery depletion curve / remaining percentage.
+    // Assuming similarity to the WH51 sensor and reusing its mapping for a start.
+    // Review and correct when more ground data is available.
+    float battery_level;
+    if (battery_mv >= 1600) {
+      battery_level = 1.0;
+    } else if (battery_mv >= 1500) {
+      battery_level = 0.9;
+    } else if (battery_mv >= 1400) {
+      battery_level = 0.5;
+    } else if (battery_mv >= 1300) {
+      battery_level = 0.1;
+    } else {
+      battery_level = 0.0;
     }
-    if (batt_lvl > 1.0f) {
-        batt_lvl = 1.0f;
-    }
-    int battery_ok = battery_mv >= 1400;
 
     /* clang-format off */
     data_t *data = data_make(
-            "model",            "",                     DATA_STRING, "Fineoffset-WH52",
-            "id",               "ID",                   DATA_STRING, id,
-            "temperature_C",    "Temperature",          DATA_FORMAT, "%.1f C", DATA_DOUBLE, temp_c,
-            "moisture",         "Moisture",             DATA_FORMAT, "%u %%",  DATA_INT,    moisture,
-            "conductivity",     "Conductivity",         DATA_FORMAT, "%.0f uS/cm", DATA_DOUBLE, ec_uscm,
-            "battery_ok",       "Battery",              DATA_INT,    battery_ok,
-            "battery_pct",      "Battery level",        DATA_DOUBLE, 100.0f * batt_lvl, // Note: this might change with #3103
-            "battery_mV",       "Battery",              DATA_FORMAT, "%d mV",      DATA_INT,    battery_mv,
-            "boost",            "Transmission boost",   DATA_INT,    boost,
-            "mic",              "Integrity",            DATA_STRING, "CRC",
+            "model",                "",                     DATA_STRING, "Fineoffset-WH52",
+            "id",                   "ID",                   DATA_STRING, id,
+            "temperature_C",        "Temperature",          DATA_FORMAT, "%.1f C", DATA_DOUBLE, temp_c,
+            "moisture",             "Moisture",             DATA_FORMAT, "%u %%",  DATA_INT,    moisture,
+            "conductivity_uS_cm",   "Conductivity",         DATA_FORMAT, "%.0f uS/cm", DATA_DOUBLE, ec_uscm,
+            "battery_ok",           "Battery level",        DATA_DOUBLE, battery_level,
+            "battery_mV",           "Battery",              DATA_FORMAT, "%d mV",      DATA_INT,    battery_mv,
+            "boost",                "Transmission boost",   DATA_INT,    boost,
+            "mic",                  "Integrity",            DATA_STRING, "CRC",
             NULL);
     /* clang-format on */
 
@@ -138,9 +142,8 @@ static char const *const output_fields_wh52[] = {
         "id",
         "temperature_C",
         "moisture",
-        "conductivity",
+        "conductivity_uS_cm",
         "battery_ok",
-        "battery_pct",
         "battery_mV",
         "boost",
         "mic",


src/devices/fineoffset_wh52.c
```diff
@@ -101,31 +101,38 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+    } else if (battery_mv >= 1400) {
+      battery_pct = 50;
+    } else if (battery_mv >= 1300) {
+      battery_pct = 10;
+    } else {
+      battery_pct = 0;
     }
-    int battery_ok = battery_mv >= 1400;
+
+    int battery_ok = battery_mv >= 1300;
```
vgabor99: @gdt This deviates from WH51 and I would like to have some clarity. The comments in WH51 say that the failure threshold is 1200mV and the 13xx range is operational and lasts long. Could you please confirm: when `battery_pct` and `battery_ok` are separate, which is the correct threshold for `battery_ok`, 1300 or 1400? My WH52-s seem to work fine with NiMH batteries in the 13xx mV range.

>      * Thus, 1600 is a brand new battery (anecdata says it remains in
>      * that state for 6 minutes), changing to 1500 quickly.  There is
>      * then a long period at 1500, 1400, 1300, with 1200 indicating
>      * imminent failure -- not long enough to change it before bad
>      * things.  For reliable operation, one has to change the battery
>      * at 1300 mV

vgabor99: WH52 seems to have better resolution than the 100mV resolution of WH51.

gdt: Are you separating "voltmeter on the battery while it is not installed says X" and "device reports Y"?  There is load on the battery in the device, and the device's calibration is unclear.

What I meant is that when the device is reporting 1300 mV, all seemed ok.   As soon as there was a 1200 mV report, bad behavior happened within hours.   Thus, for reliable operation, I believe that you have to replace the battery when the device reports 1300 mV.   Or, you can wait until it gives bad reports and fails, if that's what you want to do.  But the battery level reported should cause the human to perceive that 1300 is a battery in the red zone that should be attended to, not in the yellow zone where it's fine for a while.   That's why I picked 10% for 1300.   When it reports 1400, I find that the device will operated for a very long time, and I'm fuzzy on 6 months vs 9 months, but it's not "almost dead".    That's how I chose thresholds.

The key point is that when the device reports 1300,  the actual remaining  lifetime could be months, if it just changed to 1300, or it could be 12 hours, if it is just about to change to reporting 1200.    I believe that the % should be more or less the lowest remaining life that is consistent with the report, because the point of this sensor is to guide humans (aided by integrations that only show crtiically low batteries) to replace the battery at the right time.

You could nitpick that I should say 1%, based on my remaining lifetime comment, and you'd technically be right.   I picked 10% to represent "this battery is dangerously low and should be replaced right away, but it hasn't failed", and I think that is consistent with overall culture.

gdt:
We have in docs/DATA_FORMAT.md a definition that there should be "battery_ok" which is a number from 0.0 to 1.0, and that there should be battery_V for the battery voltage in volts, with a note that mV is historical usage.    The current code is hard to follow and I don't think it's right.

I've read the checked-in code first.

First, and maybe you did this, but you should measure a battery, put it in and get the battery_mV value and then remove the battery and measure it again, and log that, for a bunch of batteries in various conditions.    You should expect lower readings from the device than with a voltmeter.     And then put in batteries that are very low and find out where it fails.

From that, develop a mapping from voltage to fraction of remaining operational life, and write code to do that.

There should not be a binary "battery low" data point.   The sensor from 0.0 to 1.0 has that meaning, and people can decide if they want to replace at < 0.25, < 0.10, or whatever.    But really, there should not be such an item because our documentation about what to add does not describe it.

(Next I'll read the proposed changes.)

examples/rtl_433_mqtt_hass.py
```diff
@@ -212,6 +212,32 @@
...
+        "config": {
+            "device_class": "battery",
+            "name": "Battery Status",
+            "payload_on": "0",
+            "payload_off": "1",
+            "value_template": "{{ value | int }}",
+            "entity_category": "diagnostic"
+        }
+    },
+
```
gdt: I don't think battery_ok_b should exist as a data object and thus don't think we should add it here.

examples/rtl_433_mqtt_hass.py
```diff
@@ -212,6 +212,32 @@
...
+        "config": {
+            "device_class": "battery",
+            "name": "Battery Level",
+            "unit_of_measurement": "%",
+            "value_template": "{{ value | int }}",
+            "state_class": "measurement",
+            "entity_category": "diagnostic"
+        }
+    },
+
```
gdt: It's a clue that we don't have battery_pct anywhere else.  The docs say battery_ok.   Instead of adding transformations for non-standard values, fix the values to follow the documented plan.

src/devices/fineoffset_wh52.c
```diff
@@ -47,8 +47,8 @@ Preamble: aa aa aa 2d d4
 - Byte 9: EC bits 15..8
 - Byte 10: EC bits 7..0; ec_raw = ((b8 & 0x0F) << 16) | (b9 << 8) | b10 (20-bit); conductivity_uS = ec_raw / 25.6 (empirical; 25.6 = 256/10)
 - Byte 11: coarse EC / range indicator (redundant, low nibble fixed 0x6)
-- Byte 15: battery voltage; battery_mV ~= b15 * 20 - 60 (0.02 V/LSB, -0.06 V offset; empirical, fit from 4 field units reading 1.56/1.58/1.58/1.62 V vs bytes 0x51/0x52/0x52/0x54; scale approximate pending a wider range). Reported as battery_ok (low below 1.4 V) + battery_pct + battery_mV.
-- Bytes 12-14, 16-21: per-unit fixed data (factory serial / calibration), constant per unit, differ between units. Not decoded.
+- Byte 20: battery voltage; battery_mV ~= b15 * 19.2 + 85.3 (0.0192 V/LSB, 0.0853 V offset; empirical. Reported as battery_ok_b (low below 1.4 V) + battery_pct + battery_mV.
```
gdt: This seems very likely to be off.    You are fitting from a tiny range.   I bet you can come up with some batteries at 1.4, 1.3, 1.2 V, perhaps by raiding other devices that use AA.   (I keep my old batteries in plastic bags sorted by voltage, to be used in devices that are ok with low values, and put < 1.3V in recycling :-)

I do think it's good to have support for the encoding in comments.

Typically the people that design the wire format are engineers and in a different context (Ford TPMS) there were approximations from fitting and I ended up deciding that the encoding was psi * 4.   That was very close to the curve fits, and car engineers would just not pick psi * 3.97 - 0.05, to pick something strange that was near a curve fit.

20 mV as a scale factor seems quite plausible.    The idea that the operational voltage is 0.06V lower than the at-rest voltage is not crazy, as is that your voltmeter and the WH52's voltmeter differ. 

But until you measure a battery that is so low as to cause the device to fail or send bogus data, we really don't have enough to get this right.

We also should not assume that the WH52's end-of-functioning battery voltage is the same.   It's actually a good product feature to eke out as much operation with very low battery voltage, and with a boost converter there's no obvious endpoint.  I have no idea what's inside.

Also, it says Byte 20 and then b15 in the comment.

src/devices/fineoffset_wh52.c
```diff
@@ -101,31 +101,38 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
+    if (battery_mv >= 1600) {
+      battery_pct = 100;
+    } else if (battery_mv >= 1500) {
+      battery_pct = 90;
+    } else if (battery_mv >= 1400) {
+      battery_pct = 50;
+    } else if (battery_mv >= 1300) {
+      battery_pct = 10;
+    } else {
+      battery_pct = 0;
```
gdt: this is ok as a guess, but note that you have no data that this is how the WH52 works.

src/devices/fineoffset_wh52.c
```diff
@@ -101,31 +101,38 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
-            "battery_mV",       "Battery",              DATA_FORMAT, "%d mV",      DATA_INT,    battery_mv,
-            "boost",            "Transmission boost",   DATA_INT,    boost,
-            "mic",              "Integrity",            DATA_STRING, "CRC",
+            "model",                "",                     DATA_STRING, "Fineoffset-WH52",
+            "id",                   "ID",                   DATA_STRING, id,
+            "temperature_C",        "Temperature",          DATA_FORMAT, "%.1f C", DATA_DOUBLE, temp_c,
+            "moisture",             "Moisture",             DATA_FORMAT, "%u %%",  DATA_INT,    moisture,
+            "conductivity_uS_cm",   "Conductivity",         DATA_FORMAT, "%.0f uS/cm", DATA_DOUBLE, ec_uscm,
+            "battery_ok_b",         "Battery",              DATA_INT,    battery_ok,
+            "battery_pct",          "Battery level",        DATA_INT,    battery_pct,
```
gdt: should be battery_level, per docs

src/devices/fineoffset_wh52.c
```diff
@@ -101,31 +101,38 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
...
-            "battery_pct",      "Battery level",        DATA_DOUBLE, 100.0f * batt_lvl, // Note: this might change with #3103
-            "battery_mV",       "Battery",              DATA_FORMAT, "%d mV",      DATA_INT,    battery_mv,
-            "boost",            "Transmission boost",   DATA_INT,    boost,
-            "mic",              "Integrity",            DATA_STRING, "CRC",
+            "model",                "",                     DATA_STRING, "Fineoffset-WH52",
+            "id",                   "ID",                   DATA_STRING, id,
+            "temperature_C",        "Temperature",          DATA_FORMAT, "%.1f C", DATA_DOUBLE, temp_c,
+            "moisture",             "Moisture",             DATA_FORMAT, "%u %%",  DATA_INT,    moisture,
+            "conductivity_uS_cm",   "Conductivity",         DATA_FORMAT, "%.0f uS/cm", DATA_DOUBLE, ec_uscm,
+            "battery_ok_b",         "Battery",              DATA_INT,    battery_ok,
```
gdt: This should be removed.  We do not document it, and it's a very odd name on top of that.   Just output the specified battery_ok as 0.0 to 1.0.

gdt (changes requested):
I wrote a lot, but:
  - we need real data on battery voltages and reported values, to come up with a believable formula for battery voltage from the on-the-wire bits
  - this should output battery_ok following the docs
  - this should not output battery_pct or battery_ok_b, because we do not document that

gdt:
I sent a long note to the mailinglist prompted by but not really about this.  Please have a look.

zuckschwerdt:
I'd say we stick with `battery_ok` and additionally, if available, the more granular `battery_pct`.

There were complaints that `battery_ok` with a simple binary battery low indicator, values 0 / 1, is not easily distinguishable from `battery_ok`with a float range, values 0.0 – 1.0. Thus future `battery_ok` will be 0/1 and `battery_pct` carries the actual level.
I.e. `battery_ok` conveys only a battery low warning and if there is a `battery_pct` the consumer can know that a level is available.

gdt:
> I'd say we stick with `battery_ok` and additionally, if available, the more granular `battery_pct`.
> 
> There were complaints that `battery_ok` with a simple binary battery low indicator, values 0 / 1, is not easily distinguishable from `battery_ok`with a float range, values 0.0 – 1.0. Thus future `battery_ok` will be 0/1 and `battery_pct` carries the actual level. I.e. `battery_ok` conveys only a battery low warning and if there is a `battery_pct` the consumer can know that a level is available.

battery_ok is super granular as  a float!

If we want to deviate, I think we need to first discuss the documented design and change it, and then change all the decoders - breaking compat.   I have started that discussion on the list.   I object to starting to deviate in drivers without having the discussion and commiting doc changes.   The should we change is not at al about this particular driver, which is not special really.

(I also don't like battery_pct.   Given that we have nothing to align to in the rtl_433 world, I think we should call it battery_level and have it be percentage as a float, so that it matches home assistant naming and semantics.  Not because HA is king, just that having more than one way to name a concept in the free software world is not helpful, if sticking to one name has no downsides.)

And if we're going to rototill, I'd have "battery_low" , 1 for bad battery, 0 for good battery, so that we'll add two fields that don't exist now, and remove the old ones, and not be changing semantics of existing fields.   that also aligns with HA.

zuckschwerdt:
We had some good arguments in #3103 already. I'm perfectly fine with the current `battery_ok` definition.
But then HA users (soundly) complained that 0->0.0 and 1->1.0 is not a good mapping as "1 = not battery low" isn't 100% battery and "0 = battery low" isn't 0% but rather 10% or such. The arguments was that battery_ok (or maybe a future battery_low) should be binary and the level should be somewhere else and preferably as percentage.
Maybe I'm mistaken and 0.0-1.0 is really the natural (HA) format, I remember 1-100 being wished for (HA)?

If we know the unit then we should affix it, using `battery_pct` is just the shortest meaningful way to express "battery_level_percentage" I'd say.

gdt:
I am also totally ok with battery_ok 0.0 to 1.0.     I don't love it, but it's existing, and it's well more than good enough since it's arbitrary anyway.

As for HA, I think we should not assume or preach that there should be any kind of 1:1 mapping from rtl_433 to HA.  It is entirely normal in the HA world to have vast numbers of devices each with their own representation and to being those into common names and common semantics.    I think HA people are complaining that a naive presentation of rtl-433 bits is not what they expect, and that's not an rtl_433 bug but a "the mapping is wrong" bug.

What HA wants is "battery level" from 0% to 100%.  which is exactly 100.0 * battery_ok.     I think that leads to "rtl_433 to HA glue should do this transformation", and that's fine.    HA is just as fuzzy as to what 0 means; I'm taking the interpretation of linear 100% to 0% over the working lifetime, exactly 1.0 to 0.0 over the lifetime.   Whether people should change at 20% or 2% is up to them.  But if the device goes offline due to battery while reporting 8%, then the report was wrong. 

(My biggest point is that we need to have written rules about what ought to be, and to follow them, declining to merge changes that don't follow the rules.)

gdt:
Could you continue on the mailingllst?  This really deserves project-wide visibility, vs people following this PR (github culture considered harmful).

src/devices/fineoffset_wh52.c
vgabor99: OK, changing it to 1400, identical to Wh51.

src/devices/fineoffset_wh52.c
vgabor99: I have made more measurements, with all the batteries I can get my hands on. Posted a plot in the PR description. The batteries are measured without load, I have no means to measure it during load. I hope this is acceptable range that covers most of the voltages the device is expected to work with. Especially considering the fact that the lowest measurement 1285mV is below the 1300mV threshold where 0% battery is reported.

src/devices/fineoffset_wh52.c
vgabor99: Right, we have to start with something. Assuming similarity to WH51 seems logical to me.

vgabor99:
I seem to have stirred up a hornet’s nest. While the larger debate about project-wide handling of battery reporting is ongoing, could you please also help to find a smaller compromise localized to this PR? I would be great if there was an agreement on how to unblock this PR:

> we need real data on battery voltages and reported values, to come up with a believable formula for battery voltage from the on-the-wire bits

I have addressed this, gathered more data

> this should output battery_ok following the docs
> this should not output battery_pct or battery_ok_b, because we do not document that

I am happy to go with either way that produces some kind of "battery low" signal in HA. Single 0.0...1.0 `battery_ok`, or splitting to 0/1 `battery_ok` and 0..100 `battery_pct`. FWIW 0/1 battery_ok + battery_pct is the status quo in this decoder, I only added HA discoverability for them (hence battery_ok_b).

src/devices/fineoffset_wh52.c
vgabor99: > Typically the people that design the wire format are engineers and in a different context (Ford TPMS) there were approximations from fitting and I ended up deciding that the encoding was psi * 4. That was very close to the curve fits, and car engineers would just not pick psi * 3.97 - 0.05, to pick something strange that was near a curve fit.

20 mV as a scale factor seems quite plausible. The idea that the operational voltage is 0.06V lower than the at-rest voltage is not crazy, as is that your voltmeter and the WH52's voltmeter differ.

@gdt I understand the part using a round scaling factor because it sounds logical. But following this line of thought, why have a fixed offset at all (and not just assume that _any_ fixed offset we see is just some kind of error)? Given the measured values I have posted, what scaling and offset values would you choose?

vgabor99: Setting 20mV scale factor and no offset, we get this:
image
This does not seem significantly off to me, the error is in the 20-40 mV range.

gdt: There is no logical reason for the designer of the device to put in an offset that is small.  Offsets are only useful to shift realistic measurement range into something representable.   Like "bits = temp_C + 40" so you can represent -40 to +87 C in 7 bits.    Adding in e.g. 45 mV is non-sensical, and no competent engineer would do that.    If the offset were 1V, to fit high resultion of 1.0 to 1.8V in field, I could see that, but clearly that's not happening here..  Your plot makes a case that bits = voltage / 20 mV is very close to what's going on, and IMHO a good working approximation.   It does make sense that the voltage under load is slightly lower than the resting voltage, so I would say your data does not falsify the 20 mV no offset hypothesis.  An alternative hypothesis that it's really 20.58 mV per step seems implausible.    Especially until we have similar measurements from a bunch of devices which all have their own construction tolerances.

Thanks for doing the experiment!   This makes me far more comfortable that the bits/voltage correspondence is correct.

src/devices/fineoffset_wh52.c
gdt: Then make sure there is a pointed comment that there is no data so far about when the WH52 ceases to operate due to low voltage, and that we're using the WH51 mapping for now, to be replace when there is knowledge.   I agree that it's a reasonable null hypothesis but it's also obviously quite a different device.

gdt:
> I seem to have stirred up a hornet’s nest. While the larger debate about project-wide handling of battery reporting is ongoing, could you please also help to find a smaller compromise localized to this PR? I would be great if there was an agreement on how to unblock this PR:

I agree we should keep the larger debate separate (and I think on the list).

For this PR, it's straightforward.   We have checked-in docs that specify battery_ok from 0.0 to 1.0.   So Implement that, and only that.  Do not add additional synthetic battery entities that are not documented.  This goes for those that are under active debate, but it is even more strongly true for those that aren't.

You can see how I did it for WH51, or how it was done before I changed the mapping.  There's battery_mV and there's battery_ok, scale 0.0 to 1.0.    This is the same except decoding battery_mV is totally different.

gdt:
> I am happy to go with either way that produces some kind of "battery low" signal in HA. Single 0.0...1.0 `battery_ok`, or splitting to 0/1 `battery_ok` and 0..100 `battery_pct`. FWIW 0/1 battery_ok + battery_pct is the status quo in this decoder, I only added HA discoverability for them (hence battery_ok_b).

Sorry, I didn't realize the _b and _pct were existing.   So by normal doctrine, it's ok for you not to remove them, but also it's not ok to add them to the ha discovery.    My opinion is that _b and _pct are wrong (contrary to docs) and we should clean that up sooner rather than later.

As for producing entities in HA, we have battery_ok and it's up to whatever translates rtl_433 to HA to map that to "battery level" (0-100%) and if wanted a synthetic "battery low" binary sensor (done in HA template, or in the translation layer).  We already have that situation in the other decoders; it's not special to this one.

vgabor99:
> Sorry, I didn't realize the _b and _pct were existing. So by normal doctrine, it's ok for you not to remove them, but also it's not ok to add them to the ha discovery. 

`_b` is not existing (I invented it to support a better autodiscovery), `_pct` is (and when used, `battery_ok` becomes 0/1).

> My opinion is that _b and _pct are wrong (contrary to docs) and we should clean that up sooner rather than later.

I am caught in a crossfire between you and other maintainers ([here](https://github.com/merbanan/rtl_433/pull/3636#issuecomment-5105017727), [here](https://github.com/merbanan/rtl_433/pull/3636#issuecomment-5373363974)). 

> As for producing entities in HA, we have battery_ok and it's up to whatever translates rtl_433 to HA to map that to "battery level" (0-100%) and if wanted a synthetic "battery low" binary sensor (done in HA template, or in the translation layer). We already have that situation in the other decoders; it's not special to this one.

One thing is unclear to me. Given that `battery_ok` is already "variant", why is it wrong to make the hass script to handle this by observing the value? The damage is already done, so to say, at least harvest what we can from it. Value is 0/1 -> binary sensor (usually accompanied by a `_pct`). Value is 0.0...1.0 -> analog sensor.

I only added the `_b` as a decoder-side alternative to hass-side variant type detection. I.e. let the decoders be explicit about exposing binary state.

vgabor99:
The cherry on top: hass autodiscovery uses `"value_template": "{{ ((float(value) * 99)|round(0)) + 1 }}"`, which will never be 0.

0, ,0.0 -> 1.0
0.5 -> 51.0
1, 1.0 -> 100.0

Is there any reason to not use e.g. `"value_template": "{{ (float(value) * 100) | round(0) }}"`?

zuckschwerdt:
> I am caught in a crossfire between you and other maintainers

True ;) Don't mind me though. All is well. There will be changes to a bunch of decoders once we find a nice solution anyway.

Re x99+1: I seem to remember a problem with HA where a value of 0 did not display well. But things might have changed since creating that mapping with #1016 in 2019.

gdt:
I've seen values of 0% in HA for batteries (not saying it's from rtl_433).

In general, I lean strongly to doing things right per the specs and if a bug fix is needed elsewhere, then so be it.   Adding accomodations in other software just builds up layers of code that we'd be better off without.

In your case, if the current fromula is a little goofy, that's not a good reason to do otherwise than 0.0 to 1.0 as documented.  The whole concept of 100% vs 99% of battery life is a bit theoretical and an estimate anyway.

gdt:
> One thing is unclear to me. Given that `battery_ok` is already "variant", why is it wrong to make the hass script to handle this by observing the value? The damage is already done, so to say, at least harvest what we can from it. Value is 0/1 -> binary sensor (usually accompanied by a `_pct`). Value is 0.0...1.0 -> analog sensor.

The plan as documented is that battery_ok, despite using the word ok, is 0.0 for a dead battery to 1.0 for a brand new one, as best as can be figured out.    With my HA user hat on, that should obviously map to 100x as "battery level" in HA.   A battery level sensor in HA that reads either 100 or 0 is perfectly easy to deal with; all the code that reads it will not be confused and will not do the wrong thing.    The real enemy is variant encodings, so that code processing batteries has to be written twice and you have to hope both ways work, since there are 2x the chances for bugs.   **We are doing protocol design here and less is more.**

If rtl_433->HA code exposed battery_ok 1) with the name ok or 2) in the 0.0 to 1.0 scale, that would be a bug.   We should be translating to words and scales that are consistent with HA best practices.

> I only added the `_b` as a decoder-side alternative to hass-side variant type detection. I.e. let the decoders be explicit about exposing binary state.

Yes, but you are adding things that are not documented as part of the plan, and IMHO we should be cleaning up those things, not making them more entrenched.   Adding options is a bug, not a feature, unless they are part of a coherent architecture, agreed on, and documented.  (As opposed to exposing things that are really received over the air, which is fine, if they can't be mapped solidly to standard names.)

I also don't think the "be explicit about binary" makes sense.  We aren't saying if we have 3 states, 4, 10, or if we are doing some mapping from voltage, and if so what the voltage resolution is, what the assumptions about battery chemistry are, and whether the mapping has been validated.   All of that is just as important.   The point is that we have a protocol spec that says 0.0 is dead and 1.0 is brand new and we are trying to do our best.    Anything more becomes a quagmire pretty much immediately as every question about why is equally valid.

gdt:
> Is there any reason to not use e.g. `"value_template": "{{ (float(value) * 100) | round(0) }}"`?

I see the current code as a bug.    Separately, it's a bug (that I see as serious) to have such code without an explanation of why.    I am in favor of cleaning it up and your suggestion looks good.

This belongs in a separate PR, though - it's not about WH52, and it will affect everything (detectably, not meaningfully).

vgabor99:
> There will be changes to a bunch of decoders once we find a nice solution anyway.

That sounds good. I would absolutely love if all decoders would consistently expose a binary ok sensor, an optional percentage sensor and an optional voltage sensor. Until then, in this PR, I will go the documented way which is also consistent with WH51.

Thanks for all the input!

src/devices/fineoffset_wh52.c
vgabor99: battery_pct removed

vgabor99:
Done. The changes are quite small this way. I will update the tests as well (this will break existing tests in open PR https://github.com/merbanan/rtl_433_tests/pull/514).

vgabor99:
Closing in favor of https://github.com/merbanan/rtl_433/pull/3668.
This PR sparked a (mostly unrelated) debate, accumulated a lot of comments that are irrelevant to the final result, which is quite simple.
To keep the actual PR readable, I have created a new one with the relevant changes only.

PR #3668, opened 2026-08-23
Improve WH52 battery and conductivity reporting

vgabor99:
WH52 battery voltage calculated from byte 20, not byte 15 (based on empirical observations).
Report battery data according to documented data format (single `battery_ok` in 0.0 .. 0.1 range).
Renamed `conductivity` to `conductivity_uS_cm` for consistency and clarity.

The multiplier / scale factor is based on 7 measurements x 2 physical devices in the 1.282...1.617 voltage range, rounded to the nearest plausible values (20mV scale, no offset):

image

Related regression test PR: https://github.com/merbanan/rtl_433_tests/pull/519

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -284,7 +284,7 @@ mappings = {
         }
     },
 
-    "conductivity": {
+    "conductivity_uS_cm": {
         "device_type": "sensor",
         "object_suffix": "EC",
         "config": {
--- a/src/devices/fineoffset_wh52.c
+++ b/src/devices/fineoffset_wh52.c
@@ -47,8 +47,8 @@ Preamble: aa aa aa 2d d4
 - Byte 9: EC bits 15..8
 - Byte 10: EC bits 7..0; ec_raw = ((b8 & 0x0F) << 16) | (b9 << 8) | b10 (20-bit); conductivity_uS = ec_raw / 25.6 (empirical; 25.6 = 256/10)
 - Byte 11: coarse EC / range indicator (redundant, low nibble fixed 0x6)
-- Byte 15: battery voltage; battery_mV ~= b15 * 20 - 60 (0.02 V/LSB, -0.06 V offset; empirical, fit from 4 field units reading 1.56/1.58/1.58/1.62 V vs bytes 0x51/0x52/0x52/0x54; scale approximate pending a wider range). Reported as battery_ok (low below 1.4 V) + battery_pct + battery_mV.
-- Bytes 12-14, 16-21: per-unit fixed data (factory serial / calibration), constant per unit, differ between units. Not decoded.
+- Byte 20: battery voltage; battery_mV ~= b20 * 20 (0.02 V/LSB, no offset; empirical. Reported as battery_ok + battery_mV.
+- Bytes 12-19, 21: per-unit fixed data (factory serial / calibration), constant per unit, differ between units. Not decoded.
 - Byte 22: CRC-8, poly 0x31, init 0x00, over bytes 0..21
 - Byte 23: checksum = sum(bytes 0..22) & 0xFF
 
@@ -101,9 +101,9 @@ static int fineoffset_wh52_decode(r_device *decoder, bitbuffer_t *bitbuffer)
     int moisture   = b[6];
     int ec_raw     = ((b[8] & 0x0f) << 16) | (b[9] << 8) | b[10];
     float ec_uscm  = ec_raw / 25.6f;  // empirical calibration, see notes above
-    int battery_mv = b[15] * 20 - 60; // 0.02 V/LSB, -0.06 V offset (empirical, see notes above)
+    int battery_mv = b[20] * 20; // 0.02 V/LSB, no offset (empirical, see notes above)
 
-    // Battery reported as an OK flag + a percentage of the usable ~1.3-1.6 V range
+    // Battery reported as a percentage of the usable ~1.3-1.6 V range
     // (WH51 sibling breakpoints), following the standard battery convention.
     float batt_lvl = (battery_mv - 1300) * (1.0f / 300.0f);
     if (batt_lvl < 0.0f) {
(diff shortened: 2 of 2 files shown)


Verstreubulator:
Thank you for digging into this, @vgabor99 — great catch, and I went and checked it against my own units. I read six WH52s off-air (CRC + checksum clean, decode cross-checks), and four of them are the same units that produced the July test captures, so I have both a snapshot across units and a ~6-week time-series. Both point the same way as your result:

**Across 6 units, same day** — byte15 is spread wide (0x51–0x5e) while byte20 is clustered tight (0x4b–0x4d ≈ 1.50–1.54 V). If byte15 were voltage, six similar-age units would span 1.56–1.82 V, which makes no sense; byte20 gives the sensible clustered value.

**Same 4 units, July → now** — byte15 is *frozen* at its day-1 value on every unit while byte20 has *declined* on every unit as the cells aged:

| Unit | byte15 Jul → now | byte20 Jul → now |
|------|------------------|------------------|
| 0058ac | 0x51 → 0x51 (no change) | 79 → 75 |
| 005b45 | 0x52 → 0x52 (no change) | 78 → 76 |
| 005cb5 | 0x52 → 0x52 (no change) | 79 → 76 |
| 005995 | 0x54 → 0x54 (no change) | 81 → 77 |

So byte15 is a fixed per-unit constant and byte20 is the live voltage — reproducing your finding on 915 MHz too (no US/EU layout difference). It also explains my original mistake: with four fresh units all near one voltage, a fixed per-unit byte15 happened to be monotonic with voltage, so it correlated by luck; only within-unit change over time or a wider unit spread separates them, neither of which my single snapshot had. Full per-unit data (all six units, raw frames, meter table) is at the bottom of this comment.

One caveat on the voltages I gave earlier: my July "ground-truth" app readings (1.56/1.58/1.58/1.62 V) turn out to equal `byte15×0.02−0.06` *exactly* for all four units. Since byte15 is a fixed per-unit constant, I now suspect the Ecowitt app's WH52 battery figure was derived from byte15 as well — a near-static per-unit number, not the true cell voltage. So please don't lean on those app values as battery ground truth; a meter is the right reference, which is what you used.

A note on my situation, so you know what to expect from me: I'm caring for elderly parents right now and can't give this project the time it deserves. My decoding was only ever meant as a starting point, never a definitive solution, so I'm genuinely glad to see others stepping in and building on it — thank you for the effort you're putting in. I'll help where I can.

On calibration: I metered all six cells with a Fluke 17B+ (resting) as an app-independent check. Against the current byte20 the residual is small and positive — mean +11.5 mV, all within one count — and a no-offset fit to the six points gives 20.15 mV/count, essentially your 20.0. So a meter independent of the app confirms `V = byte20 × 0.02`, and the small positive gap is consistent with the device reading a touch below the resting voltage under load.

The one gap I can't fill right now: my cells are all still fairly fresh (1.50–1.54 V), so I have nothing at the low end, and the low-voltage behaviour / the WH51-style `battery_ok` thresholds are still unverified on the WH52 — worth keeping the "no data yet on when the WH52 quits" caveat you and @gdt discussed. I'll watch for a chance to capture a genuinely low battery. Thanks again for carrying this forward.

---

**Full data — six 915 MHz units, 2026-08-23.** All frames pass CRC-8 (poly 0x31) and the trailing sum; moisture/EC/temp cross-check the decoder.

| ID | byte15 | byte20 → ×0.02 | Fluke 17B+ resting | moisture |
|----|--------|----------------|--------------------|----------|
| 0058ac | 0x51 | 0x4b → 1.50 V | 1.535 V | 0 % |
| 005b45 | 0x52 | 0x4c → 1.52 V | 1.518 V | 0 % |
| 005cb5 | 0x52 | 0x4c → 1.52 V | 1.536 V | 0 % |
| 005995 | 0x54 | 0x4d → 1.54 V | 1.538 V | 0 % |
| 0057a2 | 0x58 | 0x4c → 1.52 V | 1.534 V | 33 % |
| 005c90 | 0x5e | 0x4c → 1.52 V | 1.528 V | 12 % |

byte15 spans 0x51–0x5e while byte20 stays 0x4b–0x4d. Meter − byte20×0.02: mean +11.5 mV, all within one count; no-offset fit = 20.15 mV/count.

Raw frames (24 bytes, after preamble AA 2D D4):
```
0058ac  a20058ac228c004520007916593e93517e637b454b080dc4
005b45  a2005b4502ce004d200084167743935277837ba04c088dae
005cb5  a2005cb5028e005c20008516565393527e6f7b6a4c08e9f7
005995  a20059950289003a20008416473293547e807b884d086429
0057a2  a20057a2026a21c53000841669459358774d7b364c08465f
005c90  a2005c90027e0ce6200085169d32935e82787b684c0814c0
```
(0057a2 and 005c90 are two units not in the current test set; I can capture `.cu8` for them later if that's useful for rtl_433_tests.)

vgabor99:
Thanks for confirming my findings, @Verstreubulator!

> four fresh units all near one voltage, a fixed per-unit byte15 happened to be monotonic with voltage, so it correlated by luck

I wonder if this is truly constant. Thinking aloud, what if, for example, the sensor stores its insertion-time voltage to handle different battery chemistries better. I plan to investigate this more as my time allows.

> So please don't lean on those app values as battery ground truth; a meter is the right reference, which is what you used

Not having an Ecowitt gateway played a significant role in this ;D

> A note on my situation, so you know what to expect from me

I don't think you have to explain yourself at all. Thanks for putting your time and effort into supporting the community!

> worth keeping the "no data yet on when the WH52 quits" caveat you and @gdt discussed

I agree. Initially I had the naive idea of WH52 being similar to WH51 (and adopt the battery reporting), but now I see that this was a wrong assumption. Apart from the protocol being different, their radio performance is different. Got a few WH51 units and they pale in comparison. In problem spots, I had to swap WH51-s with WH52-s to be able to get an signal at all. Could be an improved radio/antenna or increased transmit power. We need to learn the battery depletion curve from field data.

gdt:
I read the diff.  Looks fine, and it built without warnings on my Debian 13 box.    I have kicked off CI (sorry, should have hit that button earlier, and didn't realiize it was watiing for somebody with maintainer privs to allow a CI run) and intend to merge if it passes.

PR #3673, opened 2026-08-28
Fix crash decoding DeltaDore X3D sensor messages

peio42:
## Summary

Fix a segmentation fault in the DeltaDore X3D decoder when decoding
sensor messages (type 0x00).

The decoder previously parsed every message with a payload as a
standard message (type 0x01). The standard payload parser consumes
12 bytes unconditionally.

For shorter sensor frames this advanced `bytes_read` past the end of
the frame. The resulting `len - bytes_read - 2` underflow was then
passed to `data_hex()`, causing an out-of-bounds read and a crash.

A valid 20-byte X3D sensor frame (type 0x00) triggered the issue.

For this frame, bytes_read is 16 before attempting to parse the
12-byte standard payload, which advances it past the 20-byte frame.
The subsequent raw payload length calculation underflows and causes
an out-of-bounds read.

## Fix

- Parse the standard payload only for `DELTADORE_X3D_MSGTYPE_STANDARD`
  messages.
- Validate that enough bytes remain before parsing a standard payload.
- Keep valid sensor messages partially decoded instead of rejecting them.

### DIFF
--- a/src/devices/deltadore_x3d.c
+++ b/src/devices/deltadore_x3d.c
@@ -157,6 +157,7 @@ To get raw data:
 #define DELTADORE_X3D_HEADER_TEMP_OUTDOOR     0x01
 
 #define DELTADORE_X3D_MAX_PKT_LEN             (64U)
+#define DELTADORE_X3D_STANDARD_PAYLOAD_LEN    (12U)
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks!

PR #3418, opened 2025-12-11
Add support for ThermoPro TX-2B Outdoor Thermometer and Humidity Sensor

philippem:
```
$ ./build/src/rtl_433 -f 915M -s 250k -R 291 -v -vv
rtl_433 version 25.02-72-ge24bf0a2 branch feat-thermopro-tx2b at 202512102017 inputs file rtl_tcp RTL-SDR

New defaults active, use "-Y classic -s 250k" if you need the old defaults

[Protocols] Registered 1 out of 291 device decoding protocols
[Input] The internals of input handling changed, read about and report problems on PR #1978
[SDR] Found 1 device(s)
[SDR] trying device 0: Nooelec, NESDR SMArt v5, SN: 06064698
Found Rafael Micro R820T tuner
[SDR] Using device 0: Nooelec, NESDR SMArt v5, SN: 06064698, "Generic RTL2832U OEM"
Exact sample rate is: 250000.000414 Hz
[R82XX] PLL not locked!
[SDR] Sample rate set to 250000 S/s.
[Input] Bit detection level set to 0.0 (Auto).
[SDR] Tuner gain set to Auto.
[Input] Reading samples in async mode...
[SDR] rtlsdr_set_center_freq 915000000 = 0
[SDR] Tuned to 915.000MHz.
[acquire_thread] acquire_thread enter...
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
time      : 2025-12-10 20:23:52
model     : Thermopro-TX2B                         Id        : 149
Channel   : 2            Battery   : 1             Temperature: 21.3 C       Humidity  : 49 %          Button    : 0
[...]

PR #3675, opened 2026-09-01
Add support for ThermoPro TX-2B to Thermopro-TX2C

philippem:
The TX-2B is the 915 MHz (North America ISM band) variant of the TX-2C and is
protocol-identical to it, so this adds it to the existing decoder rather than
adding a second, near-duplicate one. There are no decoding logic changes.

### The bug

`gap_limit` was 3829 us, only 4 us above `long_width` 3825. Measured across
five units the long gap runs 3810-3920 us, so on units at the top of that
spread the long gaps were classified as row boundaries instead of data bits.
The transmission shatters into ~50 rows of 0-9 bits,
`bitbuffer_find_repeated_row(bitbuffer, 4, 36)` finds nothing, and the decode
fails silently — and only on some units, which is why this wasn't noticed.

Widening `gap_limit` to 7000 us clears the 3920 us worst case and stays well
below the ~8550 us packet gap and the unchanged `reset_limit` of 8643. It is
also the value `thermopro_tx2` already uses.

### Verification

- No TX-2C regression: the existing `tests/thermopro-tx2c` fixture (868 MHz)
  passes against the patched build, 2 records, 0 failed.
- The `thermopro-tx2` samples still decode only as Thermopro-TX2; no
  cross-decode.
- Running live, all five of my units decode simultaneously, including the one
  whose long gap motivated the change. Two of the IDs that decoded appear in
  none of the captures, so the fix generalises beyond the recorded samples.
- `maintainer_update.py` reproduces the two protocol 245 lines in `README.md`
  and `conf/rtl_433.example.conf` exactly as committed.

The hardware is a TX-2B: the case label reads `FCC ID: 2AATP-TX-2B` /
`Model No.: TX-2B`, photographed alongside the samples.

### Test signals

Companion fixtures: merbanan/rtl_433_tests#522

They require this change in order to decode, so they need to land after or
with it.

Supersedes #3418.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [242]* Baldr / RainPoint rain gauge.
     [243]  Celsia CZC1 Thermostat
     [244]  Fine Offset Electronics WS90 weather station
-    [245]* ThermoPro TX-2C Thermometer and Humidity sensor
+    [245]* ThermoPro TX-2C, TX-2B Thermometer and Humidity sensor
     [246]  TFA 30.3151 Weather Station
     [247]  Bresser water leakage
     [248]  Nissan TPMS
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
Looks minimal and clean, thanks!

PR #3679, opened 2026-09-05
Add support for Fineoffset WN35 Leaf Wetness sensor

zuckschwerdt:
Adds a decoder for Fineoffset WN35, template copied from WN34 etc.

See also #3665

### DIFF
--- a/README.md
+++ b/README.md
@@ -478,6 +478,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [382]* Cotech 36-7900 rain gauge
     [383]  Silver Spring Networks mesh endpoint (-s 1600k)
     [384]  Bresser SmartHome Garden soil moisture and water timer valve (Baldr Homgar, RainPoint)
+    [385]  Fine Offset Electronics WN35 Leaf Wetness sensor
 
 * Disabled by default, use -R n or a conf file to enable
 
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -624,6 +624,7 @@ convert si
 # protocol 382 # Cotech 36-7900 rain gauge
   protocol 383 # Silver Spring Networks mesh endpoint (-s 1600k)
   protocol 384 # Bresser SmartHome Garden soil moisture and water timer valve (Baldr Homgar, RainPoint)
+  protocol 385 # Fine Offset Electronics WN35 Leaf Wetness sensor
 
 ## Flex devices (command line option "-X")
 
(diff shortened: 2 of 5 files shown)


hansniekus:
WN35 Leaf Wetness Sensor

` 0  1  2  3  4  5  6  7  8  9 10 11 ?`
`35 00 30 5E 01 06 55 07 55 70 EB 40 8`
`YY II II II .. RR RR WW VV XX AA`

WN35 ID=305E Wetness=7% Voltage=1.7V

Y (byte 0): sensor-type-byte = 0x35
I (bytes 1-3): ID (24 bit)
R (bytes 5-6): Raw AD value
W (byte 7): Wetness
V (byte 8): V * 20mV 
X (byte 9): CRC
A (byte 10): checksum

I used:
RTL-SDR Blog V4
RTL_433 Release 25.12
Widows 11

At first, I used this string:
-f 868.35M -s 1024k -F json -M utc -X n=sensor,m=FSK_PCM,s=58,l=58,t=5,r=5000,g=4000,preamble=aa2dd4

One of the WN35s would connect occasionally but often encountered data errors.
The WN34s never connected either, though the WN34L did.
It only started working flawlessly when I switched to -Y classic -s 250 (instead of -s 1024).

-f 868.35M -Y classic -s 250k -F json -M utc -X n=sensor,m=FSK_PCM,s=58,l=58,t=5,r=5000,g=4000,preamble=aa2dd4

some exampels from the 2 units:
18:28:58 WN35  1 305E 0%  1.68V CS=True(43) 3500305e01061c00540943a028
18:30:18 WN35  1 305E 0%  1.68V CS=True(43) 3500305e01061c00540943a010
18:31:39 WN35  1 305E 0%  1.68V CS=True(8A) 3500305e01061d00544f8a2010
18:32:58 WN35  1 305E 0%  1.68V CS=True(8A) 3500305e01061d00544f8a2010
18:29:03 WN35  2 4139 0%  1.72V CS=True(F4) 35004139010588005661f42014
18:30:23 WN35  2 4139 14%  1.72V CS=True(6C) 3500413901074c0e56056c2010
18:31:43 WN35  2 4139 35%  1.72V CS=True(76) 350041390109682356dc762020
18:33:02 WN35  2 4139 39%  1.72V CS=True(AC) 350041390109d62756a0ac2012
 
The voltage matches what the GW2000 sees. The wetness reading matches the display and the GW2000.

When you insert the battery into the WN35, you will see a number on the display fluctuating for a while. The final value shown on the display corresponds to bytes 5 end 6. So you must do that when your sensor is dry or else the calibration is not correct. That value rises with the wetness. 

The used command string also could detect the folling Ecowitt sensors: 
WS90/WH40/WN31/WN32/WH57/WH51/WH52/WN34S/WN34L

(My second unit must have other harware becouse it still works when the voltage is 0.72 volt.)

Hans Niekus

PR #3685, opened 2026-09-07
Add support for Olympia Protect 9571 alarm system sensors

pzlabuda:
Add support for the Olympia Protect 9571 alarm system sensors.

- 868.5 MHz, FSK-PCM, 366 µs bit width
- Preamble `aa aa 2d d4`, 48-bit payload
- Checksum: sum of bytes 0-4 & 0xFF (mic "CHECKSUM")
- Output: subtype (Contact / PIR-Motion / Keyfob), 24-bit id,
  state (open/closed, motion/idle) for sensors, cmd + cmd_id for the
  keyfob (Disarm, Arm Home + Siren, Arm Siren Only, Arm Home Only, SOS),
  battery_ok

Test captures: https://github.com/merbanan/rtl_433_tests/pull/523

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,6 +479,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [383]  Silver Spring Networks mesh endpoint (-s 1600k)
     [384]  Bresser SmartHome Garden soil moisture and water timer valve (Baldr Homgar, RainPoint)
     [385]* Neptune R900 BCD flow meters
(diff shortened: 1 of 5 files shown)


zuckschwerdt:
We always need to carefully check that "subtype" is [properly used](https://triq.org/rtl_433/DATA_FORMAT.html#message-data), it aligns well with other security decoders here.

Code and comments look good, thanks!

PR #3688, opened 2026-09-08
Add support for ThermoPro TP212B

uber027:
Adds support for the ThermoPro TP212B 915 MHz thermometer.

The TP212B uses the same modulation, sync word, packet layout, and
temperature encoding observed for the existing TP211B decoder.

The TP212B uses a different final checksum XOR constant:
- TP211B: 0x411B
- TP212B: 0x455B

The existing TP211B decoder now recognizes both checksum variants and
reports the appropriate model.

Tested with a TP212B sensor at 915 MHz.

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,7 +390,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
     [294]  ThermoPro TempSpike XR TP862b / TP863b Wireless Dual-Probe Meat Thermometer
     [295]  Airpuxem TPMS TYH11_EU6_ZQ
     [296]  Apator Metra E-ITN 30 heat cost allocator
-    [297]  ThermoPro TP211B Thermometer
+    [297]  ThermoPro TP211B/TP212B Thermometer
     [298]  TRW TPMS OOK OEM and Clone models
     [299]  TRW TPMS FSK OEM and Clone models
     [300]  Govee Water Leak Detector H5059
--- a/conf/rtl_433.example.conf
+++ b/conf/rtl_433.example.conf
@@ -536,7 +536,7 @@ convert si
   protocol 294 # ThermoPro TempSpike XR TP862b / TP863b Wireless Dual-Probe Meat Thermometer
   protocol 295 # Airpuxem TPMS TYH11_EU6_ZQ
   protocol 296 # Apator Metra E-ITN 30 heat cost allocator
-  protocol 297 # ThermoPro TP211B Thermometer
+  protocol 297 # ThermoPro TP211B/TP212B Thermometer
   protocol 298 # TRW TPMS OOK OEM and Clone models
   protocol 299 # TRW TPMS FSK OEM and Clone models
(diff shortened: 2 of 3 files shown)


ProfBoc75:
@uber027 Nice, is there any possibility to identify the model with the first byte of the ID ?
The issue is false positive result based on wrong checksum value.

On current TP211B, I saw 0x00 or 0x01 for the first byte, when I worked on the decoder.
On your TP212B, you have 0x03.
But may be else where into the ID there is bit flags related to the model.
Based on that flags + checksum, could improve the validation, reduce false positive answers.

Simple question, since you have the TP212B, I guess you are using your code since a while and did you notice, time to time, your sensor was decoded as TP211B instead of 212B ?

uber027:
Thanks, that makes sense. I haven’t seen my TP212B get decoded as a TP211B so far. Every valid packet I’ve seen from it has matched the TP212B checksum.
My unit’s ID is 03 89 71, so the 0x03 is interesting if the TP211B units you’ve seen were 0x00 or 0x01. I’m just not sure I’d want to rely on that yet based on one TP212B sample. It could be model-related, but it could also just be part of the ID range.
I have had some trouble decoding weaker signals from this unit at longer range. The factory receiver is still able to receive it in cases where rtl_433 doesn’t, so I’ve been doing some testing around the FSK detection/slicing side as well. I haven’t seen that result in the sensor being misidentified as a TP211B, though — it just fails to decode.

ProfBoc75:
> Thanks, that makes sense. I haven’t seen my TP212B get decoded as a TP211B so far. Every valid packet I’ve seen from it has matched the TP212B checksum. My unit’s ID is 03 89 71, so the 0x03 is interesting if the TP211B units you’ve seen were 0x00 or 0x01. I’m just not sure I’d want to rely on that yet based on one TP212B sample. It could be model-related, but it could also just be part of the ID range. I have had some trouble decoding weaker signals from this unit at longer range. The factory receiver is still able to receive it in cases where rtl_433 doesn’t, so I’ve been doing some testing around the FSK detection/slicing side as well. I haven’t seen that result in the sensor being misidentified as a TP211B, though — it just fails to decode.

Agree. We can deploy your code, add the remark about model discrepancy and may be, later, someone could confirm / rafine the decoder with another ID.

Regarding your decoding and reception issues:
- Move the USB dongle as far away from the PC as possible using a shielded, good-quality USB extension cable. This reduces interference from the computer 
- Take care of the antenna too; similarly, keep it away from any sources of electronic interference, such as a monitor, TV or computer...
- There should be as few obstacles as possible between the dongle and the sensors; avoid metal or thick walls.
- The dongle itself, the best are with Rafael Micro r828d or r860 tuner (formerly r820t or r820t2), avoid Fitipower fc001x, they bad sensitivity.

Last point in order to add the support of this TP212B, add some cu8 samples at https://github.com/merbanan/rtl_433_tests could help also to improve the decoder.

PR #3689, opened 2026-09-08
Add La Crosse TX63U-IT wind sensor support

kbloxsom:
Adds support for the La Crosse TX63U-IT solar wind sensor (FCC ID OMO-M-12) and introduces a reusable DBPSK demodulation path for raw IQ input.

### Highlights

* Adds TX63U-IT protocol decoder with CRC-16/X-25 validation.
* Decodes average wind speed, gust speed, wind direction, and gust direction.
* Adds reusable PSK candidate detection and differential-BPSK symbol recovery.
* Integrates PSK demodulation into the streaming IQ path with duplicate suppression.
* Does not invent a sensor ID; currently observed fixed bytes are treated only as protocol constants.

### Protocol details validated from captures

* Approximately 36 ksymbol/s differential phase-coded signal.
* HDLC-style `0x7E` framing with zero-bit stuffing.
* Bytes transmitted LSB-first.
* CRC-16/X-25:

  * polynomial `0x1021`
  * init `0xFFFF`
  * reflected input/output
  * xorout `0xFFFF`

### Validation performed

* All 5 available TX63U-IT captures decode correctly.
* Verified at both 1 Msps and 250 ksps.
* 17/17 shifted stream-boundary tests passed.
* Multiple distinct packets in one continuous stream decode correctly.
* 6/6 short-gap tests passed, including a synthetic 0 ms gap.
* Existing CTest suite passes.

Related test-data PR: `merbanan/rtl_433_tests#524`

### DIFF
--- a/examples/rtl_433_mqtt_hass.py
+++ b/examples/rtl_433_mqtt_hass.py
@@ -463,6 +463,17 @@ mappings = {
         }
     },
 
+    "wind_max_dir_deg": {
+        "device_type": "sensor",
+        "object_suffix": "GWD",
+        "config": {
+            "name": "Gust Direction",
+            "unit_of_measurement": "°",
+            "value_template": "{{ value|float }}",
+            "state_class": "measurement"
+        }
+    },
+
     "rain_mm": {
         "device_type": "sensor",
         "object_suffix": "RT",
--- a/include/psk_demod.h
+++ b/include/psk_demod.h
@@ -1,0 +1,37 @@
+#ifndef INCLUDE_PSK_DEMOD_H_
+#define INCLUDE_PSK_DEMOD_H_
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "bitbuffer.h"
+
+typedef struct {
+    double offset_s;
+    double signature_db;
+    double carrier_offset_hz;
+    double symbol_rate_hint;
+} psk_candidate_t;
+
+typedef int (*psk_validate_fn)(
+        bitbuffer_t *bitbuffer,
+        void *context);
+
+int psk_find_candidate(
+        uint8_t const *iq_buf,
+        size_t sample_count,
+        unsigned sample_size,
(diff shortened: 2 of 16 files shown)


kbloxsom:
Just following up on this PR. The decoder has been validated against all available TX63U-IT captures at 1 Msps and 250 ksps, including stream-boundary and short-gap tests. The related test-data PR is merbanan/rtl_433_tests#524. Please let me know if any additional tests or changes would help with review.

zuckschwerdt:
First need for a PSK demod and we get the demod as bonus, thanks!
This will need some time to read and test through, but already looks very nice.

One thing we likely need is clear copyright and license on each file.

Clang analyzer seems to detect a logic problem where c_sub() is call with an uninitialized value? Use longer lines where you can, e.g. lacrosse_tx63_decode() needs to be one line for our scripts to recognise. Some EOF are missing.

zuckschwerdt:
Does the PSK demod really need to run without packet framing? Or was that just easier to implement?

kbloxsom:
PSK does not inherently need to run without packet framing. I used overlapping raw-IQ windows because the existing packet path exposes pulse data rather than the corresponding raw-IQ region, and the rolling buffer also handles TX63 bursts crossing SDR input-buffer boundaries.

A framed implementation should be possible: retain a short raw-IQ history, use the existing envelope detector to identify the completed RF packet, then run candidate estimation and DBPSK demodulation only on that padded packet region. That would avoid continuously running the FFT, although I would first need to verify reliable envelope framing at both 250 ksps and 1 Msps.

I’ll separately fix the current CI issues: make the cumulative buffer initialization explicit for Clang, put the decoder signature on one line for `symbolizer.py`, add clear copyright/license headers, and restore the missing final newlines.

Would you prefer that I also convert the PSK path to packet-triggered demodulation in this PR, or keep the validated rolling-window implementation for now?

PR #3691, opened 2026-09-10
Fix vevor bytewise offset

zaubara:
## Summary

Fix decoding of multi-byte values in the Vevor 7-in-1 weather station protocol.

The protocol applies an offset of one to each encoded byte independently. Subtracting `0x0101` after combining the bytes causes an unintended borrow when the low byte wraps from `0xff` to `0x00`. This produced large discontinuities, most visibly in the cumulative rainfall value.

This change:

- decodes wind speed, wind direction, rainfall, and illuminance by removing the offset from each byte before combining them;
- documents the observed rainfall scale of 0.233 mm per tip;
- validates that the trailing TX counter is one greater than the first TX counter, including the `0xff` to `0x00` wrap.

## Verification

Tested with a Vevor YT60231-compatible 868 MHz weather station through `rtl_433_ESP` and OpenMQTTGateway.

Nine observed checksum-valid frames satisfied the TX counter relationship.

Synthetic checksum-valid boundary vectors decode as:

- encoded rain `0x1000`: 954.135 mm
- encoded rain `0x1101`: 954.368 mm

The transition is therefore one rain-gauge tip of 0.233 mm instead of the previous large discontinuity.

An otherwise checksum-valid frame with an invalid trailing TX counter is rejected.

## Tests

- Full CMake build
- Complete CTest suite: 16/16 passed
- Project style check
- Existing documented Vevor test vector
- Synthetic low-byte wrap boundary vectors

### DIFF
--- a/src/devices/vevor_7in1.c
+++ b/src/devices/vevor_7in1.c
@@ -13,7 +13,7 @@
 
 #include "decoder.h"
 
-/**
+/** @fn static int vevor_7in1_decode(r_device *decoder, bitbuffer_t *bitbuffer)
 Vevor Wireless Weather Station 7-in-1.
 
 Manufacturer : Fujian Youtong Industries Co., Ltd. rebrand under Vevor name.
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Interesting find. So each byte would independently wrap? With 0xff data encoded as 0x00 and not as 0x100, right? Does make sense.

The decoder doc-comment needs an `@fn` annotation to attach the function.

zuckschwerdt:
Another simpler idea might be to just put a decoding block something like this before extracting the fields?
```
b[8] = b[8] - 1;
b[9] = b[9] - 1;
b[11] = b[11] - 1;
b[12] = b[12] - 1;
b[13] = b[13] - 1;
b[14] = b[14] - 1;
b[16] = b[16] - 1;
b[17] = b[17] - 1;
```

zaubara:
Yes, exactly: the offset is applied separately to each byte, wrapping at 256. So 0xff becomes 0x00 without affecting the next byte.
Sounds good, I included your suggestions in the PR.

PR #3696, opened 2026-09-12
Fix FSK pulse detection after short leading pulses

pzlabuda:
A transmission leading with a short carrier blip was always classified as OOK, so FSK bursts in later pulses were never decoded (e.g. Olympia Protect 9571, whose ~84 us blip precedes 5x 38 ms FSK-PCM bursts).

This evaluates later pulses with fresh FSK state, but only for the first two pulses and only trusting sustained (> 1 ms) ones, keeping first-pulse behavior unchanged. Incidental FM chatter on short OOK pulses can therefore no longer steal a package as a garbage FSK package.

Test fixtures: https://github.com/merbanan/rtl_433_tests/pull/525 (add it as the first reviewer check — the 3 new samples decode to 5 packets each with this change and to nothing without it).

Verification with rtl_433_tests (5554 records, run_test.py -I time --first-line): master baseline 102 fails -> 99 fails, the only delta being the 3 fixed Olympia fixtures; no new failures, false-positive warnings unchanged (Bresser pair only).

### DIFF
--- a/src/pulse_detect.c
+++ b/src/pulse_detect.c
@@ -235,8 +235,15 @@ int pulse_detect_package(pulse_detect_t *pulse_detect, int16_t const *envelope_d
 
         case PD_OOK_STATE_GAP_START:
             s->ook_state = PD_OOK_STATE_GAP;
-            // Determine if FSK modulation is detected
-            if (fsk_pulses->num_pulses > PD_MIN_PULSES) {
+            // Determine if FSK modulation is detected (same sustained-pulse
+            // gate for non-first pulses as in the live path above; samp_rate
+            // is used here as samples_per_ms is not in scope yet).
+            int fsk_detected = fsk_pulses->num_pulses > PD_MIN_PULSES;
+            if (fsk_detected && pulses->num_pulses > 0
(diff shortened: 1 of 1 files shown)


merbanan:
There are some pending work to re-architect the fsk demod flow. It might take a long time before that is finally merged. When that is done we can look at this PR.

pzlabuda:
That's ok, I will rework this PR once that changes are merged.

Also a quick update to my previous note: the numbers I posted (master 102 → 99 failures) came from `run_test.py -I time --first-line`.

I’ve also rerun in full-line mode:
* master baseline: 5554 records, 110 failures
* this branch: 5554 records, 111 failures

This is due to more packets being decoded:
* 3 Olympia-9571 fixtures are now fully decoded.
* 4 captures produce one additional decode each (Cavius Door CLOSE/OPEN, Steelmate TPMS leak/normal1). These are genuine packets that were previously missed.

In case of the Cavius test files, I can see the exact same carrier blip before each packet as in Olympia's case. Might be a quirk of the chip being used (Si4010-C2 in case of Olympia).

pzlabuda:
I refactored the fix into what (I think) is a more elegant solution.

PR #3700, opened 2026-09-17
Add generated file name to w option

zuckschwerdt:
Adds a template name (`"rec%03u_%gM_%gk.%s"`) to the `-w` option to automatically generate a file name for the recording.

I.e. `-w myfile_868M_1000k.cu8` can become `-w REC` and a file name of `rec001_868M_1000k.cu8` will be used.

Is the magic argument of `REC` good enough? Should we use a real template with e.g. user defined prefix (e.g. `-w CU8:PREFIX:myfile`)?


merbanan:
LGTM

zuckschwerdt:
Thanks! I think I'll refine the condition to be an empty base filename. So either of these will work:
`-w ""` `-w CU8:` `-w CS8:` `-w AM:S16:` `-w FM:S16:` `-w CS16:` `-w CF32:` `-w AM:F32:` `-w FM:F32:` `-w I:F32:` `-w Q:F32:` `-w LOGIC:U8:` `-w VCD:` `-w OOK:` `-w SIGMF:`
I figure this is the most plain and still allows for some flexibility. And it aligns nicely with e.g. the `-w OOK:-` we already have to write to stdout.

PR #3705, opened 2026-09-20
Add SoapySDR CF32 input for RedPitaya

zuckschwerdt:
Adds detection for SoapySDR CF32 format, conversion to internal CS16.
The order now is native CU8, native CS16, forced CS16, forced CF32, error.

Also prints an "no supported sample formats" error otherwise.


zuckschwerdt:
Closes #3701

Issue #3611, opened 2026-07-16
Multiple decoders are default on and produce execssive false decodes, rendering HA integration unusable


gdt:
I'm trying @deviantintegral's https://rtl-433-hass.github.io/rtl_433/latest/ in a staging server, and getting vast numbers of "new device" notifications, almost all of which (after the real devices were discovered quickliy) I believe do not exist.

We have a longstanding norm that decoders should be default on only if they have enough CRC/checksum/etc. to reasonably reliably not produce decodes when their device has not actually transmitted.

At a quick scan of HA-discovered devices, the most problematic devices are:
  - Schrader-MRXBC5A4 
  - Megacode Remote
  - Microchip HCS361
  - Compustar-1WG3R

From using jq .payload.model on my log, and finding outliers, these are all things that have showed up a tiny bit and which I believe are not right.   I will write better analysis code.  The 4 above are bigger offenders and fixing them and iterating would be a good start.

   1 "BM5-v2"
   1 "Oil-SonicStd"
   2 "Chuango-Security"
   2 "FS20"
   2 "HT680-Remote"
   3 "Efergy-e2CT"
   3 "Interlogix-Security"
   3 "Markisol"
   3 "Secplus-v1"
   3 "Waveman-Switch"
   6 "Secplus-v2"

 

 
I haven't looked at source code yet.

I would like to take a look to see what kind of CRC/etc. there is, and make them not default enabled if they are indeed prone to falsing.

There's a separate not-yet-feature consideration, which is that the HA integration should have some kind of "is this real" detection in terms of multiple events before auto-adding, but that's out of scope for rtl_433.  This issue is about decodes being emitted by default, from decoders that unreasonably emit false decodes.

merbanan:
Please add more to the list and I'll make sure they get better validation or we disable them by default.

zuckschwerdt:
This might concern the batch of button-remotes / keyfobs from #3513, which picked up #2764 (no issue in the git log as it was a rebase merge, not a squash merge).

We could perhaps expand the enable list to: default enabled / extended enabled (keyfobs…) / default disabled (no mic) / hard disabled (template decoder)? I.e. insert an "extended" category of decoders.

zuckschwerdt:
Some time ago I had the idea of a new output that could be of help to spot these false positives and also give us feedback which decoders are of core concern: a "crowd source" output.

I.e. an easy way for users to donate their received signals to the project. There would be obvious privacy concerns to address.
[...]

PR #3708, opened 2026-09-21
Add support for Schrader AG6SP4

kssmll:
Modified schrader_EG53MA4_decode to process packets with a variable number of sync bytes. 
The function now searches for a hardcoded preamble. 
Tested with GM 13581558 (EG53MA4) and GM 13598773 (AG6SP4).


TheButterZone:
Think the preamble search will kill a false decode like this "I'm in boiling pure ethanol"?

time      : 2026-09-22 09:06:23
model     : Schrader-EG53MA4                       type      : TPMS          flags     : 23495f17      ID        : 39FB86
Pressure  : 72.2 PSI     Temperature: 174.0 F      Integrity : CHECKSUM

Heckuva coincidence...

kssmll:
> Think the preamble search will kill a false decode like this "I'm in boiling pure ethanol"?

Are these real data? Which sensor are these data from? With my changes, this packet will be dropped.

TheButterZone:
> > Think the preamble search will kill a false decode like this "I'm in boiling pure ethanol"?
> 
> Are these real data? Which sensor are these data from? With my changes, this packet will be dropped.

All I ever do is copy-paste terminal prints... tonight that was under

./rtl_433 -f 315M -f 345M -f 433.92M -f 868.3M -f 915M -H 5 -s 1024k -g 42.1 -C customary -R -40 2>&1 

I guess I could apply your commit, but that may have been a one-and-done falsey, unlike the multitude of [what I think are HCS361 falseys](https://github.com/merbanan/rtl_433/issues/3611#issuecomment-5569645930).

As a user I'll probably see it in the next "nightly" like some/most? others. I wish I had regular decodes to justify applying it immediately.

kssmll:
> I guess I could apply your commit, but that may have been a one-and-done falsey, unlike the multitude of [what I think are HCS361 falseys](https://github.com/merbanan/rtl_433/issues/3611#issuecomment-5569645930).

Maybe, but the main goal of these changes is to support new sensors with varying numbers of sync bits at the beginning of the packet.

